PR #43, opened 2017-06-02
release notes for 1.1.4-2 hotfix

azhebel:


azhebel:
@dbpercona @rnovikovP I am not sure what to do with this. It seems we updated the AMIs but new Docker images and OVA were not uploaded. Should I publish this or will this fix be released with 1.1.5?

delgod:
1.1.4-2 is only AMI affected

PR #52, opened 2017-07-06
DOC-533: Added mongodb:queries and --dev-enable

azhebel:


doc/source/pmm-admin.rst
```diff
@@ -378,6 +381,49 @@ use a command similar to the following:
...
+   To add it in a non-production environment,
+   use the ``--dev-enable`` option for ``pmm-admin add``.
+
+**USAGE**
+
+.. code-block:: text
+
+   pmm-admin add mysql:queries [NAME] [OPTIONS]
+
+This creates the ``pmm-mongodb-queries-0`` service
```
michaelcoburn: @azhebel Why is this named pmm-mongodb-queries-0 and not the actual port number that the service uses, for example pmm-mysql-queries-42001 ?

azhebel: MySQL QAN service used to be with no port. Not sure when that changed. I will update accordingly.

azhebel: Confirmed that no port is used for QAN (both MySQL and MongoDB). It used to have a port number in the service name, but not any more.

PR #189, opened 2018-06-19
Change the representation of release notes; Update referencing of rep…

borys-belinsky:
…lacement directives


dbpercona:
There is some concern that these changes will negatively affect our search engine results and rankings.  Before we push these changes,  we need to discuss redirect mappings which can be requested by IT.

PR #233, opened 2018-10-18
PMM-3118 fix removing warning about QRT plugin making very high overhead

fiowro:


JNKPercona:
http://docs-test.cd.percona.com/PR-233/

PR #282, opened 2019-03-06
PMM-3478: Change runs_on_node_id to pmm_agent_id

DexterHD:


api/inventory/agents.proto
```diff
@@ -51,7 +51,7 @@ message MySQLdExporter {
   // Unique randomly generated instance identifier.
   string agent_id = 1;
   // Node identifier where this instance runs.
```
AlekSi: Comment should be updated everywhere.

JNKPercona:
http://docs-test.cd.percona.com/PR-282/

api/inventory/agents.proto
```diff
@@ -158,8 +158,8 @@ message QANMySQLPerfSchemaAgent {
 
 message ListAgentsRequest {
   oneof filter {
-    // Return only Agents running on that Node.
-    string runs_on_node_id = 1;
+    // Return only Agents running near concrete PMMAgent.
```
AlekSi: Not "near". pmm-agent starts other Agents.
And not "concrete". That means "solid" and "physical", but it's not solid or physical. (It does not mean "specific" in the same sense as word "конкретный").

JNKPercona:
http://docs-test.cd.percona.com/PR-282/

api/inventory/agents.proto
```diff
@@ -28,8 +28,8 @@ message PMMAgent {
 message NodeExporter {
   // Unique randomly generated instance identifier.
   string agent_id = 1;
-  // Node identifier where this instance runs.
-  string node_id = 2;
+  // pmm-agent identifier where this instance runs.
```
AlekSi:
```suggestion
  // The pmm-agent identifier which runs this instance.
```

Do the same everywhere.

DexterHD: This is your comment just copied from https://github.com/percona/pmm/blob/0efdbf9d6a2a97301af3849456dacbd0189839be/api/inventory/agents.proto#L151 but nevermind, I'll fix that.

AlekSi: Yes, but it makes less sense. :) "where" points to the place. The place stays the same – some Node (host). pmm-agent is not a place – it is a way to run agent.

AlekSi:
For you to merge together with PRs for pmm-managed and (maybe) pmm-agent.

JNKPercona:
http://docs-test.cd.percona.com/PR-282/

PR #287, opened 2019-03-07
PMM-3595 Add MySQL username and password.

AlekSi:


api/inventory/agents.proto
```diff
@@ -154,6 +154,20 @@ message QANMySQLPerfSchemaAgent {
   string pmm_agent_id = 2;
   // Service identifier.
   string service_id = 3;
+  // MySQL username.
```
DexterHD:
```suggestion
  // MySQL username for scraping metrics.
```

api/inventory/agents.proto
```diff
@@ -154,6 +154,20 @@ message QANMySQLPerfSchemaAgent {
...
+  string username = 4;
+  // MySQL password.
```
DexterHD:
```suggestion
  // MySQL password for scraping metrics.
```

api/inventory/agents.proto
```diff
@@ -379,6 +393,14 @@ message AddQANMySQLPerfSchemaAgentRequest {
       string_not_empty: true
     }
   ];
+  // MySQL username.
```
DexterHD:
```suggestion
  // MySQL username for scraping metrics.
```

api/inventory/agents.proto
```diff
@@ -379,6 +393,14 @@ message AddQANMySQLPerfSchemaAgentRequest {
...
+  string username = 4 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  // MySQL password.
```
DexterHD:
```suggestion
  // MySQL password for scraping metrics.
```

api/inventory/agents.proto
AlekSi: But it's not for scraping metrics! 😆

JNKPercona:
http://docs-test.cd.percona.com/PR-287/

api/inventory/agents.proto
DexterHD: Ok, so for what?? 😕

JNKPercona:
http://docs-test.cd.percona.com/PR-287/

api/inventory/agents.proto
AlekSi: We want to keep using the word "metrics" only for Prometheus metrics. QAN is about performance data / query statistics / performance events… Terminology is a bit in flux. But not "metrics".

AlekSi: Oh, and "scraping" is very Prometheus-specific.

DexterHD: It's confusing because all these things called "agents" ... We have an `agent` to all life situations 😄

AlekSi: > An **Agent** represents something that runs on the Node which is not useful itself but instead provides insights (metrics, query performance data, etc) about Nodes and/or Services.

DexterHD: Yep, I'm saying about this. It's a big and blurred abstraction I guess. I mean agent doesn't have a single responsibility. You can say that agents responsibility is `provide insights`, but it also very abstract and in fact, it can be different things with different business logic and meaning. So there is a reason I a little bit confusing sometimes.

PR #290, opened 2019-03-12
PMM-3618 update docs with pmm-server:1 tag instead of latest

fiowro:


JNKPercona:
http://docs-test.cd.percona.com/PR-290/

PR #295, opened 2019-03-15
PMM-3576 Sparklines+column selection

askomorokhov:
PMM-3576 Sparklines+column selection


rnovikovP:
Is this only for sparklines?  Looks like this is "Profile" endpoint API 
@askomorokhov  ?

JNKPercona:
http://docs-test.cd.percona.com/PR-295/

api/qanpb/profile.proto
```diff
@@ -39,32 +40,37 @@ message ReportMapFieldEntry {
...
-message Stats {
-  float num_queries = 1;
-  float m_query_time_sum = 2;
-  float m_query_time_min = 3;
-  float m_query_time_max = 4;
-  float m_query_time_p99 = 5;
+// Metric cell.
+message Metric {
+  Stat stats = 1;
+}
```
AlekSi: Do we really need that type with a single field? Will it grow?

api/qanpb/profile.proto
```diff
@@ -39,32 +40,37 @@ message ReportMapFieldEntry {
...
+  float sum = 3;
+  float min = 4;
+  float max = 5;
+  float p99 = 6;
+}
+
+// Point is a values of all selected metrics for one point of sparkline.
+// points: x -> timestamp or point, y -> values of any metric.
+message Point {
+  map<string, float> values = 1;
```
AlekSi: Why not two separate float fields?

api/qanpb/profile.proto
askomorokhov: Yes, it will grow up with generic metrics like QPC, percent of total, etc.

api/qanpb/profile.proto
askomorokhov: it is not two separated fields. it will looks like:
```json
 "values": {
            "m_lock_time_sum": 0.00818,
            "m_query_load": 0.002203,
            "m_query_time_avg": 5.6115473e-06,
            "m_query_time_sum": 0.13218,
            "num_queries_sum": 23555,
            ...
            "point": 59,
            "time_frame": 60,
            "timestamp": 1546300800
          }
```

AlekSi: Well, then comment is confusing. It probably should say "key" and "value" instead of "x" and "y".

JNKPercona:
http://docs-test.cd.percona.com/PR-295/

api/qanpb/profile.proto
askomorokhov: Improved comment. But not sure how this helps. Almost avery key can represent values for y coordinate.

PR #308, opened 2019-03-22
PMM-3575 Filters.

askomorokhov:
https://jira.percona.com/browse/PMM-3575


api/qanpb/filters.proto
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// FiltersReply is map of labels for given period by key.
+message FiltersReply {
+  map<string, ListLabels> labels = 1;
+}
+
+// ListLabels is list of label values.
+message ListLabels {
+  repeated NameAndCount values = 1;
```
AlekSi: Can it contain duplicate names? If not, can it be map too?

api/qanpb/filters.proto
```diff
@@ -0,0 +1,41 @@
...
+    option (google.api.http) = {
+      post: "/v1/qan/GetFilters"
+      body: "*"
+    };
+  }
+}
+
+// FiltersRequest contains period for which we need filters.
+message FiltersRequest {
+  google.protobuf.Timestamp from = 1;
```
AlekSi: Can those values be empty? If not, can we add validators?

JNKPercona:
http://docs-test.cd.percona.com/PR-308/

api/qanpb/filters.proto
askomorokhov: No, it cannot contains duplicates. But I do not see much benefits in smaller response versus more complex code: 5 additional loops into Go code to convert data from DB rows and on UI need an additional workaround to iterate over the map (https://github.com/angular/angular/issues/2246)

api/qanpb/filters.proto
askomorokhov: A good point. Added validation in qan-api2 https://github.com/percona/qan-api2/blob/ce300d1f12e4c7f3ea07c26699ac34a9d205baf1/services/analytics/filters.go#L30

JNKPercona:
http://docs-test.cd.percona.com/PR-308/

JNKPercona:
http://docs-test.cd.percona.com/PR-308/

api/qanpb/filters.proto
AlekSi: Then we need to document the fact that duplicates are not possible.

api/qanpb/filters.proto
```diff
@@ -0,0 +1,41 @@
...
+
+option go_package = "qanpb";
+
+import "google/api/annotations.proto";
+import "google/protobuf/timestamp.proto";
+
+// Filters serves list of names of metrics.
+service Filters {
+  // GetFilters gets map of metrics names.
+  rpc GetFilters(FiltersRequest) returns (FiltersReply) {
```
AlekSi: Method `XXX` should use `XXXRequest` and `XXXResponse`.
I propose to call method `Get` (`Filters` is already present in service name) and use URL `/v1/qan/Filters/Get`. That matches other APIs.

api/qanpb/filters.proto
```diff
@@ -0,0 +1,41 @@
...
+
+// FiltersRequest contains period for which we need filters.
+message FiltersRequest {
+  google.protobuf.Timestamp period_start_from = 1;
+  google.protobuf.Timestamp period_start_to = 2;
+}
+
+// FiltersReply is map of labels for given period by key.
+message FiltersReply {
+  map<string, ListLabels> labels = 1;
```
AlekSi: What is key there? It should be documented.

api/qanpb/filters.proto
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// ListLabels is list of label values.
+message ListLabels {
+  repeated NameAndCount values = 1;
+}
+
+// NameAndCount is label values and how many times this value occur.
+message NameAndCount {
```
AlekSi: But it doesn't contain name, isn't it? Then why it is called *Name* and Count?

JNKPercona:
http://docs-test.cd.percona.com/PR-308/

JNKPercona:
http://docs-test.cd.percona.com/PR-308/

PR #311, opened 2019-03-27
PMM-3191 Metadata constants.

DexterHD:


JNKPercona:
http://docs-test.cd.percona.com/PR-311/

api/agentpb/agent.go
```diff
@@ -72,3 +80,21 @@ func GetAgentConnectMetadata(ctx context.Context) AgentConnectMetadata {
 	}
 	return res
 }
+
+// AddAgentServerMetadata adds metadata to pmm-managed's Connect RPC call.
+// Used by pmm-managed.
+func AddAgentServerMetadata(ctx context.Context, md *AgentServerMetadata) context.Context {
+	return metadata.AppendToOutgoingContext(ctx, mdNodeID, md.RunsOnNodeID, mdServerVersion, md.PmmManagedVersion)
```
AlekSi: `AppendToOutgoingContext` works only on client side. See https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#streaming-call-2

DexterHD:
> Can we also do the next step and add `struct` and methods like the one below?

@AlekSi Fixed

api/agentpb/agent.go
```diff
@@ -72,3 +80,21 @@ func GetAgentConnectMetadata(ctx context.Context) AgentConnectMetadata {
...
+}
+
+// GetAgentServerMetadata returns pmm-managed's metadata.
+// Used by pmm-agent.
+func GetAgentServerMetadata(ctx context.Context) AgentServerMetadata {
+	var res AgentServerMetadata
+	md, ok := metadata.FromIncomingContext(ctx)
```
AlekSi: `FromIncomingContext` works only on server side. See https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#streaming-call

api/agentpb/agent.go
```diff
@@ -37,8 +37,10 @@ func (*SetStateRequest_AgentProcess) agentParams() {}
 func (*SetStateRequest_BuiltinAgent) agentParams() {}
 
 const (
-	mdID      = "pmm-agent-id"
-	mdVersion = "pmm-agent-version"
+	mdID            = "pmm-agent-id"
+	mdVersion       = "pmm-agent-version"
+	mdNodeID        = "pmm-agent-node-id"
+	mdServerVersion = "pmm-managed-version"
```
AlekSi: I think we should do two things there:
1. We care about PMM Server version, not each individual component's version (they are the same). So let's use `pmm-server-version` for key.
2. Let's sync constant names with keys. For example, rename `mdID` to `mdAgentID`.

api/agentpb/agent.go
```diff
@@ -47,6 +49,12 @@ type AgentConnectMetadata struct {
 	Version string
 }
 
+// AgentServerMetadata represents metadata sent by pmm-managed to pmm-agent.
+type AgentServerMetadata struct {
+	RunsOnNodeID      string
+	PmmManagedVersion string
```
AlekSi: ServerVersion

api/agentpb/agent.go
```diff
@@ -47,6 +49,12 @@ type AgentConnectMetadata struct {
 	Version string
 }
 
+// AgentServerMetadata represents metadata sent by pmm-managed to pmm-agent.
+type AgentServerMetadata struct {
+	RunsOnNodeID      string
```
AlekSi: Maybe `AgentRunsOnNodeID`? To make it clear that is Agent's Node, not PMM Server's Node.

JNKPercona:
http://docs-test.cd.percona.com/PR-311/

JNKPercona:
http://docs-test.cd.percona.com/PR-311/

api/agentpb/agent.go
DexterHD: 1) How can I get `pmm-server-version`? PMM Server is a "Software System" which consists of `pmm-managed, Prometheus, Grafana, nginx, qan-api2, and qan-app` according to our architecture. There is no info about PMM Server version in `pmm-managed` basically. There is only info about `PMMVersion` and `pmm-managed` version.
2) Ok, I'll fix it. But I think this type of review suggestions have no sense as a conversation about where to add new private methods (after public one or before). At least at this moment. I mean we spend 2 times the longer time for code review than for implement something. If we all will doing code review with the same principle, it stops our development at all.

AlekSi: >they are the same
PMM Server Version == PMM Version == pmm-managed version

2. Sure. But that PR was rejected not because of this, but due to incorrect working with metadata.

api/agentpb/agent.go
```diff
@@ -81,10 +81,13 @@ func GetAgentConnectMetadata(ctx context.Context) AgentConnectMetadata {
 	return res
 }
 
-// AddAgentServerMetadata adds metadata to pmm-managed's Connect RPC call.
+// AddAgentServerMetadata adds metadata to pmm-managed's Connect RPC call and returns it for sending.
```
AlekSi: It doesn't _adds_ anything. Either comment and method name should be updated, or it should become `SendAddAgentServerMetadata` and actually send it. I think the latter variant is better – that way caller doesn't have to think about headers vs trailers distinction.

DexterHD:
@AlekSi Fixed

api/agentpb/agent.go
```diff
@@ -67,8 +78,36 @@ func GetAgentConnectMetadata(ctx context.Context) AgentConnectMetadata {
...
+	return nil
+}
+
+// GetAgentServerMetadata receives metadata from pmm-managed.
+// Used by pmm-agent.
+func GetAgentServerMetadata(stream grpc.ClientStream) (AgentServerMetadata, error) {
+	var res AgentServerMetadata
+	md, err := stream.Header()
+	if err != nil {
+		return res, err
```
AlekSi: Note that we return original error there, not `status.Errorf(…)`.

api/agentpb/agent.go
```diff
@@ -67,8 +78,36 @@ func GetAgentConnectMetadata(ctx context.Context) AgentConnectMetadata {
...
+	if err := stream.SendHeader(header); err != nil {
+		return status.Errorf(codes.Internal, "Can't send server metadata to client.")
+	}
+
+	return nil
+}
+
+// GetAgentServerMetadata receives metadata from pmm-managed.
+// Used by pmm-agent.
+func GetAgentServerMetadata(stream grpc.ClientStream) (AgentServerMetadata, error) {
```
AlekSi: Existing code in this package accepts and returns structs by pointer.

JNKPercona:
http://docs-test.cd.percona.com/PR-311/

JNKPercona:
http://docs-test.cd.percona.com/PR-311/

DexterHD:
@AlekSi Can I fix all those inconsistent things and in `PMM-2.0` branch?

AlekSi:
There is no need :)

DexterHD:
@AlekSi We are in agreement that it sometimes doesn't important, but I should fix that, as beautiful code matters for me too. https://github.com/percona/pmm/pull/312 So ¯\_(ツ)_/¯

PR #327, opened 2019-04-11
PMM-3715 Hot reload API.

DexterHD:


JNKPercona:
http://docs-test.cd.percona.com/PR-327/

api/agentlocalpb/agentlocal.proto
```diff
@@ -57,6 +57,12 @@ message StatusResponse {
   repeated AgentInfo agents_info = 4;
 }
 
+// RestartRequest describes request to restart pmm-agent.
```
AlekSi: Let's call it reload. We do not actually restart pmm-agent process. Emphasis should be on configuration reloading.

JNKPercona:
http://docs-test.cd.percona.com/PR-327/

PR #345, opened 2019-05-07
PMM-3978 Actions.

DexterHD:
It's too early to say that API is ready, as we don't have a lot of requirements at this moment, but at least basic architecture was laid, and `pt-summary` action is working. So I think we can review and merge the current API version.


JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

api/agentpb/agent.proto
```diff
@@ -93,6 +122,9 @@ message AgentMessage {
     // responses from agent
     Pong pong = 8;
     SetStateResponse set_state = 9;
+    ActionRunResponse action_run_response = 10;
+    ActionCancelResponse action_cancel_response = 11;
+    ActionResult action_result = 12;
```
AlekSi: It is not possible to have Response without Request. There should be `CancelActionRequest`/`ActionCancelRequest` from pmm-agent to pmm-managed. The response should be sent from pmm-managed to pmm-agent.

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

api/agentpb/agent.proto
```diff
@@ -81,6 +81,48 @@ message SetStateResponse {
...
+  PT_SUMMARY = 1;
+  PT_MYSQL_SUMMARY = 2;
+  MYSQL_EXPLAIN = 3;
+}
+
+// StartActionRequest represents request to run an action by name.
+message StartActionRequest {
+  string action_id = 1;
+  ActionName name = 2;
+  repeated string parameters = 3;
```
BupycHuk: is only slice of strings enough to run any action?

DexterHD: We have discussed this a little with @AlekSi and @rnovikovP. But I'm not sure. I think at this moment it's enough, at least for all actions running in the `shell`. It can be something like:

```golang
[]string{"-flag1","value_of_flag1","-flag2","-flag3","value_of_flag3"}
```

The same thing, in theory, can be used for other actions. If we need `K=>V`, we can use `odd`
elements for `Keys` and `even` for `Values`. For example:
```golang
[]string{"key1","value1","key2","value2"}
```

api/managementpb/actions.proto
```diff
@@ -0,0 +1,106 @@
...
+  string pmm_agent_id = 1;
+  // Action ID returned by pmm-agent.
+  string action_id = 2;
+}
+
+// CancelActionRequest describes request to cancel an action.
+message CancelActionRequest {
+  // PMM agent instance identifier.
+  string pmm_agent_id = 1;
+  // Action name.
```
BupycHuk: wrong comment

api/managementpb/actions.proto
```diff
@@ -0,0 +1,106 @@
+syntax = "proto3";
+
+package management;
```
BupycHuk: So it will be part of management API? I have been thinking that we decided it will be separate API package.

api/agentpb/agent.proto
AlekSi: By that logic, we can also push action_id and type into nameless strings:
```go
[]string{"Action_id","pt-summary","key1","value1","key2","value2"}
```
But that hardly a good API.

Let's use common parameters for all process actions and type-specific parameters for other action. Something like:
```proto
message ProcessActionParams {
  repeated string args = 1;
}

message StartActionRequest {
  string action_id = 1;
  ActionType type = 2;
  oneof params {
    ProcessActionParams process_action_params = 3;
    MySQLExplainActionParams …
    MySQLShowCreateTableParams …
  }
}
```

I will have more comments, but that's something we should start changing now.

api/managementpb/actions.proto
DexterHD: First I've made `actionspb` directory and actions service, then I realized that it will be something like `/v0/actions/Actions/RunAction` and it seems looks ugly isn't it? So I have decided to move it to the management API in the first iteration. I think it's ok, but I'm not sure so we can discuss it. Also, I don't really like our implementation of management API in `pmm-managed` where we have a lot of services in one package (in one `namespace` in fact). Maybe we should separate those services, or it should be one service implements all functions.

api/agentpb/agent.proto
DexterHD: I agree except the concept of `ProcessActionParams` it looks bad as we broke encapsulation and give the user some knowledge about internal action nature, as some actions, in fact, are processes. It's also a bad API I think.

api/managementpb/actions.proto
BupycHuk: yeah, doesn't look good.

api/agentpb/agent.proto
BupycHuk: Do you suggest to make a separate `params` message for each action type? even if they are the same?

DexterHD: > Do you suggest to make a separate `params` message for each action type? even if they are the same?

Yep, maybe. Also, we can use unique params for all actions, or some generalized structure for all actons. Maybe map for example.

In fact, they are no the same (I mean nevermind `pt-summary` and `pt-mysql-summary` is shell scripts. They have different parameters, they also have different meanings). They will the same only if you know that they are `shell scripts` but this breaks encapsulation at once.

If we made an abstraction, we should hide the implementation details. If we show implementation details, abstraction is useless.

BupycHuk: I agree that it should be unique params for all actions. So it can help us to not let users do whatever they want.

AlekSi: That's not a user-visible API. Arguments about encapsulation miss that point.
Let's also not redefine the scope of that task, and definition of action ATM.

> I agree that it should be unique params for all actions.

My only concern there that all process actions are not, in fact, different at that level (pmm-managed<->pmm-agent protocol). All process actions parameters will have the same structure. We will just add more types without benefits.

api/managementpb/actions.proto
```diff
@@ -0,0 +1,106 @@
...
+  // PMM agent instance identifier.
+  string pmm_agent_id = 1;
+  // Service ID for which action is running.
+  string service_id = 2;
+  // Node ID for which action is running.
+  string node_id = 3;
+  // Action name to run.
+  agent.ActionName action_name = 4;
+  // action parameters.
+  repeated string action_params = 5;
```
AlekSi: _There_ we should have type-specific parameters for all Actions, both built-in and processes.

api/agentpb/agent.proto
DexterHD: > Let's also not redefine the scope of that task, and definition of action ATM.

The scope of this task isn't clean, so this task is discussable I think.

> My only concern there that all process actions are not, in fact, different at that level (pmm-managed<->pmm-agent protocol)

So, if we know about internal actions structure at this level we shouldn't use actions at all here. In this case, this abstraction need only in `pmm-managed<->UI`? In this case, I think we should rethink an API between `pmm-managed<->pmm-agent`. As I said above we probably should have an API like `RunShellScript, RunSQLQuery, RunMongoDBCommand` in that level, why not?

DexterHD: One more thing. This API does not interfere to merge all this stuff as we haven't other actions like `EXPLAIN` and we have a clean separation in code between `pmm-managed<->UI` and `pmm-managed<->pmm-agent`. So we can merge this and then rethink and refactor `pmm-managed<->pmm-agent` side without doing changes in Public API.

api/managementpb/actions.proto
DexterHD: Let's define which parameters for each action do we need here?

api/agentpb/agent.proto
AlekSi: > Also when I say user, I always mean person/program who call this API.

I'm sorry, but you said "end-user", and that's not what it means.

> As I said above we probably should have an internal API like RunShellScript, RunSQLQuery, RunMongoDBCommand in that level, why not?

We can't due to our security model: we should not provide a way to run arbitrary commands or SQL queries.
Again, let's not redefine the scope. It was clear last time we talked about.

There is no way for `repeated string parameters` for all Actions to be merged. That's my call.

api/managementpb/actions.proto
AlekSi: Let's not. We have separate tickets for that.

api/agentpb/agent.proto
DexterHD: > I'm sorry, but you said "end-user", and that's not what it means.
Ok. I'll be more accurate next time and will separate user and `end-user`. 
In this case, I mean `person/program`.

> We can't due to our security model: we should not provide a way to run arbitrary commands or SQL queries.

We can if commands will be predefined, not arbitrary. We are talking about clean API and good separation not about having arbitrary commands and queries.

All I'm trying to say is that suggested separation on `ProcessActionParams` and `AnythingElseActionParams` looks awkward and also breaks encapsulation. 

In fact, all MySQL parameters can be represented as a `map`, or as a string slice, like in SQL libraries. Why shell commands are so different and special? Why we can group process actions` but can't group `MySQL query actions` for example? I think we could group it better.

> There is no way for repeated string parameters for all Actions to be merged. That's my call.

I agree.

api/managementpb/actions.proto
DexterHD: Give me the number, please? I have checked all `epic` and didn't find anything. Or what we should do with this? Should we change this API in the context of this task?

AlekSi: image

DexterHD: Ok, so at least at this moment, current params here is OK, as we haven't any description or requirements yet for other commands. So when we will implement those commands we will have ` type-specific parameters for all Actions, both built-in and processes.` here. Or I can make `pt-summary params` here as an example and as a foundation.

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

JNKPercona:
http://docs-test.cd.percona.com/PR-345/

api/managementpb/actions.proto
```diff
@@ -0,0 +1,152 @@
...
+  string action_id = 1;
+  // pmm-agent ID where to this Action was started.
+  string pmm_agent_id = 2;
+}
+
+message StartPTMySQLSummaryActionRequest {
+  // pmm-agent ID where to run this Action.
+  string pmm_agent_id = 1;
+  // Service ID for this Action.
+  string service_id = 2;
```
BupycHuk: this should be required param, isn't it?

BupycHuk: oh it should not be part of this PR

BupycHuk: did you implement this action?

DexterHD: No, I don't there is a stub. I just merged @AlekSi API as he wants in my branch, so this is not my API at all.

AlekSi: >this should be required param, isn't it?

Yes, it should be. Good catch.

PR #355, opened 2019-05-14
PMM-3978 Actions APIs cleanup.

AlekSi:


api/agentpb/agent.proto
```diff
@@ -81,46 +82,49 @@ message SetStateResponse {
...
-  PT_SUMMARY = 1;
-  PT_MYSQL_SUMMARY = 2;
-  MYSQL_EXPLAIN = 3;
-}
-
-// StartActionRequest represents request to run an action by name.
+// StartActionRequest is a ServerMessage asking pmm-agent to start action.
 message StartActionRequest {
-  string action_id = 1;
-  ActionName name = 2;
```
AlekSi: Name is a unique identifier for Nodes and Service, similar to ID, but human-readable and -assignable.
This is type.

api/managementpb/actions.proto
```diff
@@ -20,86 +20,132 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
 };
 option go_package = "managementpb";
 
-import "agentpb/agent.proto";
```
AlekSi: We should not leak internal APIs into public ones.

JNKPercona:
http://docs-test.cd.percona.com/PR-355/

api/managementpb/actions.proto
```diff
@@ -20,86 +20,132 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
...
 import "github.com/mwitkow/go-proto-validators/validator.proto";
 import "google/api/annotations.proto";
 import "protoc-gen-swagger/options/annotations.proto";
 
-// RunActionRequest describes request to run an action.
-message RunActionRequest {
```
AlekSi: "Run" is the whole interval from start to finish. We do not wait for the action to finish, so we call this API "start".

api/managementpb/actions.proto
```diff
@@ -20,86 +20,132 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
...
-  // PMM agent instance identifier.
-  string pmm_agent_id = 1;
-  // Service ID for which action is running.
-  string service_id = 2;
-  // Node ID for which action is running.
-  string node_id = 3;
-  // Action name to run.
-  agent.ActionName action_name = 4;
```
AlekSi: In all other public APIs for creating/changing something, we use type in the API name.

JNKPercona:
http://docs-test.cd.percona.com/PR-355/

api/agentpb/agent.proto
DexterHD: - Nodes and Services are placed in different contexts, so we shouldn't define actions the same way as for nodes and services. Every time we are talking about things we should talk about this in some context.
- I called this `ActionName` as in my mind `type` is something more generalized, like `shell-action, sql-action, no-sql-action, etc...` those thing looks like types, but the current is not IMHO.

If you think the word `Type` describes this thing more clearly, ok, it's up to you. I don't think so.

AlekSi: Sure. Can we not overload the words "type" and "name" in the context of PMM? Make "type" mean the same thing in Inventory and Management APIs.

PR #357, opened 2019-05-15
PMM-4044 Change ServerInfo struct.

BupycHuk:
Change ServerInfo struct to return clock drift instead of last ping time.


JNKPercona:
http://docs-test.cd.percona.com/PR-357/

PR #359, opened 2019-05-20
PMM-4046 Check connection to Service.

BupycHuk:
Add flag to skip connection check.


JNKPercona:
http://docs-test.cd.percona.com/PR-359/

api/agentpb/agent.proto
```diff
@@ -127,6 +128,22 @@ message ActionResultRequest {
...
+  // Service type.
+  inventory.ServiceType type = 1;
+  // DSN for the service.
+  string dsn = 2;
+}
+
+// CheckConnectionResponse is an AgentMessage containing a result of check request.
+message CheckConnectionResponse {
+  // Successfully connected to the service
+  bool success = 1;
```
AlekSi: @BupycHuk, btw, I think `error != ""` would be enough. We don't want an empty error message, and we don't have a case where we can't make one if connection check failed.

PR #362, opened 2019-05-22
PMM-3832 MySQL EXPLAIN action API.

DexterHD:


JNKPercona:
http://docs-test.cd.percona.com/PR-362/

api/managementpb/actions.proto
```diff
@@ -93,9 +93,10 @@ message StartMySQLExplainActionRequest {
   string pmm_agent_id = 1;
   // Service ID for this Action.
   string service_id = 2;
-  //
-  // TODO specific parameters if we need them.
-
+  // database for query.
+  string db = 3;
+  // query example.
+  string query = 4;
```
BupycHuk: I think these params aren't enough. 
There are no credentials (username and password) to prepare dsn.

BupycHuk: And in StartMySQLJsonExplainActionRequest as well

DexterHD: Do you mean we can't get credentials from service and agent?

DexterHD: I'm not sure about `db` parameter on UI but it explained in old implementation:
> // If the query has a default db, use it; else, all tables need to be db-qualified
> // or EXPLAIN will throw an error.

BupycHuk: I mean there is no agent_id param to get credentials from mysqld_exporter and no username and password params to use them. 
are you going to use some logic inside of pmm-managed to get credentials from mysqld_exporter for current service? like `findAgentForService(MYSQLD_EXPORTER, req.ServiceID)`?

DexterHD: > are you going to use some logic inside of pmm-managed to get credentials from mysqld_exporter for current service? like findAgentForService(MYSQLD_EXPORTER, req.ServiceID)?
It seems to. We haven't requirements and didn't discuss this. 

@AlekSi @rnovikovP Need your help with that. What should I do?

AlekSi: Before GA we should add default username/password to Service itself. They should be used by Agents and Actions when explicit username/password are not given. It is currently tracked there: https://jira.percona.com/browse/PMM-3786, but will be extracted into separate one a bit later.

In the meantime, let's do `findAgentForService(MYSQLD_EXPORTER, req.ServiceID)` with a comment indicating that is a temporary hack.

api/agentpb/agent.proto
```diff
@@ -91,8 +98,10 @@ message StartActionRequest {
   }
   // MySQLExplainParams describes MySQL Explain action parameters.
   message MySQLExplainParams {
-    // TODO
-
+    string dsn = 1;
+    string db = 2;
```
AlekSi: Isn't database name a part of DSN?

api/managementpb/actions.proto
```diff
@@ -105,6 +106,24 @@ message StartMySQLExplainActionResponse {
   string pmm_agent_id = 2;
 }
 
+message StartMySQLJsonExplainActionRequest {
```
AlekSi: StartMySQLExplainJSONActionRequest. Common prefix "StartMySQLExplain" for grouping, JSON for acronym.

api/agentpb/agent.proto
DexterHD: Yes, but I'm not sure it enough:

https://github.com/percona/qan-agent/blob/9399e1fe46df02e988838816c0578282918b0b8c/query/plugin/mysql/explain/explain.go#L62-L69

AlekSi: We will extract it from DSN.
If we have two options to pass database name, we will have 4 combinations we will need to handle. I don't think we should.

JNKPercona:
http://docs-test.cd.percona.com/PR-362/

JNKPercona:
http://docs-test.cd.percona.com/PR-362/

JNKPercona:
http://docs-test.cd.percona.com/PR-362/

JNKPercona:
http://docs-test.cd.percona.com/PR-362/

PR #380, opened 2019-06-12
PMM-4013 Add request validators.

AlekSi:
In some cases, QAN UI doesn't know database or table name. That results in awful errors like:
```
Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 1
```
That PR adds automatic requests validators to make them somewhat less awful. We still need to add more checks with code, and also fully understand why UI doesn't know things it should, but that is the first step.


JNKPercona:
http://docs-test.cd.percona.com/PR-380/

PR #381, opened 2019-06-14
PMM-4185 Remove validation for database field.

AlekSi:
MySQL supports both `EXPLAIN query` and `EXPLAIN database.query` formats. We want to send them as is from pmm-managed to pmm-agent, not try to parse table_name into table name and database name. So just remove validator and expand descriptions.


JNKPercona:
http://docs-test.cd.percona.com/PR-381/

PR #426, opened 2019-09-09
PMM-4648 Add flush slow logs.

askomorokhov:
https://jira.percona.com/browse/PMM-4437

depends on:
https://github.com/percona/pmm/pull/426
https://github.com/percona/pmm-agent/pull/116
https://github.com/percona/pmm-admin/pull/59
https://github.com/percona/pmm-managed/pull/264

FB:
https://github.com/Percona-Lab/pmm-submodules/pull/444

usage ex.:
`pmm-admin add mysql --username=root --password=secret ps:3306 MySQLSlowLog --size-slow-logs=10MB`

PR #116, opened 2018-01-16
Added Jenkinsfile

delgod:

PR #429, opened 2019-09-11
 PMM-4437 Unify field name to be MaxSlowlogFileSize

askomorokhov:
https://jira.percona.com/browse/PMM-4459
https://jira.percona.com/browse/PMM-4437

depends on:
percona/pmm#426
#116
percona/pmm-admin#59
percona/pmm-managed#264
https://github.com/percona/qan-app/pull/322

FB:
Percona-Lab/pmm-submodules#444


api/agentpb/agent.proto
```diff
@@ -72,7 +72,7 @@ message SetStateRequest {
     // Disables query examples for QAN Agents if true.
     bool disable_query_examples = 3;
     // Instructs QAN Agents to rotate query log file or table at this size if > 0.
-    int64 max_query_log_size = 4;
+    int64 max_slowlog_file_size = 4;
```
AlekSi: That field will be used for other logs sources like MySQL slowlog table (a number of rows) or PostgreSQL query log (it doesn't call it "slowlog"). So a different name there is intensional.

AlekSi: Note that comment above now also mentions "table"

JNKPercona:
http://docs-test.cd.percona.com/PR-429/

PR #452, opened 2019-11-05
PMM-4926 Add Stats for connection checker.

AlekSi:


JNKPercona:
http://docs-test.cd.percona.com/PR-452/

api/agentpb/agent.proto
```diff
@@ -160,8 +160,14 @@ message CheckConnectionRequest {
 
 // CheckConnectionResponse is an AgentMessage containing a result of connection check.
 message CheckConnectionResponse {
+  // Stats contains various Service statistics.
+  message Stats {
+    // A number of tables. 0 means "unknown" (for compatibility), -1 means "there are really 0 tables".
```
BupycHuk: let's use extra boolean field instead of using -1

AlekSi: Actually, it is not even possible: `SELECT COUNT(*) FROM information_schema.tables` always returns non-zero value – `information_schema.tables` is a table itself. 😆

JNKPercona:
http://docs-test.cd.percona.com/PR-452/

JNKPercona:
http://docs-test.cd.percona.com/PR-452/

PR #454, opened 2019-11-06
PMM-4317 Upload SSH key.

BupycHuk:


JNKPercona:
http://docs-test.cd.percona.com/PR-454/

api/serverpb/server.proto
```diff
@@ -137,6 +139,16 @@ message ChangeSettingsResponse {
   Settings settings = 1;
 }
 
+message UploadSSHKeyRequest {
```
askomorokhov: From my point of view `AddSSHKeyRequest` is a better fit as a name for this message.
The prefix "Upload" confusing a bit here.
An "Upload" gives some connotation that request has to have binary body and filename.
But we have here only string with a body of the key.

@BupycHuk, what do you think?

BupycHuk: The think about add is that we don't add ssh key, we replace ssh key. Maybe it's better to call it SetSSHKeyRequest.

askomorokhov: @BupycHuk, yes, it sounds even better.

JNKPercona:
http://docs-test.cd.percona.com/PR-454/

api/serverpb/server.proto
```diff
@@ -137,6 +139,16 @@ message ChangeSettingsResponse {
   Settings settings = 1;
 }
 
+message SetSSHKeyRequest {
+  string ssh_key = 1 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+}
```
AlekSi: Why this is not a part of `ChangeSettingsRequest`?

BupycHuk: Because SSH key replaces all current keys each time we call it. If user adds new ssh key manually it will rewrite file.

JNKPercona:
http://docs-test.cd.percona.com/PR-454/

api/serverpb/server.proto
BupycHuk: We have another option to check equality of current ssh key and saved ssh key. But in this case if user manually changed ssh keys file we won't be able to restore previously uploaded ssh key.

AlekSi: If `ssh_key` is empty, do not replace it. That's what what we do for all settings.
Otherwise, there is no symmetry between setting a value (separate method) and getting a value (common method).

AlekSi: > in this case if user manually changed ssh keys file we won't be able to restore previously uploaded ssh key

Why we don't read it from file and return in GetSettings?

BupycHuk: but how we will handle it on frontend side? we just moving this problem from backend side to frontend side.
We should show current key on frontend side.

BupycHuk: > Why we don't read it from file and return in GetSettings?
That's a good option.

BupycHuk: About why did we separated method. I've never seen keeping ssh keys as part of other settings. Usually it's separated.

AlekSi: But we do return it as a part of Settings.

JNKPercona:
http://docs-test.cd.percona.com/PR-454/

JNKPercona:
http://docs-test.cd.percona.com/PR-454/

JNKPercona:
http://docs-test.cd.percona.com/PR-454/

JNKPercona:
http://docs-test.cd.percona.com/PR-454/

PR #464, opened 2019-11-21
PMM-4966 Filter sensitive information.

BupycHuk:


JNKPercona:
http://docs-test.cd.percona.com/PR-464/

JNKPercona:
http://docs-test.cd.percona.com/PR-464/

api/agentpb/agent.proto
```diff
@@ -47,6 +47,7 @@ message SetStateRequest {
     repeated string args = 4;
     repeated string env = 5;
     map<string, string> text_files = 6;
+    repeated string hide_keywords = 7;
```
AlekSi: We use several synonyms for the same thing (there and in implementation). Let's use the same term. For example, "redact":

> Redaction in its sanitization sense (as distinguished from its other editing sense) is the blacking out or deletion of text in a document, or the result of such an effort. It is intended to allow the selective disclosure of information in a document while keeping other parts of the document secret.

(from https://en.wikipedia.org/wiki/Sanitization_(classified_information))

BupycHuk: ok

JNKPercona:
http://docs-test.cd.percona.com/PR-464/

PR #501, opened 2020-02-19
Update API descriptor.

AlekSi:


AlekSi:
To make sure our changes in API are backward compatible. Try to remove a field, for example.
I removed version because there is no point of having it – we use git tags now.

AlekSi:
@BupycHuk please check and merge

PR #509, opened 2020-04-07
PMM-5559 Add annotation.

askomorokhov:
https://jira.percona.com/browse/PMM-5559
https://jira.percona.com/browse/PMM-3812

FB: https://github.com/Percona-Lab/pmm-submodules/pull/785


api/managementpb/annotation.proto
```diff
@@ -0,0 +1,37 @@
...
+      string_not_empty: true
+    }
+  ];
+  // Tags are used to filter annotations.
+  repeated string tags = 2;
+}
+
+message AddAnnotationResponse {
+  // Status message.
+  string message = 1;
```
AlekSi, line 25: What is that and how it is used?

In other APIs, we either return error or useful data.

askomorokhov: It is a message forwarded from grafana, used to display that the annotation was added.

AlekSi: Ok, but why we need it? For example, when you run `pmm-admin add mysql`, we generate "Service added" message in pmm-admin, not return it from API.

askomorokhov: Here API plays a bit different role being a proxy to third party app - Grafana. We can earlier detect any changes/issues in communication between pmm-admin and Grafana by not hide the real messages from a user.

AlekSi: Our other APIs are proxies too – for Prometheus, for example.

It doesn't look like we can add any meaningful meaning to this field. Let's remove it; we can always add it later if we need it, but we can't remove it once it is there.

askomorokhov: @AlekSi, one more question.
Why you are considering that the response with a message less meaningful than the empty response? Receiving in Swagger empty response with status code OK, when expecting Accepted looks not obvious and message helps.
So why an empty response is preferable here? Just to fit other implementations?

AlekSi: Because in our API responses we return data or errors. We don't return success messages anywhere – this is UI problem, not API problem. UI (web our CLI) generates that message based on returned data – and there we don't have any data.

> Receiving in Swagger empty response with status code OK, when expecting Accepted

Why do you expect Accepted? That's not a REST API, that's RPC-style API where any non-zero response code is an error. The fact that we map it to HTTP status code 200 is irrelevant.

askomorokhov: Sounds OK. Removed.

PR #515, opened 2020-04-24
PMM-5819 settings for stt enable

percona-csalguero:
Added fields to control STT enable/disable


api/serverpb/server.proto
```diff
@@ -117,6 +117,8 @@ message Settings {
   // Prometheus AlertManager URL (e.g., https://username:password@1.2.3.4/path).
   string alert_manager_url = 7;
   string alert_manager_rules = 8;
+  // Enable Security Threat Tool
+  bool enable_stt = 9;
```
BupycHuk, line 121:
```suggestion
  bool stt_enabled = 9;
```

api/serverpb/server.proto
```diff
@@ -117,6 +117,8 @@ message Settings {
   // Prometheus AlertManager URL (e.g., https://username:password@1.2.3.4/path).
   string alert_manager_url = 7;
   string alert_manager_rules = 8;
+  // Enable Security Threat Tool
```
BupycHuk, line 120: please update comment

api/serverpb/server.proto
```diff
@@ -138,6 +140,9 @@ message ChangeSettingsRequest {
   bool remove_alert_manager_url = 8;
   string alert_manager_rules = 9;
   bool remove_alert_manager_rules = 10;
+  // Enable/disable Security Threat Tool
```
AlekSi, line 143: This comment is added to Swagger documentation only for this field (you can see in generated serverpb.json above and in UI), and there is not way to disable STT with it. So let's add a proper comment to both fields.

PR #516, opened 2020-04-24
PMM-5777 Inventory add external exporter endpoint.

BupycHuk:


api/inventorypb/agents.proto
```diff
@@ -293,7 +293,7 @@ message QANPostgreSQLPgStatementsAgent {
   AgentStatus status = 9;
 }
 
-// RDSExporter runs on Generic or Container Node and exposes RemoteRDS Node metrics.
+// RDSExporter runs on RemoteRDS Node and exposes RemoteRDS Node metrics.
```
AlekSi, line 296: No, it really runs on Generic or Container node – we can't run processes on remote nodes.

api/inventorypb/agents.proto
```diff
@@ -325,6 +325,28 @@ message RDSExporter {
   bool enhanced_metrics_disabled = 10;
 }
 
+// ExternalExporter runs on any Node.
```
AlekSi, line 328: Not on any – it can't run on remote nodes. So – Generic or Container.

api/inventorypb/agents.proto
```diff
@@ -325,6 +325,28 @@ message RDSExporter {
...
+message ExternalExporter {
+  // Unique randomly generated instance identifier.
+  string agent_id = 1;
+  // Node identifier where this instance runs.
+  string runs_on_node_id = 2;
+  // Desired Agent status: enabled (false) or disabled (true).
+  bool disabled = 3;
+  // Service identifier.
+  string service_id = 4;
+  // ProxySQL username for scraping metrics.
```
AlekSi, line 338: huh?

api/inventorypb/agents.proto
```diff
@@ -893,6 +917,57 @@ message ChangeRDSExporterResponse {
...
+  string metric_path = 7 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  // Listen port for scraping metrics.
+  uint32 listen_port = 8 [
+    (validator.field) = {
+      int_gt: 0
+      int_lt: 65535
```
AlekSi, line 947: 65536

api/inventorypb/agents.proto
```diff
@@ -893,6 +917,57 @@ message ChangeRDSExporterResponse {
...
+  ];
+  // Service identifier.
+  string service_id = 2;
+  // HTTP basic auth username for collecting metrics.
+  string username = 3;
+  // HTTP basic auth password for collecting metrics.
+  string password = 4;
+  // Scheme to generate URI to exporter metrics endpoints.
+  bool scheme = 6;
+  // Disable basic metrics.
```
AlekSi, line 937: huh?

api/inventorypb/agents.proto
BupycHuk: Actually it can, based on next task related to external exporters it will be able to work on remote task too.

AlekSi: What is the definition of remote node?

api/inventorypb/agents.proto
BupycHuk: I set this based on readme https://github.com/mwitkow/go-proto-validators, but I checked generated code and you are right

api/inventorypb/agents.proto
BupycHuk: Some node to which we don't have access.

api/inventorypb/agents.proto
```diff
@@ -893,6 +917,57 @@ message ChangeRDSExporterResponse {
   RDSExporter rds_exporter = 1;
 }
 
+// Add/Change ExternalExporter
+
```
askomorokhov, line 921: Rudimental blank line?

artemgavrilov: And missing dot at the end

BupycHuk: @askomorokhov We need this blank line to not include this to swagger json.
@artemgavrilov We don't have dot at the end of other comments of this type.

artemgavrilov: Um, what type of that comment? Is there some document that describes how properly write such comments(and other types)? Because it's not obvious and I can accidentally refactor that code fragment.

BupycHuk: comments like `Add/Change`, `List`, `Get` in this file which was made just to separate different models related to different endpoints.
It's not documented anywhere.

api/inventorypb/agents.proto
```diff
@@ -325,6 +325,28 @@ message RDSExporter {
...
 }
 
+// ExternalExporter runs on any Node.
+message ExternalExporter {
+  // Unique randomly generated instance identifier.
+  string agent_id = 1;
+  // Node identifier where this instance runs.
+  string runs_on_node_id = 2;
+  // Desired Agent status: enabled (false) or disabled (true).
+  bool disabled = 3;
```
AlekSi, line 335: What does it mean for external exporter?

api/inventorypb/agents.proto
AlekSi: We can't easily split this file into several due to protobuf importers, etc. Those comments just separate different sections of this huge file.

api/inventorypb/agents.proto
```diff
@@ -893,6 +917,53 @@ message ChangeRDSExporterResponse {
   RDSExporter rds_exporter = 1;
 }
 
+// Add/Change ExternalExporter
+
+message AddExternalExporterRequest {
+  // The node identifier where this instance is run.
+  string runs_on_node_id = 1 [
+    (validator.field) = {
+      string_not_empty: true
```
AlekSi, line 926: Our internal docs still say that zero Nodes is possible.

api/inventorypb/agents.proto
BupycHuk: Will we scrape external exporter or not. Good point, I ignored it in pmm-managed.

api/inventorypb/agents.proto
BupycHuk: Updated. Will update other parts of documentation after implementing external exporters. External exporters have higher priority then updating documentation now.

api/inventorypb/agents.proto
AlekSi: For other exporters, it means that it will not be started by pmm-agent. That's why it is called desired – it may be running now, but pmm-agent will stop it soon.

I think we should remove the whole field for external exporters.

api/inventorypb/agents.proto
AlekSi: It is the same priority – we don't have a separate ticket for updating documentation just like we don't have separate tickets for writing unit tests. Existing page in the Confluence _must_ be in sync with code.

The easiest way to ensure that is to remove everything from the Confluence that can be seen in proto files. For example, a list of non-id fields could be removed. But Confluence documentation _must_ be updated to reflect the changes in relations between types: that Node is required even for external exporters, that we have external services, etc.

api/inventorypb/agents.proto
BupycHuk: What if user would like to pause collecting metrics for some time? We let him to do it for other exporters. I can fix comment to something like `Desired metrics collection status`

AlekSi: ok, let's keep the name, but make a comment/documentation distinctively different

api/inventorypb/agents.proto
BupycHuk: Updated documentation related to external exporters

api/inventorypb/agents.proto
AlekSi: @BupycHuk ^^^

BupycHuk: Updated

api/inventorypb/agents.proto
```diff
@@ -1070,6 +1141,20 @@ service Agents {
       body: "*"
     };
   }
+  // AddExternalExporter adds rds_exporter Agent.
```
AlekSi, line 1144: rds_exporter?

PR #518, opened 2020-04-29
Remove old comments.

AlekSi:


api/inventorypb/agents.proto
```diff
@@ -341,16 +341,12 @@ message ChangeCommonAgentParams {
 
 message ListAgentsRequest {
   // Return only Agents started by this pmm-agent.
-  // Exactly one of these parameters should be present: pmm_agent_id, node_id, service_id.
   string pmm_agent_id = 1;
   // Return only Agents that provide insights for that Node.
-  // Exactly one of these parameters should be present: pmm_agent_id, node_id, service_id.
   string node_id = 2;
   // Return only Agents that provide insights for that Service.
-  // Exactly one of these parameters should be present: pmm_agent_id, node_id, service_id.
```
BupycHuk, line 350: We haven't removed validation for that yet

AlekSi: Then should we? It seems it exist only in API and only for Agents. Do we have a ticket for that.

PR #524, opened 2020-05-01
SAAS-30 Make exposed API less awkward.

AlekSi:


artemgavrilov:
Why `int64` is awkward?

AlekSi:
`int64` is a bit awkward in public API because its usage often requires conversions.

But that removed comment was very awkward – how Starlark is related to this package?! And there is no real need for that – https://github.com/percona/pmm-managed/pull/372/commits/9ac915abd63af6a2bf5ca8e28508678fa0d916fa

PR #529, opened 2020-05-14
SAAS-67  STT API

artemgavrilov:
- https://jira.percona.com/browse/SAAS-67

A simple method to trigger STT checks execution, no parameters, no response.


api/serverpb/server.proto
```diff
@@ -197,6 +201,13 @@ service Server {
       body: "*"
     };
   }
+  // StartSecurityChecks start Security Thread Tool checks.
+  rpc StartSecurityChecks(StartSecurityChecksRequest) returns (StartSecurityChecksResponse) {
+    option (google.api.http) = {
+      post: "/v1/SecurityChecks/Start"
```
AlekSi, line 207: So here is a thing. PMM has three sets of stable public APIs:
* inventory API for low-level inventory data manipulation;
* management API for high-level inventory data manipulation and future management capabilities;
* server API for server settings, updated, etc.

(QAN API is not public/stable yet, and API for pmm-agent connectivity is explicitly private)

Inventory APIs start from `/v1/inventory/`, management – `/v1/management/`, and server APIs start from `/v1/` for somewhat unfortunate backward-compatibility reasons. We then map those APIs to three services in pmm-managed: inventory, management, and server.

That PR introduces something new: server API that is mapped to a different pmm-managed service. Is it okay? Should we do something about it now? Not now, but soon? And what should be done once we add DBaaS APIs?

Thoughts?

artemgavrilov: Hm, as I remember there will be other STT related methods (schedule setup as a minimum) so probably it makes sense to make separated API. Especially if it's a common approach in that service.    
How about  `/v1/stt/` for example?

BupycHuk: From my side this one looks close to management API like actions. I think DBaaS API endpoints also will be in Management API part since a lot of calls in DBaaS will be about management.

AlekSi: Ok, let's switch it to `/v1/management/SecurityChecks/Start`.
But what we will do about pmm-managed?

AlekSi: Ah, yes, and we should move that API out of serverpb/server.proto to managementpb/

artemgavrilov: Done

PR #536, opened 2020-06-08
PMM-5563 Annotation service, node.

JiriCtvrtka:


BupycHuk:
do not update api descriptor in feature PRs.

api/inventorypb/nodes.proto
```diff
@@ -283,6 +283,16 @@ message RemoveNodeRequest {
 
 message RemoveNodeResponse {}
 
+// Check
```
BupycHuk, line 286: We don't need check node on pmm-admin side, it should be done on pmm-managed side while adding annotation.

PR #538, opened 2020-06-11
Fix grammar, add deprecation comment.

AlekSi:


AlekSi:
So much for assigning PR…

PR #541, opened 2020-06-25
PMM-5800 QAN explain.

JiriCtvrtka:


api/qanpb/object_details.proto
```diff
@@ -103,6 +103,7 @@ message QueryExample {
   string service_type = 7;
   string schema = 8;
   repeated string tables = 9;
+  string agent_id = 10;
```
BupycHuk, line 106: it's a good idea, but it's not part of this task

PR #549, opened 2020-07-06
PMM-6130 Basic DBaaS APIs

AlekSi:
https://github.com/percona/pmm-managed/pull/440
https://github.com/percona/pmm/pull/549
https://github.com/Percona-Lab/pmm-submodules/pull/1048
https://github.com/Percona-Lab/pmm-api-tests/pull/121


api/managementpb/dbaas/common.proto
```diff
@@ -0,0 +1,15 @@
...
+option go_package = "dbaasv1beta1";
+
+// TODO move it to parent directory / package once API is v1-stable
+
+// RunningOperation respresents a long-running operation.
+message RunningOperation {
+  // Progress from 0.0 to 1.0; can decrease compared to the previous value.
+  float progress = 1;
+  // Text describing the current operation progress step.
+  string message = 2;
```
askomorokhov, line 14: Just wondering whether timestamp will be useful here. As an example: for logging, delaying progress, displaying time diff, etc.

AlekSi: There is no intention to use that API for logs – it is a current status indicator

api/managementpb/dbaas/kubernetes.proto
```diff
@@ -0,0 +1,86 @@
...
+      msg_exists: true
+    }
+  ];
+}
+
+message RegisterKubernetesClusterResponse {}
+
+message UnregisterKubernetesClusterRequest {
+  // Kubernetes cluster name.
+  string kubernetes_cluster_name = 1 [
```
BupycHuk, line 54: @AlekSi why do we use kubernetes_cluster_name instead of kubernetes_cluster_id like in other places where we remove things?

AlekSi: It is k8s resource name, not PMM Inventory object

api/managementpb/dbaas/kubernetes.proto
```diff
@@ -19,11 +19,17 @@ message KubeAuth {
   ];
 }
 
+// KubernetesCluster contains public info about kubernetes cluster.
+message KubernetesCluster {
```
AlekSi, line 23: Then we need to use it in RegisterKubernetesClusterRequest too

BupycHuk: In RegisterKubernetesClusterRequest might be more fields, not only public info. We used different messages in other endpoints.

PR #553, opened 2020-07-13
SAAS-175 Auth API

artemgavrilov:
- https://jira.percona.com/browse/SAAS-175

Implemented in: https://github.com/percona/pmm-managed/pull/420
Feature build: Percona-Lab/pmm-submodules#967


AlekSi:
image

I really hope CI will catch that…

api/serverpb/server.proto
```diff
@@ -218,4 +256,25 @@ service Server {
...
+  rpc SignIn(SignInRequest) returns (SignInResponse) {
+    option (google.api.http) = {
+      post: "/v1/Auth/SignIn"
+      body: "*"
+    };
+  }
+  // RefreshSession refreshes session timeout.
+  rpc RefreshSession(RefreshSessionRequest) returns (RefreshSessionResponse) {
+    option (google.api.http) = {
+      post: "/v1/Auth/RefreshSession"
```
AlekSi, line 276: There shouldn't be a need to expose that API – once user signed in, pmm-managed should refresh session in the background.

AlekSi:
And yes, CI fails. Run `make init gen`

api/serverpb/server.proto
artemgavrilov: Removed

artemgavrilov:
@AlekSi sorry, fixed.

api/serverpb/server.proto
```diff
@@ -218,4 +252,18 @@ service Server {
       body: "*"
     };
   }
+  // SignUp creates user Percona profile.
+  rpc SignUp(SignUpRequest) returns (SignUpResponse) {
+    option (google.api.http) = {
+      post: "/v1/Auth/SignUp"
```
AlekSi, line 258: Let me think about that URL…

AlekSi:
> We need to add SaaS user email to Settings message

I will do that

artemgavrilov:
Ok

api/serverpb/server.proto
```diff
@@ -161,6 +163,32 @@ message AWSInstanceCheckRequest {
...
 
+message PlatformSignUpRequest {
+  // New Percona Platform user's email.
+  string email = 1 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  // New Percona Platform user's password.
+  string password = 2;
```
AlekSi, line 174: The presence of that field should be validated in code. In the future, we will relax that requirement. Removing a validator would be a breaking change from prototool perspective.

api/serverpb/server.proto
```diff
@@ -161,6 +163,32 @@ message AWSInstanceCheckRequest {
...
+
+message PlatformSignInRequest {
+  // Existing Percona Platform user's email.
+  string email = 1 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  // Existing Percona Platform user's password.
+  string password = 2;
```
AlekSi, line 187: The presence of that field should be validated in code. In the future, we will relax that requirement. Removing a validator would be a breaking change from prototool perspective.

api/serverpb/server.proto
```diff
@@ -218,4 +246,18 @@ service Server {
...
+      body: "*"
+    };
+  }
+  // PlatformSignIn links that PMM instance to Percona Platform user.
+  rpc PlatformSignIn(PlatformSignInRequest) returns (PlatformSignInResponse) {
+    option (google.api.http) = {
+      post: "/v1/Platform/SignIn"
+      body: "*"
+    };
+  }
```
AlekSi, line 262: Changed both URLs and method names

api/nginx/nginx.conf
```diff
@@ -8,88 +11,13 @@ events {
 }
 
 http {
-  upstream managed-grpc {
```
BupycHuk, line 11: Where it was moved?

AlekSi: See the added comment on top of the file

PR #569, opened 2020-08-24
PMM-6130 API for Kubernetes clusters.

BupycHuk:


AlekSi:
Can we have a better PR name, please?

BupycHuk:
@AlekSi done

api/managementpb/dbaas/kubernetes.proto
```diff
@@ -0,0 +1,86 @@
...
+  // Kubeconfig file content.
+  string kubeconfig = 1 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+}
+
+// KubernetesCluster contains public info about kubernetes cluster.
+message KubernetesCluster {
```
AlekSi, line 23: Let's move that inside `ListKubernetesClustersResponse` and rename to `Cluster`

AlekSi (changes requested):

api/managementpb/dbaas/kubernetes.proto
AlekSi: > and rename to Cluster

@BupycHuk ^^^

AlekSi (approved):

api/Makefile
```diff
@@ -46,7 +46,8 @@ gen: clean            ## Generate files.
 		swagger/header.json \
 		serverpb/json/serverpb.json \
 		inventorypb/json/inventorypb.json \
-		managementpb/json/managementpb.json
+		managementpb/json/managementpb.json \
+		managementpb/dbaas/json/dbaas.json
```
AlekSi, line 50: No, that line should not be there – the comment above methods that

AlekSi (changes requested):

api/Makefile
AlekSi: 9fb795e5678bb5edb3a4c0d126592563c2a9edac

BupycHuk: Why dbaas is private?

AlekSi: > the comment above mentions that

That comments reads (in my branch):

> 	# generate public API spec, omit agentlocalpb (always private),
>	# and backups, dbaas, and qanpb (not v1 yet)

BupycHuk: I see that it's mentioned, but why?

AlekSi: Why it is not v1 stable yet? Because it is not stable yet – we reserve the right to make breaking changes. Think as version 0 in SemVer

AlekSi (approved):

PR #572, opened 2020-08-28
PMM-6460 Add pg stat agent type.

JiriCtvrtka:


AlekSi:
There seems to be something wrong with the workflow – @JiriCtvrtka keeps updating api/api.descriptor file. Why? Why do you run this Makefile target?

JiriCtvrtka:
@AlekSi Without that I getting error for make gen. I wanted discard it then. So please what I am doing wrong?

AlekSi:
Why this file exist, what it does, and what error you get?

JiriCtvrtka:
It contains info from proto files. 

For example when I change number for AgentStatus I getting:
Message field "10" on message "inventory.QANPostgreSQLPgStatMonitorAgent" was deleted

AlekSi:
You answered the question "what is it?", bit not "what it does?" and "why we need it?"

JiriCtvrtka:
Propably for python files?

AlekSi:
So you update that file because it stands in your way without knowing what it does. Got it. But why you don't do the same with the Go compiler, for example? If some code does not compile, you can disable compilation instead of fixing the problem.

Please read [this](https://developers.google.com/protocol-buffers/docs/proto#updating), [this](https://github.com/uber/prototool#prototool-break-check), [this](https://github.com/uber/prototool#prototool-descriptor-set), and [this](https://github.com/uber/prototool/blob/dev/docs/breaking.md).

api/inventorypb/agents.proto
```diff
@@ -21,8 +21,9 @@ enum AgentType {
   QAN_MYSQL_SLOWLOG_AGENT = 8;
   QAN_MONGODB_PROFILER_AGENT = 9;
   QAN_POSTGRESQL_PGSTATEMENTS_AGENT = 10;
-  RDS_EXPORTER = 11;
-  EXTERNAL_EXPORTER = 12;
+  QAN_POSTGRESQL_PGSTATMONITOR_AGENT = 11;
```
BupycHuk, line 24: It'll break backward compatibility, that's why make gen is failing.

BupycHuk (changes requested):
You are making to much things out of scope of your task.

JiriCtvrtka:
https://github.com/percona/pmm/pull/577

PR #585, opened 2020-09-24
PMM-6576 PMM-6972 Integrated Alerting API

AlekSi:


api/managementpb/ia/rules.proto
```diff
@@ -0,0 +1,123 @@
...
+  }
+  // Rule name.
+  string name = 1 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  // Sets new rule status: enabled or disabled. Must always be sent.
+  // FIXME ?
+  bool enabled = 2;
```
AlekSi, line 92: @BupycHuk WDYT?

BupycHuk: it's a little bit different from what we have now and it's better to discuss with a team. With frontend devs as well. I understand that we should change our logic and don't pass two fields instead of one, but not sure how we will handle two different approaches and keep backward compatibility in the same time.

AlekSi:
https://github.com/percona/pmm/pull/617

PR #587, opened 2020-10-05
PMM-6629 Protobuf 1.4

askomorokhov:
https://jira.percona.com/browse/PMM-6629

Build: https://github.com/Percona-Lab/pmm-submodules/pull/1155

1. https://github.com/percona/pmm/pull/587
2. https://github.com/percona/pmm-managed/pull/476
3. https://github.com/percona/pmm-agent/pull/190
4. https://github.com/Percona-Lab/pmm-api-tests/pull/129
5. https://github.com/percona/pmm-admin/pull/119
6. https://github.com/percona/qan-api2/pull/101


BupycHuk (approved):
Looks good to me, but are we sure that we should merge it right before release?

Gopkg.toml
```diff
@@ -41,13 +41,13 @@ required = [
   version = "=0.9.1"
 [[constraint]]
   name = "github.com/golang/protobuf"
```
AlekSi, line 43: Let's add "google.golang.org/protobuf" to this list of key deps

AlekSi:
> Looks good to me, but are we sure that we should merge it right before release?

I'm afraid, yes – otherwise, SaaS and PMM code will use different versions, and debugging incompatibilities between them will be much harder.

And that problem was known for a month.

Gopkg.toml
askomorokhov: Added.

AlekSi (approved):

PR #593, opened 2020-10-13
PMM-6766 Revert "Remove qan-app from list"

AlekSi:
Reverts percona/pmm#574

We should keep it in the list until we _really_ remove it everywhere.


BupycHuk:
What is the impact for user(contributor)? Do we really use qan-app in our build process?

AlekSi:
> Do we really use qan-app in our build process?

Yes, that's why we need to keep it for now.

That repository is archived now, but it is still referenced.

BupycHuk (approved):

atymchuk (approved):

atymchuk:
Should we ignore Travis build error?

PR #600, opened 2020-10-22
PMM-4172 pt-mysql-summary

askomorokhov:


api/agentpb/agent.proto
```diff
@@ -167,6 +167,12 @@ message StartActionRequest {
   // PTSummaryParams describes parameters for PT summary.
   message PTSummaryParams {}
 
+  // PTMySQLSummaryParams describes parameters for PT MySQL summary.
+  message PTMySQLSummaryParams {
+    string address = 1;
+    string username = 2;
+    string password = 3;
```
AlekSi, line 174: Why we need address, username, and password instead of DSN as other actions above?

AlekSi (changes requested):
Ci fails

api/agentpb/agent.proto
yareach: The pt-mysql-summary execution requires those params.

AlekSi: Okay, but can we extract them from DSN?

yareach: It takes them from DSN.

AlekSi: Who is "it"?
pt-mysql-summary does accept address – it accepts host and port, and we have to parse address anyway. So let's pass  dsn for consistency with other actions above and for socket support.

percona-csalguero (approved):

api/agentpb/agent.proto
BupycHuk: @AlekSi, For agents we decide on pmm-managed side what flags should be passed to processes, why do we let pmm-agent to decide how to run process? We can send list of flags like for exporters or we can send address and port separately since we store them separately in our db. So what's the reason to generate dsn and parse it later? Do we expect that pt-mysql-summary will accept dsn with additional query parameters in future?

AlekSi: > what flags should be passed to processes

For Agents, not for Actions. We don't send CLI flags with dashes there, we send parameters.

I think we should pass DNS because we use DSN for all other Actions there. In the current implementation, we have to do parameters parsing anyway (split address into host and port), and socket support is missing.

BupycHuk: I don't suggest to use current implementation. I'm suggesting to use separate fields for socket, host and port to not overcomplicate code on pmm-agent side.
The only reason why we should use dsn is possible backward compatibility in future if we decide to add some query parameters, are we going to implement that later?

AlekSi: @yareach Discussed that with @BupycHuk. Ok, let's pass separate fields for address (what pt-mysql-summary calls host), port, UNIX socket (that was missing, AFAIU), username and password

percona-csalguero (approved):

api/agentpb/agent.proto
```diff
@@ -167,6 +167,14 @@ message StartActionRequest {
   // PTSummaryParams describes parameters for PT summary.
   message PTSummaryParams {}
 
+  // PTMySQLSummaryParams describes parameters for PT MySQL summary.
+  message PTMySQLSummaryParams {
+    string addr = 1;
```
BupycHuk, line 172: why are did you cut field name?

BupycHuk (changes requested):
what is the reason of shorting field names?

api/agentpb/agent.proto
yareach: I just did. Was there any requirement specifying the members' names?

yareach:
> what is the reason of shorting field names?

There is no big reason to use shorten those names. Linux tends to have shorten names. While in Windows for sure I'd used a complete name. To avoid these situations, I suggest having these details specified at the beginning. Are there any difficulties of understanding their meanings?

BupycHuk:
there is no any other place in api where we use shorten names, so let's keep using the same style.

AlekSi:
> I suggest having these details specified at the beginning

We talked about consistency, aren't we? Let's use the same field names as other APIs in that repo: address, port, socket, username, password

api/agentpb/agent.proto
```diff
@@ -167,6 +167,14 @@ message StartActionRequest {
   // PTSummaryParams describes parameters for PT summary.
   message PTSummaryParams {}
 
+  // PTMySQLSummaryParams describes parameters for PT MySQL summary.
+  message PTMySQLSummaryParams {
+    string address = 1;
+    string port = 2;
```
AlekSi, line 173: uint32

yareach: Are you sure? It is defined as uint16 in the Service struct.

AlekSi: I'm sure that there is no uint16 in protobuf :)

yareach: ok

AlekSi (approved):

BupycHuk (approved):

BupycHuk:
@yareach in future please merge all PRs related to the same task in the same time. If any of us will update `pmm` repo for `pmm-managed` now build will fail.

PR #603, opened 2020-10-27
PMM-6397 Adds metrics flow with push and pull model

f41gh7:
inventory api - option to enable push metrics flow,
management api - 3 possible ways to define metrics_mode - auto, push and pull.

https://github.com/Percona-Lab/pmm-submodules/pull/1185
https://github.com/percona/pmm/pull/603
https://github.com/percona/pmm-admin/pull/122
https://github.com/percona/pmm-agent/pull/201
https://github.com/percona/pmm-managed/pull/501
https://github.com/percona/grafana-dashboards/pull/1034
https://github.com/Percona-Lab/pmm-api-tests/pull/140


api/agentpb/agent.proto
```diff
@@ -60,6 +60,9 @@ message SetStateRequest {
     int64 max_query_log_size = 4;
   }
   map<string, BuiltinAgent> builtin_agents = 2;
+  // Scrape configuration for vmagent, that runs at pmm-agent,
+  // built and validated at a pmm-server.
+  bytes vmagent_scrape_config = 3;
```
AlekSi, line 65: Why not string?

api/prototool.yaml
```diff
@@ -11,6 +11,13 @@ protoc:
     - ../vendor
 
 lint:
+  ignores:
+    - id: ENUM_ZERO_VALUES_INVALID
+      files:
+        - managementpb/metrics.proto
+    - id: ENUM_ZERO_VALUES_INVALID_EXCEPT_MESSAGE
+      files:
+        - managementpb/metrics.proto
```
AlekSi, line 20: That's very nice that we can do it only for that file

api/agentpb/agent.proto
f41gh7: to avoid unneeded  []byte-> string,string -> []byte conversions.

 Pmm-server renders config with yaml.marshal to []byte, pmm-agent writes this config to a file.

api/prototool.yaml
f41gh7: also, i've tried to use struct comments from documentation, but it didn't work for some reason.

api/inventorypb/agents.proto
```diff
@@ -52,6 +52,8 @@ message NodeExporter {
   bool disabled = 3;
   // Custom user-assigned labels.
   map<string, string> custom_labels = 4;
+  // True if exporter use pull metrics mode.
+  bool push_metrics_disabled = 7;
```
BupycHuk, line 56: so maybe it's better to call it `pull_metrics_enabled` or `push_metrics_enabled` then?
or even better `metrics_mode` since we already have enum for it.
@AlekSi what do you think?

AlekSi: Enum exists only in Management API because only in it we have "auto". In Inventory API and database layer we have a single boolean toggle. See ticket description

BupycHuk: We should keep previous behaviour, so for existed agents we should return field with `false` value. that's why I'm suggesting to use `push_metrics_enabled`

f41gh7: done

api/inventorypb/agents.proto
```diff
@@ -52,6 +52,8 @@ message NodeExporter {
   bool disabled = 3;
   // Custom user-assigned labels.
   map<string, string> custom_labels = 4;
+  // True if exporter use pull metrics mode.
```
BupycHuk, line 55:
```suggestion
  // True if exporter uses push metrics mode.
```
Here and in other places

api/inventorypb/agents.proto
```diff
@@ -477,6 +492,7 @@ message AddNodeExporterRequest {
   ];
   // Custom user-assigned labels.
   map<string, string> custom_labels = 2;
+  bool push_metrics = 3;
```
BupycHuk, line 495: comment missed

BupycHuk (changes requested):

api/inventorypb/agents.proto
f41gh7: thx, updated.

api/inventorypb/agents.proto
f41gh7: added

BupycHuk (approved):

api/agentpb/agent.proto
AlekSi: Well. Many conversions are actually being optimised by the compiler. But is also not like we care about conversion performance there – that place is far from being a hot. What we care about is debug-ability: bytes will not be logged properly in debug/trace modes, for example.

But you know, I actually think we should use `AgentProcess vmagent = 3` there, and pass everything, including templated flags, environment, etc from the PMM Server (`type` should be invalid in this case, but everything else should be used). That will allow use to enable debug mode from the server, for example, or send different flags for different versions of pmm-agent/vmagent.

/cc @BupycHuk

api/inventorypb/agents.proto
AlekSi: But it is copied from the wrong place now :) It is copied from `push_metrics_enabled` above (describes state), not from `push_metrics` below (describes action).

api/inventorypb/agents.proto
```diff
@@ -629,6 +650,8 @@ message AddPostgresExporterRequest {
   map<string, string> custom_labels = 7;
   // Skip connection check.
   bool skip_connection_check = 8;
+  // Enables push metrics mode for exporter.
+  bool push_metrics = 9;
 }
```
AlekSi, line 655: It should also be added to proxysql_exporter, postgres_exporter rds_exporter at least. But please also check with @BupycHuk other types.

AlekSi (changes requested):

api/inventorypb/agents.proto
f41gh7: as far as i understand, there is no need for rds_exporter and it will be scrapped by pmm-server. 

 for proxysql_exporter needs guidance, dont know what it is.

api/agentpb/agent.proto
f41gh7: Ok, get it, its not hard to change for string. 

 Need to think how to implement it better, you rejected this suggestion, last time we've discussed it. If we can simply add new agentType for vmagent it solves a lot of problems (scrape config generation, configuration, error logging and etc). For instance, you would be able to manage it from pmm-admin for "debug mode".

AlekSi: Let's talk about it on or after our call tomorrow

api/inventorypb/agents.proto
AlekSi: All those exporters are in this file.
rds_exporter is _typically_ started on PMM Server, but it is a normal exporter and it _can_ be started anywhere, so no reason to exclude it.

api/inventorypb/agents.proto
f41gh7: yep, changed

api/inventorypb/agents.proto
f41gh7: added missing exporters.

api/agentpb/agent.proto
f41gh7: added vmagent type.

AlekSi (approved):

PR #609, opened 2020-11-11
Pmm 6397 vmagent type

AlekSi:


AlekSi:
@f41gh7 Do we need that branch, that was pushed without PR? If not, please close and delete branch

PR #613, opened 2020-11-11
PMM-6826 Add disk size to create API.

BupycHuk:
https://github.com/percona/pmm-managed/pull/519
https://github.com/percona-platform/dbaas-controller/pull/91
https://github.com/percona/pmm/pull/613
https://github.com/percona-platform/dbaas-api/pull/62
https://github.com/Percona-Lab/pmm-submodules/pull/1220


api/managementpb/dbaas/psmdb_cluster.proto
```diff
@@ -28,6 +28,8 @@ message PSMDBClusterParams {
   message ReplicaSet {
     // Requested compute resources.
     ComputeResources compute_resources = 1;
+    // Disk size in megabytes.
+    int64 disk_size = 2;
```
BupycHuk, line 32: I made it separate field because it's not a compute resource.

api/managementpb/dbaas/psmdb_cluster.proto
```diff
@@ -28,6 +28,8 @@ message PSMDBClusterParams {
   message ReplicaSet {
     // Requested compute resources.
     ComputeResources compute_resources = 1;
+    // Disk size in megabytes.
```
AlekSi, line 31: Megabytes or mebibytes?

Using bytes (and SI units in general) would solve that problem :)

BupycHuk: Good point, I though that int64 won't be able to store that value, but checked again and looks like bytes is the best solution.

askomorokhov (approved):

AlekSi (approved):

JiriCtvrtka (approved):

percona-csalguero (approved):

PR #618, opened 2020-11-19
PMM-5364 SSL support for Mongo.

JiriCtvrtka:


api/inventorypb/agents.proto
```diff
@@ -623,6 +623,12 @@ message AddMongoDBExporterRequest {
   bool tls = 5;
   // Skip TLS certificate and hostname validation.
   bool tls_skip_verify = 6;
+  // Certificate file.
+  string tls_certificate_key_file = 10;
```
AlekSi, line 627: Not file – file's content. So API field (unlike pmm-admin CLI flag, for example) should not have "_file".

api/inventorypb/agents.proto
```diff
@@ -623,6 +623,12 @@ message AddMongoDBExporterRequest {
...
+  // Certificate password.
+  string tls_certificate_key_file_password = 11;
```
AlekSi, line 629: TBH, not sure about that one. Storing encrypted content next to the decryption password doesn't achieve anything. It might be better to decrypt file on pmm-admin side.

AlekSi (commented):
@BupycHuk PTAL

AlekSi (commented):
Another note – I _think_ that QAN's MongoDB profiler is in the scope of this ticket, but it should be mentioned explicitly

api/inventorypb/agents.proto
JiriCtvrtka: Yes, you are right. Will remove "_file" from fields, where I have already loaded certificate.

api/inventorypb/agents.proto
```diff
@@ -882,6 +888,12 @@ message AddQANMongoDBProfilerAgentRequest {
   bool tls = 5;
   // Skip TLS certificate and hostname validation.
   bool tls_skip_verify = 6;
+  // Certificate.
+  string tls_certificate_key = 9;
+  // Certificate password.
+  string tls_certificate_key_file_password = 10;
+  // Certificate authority.
+  string tls_ca = 11;
```
percona-csalguero, line 896: This is repeated? at the beginning of the block there is a tls_cs field with number 12

api/inventorypb/agents.proto
percona-csalguero: :+1:

percona-csalguero (changes requested):

api/inventorypb/agents.proto
JiriCtvrtka: No, one for AddMongoDBExporterRequest, second one for AddQANMongoDBProfilerAgentRequest and diffrent numbers because of diffrent count of properties.

api/inventorypb/agents.proto
BupycHuk: It will require including mongodb lib into pmm-admin and as far as I remember we are trying to keep pmm-admin as simple as possible. and it's simpler to just use `tlsCertificateKeyFile` and `tlsCertificateKeyFilePassword` in DSN

BupycHuk (approved):

percona-csalguero (approved):

api/inventorypb/agents.proto
AlekSi: X.509 certificate decryption can be done without mongolib for sure – all requires package are in stdlib.

I will ask around if we have use case for that

JiriCtvrtka: Link to Roma response:
https://percona.slack.com/archives/G2HEW592P/p1606391990252400?thread_ts=1606376788.238000&cid=G2HEW592P

askomorokhov (approved):

percona-csalguero (approved):

api/agentpb/agent.proto
```diff
@@ -11,6 +11,13 @@ import "inventorypb/agent_status.proto";
 import "inventorypb/agents.proto";
 import "inventorypb/services.proto";
 
+// TextFiles contains files which can be used to connect to DB (certificates, keys and etc).
+message TextFiles {
+  map<string, string> files = 1;
+  string template_left_delim = 2;
+  string template_right_delim = 3;
```
artemgavrilov, line 18: What is this? What is template and for what this delimiters required for?

Can we some comments here?

BupycHuk: template_left_delim and template_right_delim are used to replace placeholders in templates(DSN and etc.) with paths to files.

artemgavrilov: IMHO better to add some comments with explanation of purpose of these fields.

artemgavrilov (approved):

api/agentpb/agent.proto
BupycHuk: Done

api/agentpb/agent.proto
```diff
@@ -11,6 +11,15 @@ import "inventorypb/agent_status.proto";
 import "inventorypb/agents.proto";
 import "inventorypb/services.proto";
 
+// TextFiles contains files which can be used to connect to DB (certificates, keys and etc).
+message TextFiles {
+  map<string, string> files = 1;
+  // template_left_delim is used to replace placeholders in templates(DSN and etc.) with paths to files.
```
AlekSi, line 17:
```suggestion
  // template_left_delim is used to replace placeholders in templates (DSN and etc.) with paths to files.
```

api/agentpb/agent.proto
```diff
@@ -11,6 +11,15 @@ import "inventorypb/agent_status.proto";
 import "inventorypb/agents.proto";
 import "inventorypb/services.proto";
 
+// TextFiles contains files which can be used to connect to DB (certificates, keys and etc).
+message TextFiles {
+  map<string, string> files = 1;
```
AlekSi, line 16: Comment there?

api/managementpb/actions.proto
```diff
@@ -250,6 +250,13 @@ message StartPostgreSQLShowIndexActionResponse {
   string pmm_agent_id = 2;
 }
 
+// MongoDBOptions is used to passing TLS certificates and password.
+message MongoDBOptions {
+  string tls_certificate_key = 1;
+  string tls_certificate_key_file_password = 2;
+  string tls_ca = 3;
+}
```
AlekSi, line 258: Comments?

And why the are not inline in StartMongoDBExplainActionRequest?

BupycHuk: good point, will update

BupycHuk: actually we don't need it at all.

BupycHuk: Removed

api/agentpb/agent.proto
BupycHuk: Added

api/agentpb/agent.proto
BupycHuk: Fixed

PR #625, opened 2020-11-25
PMM-6827 DB cluster monitoring.

BupycHuk:


api/serverpb/server.proto
```diff
@@ -147,6 +149,9 @@ message ChangeSettingsRequest {
   bool enable_stt = 11;
   // Disable Security Threat Tool
   bool disable_stt = 12;
+  // PMM Server public address.
+  string pmm_public_address = 13;
```
askomorokhov, line 153: What about pmm external address? :)

askomorokhov (approved):

api/serverpb/server.proto
```diff
@@ -122,6 +122,8 @@ message Settings {
   string platform_email = 10;
   // DBaaS enabled
   bool dbaas_enabled = 11;
+  // PMM Server public address.
+  string pmm_public_address = 13;
```
JiriCtvrtka, line 126: Number 12 skipped because of consistence with Change Settings?

percona-csalguero (approved):

AlekSi (commented):
Why is that address and not URL? Please add example value to the comment

BupycHuk:
@AlekSi operators expects address, the same we use in PMM_AGENT_SERVER_ADDRESS

AlekSi:
I'm pretty sure they also need username and password for `pmm-agent setup`

BupycHuk:
@AlekSi at the moment they use secrets for the password in xtradb and for the username and password in psmdb. Maybe we can leave it for the time when we will handle secrets creation? Just to not confuse users who will expect that if they set url with username and password it should work. WDYT?

AlekSi:
Sure

api/serverpb/server.proto
BupycHuk: isn't it the same? public sounds better for me :)

api/serverpb/server.proto
BupycHuk: Will update before merge

api/serverpb/server.proto
askomorokhov: This is not critical. :)
The word "public" has the connotation "public place". pmm-server not necessary to be publicly available for everybody.  It only must be available for external to its services.

PR #638, opened 2020-12-08
PMM-7108 Fix reform version

AlekSi:


BupycHuk (commented):
@AlekSi Why do we use reform in pmm repository?

AlekSi:
For Prometheus metrics: https://github.com/percona/pmm/blob/PMM-2.0/utils/sqlmetrics/reform.go This code is used by pmm-managed and pmm-agent.

BupycHuk (approved):

PR #639, opened 2020-12-08
PMM-4306 Pt-mongodb-summary

yareach:


api/agentpb/agent.proto
```diff
@@ -167,6 +167,13 @@ message StartActionRequest {
   // PTSummaryParams describes parameters for PT summary.
   message PTSummaryParams {}
 
+  // PTMgDbSummaryParams describes parameters for PT MongoDB summary.
+  message PTMgDbSummaryParams {
```
BupycHuk, line 171: Command is called pt-mongodb-summary, everywhere in this file we MongoDB, so why should we start using new MgDb here? Next time before creating new variable, please check at least the same file for usages.

api/agentpb/agent.proto
```diff
@@ -167,6 +167,13 @@ message StartActionRequest {
...
+    string address = 1;
```
BupycHuk, line 172: it should be renamed to host.

BupycHuk: please do the same in other PRs.

BupycHuk (changes requested):

api/agentpb/agent.proto
yareach: Sometimes it uses "mongodb", sometimes "mongo". The same for postgres - "posgresSQL", "pg", "pgSQL". I haven't found any clear key to name it. If there was no problem with "pg" or "pgSQL", I don't understand why there is with "MgDb".

api/agentpb/agent.proto
yareach: All of you have approved the "address" field in previous two PRs. There was a discussion at the beginning on these fields' names, it was also changed from "addr" to "address".
I will change it in this PR. The others were approved, therefore in my opinion a correct action would be to file a ticket (two) to change it. What do you think?

api/agentpb/agent.proto
BupycHuk: Could you point where we used pgsql?
`pg` is used for `pg_stat_monitor`, `pg_stat_statements`, `pt-pg-summary` as you can see it's used because of naming in 3rd party tools.
`postgres` used only with `postgres_exporter` for the same reason.
`PostgreSQL` in all other places.

yareach: I used pgSQL here: https://github.com/percona/pmm/pull/614/files
And it has two approvals, one by yourself.
I can use full words, I don't have problem with that. But I wouldn't like to read then comments like "this is Linux, not Windows".

BupycHuk: in that case it should be `pg`, not `pgSQL`. So we had 3 options and you decided to add 4th one? Why did you decide to use `pgSQL` there? We missed that and it should be fixed there.

api/agentpb/agent.proto
BupycHuk: We discussed it with Marcos and agreed that it should be done as part of these PRs, because we won't be able to change it in later releases because of backward compatibility and there is high possibility that we won't deliver your new changes in the same release with current PRs.

percona-csalguero (changes requested):
Please change `PTMgDb` to `PTMongoDB` in all places.

api/agentpb/agent.proto
yareach: Maybe my sense for uniformity: mySQL - pgSQL - mgDQ. Maybe a question for Freud. 
Why? Why not? You have the same deal of guilt as I do since you've approved. Don't try to escape from this...

api/agentpb/agent.proto
yareach: I'm on it.

BupycHuk (approved):

PR #651, opened 2021-01-12
Docs moved to pmm-doc

PaulJacobs-percona:
Sphinx/rst migrated to MkDocs/md and copied to 1.x branch of pmm-doc repo.
doc/ dir renamed to .doc until confirmation of content. Then it can be deleted.


rnovikovP (approved):

BupycHuk (approved):

BupycHuk:
I think we should remove these checks which are waiting for status to be reported too

PaulJacobs-percona:
> I think we should remove these checks which are waiting for status to be reported too

Who knows how?

atymchuk:
> > I think we should remove these checks which are waiting for status to be reported too
> 
> Who knows how?

We can't remove them since they are triggering updates for the feature builds when you modify the protobufs, for example.

PaulJacobs-percona:
> > > I think we should remove these checks which are waiting for status to be reported too
> > 
> > 
> > Who knows how?
> 
> We can't remove them since they are triggering updates for the feature builds when you modify the protobufs, for example.

Then it's catch-22. Can't merge until checks finish, which they never will because they depend on docs.

PR #652, opened 2021-01-12
fix: do not build the docs

atymchuk:
Since the doc directory is gone, we should no longer run `Build docs` stage

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,17 +15,6 @@ pipeline {
                 '''
             }
         }
-        stage('Build Docs') {
-            steps {
-                sh '''
-                    sudo make -C doc clean
-                    make -C doc theme
-                    sg docker -c "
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -15,17 +15,6 @@ pipeline {
...
-                sh '''
-                    sudo make -C doc clean
-                    make -C doc theme
-                    sg docker -c "
-                        docker run --rm -v $(pwd -P)/doc:/doc florianholzapfel/docker-alpine-sphinx make -C /doc html
-                    "
-                '''
-            }
-        }
         stage('Publish Docs') {
```
BupycHuk, line 18: if we are removing build docs step, then we should also remove publish docs step

atymchuk: yes, and then the whole pipeline doesn't make sense )

BupycHuk: Then let's remove whole pipeline :)

atymchuk:
Closing this since the pipeline has to be deleted as part of the PR https://github.com/percona/pmm/pull/651

PR #654, opened 2021-01-14
Add dev swagger.json

askomorokhov:
https://jira.percona.com/browse/PMM-7287

FB: https://github.com/Percona-Lab/pmm-submodules/pull/1385

image


api/Makefile
```diff
@@ -53,6 +53,17 @@ gen: clean            ## Generate files.
 
 	swagger-order --output=swagger/swagger.json swagger/swagger.json
 
+	# generate dev API spec
+	swagger mixin --output=swagger/swagger-dev.json \
+		swagger/header-dev.json \
```
BupycHuk, line 58: can we try to mixin `swagger.json` with `dbaas.json` and check if the output is the same?
Also let's include `ia` and `qan` packages.

BupycHuk (changes requested):

api/Makefile
askomorokhov: Mixin swagger.json will not work as we need a custom header.
Added is and qan.

BupycHuk (approved):

JiriCtvrtka (approved):

percona-csalguero (approved):

PR #659, opened 2021-01-20
PMM-7145 Push mode by default.

JiriCtvrtka:


api/managementpb/rds.proto
```diff
@@ -126,6 +126,7 @@ message AddRDSRequest {
   // Defines metrics flow model for this exporter.
   // Metrics could be pushed to the server with vmagent,
```
AlekSi, line 127: No, they could not be with that API

api/managementpb/rds.proto
```diff
@@ -126,6 +126,7 @@ message AddRDSRequest {
...
   // pulled by the server, or the server could choose behavior automatically.
+  // Push metrics mode is not allowed for exporters running on pmm-server.
```
AlekSi, line 129: There no way to choose a different Node in that API

AlekSi (changes requested):

BupycHuk (approved):

askomorokhov (approved):

PR #662, opened 2021-01-25
PMM-7296: List Alert Rules pagination

oter:
https://jira.percona.com/browse/PMM-7296


api/managementpb/ia/pagination.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package ia.v1beta1;
+
+option go_package = "api/managementpb/ia;iav1beta1";
+
+// PaginationRequest represents request values for page by page pagination.
+message PaginationRequest {
```
artemgavrilov, line 8: Typically `*Request/Response` are top-level entities, used as parameter/response for gRPC methods. Probably we should think about naming.

api/managementpb/ia/pagination.proto
```diff
@@ -0,0 +1,21 @@
...
+  int32 page = 2;
+}
+
+// PaginationResponse represents response values for page by page pagination.
+message PaginationResponse {
+  // Total number of results.
+  int32 total_items = 1;
+  // Total number of pages.
+  int32 total_pages = 2;
+}
```
artemgavrilov, line 21: How client should understand what page index to pass to the next request? Calculate it each time form `page_size` and total returned items?

api/managementpb/ia/pagination.proto
```diff
@@ -0,0 +1,21 @@
...
+package ia.v1beta1;
+
+option go_package = "api/managementpb/ia;iav1beta1";
+
+// PaginationRequest represents request values for page by page pagination.
+message PaginationRequest {
+  // Maximum number of results per page.
+  int32 page_size = 1;
+  // Index of the requested page, starts from 0.
+  int32 page = 2;
```
artemgavrilov, line 12: Typically `cursor` gives a more slick UX than `offset` approach.  Maybe we can think about it?

Dasio: Cursor is better also performance-wise but I don't think it is suitable for us. There won't be so many records >10000.
Also implementation is based on UI and there are pages. With cursor approach we can't go to exact page, just to the next.

AlekSi:
(PR should not be a draft if reviews are already requested)

api/managementpb/ia/pagination.proto
oter: naming reconsidered

api/managementpb/ia/pagination.proto
oter: No need to calculate, the user chooses the page he wants to go to and this is a page index that passed in the request. 
The FE part knows the number of pages available from the response, so the appropriate number of page links is shown.

artemgavrilov: Ok, thanks for the explanation!

api/managementpb/ia/pagination.proto
artemgavrilov: Yes, my message was about the representation approach, but it's out of code review scope.

artemgavrilov (approved):

Dasio (approved):

api/managementpb/ia/pagination.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package ia.v1beta1;
+
+option go_package = "api/managementpb/ia;iav1beta1";
+
+// Page represents page request values for pagination.
+message Page {
```
BupycHuk, line 8: maybe `PageParams` fits better here

BupycHuk (commented):
just a few comments about namings

api/managementpb/ia/pagination.proto
oter: changed

api/managementpb/ia/pagination.proto
```diff
@@ -0,0 +1,21 @@
...
+  // Maximum number of results per page.
+  int32 size = 1;
```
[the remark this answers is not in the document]
oter: renamed to `page_size`

BupycHuk (approved):

PR #663, opened 2021-01-26
PMM-7119 Add field name to UpdateTemplateRequest

Dasio:


AlekSi (approved):

api/managementpb/ia/templates.proto
```diff
@@ -132,8 +133,14 @@ message CreateTemplateRequest {
 message CreateTemplateResponse {}
 
 message UpdateTemplateRequest {
+  // Machine-readable name (ID).
+  string name = 1 [
```
BupycHuk, line 137: Probably it's what we discussed about changing ids in beta APIs, right?

BupycHuk (approved):

api/managementpb/ia/templates.proto
Dasio: Yeah. Even without changing ids, it is breaking change. Because name field is new and required.

PR #678, opened 2021-02-08
Add test descriptions

artemgavrilov:
Screen Shot 2021-02-11 at 11 47 38

That's how we can add custom descriptions to the swagger.

Most likely we need only method descriptions.


api/serverpb/server.proto
```diff
@@ -181,7 +182,18 @@ message GetSettingsResponse {
 }
 
 message ChangeSettingsRequest {
-  bool enable_telemetry = 1;
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: {
+      title: "Super puper message title!"
+      description: "Awesome message description"
+    }
+  };
```
artemgavrilov, line 190: Message title

artemgavrilov, line 190: Message title and descriptions

api/serverpb/server.proto
```diff
@@ -181,7 +182,18 @@ message GetSettingsResponse {
...
+  bool enable_telemetry = 1 [
+    (grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {
+      title: "Super puper field title!"
+      description: "Awesome field description"
+    }
+  ];
```
artemgavrilov, line 196: Field title and descriptions

p.s. Descriptions already present in our docs

api/serverpb/server.proto
```diff
@@ -314,6 +326,10 @@ service Server {
       post: "/v1/Settings/Change"
       body: "*"
     };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      summary: "super puper method summary!"
+      description: "Awesome method description"
+    };
```
artemgavrilov, line 332: Method summary and description

p.s. Summaries already present in our docs

api/serverpb/server.proto
AlekSi: But do we need them? Messages don't really exist in JSON APIs

api/serverpb/server.proto
AlekSi: I vaguely remember that both title and description can be set without annotation if you end sentences with dots

api/serverpb/server.proto
AlekSi: Aren't descriptions too? Try to add another sentence (ending it with a dot) to the comment.

artemgavrilov: `/v1/readyz` has such comment, but on swagger UI it's visualized as the summary
```
  // Readiness returns an error when some PMM Server component is not ready yet or is being restarted.
  // It can be used as for Docker health check or Kubernetes readiness probe.
  rpc Readiness(ReadinessRequest) returns (ReadinessResponse) {
    option (google.api.http) = {
      get: "/v1/readyz"
    };
  }
```

Screen Shot 2021-02-11 at 12 36 52

api/serverpb/server.proto
artemgavrilov: That's more like a comment to the request body (see screenshot), and I suppose we don't need them. I just tried what we can to specify.

Most likely we need only method descriptions.

api/serverpb/server.proto
artemgavrilov: The whole comment visualized as a description, but I think titles are useless for fileds.
```
  // A number of full days for Prometheus and QAN data retention. Should have a suffix in JSON: 2592000s, 43200m, 720h.
  google.protobuf.Duration data_retention = 4;
  ```
  
Screen Shot 2021-02-11 at 12 43 59

PR #688, opened 2021-02-18
PMM-5778 Check exporter format

askomorokhov:
https://jira.percona.com/browse/PMM-5778

FB: https://github.com/Percona-Lab/pmm-submodules/pull/1521

Ref:
 - https://github.com/percona/pmm/pull/688
 - https://github.com/percona/pmm-agent/pull/237
 - https://github.com/percona/pmm-managed/pull/670


AlekSi (changes requested):
Unclear field name, no documentation comment

api/agentpb/agent.proto
```diff
@@ -310,6 +310,8 @@ message CheckConnectionResponse {
   // Error message if connection check failed.
   string error = 1;
   Stats stats = 2;
+  // Contains the body of HTTP response of connection check to an exporter.
+  string exporter_response_body = 3;
```
AlekSi, line 314: `CheckConnectionResponse` is about checking connectivity with Services, not with exporters. The word "exporter" is not used in this file at all.

Additionally, protobuf `string` can contain only UTF-8 data. I don't think we want limit us it it there

AlekSi (changes requested):

api/agentpb/agent.proto
askomorokhov: @AlekSi, and what is your suggestion?

AlekSi: https://github.com/percona/pmm/pull/689

askomorokhov: Thanks, elegant detour :)

api/agentpb/agent.proto
```diff
@@ -310,6 +310,8 @@ message CheckConnectionResponse {
   // Error message if connection check failed.
   string error = 1;
   Stats stats = 2;
+  // Contains the body of HTTP response of connection check to an exporter.
```
BupycHuk, line 313: to be honest it looks weird for me that all other connections we check on pmm-agent side (even returning table_count), but for exporters we move connection check to pmm-server side.

PR #702, opened 2021-03-03
PMM-7590 Jobs API

artemgavrilov:
https://jira.percona.com/browse/PMM-7590

https://github.com/percona/pmm-agent/pull/243
https://github.com/percona/pmm-managed/pull/700


api/jobspb/jobs.proto
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// StartJob asks pmm-agent to start job.
+message StartJob {
+  // Echo is a simple echo job that can be delayed.
+  message Echo {
+    string message = 1;
+    google.protobuf.Duration delay = 2;
+  }
+  string job_id = 1;
```
Dasio, line 26: this probably should generate agent not and not server

artemgavrilov: Jobs API works as a bidirectional stream, so there is no explicit response where agent can notify server about what ID was assigned to the job. The server just throws StartJob message and waits for result/status/progress with the same ID.

api/jobspb/jobs.proto
```diff
@@ -0,0 +1,81 @@
...
+    string status = 1;
+  }
+  google.protobuf.Timestamp timestamp = 1;
+  oneof result {
+    Echo echo = 10;
+  }
+}
+
+// AgentMessage is a container for the agent messages.
+message AgentMessage {
```
BupycHuk, line 58: Maybe it's better to include it here? https://github.com/percona/pmm/blob/release%2F2.16/api/agentpb/agent.proto#L336

artemgavrilov: I don't get it, include what? Message payload on line 61, but it's all about jobs only.

api/agentpb/agent.proto
```diff
@@ -329,6 +329,73 @@ message CheckConnectionResponse {
   Stats stats = 2;
 }
 
+// JobStatusRequest is a ServerMessage asking pmm-agent for job status.
+message JobStatusRequest {
```
BupycHuk, line 333: Do we use it somewhere?

BupycHuk (changes requested):

Dasio (approved):

api/agentpb/agent.proto
```diff
@@ -329,6 +329,73 @@ message CheckConnectionResponse {
...
+  string job_id = 1;
+}
+
+// JobStatusResponse is an AgentMessage containing job status.
+message JobStatusResponse {
+  bool alive = 1;
```
oter, line 339: maybe use enum? since we could have more statuses like stopped, stopped with error etc.

api/managementpb/jobs.proto
```diff
@@ -0,0 +1,98 @@
...
+    // Error result.
+    Error error = 10;
+    // Echo job result.
+    Echo echo = 11;
+  }
+}
+
+message StartEchoJobRequest {
+  // pmm-agent ID where to run this Action.
+  string pmm_agent_id = 1;
```
oter, line 45: should it be required?

oter (approved):

api/agentpb/agent.proto
artemgavrilov: Managed doesn't request it anywhere now, but it allows to check if the job alive or dead.

api/agentpb/agent.proto
artemgavrilov: If job stoped or terminated its result will be delivered to the server. We need this mostly for the case when pmm-agent crashed or restarted and we want to check if our job is alive.

api/managementpb/jobs.proto
artemgavrilov: Hm, yes, for echo job it's required. But in general - no.

Fixed.

api/managementpb/jobs.proto
```diff
@@ -0,0 +1,98 @@
...
+  }
+  // Echo contains result for echo job.
+  message Echo {
+    string message = 1;
+  }
+  // Unique Action ID.
+  string job_id = 1;
+  // pmm-agent ID where this Action is running / was run.
+  string pmm_agent_id = 2;
+  // True if Action is finished.
```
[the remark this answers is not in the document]
artemgavrilov: Fixed

api/managementpb/jobs.proto
```diff
@@ -0,0 +1,98 @@
...
+  message Error {
+    string message = 1;
+  }
+  // Echo contains result for echo job.
+  message Echo {
+    string message = 1;
+  }
+  // Unique Action ID.
+  string job_id = 1;
+  // pmm-agent ID where this Action is running / was run.
```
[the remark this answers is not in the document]
artemgavrilov: Fixed

api/managementpb/actions.proto
```diff
@@ -151,7 +151,11 @@ message StartMySQLShowCreateTableActionResponse {
 
 message StartMySQLShowTableStatusActionRequest {
   // pmm-agent ID where to run this Action.
-  string pmm_agent_id = 1;
+  string pmm_agent_id = 1 [
```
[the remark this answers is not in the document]
artemgavrilov: It was mistake. Fixed.

BupycHuk (approved):

PR #709, opened 2021-03-17
PMM-7386 Start backup

Dasio:
https://jira.percona.com/browse/PMM-7386


api/managementpb/backup/backups.proto
```diff
@@ -0,0 +1,47 @@
...
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  string location_id = 2 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  repeated string database_names = 3;
```
BupycHuk, line 28: as far as I know currently we don't support backup for exact databases.

api/managementpb/backup/backups.proto
```diff
@@ -0,0 +1,47 @@
...
+  string name = 4;
```
BupycHuk, line 29: can we add comments to these fields to make them clear?

api/managementpb/backup/backups.proto
Dasio: It's in arch design.
```
databaseNames: string[], required: false
note: "false" means it should backup all schemas found in the service
```
Everything outside of service_id and location_id is out of scope for this ticket. 
But I would keep it in API, even though is doesn't do anything. Maybe we can comment it, that for now it is unused.

BupycHuk: Yes, let's [reserve](https://developers.google.com/protocol-buffers/docs/proto3#reserved) the id = 3 and add a comment. I don't want to confuse people with a field which is not used actually.

api/managementpb/backup/backups.proto
```diff
@@ -0,0 +1,58 @@
...
+
+message PerformBackupResponse {
+  // Unique identifier.
+  string backup_id = 1;
+}
+
+// Backups service handles backup operations to DB.
+service Backups {
+  // PerformBackup request backup specified service to location.
+  rpc PerformBackup(PerformBackupRequest) returns (PerformBackupResponse) {
```
artemgavrilov, line 52: Maybe it's better to name it `StartBackup`, `perform` a quite unusual verb for the method name.

api/managementpb/backup/backups.proto
Dasio: I commented it out. Is that reserved keyword good idea? When I read docs, it's said, that it should be used for deleted fields.

api/managementpb/backup/backups.proto
Dasio: On arch design there is used verb `perform`.

oter (approved):

atymchuk (approved):

api/agentpb/agent.proto
```diff
@@ -339,18 +339,36 @@ message JobStatusResponse {
...
 message StartJobRequest {
   // Echo is a simple echo job that can be delayed.
   message Echo {
     string message = 1;
     google.protobuf.Duration delay = 2;
   }
+  // MySQLBackup is job for backup MySQL service.
+  message MySQLBackup {
+    // DSN for the service.
+    string dsn = 1;
```
artemgavrilov, line 360: I think it should be service_id, not `dsn`. `dsn` we can find on managed side by service id.

artemgavrilov: Ah, wrong file. Nevermind.

oter (approved):

api/managementpb/backup/backups.proto
```diff
@@ -0,0 +1,58 @@
...
+  ];
+  // Machine-readable location ID.
+  string location_id = 2 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  // Database names to backup, If empty then each database is backed up.
+  // repeated string database_names = 3;
+  // If empty then name is auto-generated.
```
BupycHuk, line 32: let's add empty line between these two fields.

api/managementpb/backup/backups.proto
```diff
@@ -0,0 +1,58 @@
...
+import "google/api/annotations.proto";
+
+// RetryMode specifies how backup should retry in case of failure.
+enum RetryMode {
+  RETRY_MODE_INVALID = 0;
+  AUTO = 1;
+  MANUAL = 2;
+}
+
+message StartBackupRequest {
```
BupycHuk, line 17: let's remove commented fields

api/managementpb/backup/backups.proto
```diff
@@ -0,0 +1,58 @@
...
+  // Retry interval in seconds. Default 30s.
+  // uint32 retry_interval = 7;
+  // How many times should backup be retried in case of failure. Default 2 times.
+  // uint32 retry_times = 8;
+
+}
+
+message StartBackupResponse {
+  // Unique identifier.
+  string artifact_id = 1;
```
BupycHuk, line 46: should we return backup_id as well?

BupycHuk (changes requested):

api/managementpb/backup/backups.proto
artemgavrilov: Removed all commented fields

api/managementpb/backup/backups.proto
artemgavrilov: Done

api/managementpb/backup/backups.proto
artemgavrilov: That's a good questions. Right now there is no way to get a list of running backups. Artifacts list used for that purpose, so each artifact has a status where you can understand backups state.

BupycHuk: I think that's not the best idea to use artifacts to show status, because artifacts are result of backup process. Feels like something went wrong during creating architecture.

artemgavrilov: Well, we follow UI/architecture. We discussed this with @atymchuk yesterday, for user there is no such thing as job/backup process. On UI he sees artifacts list with statuses (In progress/done/error/etc)

artemgavrilov (approved):

PR #711, opened 2021-03-22
PMM-7697 Enable/Disable backup management

Dasio:
https://jira.percona.com/browse/PMM-7697


api/serverpb/server.proto
```diff
@@ -232,6 +234,10 @@ message ChangeSettingsRequest {
   bool remove_pmm_public_address = 20;
   // Intervals between STT check runs.
   STTCheckIntervals stt_check_intervals = 21;
+  // Enable Backup Management.
+  bool enable_backup_management = 22;
+  // Disable Backup Management.
+  bool disable_backup_management = 23;
```
oter, line 240: just interesting, why using 2 fields?

oter: we could use well known types e.g. https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.BoolValue

oter (approved):

api/serverpb/server.proto
Dasio: Yeah, I was suggesting this also.
But wanted to keep API consistent.

oter: ok then

atymchuk: @Dasio @BupycHuk What is the consequence of starting to use the https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.BoolValue to avoid the double boolean fields? They have always been a significant concern for both UI and QA engineers - difficult to understand and test as well.

Therefore, this is a good precedent to break consistency.

We don't want to migrate everywhere, I understand. But we don't want to continue developing the APIs with double booleans either.

Dasio: I can't see any consequence. With using these wrappers, we can distinguish if value was not sent.

artemgavrilov (approved):

api/serverpb/server.proto
atymchuk: Then why don't we use it?

Dasio: It's not up to me  to make this decision. And also maybe I'm missing something.

BupycHuk: Hi, the reason is simple, grpc-gateway can't differentiate between empty value and false. API was written the way that only part of values can be presented.

lunaticusgreen: Cannot agree too much with @atymchuk, current behavior is a little overcomplicated and lead to possible bugs on a frontend when we forgetting about it

fabio-silva: I agree with @atymchuk that this is a good chance for us to break the chain. After this, we'll have a set of fields that need to be refactored, but this set won't grow any larger than that.

BupycHuk: Any idea how can we change behaviour, but keep ability to pass small part of fields to change only them?

nicolalamacchia: My suggestion is not to scatter different approaches around. While I agree that it would be better to avoid this enable/disable, this is something that we should start doing consistently. So please, if we agree on a solution let's bring that solution to the other APIs as well.

atymchuk: Right now the only API endpoint (to my knowledge) that is affected is `/v1/Settings/Change`. So I'm not sure there is much scattering.

Moreover, some work has already been done as part of [PMM-4597](https://github.com/percona/pmm-managed/pull/473/files).

atymchuk: We agreed to not block this PR and to let it be merged. BE Team will further discuss the issue to find a plausible solution to ease the pain.

atymchuk (approved):

api/serverpb/server.proto
Dasio: @BupycHuk " grpc-gateway can't differentiate between empty value and false", yeah. But If we would use that wrappers instead of actual value. (BoolWrapper vs Bool) we know if value was not sent, or it is true/false. So instead of 2 states(true/false) we would have 3 (not sent/true/false). And REST API should be same.

Dasio: Also this affect any "Change/Update" endpoint, not just settings. And we can't do it everywhere now. It's breaking change, even though JSON should be same.

BupycHuk: @Dasio Yeah, that's correct. We discussed with @atymchuk and @oter. We can start implementing new approaches in new APIs (backup, ia, e.t.c.) for now and check how it works.

BupycHuk (approved):

PR #713, opened 2021-03-28
PMM-7358 New endpoint for Managing components.

BupycHuk:
I have two ideas to implement. 

1. First one is this one.
2. The second one is `ChangeKubernetesCluster` endpoint and field for each component.

What do you think which one is better to keep?


jprukner:
@BupycHuk I think this one is better.

jprukner (approved):

tiagomotasantos (approved):
This one seems best to me. There's no benefit in putting everything in the same request, we always have to separate it in some way because each operator has different components and they can also have components with the same name (e.g. backup).

api/managementpb/dbaas/components.proto
```diff
@@ -58,6 +58,32 @@ message GetPXCComponentsResponse {
...
+message SetPSMDBComponentsRequest {
+  // Kubernetes cluster name.
+  string kubernetes_cluster_name = 1;
+  SetComponent mongod = 2;
+  SetComponent backup = 3;
+}
+
+message SetPSMDBComponentsResponse {}
+
+message SetPXCComponentsRequest {
```
tiagomotasantos, line 76: Shouldn't we allow to pass pmm here too?

BupycHuk: Good point. But version server contains outdated pmm client. I'll create a ticket for it.

BupycHuk: https://jira.percona.com/browse/PMM-7814

tiagomotasantos: Got it!! 👍

jprukner (approved):

PR #714, opened 2021-03-29
PMM-7338 Add API for changing security check interval

palash25:


api/managementpb/checks.proto
```diff
@@ -37,6 +47,13 @@ message ChangeSecurityCheckParams {
   bool disable = 3;
 }
 
+// ChangeSecurityCheckIntervalParams specifies interval change parameters for a single check.
+message ChangeSecurityCheckIntervalParams {
+  // The name of the check to change interval of.
+  string name = 1;
+  SecurityCheckInterval interval = 2;
```
artemgavrilov, line 54: Missing comment

api/managementpb/checks.proto
```diff
@@ -59,6 +76,12 @@ message ChangeSecurityChecksRequest {
 
 message ChangeSecurityChecksResponse {}
 
+message ChangeSecurityChecksIntervalRequest {
+  ChangeSecurityCheckIntervalParams param = 1;
```
artemgavrilov, line 80: Why do we need this `ChangeSecurityCheckIntervalParams`? Why not just add required fields to the `ChangeSecurityChecksIntervalRequest` directly?

palash25: you are right, I must've copied it from `ChangeSecurityChecks` will fix it

api/managementpb/checks.proto
```diff
@@ -7,6 +7,14 @@ option go_package = "api/managementpb;managementpb";
 import "google/api/annotations.proto";
 import "managementpb/severity.proto";
 
+// SecurityCheckInterval represents possible execution interval values for checks.
+enum SecurityCheckInterval {
+  SECURITY_CHECK_INTERVAL_INVALID = 0;
+  STANDARD = 1;
+  FREQUENT = 2;
```
oter, line 14: Can we add a short description for each interval, e.g. from where they come?

oter (approved):

api/managementpb/checks.proto
palash25: will add it

palash25: on second thoughts all the intervals come from the same source (the UI in this case) so why do we need a comment for each of them. could you give me example of what you are looking for?

palash25:
sorry for the force push, had to do it, since the conflicts were in a generated code, it got ugly

artemgavrilov (approved):

BupycHuk (approved):

Dasio (approved):

atymchuk (approved):

PR #715, opened 2021-03-29
PMM-7767 Azure monitoring - feature flag

adivinho:
- [ ] https://github.com/Percona-Lab/pmm-submodules/pull/1642


api/serverpb/server.proto
```diff
@@ -232,6 +234,10 @@ message ChangeSettingsRequest {
   bool remove_pmm_public_address = 20;
   // Intervals between STT check runs.
   STTCheckIntervals stt_check_intervals = 21;
+  // Enable Azure Discover.
+  bool enable_azurediscover = 22;
+  // Disable Azure Discover.
+  bool disable_azurediscover = 23;
```
askomorokhov, line 240: Maybe disable_azure_discover?

askomorokhov (approved):

api/serverpb/server.proto
adivinho: IMHO we are not going to have extended azure variables. So it would require to distinguish the third part.
e.g.
disable_azure_parameter1
disable_azure_parameter2

BupycHuk (changes requested):
Should we close this PR since we have https://github.com/percona/pmm/pull/718?

PR #720, opened 2021-04-08
PMM-7847 API for MySQL TLS.

JiriCtvrtka:


api/inventorypb/agents.proto
```diff
@@ -100,6 +100,12 @@ message MySQLdExporter {
   bool tls = 6;
   // Skip TLS certificate and hostname validation.
   bool tls_skip_verify = 7;
+  // Certificate Authority certificate chain.
+  string tls_ca = 15;
+  // Client certificate.
+  string tls_cert = 16;
+  // Password for decrypting tls_cert.
+  string tls_key = 17;
```
askomorokhov, line 108: Looks like we repeated these 3 fields over several massages. Can we extract them to a separate message? 
In this case, we will keep it more DRY and it will be more easer to access it from code as a group?

api/agentpb/agent.proto
```diff
@@ -139,24 +139,32 @@ message StartActionRequest {
...
+    // Contains files and their contents which can be used in DSN.
+    TextFiles text_files = 3;
   }
   // MySQLShowIndexParams describes MySQL SHOW INDEX action parameters.
   message MySQLShowIndexParams {
     // DSN for the service. May contain connection (dial) timeout.
     string dsn = 1;
     string table = 2;
+    // Contains files and their contents which can be used in DSN.
+    TextFiles text_files = 3;
```
askomorokhov, line 167: Maybe better to call it tls_files? This could give a more understanding that inside certificates but not plain text files.
What do you think?

askomorokhov (changes requested):

api/agentpb/agent.proto
JiriCtvrtka: You want to change name everywhere, or just for all actions for mysql? I chose this name because we already using it on different places and in frame of compatibility I chose same name for this field in actions too.

api/inventorypb/agents.proto
JiriCtvrtka: If we will agree on this then no problem, but then it is good change this also for mongodb and RDS, because it is same case like here.

api/agentpb/agent.proto
askomorokhov: Probably only here, it can be a breaking changes

askomorokhov (dismissed):

percona-csalguero (approved):

api/agentpb/agent.proto
JiriCtvrtka: Done

percona-csalguero (dismissed):

askomorokhov (dismissed):

api/agentpb/agent.proto
```diff
@@ -58,6 +58,10 @@ message SetStateRequest {
     repeated string env = 5;
     map<string, string> text_files = 6;
     repeated string redact_words = 7;
+    // TLS in enabled if true.
```
BupycHuk, line 61: these fields actually is not required for processes

api/agentpb/agent.proto
```diff
@@ -139,24 +147,48 @@ message StartActionRequest {
     string dsn = 1;
     string query = 2;
     MysqlExplainOutputFormat output_format = 3;
+    // Contains files and their contents which can be used in DSN.
+    TextFiles tls_files = 4;
+    // TLS in enabled if true.
+    bool tls = 6;
```
BupycHuk, line 153: maybe non-empty tls_files should be enough for check, what do you think?

BupycHuk (changes requested):

api/agentpb/agent.proto
JiriCtvrtka: Yes, removed. Thanks.

api/agentpb/agent.proto
JiriCtvrtka: Good idea. Changed to check textfile instead.

percona-csalguero (approved):

askomorokhov (approved):

BupycHuk (approved):

PR #729, opened 2021-04-19
PMM-5607 Add printing full version in JSON

maxkondr:


atymchuk (approved):

version/release_test.go
```diff
@@ -59,4 +85,29 @@ func TestInfoExporter(t *testing.T) {
 	if expected != actual {
 		t.Errorf("expected: %q\nactual: %q", expected, actual)
 	}
+
+	expectedMap := map[string]string{
```
BupycHuk, line 89: maybe it's better to compare just string outputs, what do you think?

version/release.go
```diff
@@ -75,3 +82,26 @@ func FullInfo() string {
...
+		"FullCommit":  FullCommit,
+	}
+	if Branch != "" {
+		res["Branch"] = Branch
+	}
+
+	b := &bytes.Buffer{}
+	encoder := json.NewEncoder(b)
+
+	if err := encoder.Encode(res); err != nil {
```
BupycHuk, line 102: let's use `json.Marshal()`
and in case of any error `panic` or return the error as the second value.

BupycHuk (changes requested):

version/release.go
maxkondr: done

version/release_test.go
maxkondr: done

version/release_test.go
```diff
@@ -1,62 +1,116 @@
...
 
-	expected = strings.Join([]string{
-		"ProjectName: pmm-managed",
-		"Version: 2.1.2",
-		"PMMVersion: 2.1.2",
-		"Timestamp: 2018-12-19 13:41:48 (UTC)",
-		"FullCommit: 6559a94ab33831deeda04193f74413b735edb1a1",
-		"Branch: master",
+	expected := strings.Join([]string{
+		fmt.Sprintf("ProjectName: %s", ProjectName),
```
BupycHuk, line 41: TBH, I don't like that we use almost the same code for expected result as for the actual code. I would prefer to use strings if possible.

BupycHuk (approved):

palash25 (approved):

version/release_test.go
maxkondr: Only data is the same, need to check that passed data to func and returned are equal. The rest (formatting, Join) is just helpers to construct the expected result. To my mind it improves code maintainability - no need to look through the whole sis and find where to change the strings

PR #737, opened 2021-05-14
PMM-6295 Add UNKNOWN agent status

jprukner:


api/inventorypb/agent_status.dot
```diff
@@ -6,17 +6,29 @@ digraph {
...
 
   WAITING -> STARTING;
   WAITING -> DONE;
+  WAITING -> UNKNOWN; [style=dotted]
 
   STOPPING -> DONE;
+  STOPPING -> UNKNOWN; [style=dotted]
 
   DONE -> STARTING [style=dotted];
+  DONE -> UNKNOWN [style=dotted];
```
palash25, line 27: question: should `DONE` to `UNKNOWN` be a valid transition of states?  once an agent is done it should stay done right? or is it some edge case where an agent was set to `DONE` then maybe lost connection and has to be set to `UNKNOWN` after connection is re-established?

palash25 (approved):

api/inventorypb/agent_status.dot
jprukner: When agents stops gracefully it has status DONE, then pmm-managed stops, we don't know what happens in that period when pmm-managed is down. What if the agent that is DONE starts? So we set all agents' statuses to UNKNOWN at pmm-managed startup, and let agents change them when they reconnect. 
Does it make sense?

BupycHuk: Once agent moved to DONE it shouldn't change status without request from pmm-managed. But there might be race condition that pmm-managed goes down right after sending state request.

BupycHuk (approved):

PR #741, opened 2021-05-24
PMM-7722 Add RPC for getting latest and installed operator versions 

jprukner:
To be able to check if there is a new version available we need to be able to get latest version of operators, also we should return version installed together with operator status as part of response of ListKubernetesClusters RPC.


jprukner:
We need something different actually. We need to check which pmm is compatible with what operator version. Changes to version service needed.

PR #746, opened 2021-06-03
PMM-8172 Add rpc CheckForOperatorUpdate

jprukner:


api/managementpb/dbaas/components.proto
```diff
@@ -102,6 +102,43 @@ message ChangePXCComponentsRequest {
...
+}
+
+message CheckForOperatorUpdateResponse {
+  OperatorUpdateStatus status = 1;
+  // available_operator_version is not empty if there is an update available,
+  // for both compatible and incompatible one.
+  string available_operator_version = 2;
+  // available_pmm_server_version is not empty only if the update is available but
+  // the version of the operator is not compatible with installed PMM version.
+  string available_pmm_server_version = 3;
```
BupycHuk, line 139: I think we should not return available operator version if it's not compatible with installed PMM version at all. And having this field makes no sense for me.

api/managementpb/dbaas/components.proto
```diff
@@ -102,6 +102,43 @@ message ChangePXCComponentsRequest {
...
+  UPDATE_AVAILABLE_BUT_NOT_COMPATIBLE = 3;
+}
+
+message CheckForOperatorUpdateRequest {
+  // Kubernetes cluster name.
+  string kubernetes_cluster_name = 1 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
```
BupycHuk, line 123: When we are going to check for operator update? How often?

BupycHuk: If it's the same as list kubernetes cluster, then we can just extend ListKubernetesCluster.

BupycHuk: Or just send make one endpoint for all clusters to not send N requests to version service.

api/managementpb/dbaas/components.proto
jprukner: We should be able to notify user that there is a new version available but he has to update PMM first.

api/managementpb/dbaas/components.proto
jprukner: Calling version service only once, only one API endpoint to get information about all updates.

api/managementpb/dbaas/components.proto
jprukner: Removed.

api/managementpb/dbaas/components.proto
```diff
@@ -102,6 +102,39 @@ message ChangePXCComponentsRequest {
...
+}
+
+// OperatorsUpdateInformation is just a container for update information for PXC and PSMDB operators.
+message OperatorsUpdateInformation {
+  OperatorUpdateInformation pxc_operator = 1;
+  OperatorUpdateInformation psmdb_operator = 2;
+}
+
+message CheckForOperatorUpdateResponse {
+  // The cluster name is used as a key for this map, value stores information about oprators update availablility.
```
BupycHuk, line 134:
```suggestion
  // The cluster name is used as a key for this map, value stores information about operators update availability.
```

api/managementpb/dbaas/components.proto
```diff
@@ -102,6 +102,39 @@ message ChangePXCComponentsRequest {
 
 message ChangePXCComponentsResponse {}
 
+// OperatorUpdateStatus represents status of an update for the operator.
+enum OperatorUpdateStatus {
+  OPERATOR_UPDATE_STATUS_INVALID = 0;
+  // An update of the operator is available for the installed version of PMM server.
+  UPDATE_AVAILABLE = 1;
+  // An update of the operator is not available for the installed version of PMM server.
+  UPDATE_NOT_AVAILABLE = 2;
```
BupycHuk, line 111:
```suggestion
  OPERATOR_UPDATE_STATUS_AVAILABLE = 1;
  // An update of the operator is not available for the installed version of PMM server.
  OPERATOR_UPDATE_STATUS_NOT_AVAILABLE = 2;
```

api/managementpb/dbaas/components.proto
```diff
@@ -102,6 +102,39 @@ message ChangePXCComponentsRequest {
...
+}
+
+message CheckForOperatorUpdateRequest {}
+
+// OperatorUpdateInformation indicates if operator update is available. It includes what version of operator we can update to.
+// If we are running latest operator for installed PMM version and more up-to-date operator is available, we also get PMM
+// version we can update to to get the most up-to-date operator.
```
BupycHuk, line 118: let's update this comment

BupycHuk (changes requested):

api/managementpb/dbaas/components.proto
```diff
@@ -102,6 +102,39 @@ message ChangePXCComponentsRequest {
...
+message OperatorUpdateInformation {
+  // The status indicates operator status availablility.
+  OperatorUpdateStatus status = 1;
+  // available_operator_version is not empty if there is an update available,
+  // for both compatible and incompatible one.
+  string available_operator_version = 2;
+}
```
BupycHuk, line 125: Feels like we will be able to reuse it later for DBs. Maybe it's better to make it somehow common?
```suggestion
message ComponentUpdateInformation {
  // The status indicates operator status availablility.
  ComponentUpdateStatus status = 1;
  // available_component_version is not empty if there is an update available.
  string available_component_version = 2;
}
```
Do we really need `ComponentUpdateStatus` if `available_component_version` is not empty only when update is available?

jprukner: I removed it.

jprukner: For the common thing: Since I removed status, I could place operator specific fields right into OperatorUpdateInformation. One way to generalize it would be to have map with key as operator type and value available version. what do you think @BupycHuk ?

BupycHuk: It can work, just to not forget to mention it in comments.

BupycHuk: But since we are going to add ability to notify later, maybe it's better to keep the message

jprukner: Let's solve this when we actually decide to show updates to incompatible versions are available.

jprukner: I changed the implementation to use maps, adding postresql operator will result in a little less pain in this case.

BupycHuk: the problem with this approach that we won't be able to change already made API.

jprukner: Why not? As long as we change FE and BE in the same release, it should be fine, shouldn't it?

jprukner: I added the one field message, so we are backwards compatible.

BupycHuk (approved):

BupycHuk (approved):

palash25 (approved):

PR #748, opened 2021-06-07
PMM-8155 Backup scheduling

Dasio:
https://jira.percona.com/browse/PMM-8155

 - not sure if these endpoints should be in `backups.proto`
 - API doesn't allow edit service and location

PMM-managed: https://github.com/percona/pmm-managed/pull/778
API tests: https://github.com/Percona-Lab/pmm-api-tests/pull/221


BupycHuk (approved):

fabio-silva (commented):
Thanks @Dasio . Some questions:
- We have retry mode / interval / times here, but not for regular backups. Are we opening them there too, or just here for now?
- What's the process to toggle (activate/deactivate) a scheduled backup?
- In Figma we have a "Full logs" checkbox. Is that out of scope for now?

Dasio:
1. Just for scheduled ones. But as I can see, it should be also for on-demand. Thanks for pointing out, we need to do some refactor. But let's keep on-demand retries out-of-scope for now.
2.  https://github.com/percona/pmm/pull/748/files#diff-9b84b387462cf392f9f4cdd59886910c154081e8a5cf1e9f1b0812ae04744aa2R132
So send ID + flag true/false
3. I think so. Right now we don't have any logs anyway.

fabio-silva:
> 1. Just for scheduled ones. But as I can see, it should be also for on-demand. Thanks for pointing out, we need to do some refactor. But let's keep on-demand retries out-of-scope for now.
> 2. https://github.com/percona/pmm/pull/748/files#diff-9b84b387462cf392f9f4cdd59886910c154081e8a5cf1e9f1b0812ae04744aa2R132
>    So send ID + flag true/false
> 3. I think so. Right now we don't have any logs anyway.

Got it, thanks. Other points:
- the on-demand backups list is returning service_name and location_name, which are displayed to the user;
- what about data_model and vendor ?

api/swagger/swagger-dev.json
```diff
@@ -24479,6 +24479,376 @@
...
+                        "enum": [
+                          "RETRY_MODE_INVALID",
+                          "AUTO",
+                          "MANUAL"
+                        ],
+                        "x-order": 9
+                      },
+                      "retry_interval": {
+                        "description": "Delay between each retry.",
+                        "type": "string",
```
fabio-silva, line 24689: why is this a string?

Dasio: I used `google.protobuf.Duration` as it used also on other places. And it needs more bits than JSON int can represent.
```
// # JSON Mapping
//
// In JSON format, the Duration type is encoded as a string rather than an
// object, where the string ends in the suffix "s" (indicating seconds) and
// is preceded by the number of seconds, with nanoseconds expressed as
// fractional seconds. For example, 3 seconds with 0 nanoseconds should be
// encoded in JSON format as "3s", while 3 seconds and 1 nanosecond should
// be expressed in JSON format as "3.000000001s", and 3 seconds and 1
// microsecond should be expressed in JSON format as "3.000001s".
```

fabio-silva: But will the API return only "s", or can it return "s", "m" and "h"?

api/managementpb/backup/backups.proto
```diff
@@ -58,12 +62,130 @@ message RestoreBackupResponse {
...
+  // Machine-readable location ID.
+  string location_id = 4;
+  // Location name.
+  string location_name = 5;
+  // How often backup will be run in cron format.
+  string cron_expression = 6;
+  // First backup wouldn't happen before this time.
+  google.protobuf.Timestamp start_time = 7;
+  // Artifact name.
+  string name = 8;
```
oter, line 82: probably we should have a pattern here, so every new artifact will have a different name.

oter (approved):

api/managementpb/backup/backups.proto
Dasio: for now, i want just add timestamp after this name. Still we didn't had discussion how to handle this "autogenerate name"

api/swagger/swagger-dev.json
Dasio: I think it should return only "s" based on that description

api/managementpb/backup/backups.proto
```diff
@@ -58,12 +62,130 @@ message RestoreBackupResponse {
...
+  // Description.
+  string description = 9;
+  // Retry mode.
+  RetryMode retry_mode = 10;
+  // Delay between each retry. Should have a suffix in JSON: 1s, 1m, 1h.
+  google.protobuf.Duration retry_interval = 11;
+  // How many times to retry a failed backup before giving up.
+  uint32 retry_times = 12;
+  // If scheduling is enabled.
+  bool enabled = 13;
```
nicolalamacchia, line 92: 🙏

nicolalamacchia (approved):

artemgavrilov (approved):

api/managementpb/backup/backups.proto
```diff
@@ -58,12 +62,130 @@ message RestoreBackupResponse {
...
+    }
+  ];
+  // First backup wouldn't happen before this time.
+  google.protobuf.Timestamp start_time = 4;
+  // Name of backup.
+  string name = 5;
+  // Human-readable description.
+  string description = 6;
+  // Retry mode.
+  RetryMode retry_mode = 7;
```
Dasio, line 129: I think we can discard this field .
Retry_times == 0 => Retry Mode = MANUAL
Retry_times > 0 => Retry Mode = AUTO

Whata do you think @atymchuk

api/managementpb/backup/backups.proto
```diff
@@ -58,12 +61,127 @@ message RestoreBackupResponse {
...
+  ];
+  google.protobuf.BoolValue enabled = 2;
+  // How often backup should be run in cron format.
+  google.protobuf.StringValue cron_expression = 3;
+  // First backup wouldn't happen before this time.
+  google.protobuf.Timestamp start_time = 4;
+  // Name of backup.
+  google.protobuf.StringValue name = 5;
+  // Human-readable description.
+  google.protobuf.StringValue description = 6;
```
artemgavrilov, line 159: Why these fields are wrapped? Why not just `string`?

api/managementpb/backup/backups.proto
```diff
@@ -58,12 +61,127 @@ message RestoreBackupResponse {
...
+
+message RemoveScheduledBackupResponse {}
+
 // Backups service handles backup operations to DB.
 service Backups {
   // StartBackup request backup specified service to location.
   rpc StartBackup(StartBackupRequest) returns (StartBackupResponse) {
     option (google.api.http) = {
-      post: "/v1/management/backup/Backups/StartBackup"
+      post: "/v1/management/backup/Backups/Start"
```
artemgavrilov, line 184: 👍

api/managementpb/backup/backups.proto
```diff
@@ -74,4 +192,32 @@ service Backups {
       body: "*"
     };
   }
+  // ScheduleBackup schedules repeated backup.
+  rpc ScheduleBackup(ScheduleBackupRequest) returns (ScheduleBackupResponse) {
+    option (google.api.http) = {
+      post: "/v1/management/backup/Backups/Schedule"
```
artemgavrilov, line 198: For consistency it probably should be `/v1/management/backup/Backups/AddScheduled`

artemgavrilov (approved):

api/managementpb/backup/backups.proto
oter: Probably we allow empty description, so for the ability to distinguish if the field needs to be updated or not.

oter (approved):

api/managementpb/backup/backups.proto
artemgavrilov: I don't remember well discussion regarding this, but we use a different approach in other APIs where user should send whole object on `Change` action.

Dasio: I can't remember when, but we had several times discussion about this, also with @BupycHuk, @nicolalamacchia 
 and otehrs. And we agreed to try this approach for new API.

api/managementpb/backup/backups.proto
Dasio: We removed retries entirely for now.

api/managementpb/backup/backups.proto
artemgavrilov: Ok then  👌

api/managementpb/backup/backups.proto
Dasio: @atymchuk ^what do you think?

Hlmn-Rmr: I would go with either `Schedule` or `CreateScheduled`. Agree with @artemgavrilov, consistency is important, what's the equivalent endpoint for creating an on-deman/ad-hoc backup?

api/managementpb/backup/backups.proto
atymchuk: 💯

api/managementpb/backup/backups.proto
atymchuk: As I can see it's the `Add` verb which is used everywhere. 
On consistency:
To me, consistency is a trap in the sense that once you land a sub-optimal term or verb, you have to stick to it forever. Which is wrong, because we want people to be able to read the code and understand it.

That said, IMO the best semantics here is "As a System, I'm scheduling a backup" and therefore David's version of "/v1/management/backup/Backups/Schedule" seems to be just OK.

atymchuk (approved):

oter (approved):

api/managementpb/backup/backups.proto
oter: @fabio-silva we should update the endpoint on the FE side.

fabio-silva: Yes @oter , did it already, thanks!

nicolalamacchia (approved):

PR #749, opened 2021-06-07
PMM-7888 Added TLS fields for PostgreSQL

percona-csalguero:
https://jira.percona.com/browse/PMM-7888

FB: https://github.com/Percona-Lab/pmm-submodules/pull/1898


api/managementpb/postgresql.proto
```diff
@@ -78,6 +78,12 @@ message AddPostgreSQLRequest {
   MetricsMode metrics_mode = 21;
   // List of collector names to disable in this exporter.
   repeated string disable_collectors = 22;
+  // TLS CA certificate.
```
JiriCtvrtka, line 81: I think same fields should be also in inventory for postgresql.
https://github.com/percona/pmm/blob/PMM-2.0/api/inventorypb/agents.proto#L754:L754

askomorokhov:
CI fails, looks like we need to tidy go mod

api/inventorypb/services.proto
```diff
@@ -314,6 +314,16 @@ message AddPostgreSQLServiceRequest {
...
+  // TLS enabled
+  bool tls = 10;
+  // Skip TLS certificates validation
+  bool tls_skip_verify = 11;
+  // TLS CA certificate.
+  string tls_ca = 12;
+  // TLS Certifcate.
+  string tls_cert = 13;
+  // TLS Certificate Key.
+  string tls_key = 14;
```
BupycHuk, line 326: CRITICAL: these parameters should be for agents, not services.

api/managementpb/postgresql.proto
```diff
@@ -78,6 +78,12 @@ message AddPostgreSQLRequest {
...
+  string tls_ca = 23;
+  // TLS Certifcate.
+  string tls_cert = 24;
+  // TLS Certificate Key.
+  string tls_key = 25;
```
BupycHuk, line 86: ok, here we have a problem 😄 we were saying that we use same naming that's used on upstream, but looks like it was false for PostgreSQL, because upstream uses SSL everywhere, but we use TLS wording. Any ideas how can we fix this problem, without loosing backward compatibility?

BupycHuk (changes requested):

api/managementpb/postgresql.proto
JiriCtvrtka: Same fields should be presented for actions and connection check.

percona-csalguero: > Same fields should be presented for actions and connection check.

Where?

api/inventorypb/agents.proto
```diff
@@ -182,14 +182,16 @@ message PostgresExporter {
...
   // Actual Agent status.
   AgentStatus status = 9;
   // Listen port for scraping metrics.
   uint32 listen_port = 10;
+  // TLS CA certificate.
+  string tls_ca = 13;
+  // TLS Certifcate.
+  string tls_cert = 14;
+  // TLS Certificate Key.
+  string tls_key = 15;
```
BupycHuk, line 194: we shouldn't return them.

BupycHuk (changes requested):

askomorokhov (approved):

JiriCtvrtka (dismissed):

BupycHuk (dismissed):

BupycHuk (approved):

askomorokhov (approved):

PR #752, opened 2021-06-09
PMM-8174 Update operators API

jprukner:
This PR adds RPC for installing operators in given version.


api/managementpb/dbaas/components.proto
```diff
@@ -102,6 +103,59 @@ message ChangePXCComponentsRequest {
...
+  string operator_type = 2 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+}
+
+message InstallLatestOperatorResponse {
+  // Status tells if installation was successfull.
+  OperatorsStatus status = 1;
```
askomorokhov, line 156: Do we need the operator's type here as well - or we will know this from context?

api/managementpb/dbaas/components.proto
```diff
@@ -102,6 +103,59 @@ message ChangePXCComponentsRequest {
...
+// OperatorsUpdateInformation is just a container for update information for PXC and PSMDB operators.
+message OperatorsUpdateInformation {
+  OperatorUpdateInformation pxc_operator = 1;
+  OperatorUpdateInformation psmdb_operator = 2;
+}
+
+message CheckForOperatorUpdateResponse {
+  // The cluster name is used as a key for this map, value stores information about oprators update availablility.
+  map<string, OperatorsUpdateInformation> update_information = 1;
+}
```
askomorokhov, line 137: For PMM update we have such field as the last check timestamp. Maybe it will be useful here for UX?

askomorokhov (commented):
> installing latest version

Sounds reasonable to me.

Update operator API in general looks good to me.

api/managementpb/dbaas/components.proto
jprukner: I think we will know this from context.

api/managementpb/dbaas/components.proto
jprukner: We decided to go with most simple solution - FE calling this RPC every N days. So it should be stored on FE. :)

jprukner:
Thanks for review @askomorokhov :+1:

palash25 (dismissed):

BupycHuk (dismissed):
Looks good to me, but I'm not sure where endpoint should be: in components.proto or in kubernetes.proto?

jprukner:
I screwed this up. Since the reviews got dismissed anyway

jprukner:
> Looks good to me, but I'm not sure where endpoint should be: in components.proto or in kubernetes.proto?

I think we should keep kubernetes service to be related only to kubernetes cluster. Operators are in the end components of DBaaS.

BupycHuk (approved):

palash25 (approved):

api/managementpb/dbaas/components.proto
```diff
@@ -102,6 +103,32 @@ message ChangePXCComponentsRequest {
...
 
+message InstallOperatorRequest {
+  // Kubernetes cluster name.
+  string kubernetes_cluster_name = 1 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  // operator_type tells what operator we are interested in updating.
+  string operator_type = 2 [
```
BupycHuk, line 114: Can we use enum here?

jprukner: @BupycHuk Sorry, did not notice it until now. It would require changes in FE, too and Tiago is on vacation. Can we add this later?

BupycHuk: I'm ok if it will be fixed before 2.21.0 RC

BupycHuk: Please create tech debt task and set fix version 2.21

jprukner: https://jira.percona.com/browse/PMM-8532

BupycHuk (approved):

palash25 (approved):

PR #755, opened 2021-06-17
PMM-8160: List compatible services

oter:
https://jira.percona.com/browse/PMM-8160


BupycHuk (changes requested):
Why do we need version of LocalMySQL and RemoteMySQL?

oter:
> Why do we need version of LocalMySQL and RemoteMySQL?

just realised I need only local MySQL check, will remove the second one.

Dasio (approved):

.gitignore
```diff
@@ -5,3 +5,6 @@
 /api/nginx/*.pem
 /bin
 /third_party
+
+# System Files
+.DS_Store
```
artemgavrilov, line 10: I would say it's better to have it in the first section and in the same style.

api/agentpb/agent.proto
```diff
@@ -534,6 +534,25 @@ message JobProgress {
   }
 }
 
+// GetVersionRequest represents a request for version retrieving of different software.
+message GetVersionRequest {
+  // LocalMySQL is used for local MySQL version retrieving using `mysql --version`.
+  message LocalMySQL {}
+
+  // XtraBackup is used for xtrabackup version retrieving.
+  message XtraBackup {}
```
artemgavrilov, line 543: Can we use enums instead of empty messages? Or maybe we can just return all software versions that we have at once? I mean:
```
message GetVersionsRequest {}
message GetVersionsResponse {
     string mysql_version = 1;
     xtrabackup_version = 2;
     ...
}
```

artemgavrilov: Or we can pass a list of software that we need and get an array of versions as result.

oter: Yeah, enums could work in my case, but later we probably want to add a mongodb, and I suppose it will require some credentials, so the message suits better. Not sure we can return all the software versions at once, because some hosts could have MySQL only installed, others mongodb etc.

.gitignore
oter: Not sure I'm following regarding style.

api/agentpb/agent.proto
artemgavrilov: Why mongodb will require credentials? The version check for it will differ from `mongod --version`?

My point that typically we need versions of some set of software (mongod + pbm, MySQL + xtrabackup + xtracloud, etc), so it's better to grab it In one request.


>  Not sure we can return all the software versions at once, because some hosts could have MySQL only installed, others mongodb etc.
Why it's the problem? If agent has only MySQL then it will return only MySQL version.

.gitignore
oter: As per our call `.DS_Store` could be located in the subfolders, so leaving as is.

api/agentpb/agent.proto
oter: As per our call in case of logical backup, we need to pass DB creds, so no better solution at the moment.

api/agentpb/agent.proto
```diff
@@ -534,6 +534,25 @@ message JobProgress {
   }
 }
 
+// GetVersionRequest represents a request for version retrieving of different software.
+message GetVersionRequest {
```
BupycHuk, line 538: I would suggest to make it return all versions in one response instead of one per request, what do you think?

api/managementpb/backup/backups.proto
```diff
@@ -38,6 +39,19 @@ message StartBackupResponse {
   string artifact_id = 1;
 }
 
+message ListServicesForRestoreRequest {
+  // Artifact id used for for determining restore compatability.
```
BupycHuk, line 43:
```suggestion
  // Artifact id used to determine restore compatibility.
```

api/agentpb/agent.proto
artemgavrilov: To check the version of remote MongoDB we should pass DB creds. We couldn't find a better way, maybe you have some ideas?

BupycHuk: We can send list of messages. each message will have the same format that we use for `CheckConnectionRequest`
```
 // Service type.
  inventory.ServiceType type = 1;
  // DSN for the service. May contain connection (dial) timeout.
  string dsn = 2;
  // Timeout for the whole request.
  google.protobuf.Duration timeout = 3;
  // Contains files and their contents which can be used in DSN.
  TextFiles text_files = 4;
  // TLS certificate wont be verified.
  bool tls_skip_verify = 5;
```

And instead of checking for version right before restore we can do it periodically and store version in DB.
What do you think?

oter: we should guarantee the version of the db at the time of making a backup. That could be the issue in the case of periodical checking.

artemgavrilov: In general, I prefer to have one request to fetch all versions instead of multiple requests.

Dasio (approved):

artemgavrilov (approved):

BupycHuk (approved):

api/managementpb/backup/backups.proto
```diff
@@ -191,6 +206,13 @@ service Backups {
       body: "*"
     };
   }
+  // ListServicesForRestore lists compatible services for restoring a backup.
+  rpc ListServicesForRestore(ListServicesForRestoreRequest) returns (ListServicesForRestoreResponse) {
```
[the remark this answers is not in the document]
oter: I'm thinking about ListCompatibleServices, what do you think?

api/managementpb/backup/backups.proto
```diff
@@ -41,6 +42,20 @@ message StartBackupResponse {
...
+  // Artifact id used to determine restore compatibility.
+  string artifact_id = 1 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+}
+
+message ListArtifactCompatibleServicesResponse {
+  repeated inventory.MySQLService mysql = 1;
```
Dasio, line 55: don't we need only one type of service? so it'll be better oneof, than everything in one message?

oter: sounds reasonable, will try

oter: so I cannot put repeated in the oneof 😞

Dasio (dismissed):

Dasio (approved):

artemgavrilov (approved):

BupycHuk (approved):

PR #756, opened 2021-06-21
PMM-8278 add sha256 hash for one line script

nikita-b:


atymchuk:
@puneet0191 Can we cover this with a test? I was thinking about bats, but l leave it up to you.

atymchuk (approved):

PR #758, opened 2021-06-30
PMM-8322: Delete artifact from storage

oter:
https://jira.percona.com/browse/PMM-8322


BupycHuk (approved):

api/managementpb/backup/artifacts.proto
```diff
@@ -57,4 +66,11 @@ service Artifacts {
       body: "*"
     };
   }
+  // DeleteArtifact deletes specified artifact.
+  rpc DeleteArtifact(DeleteArtifactRequest) returns (DeleteArtifactResponse) {
+    option (google.api.http) = {
+      post: "/v1/management/backup/Artifacts/Remove"
```
artemgavrilov, line 72: Why `Remove` if gRPC method is named `DeleteArtifact`?

oter: fixed, thanks

artemgavrilov (dismissed):

api/managementpb/backup/artifacts.proto
```diff
@@ -16,6 +16,8 @@ enum BackupStatus {
   BACKUP_STATUS_PAUSED = 3;
   BACKUP_STATUS_SUCCESS = 4;
   BACKUP_STATUS_ERROR = 5;
+  BACKUP_STATUS_DELETING = 6;
```
oter, line 19: @fabio-silva we need to handle these statuses on the FE side

BupycHuk (approved):

Dasio (approved):

api/managementpb/backup/artifacts.proto
```diff
@@ -16,6 +16,8 @@ enum BackupStatus {
...
+  BACKUP_STATUS_FAILED_TO_DELETE = 7;
```
artemgavrilov, line 20: How we should handle this status? How artifact can be moved to another state from this status?

oter: I'm setting `BACKUP_STATUS_DELETING` at the beginning and then I'm setting `BACKUP_STATUS_FAILED_TO_DELETE ` if for some reason removal from S3 has failed, or removing db entries on success. In case of `BACKUP_STATUS_FAILED_TO_DELETE` user can retry removing.

artemgavrilov: If you set `BACKUP_STATUS_DELETING` and the server was restarted, on startup server will try to finish removal? Or user able to retry removal?

artemgavrilov (approved):

PR #762, opened 2021-07-08
PMM-8219 Blank PMM version in offline mode.

JiriCtvrtka:
pmm-managed part: https://github.com/percona/pmm-managed/pull/797
FB: https://github.com/Percona-Lab/pmm-submodules/pull/1899


api/serverpb/server.proto
```diff
@@ -68,6 +70,10 @@ message CheckUpdatesResponse {
   string latest_news_url = 4;
   // Last check time.
   google.protobuf.Timestamp last_check = 5;
+  // If true, PMM is working in offline mode.
+  bool offline_mode = 6;
+  // Contains error message during request, if any
+  string error_message = 7;
```
BupycHuk, line 76: I think we don't need these two fields

askomorokhov: This may be useful from a UX point of view - to display some icons - no internet connection.
What do you think?

BupycHuk: So, I think it will work this way:
Frontend requests update
Backend returns error
Frontend shows error icon
Frontend requests update with flag `only_installed`

BupycHuk (approved):

askomorokhov (approved):

PR #768, opened 2021-07-21
PMM-7914 DBaaS: update database

jprukner:
I edited database cluster listing RPC response to include available version for each db cluster.
I added RPC for triggering database version installation.


api/managementpb/dbaas/psmdb_cluster.proto
```diff
@@ -77,6 +77,8 @@ message ListPSMDBClustersResponse {
     PSMDBClusterParams params = 4;
     // DB cluster accessible outside of K8s cluster.
     bool exposed = 5;
+    // Available version is version of database use can update to. If it's empty, no update is available.
```
jprukner, line 80: a version ... one can update to.

api/managementpb/dbaas/components.proto
```diff
@@ -158,4 +189,11 @@ service Components {
       body: "*"
     };
   }
+  // InstallDatabase triggers installation of given version of database. It does not block to wait for it's instalation to finish.
+  rpc InstallDatabase(InstallDatabaseRequest) returns (InstallDatabaseResponse) {
```
BupycHuk, line 193: UpgradeDatabaseVersion?

api/managementpb/dbaas/components.proto
```diff
@@ -121,6 +121,37 @@ message CheckForOperatorUpdateResponse {
...
+    }
+  ];
+  // Database type is either "pxc" or "psmdb".
+  string database_type = 2 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  // Database cluster name.
+  string cluster_name = 3 [
```
BupycHuk, line 138: database_cluster_name

api/managementpb/dbaas/components.proto
```diff
@@ -121,6 +121,37 @@ message CheckForOperatorUpdateResponse {
...
 
+message InstallDatabaseRequest {
+  // Kubernetes cluster name.
+  string kubernetes_cluster_name = 1 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  // Database type is either "pxc" or "psmdb".
+  string database_type = 2 [
```
BupycHuk, line 132: let's use enum everywhere we use it.

api/managementpb/dbaas/components.proto
```diff
@@ -121,6 +121,37 @@ message CheckForOperatorUpdateResponse {
...
+    }
+  ];
+  // Database cluster name.
+  string cluster_name = 3 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  // Database version to install.
+  string version = 4 [
```
BupycHuk, line 144: db_version

BupycHuk (changes requested):

api/managementpb/dbaas/components.proto
```diff
@@ -9,6 +9,15 @@ import "google/api/annotations.proto";
 
 // TODO move it to parent directory / package once API is v1-stable
 
+// DatabaseType contains databases supported by DBaaS.
+enum DatabaseType {
+  DATABASE_TYPE_INVALID = 0;
```
jprukner, line 14: We don't actually need that. What we need is enum with operators. Because we already use operator_type and each operator has one and only database type.

jprukner: TODO CHANGE THIS

jprukner:
How about not introducing a new RPC but reusing already present ones @BupycHuk ?

BupycHuk:
@jprukner I'm ok with that :) But we have one thing to not forget:
For creation we use image, but for update we are going to use db_version. So it's better to use the same thing in both places.

jprukner:
@BupycHuk I included image into params so we are consistent. If the given image is the same except tag, we trigger update to the version given in the image tag, return error otherwise.

BupycHuk (dismissed):

JiriCtvrtka (dismissed):

JiriCtvrtka (approved):

JiriCtvrtka (approved):

BupycHuk (approved):

palash25 (approved):

PR #769, opened 2021-07-21
PMM-8269 MongoDB Point-in-Time backups

artemgavrilov:
[PMM-8269](https://jira.percona.com/browse/PMM-8269)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/2001
---
- [ ] https://github.com/percona/pmm-managed/pull/811
- [ ] https://github.com/percona/pmm-agent/pull/285
---


api/managementpb/backup/backups.proto
```diff
@@ -176,6 +187,17 @@ message RemoveScheduledBackupRequest {
 
 message RemoveScheduledBackupResponse {}
 
+message GetMongoPITRRangesRequest {}
```
Dasio, line 190: shouldn't be there something like service id?

api/agentpb/agent.proto
```diff
@@ -283,6 +283,24 @@ message StartActionRequest {
     // Contains files and their contents which can be used in DSN.
     TextFiles text_files = 2;
   }
+  // PBMStatusParams describes pbm status action parameters.
+  message PBMStatusParams {
+    // DSN for the service. May contain connection (dial) timeout.
+    // May contain placeholders for file paths in DSN.
+    string dsn = 1;
+    // Contains files and their contents which can be used in DSN.
+    TextFiles text_files = 2;
```
Dasio, line 292: what is this? why we need some textfiles?

artemgavrilov: That's for TLS certificates. Same field is present in all Mongo-related requests.

api/managementpb/backup/backups.proto
artemgavrilov: I will remove that method for now. There will separate ticket for that.

fabio-silva (approved):
Seems ok from my side. Just one question for the UI: is every data_model / mode combination ok, or is there any constraint?

fabio-silva (changes requested):
Also, aren't we missing `mode` on `ListScheduled` call?

fabio-silva:
One last question: when we support MySQL, what's gonna be the `mode` there? Are we gonna use `INCREMENTAL` as a generic term for PITR / Incremental?

artemgavrilov:
> One last question: when we support MySQL, what's gonna be the `mode` there? Are we gonna use `INCREMENTAL` as a generic term for PITR / Incremental?

Yes, I think so.

Dasio (dismissed):

fabio-silva (dismissed):

api/managementpb/backup/common.proto
```diff
@@ -10,3 +10,10 @@ enum DataModel {
   PHYSICAL = 1;
   LOGICAL = 2;
 }
+
+// BackupMode specifies backup mode.
+enum BackupMode {
+  BACKUP_MODE_INVALID = 0;
+  SNAPSHOT = 1;
+  INCREMENTAL = 2;
```
atymchuk, line 18: Can we add PITR here as well?
The thing is, SQL engines differentiate between `INCREMENTAL` and `PITR`, so it is by far not the same. If that is true, it'd make sense to have them all here.

atymchuk (changes requested):

api/managementpb/backup/common.proto
artemgavrilov: I thought about it, but then checked `pbm` documentation and discovered that there PITR also called "Incremental backups"

https://www.percona.com/doc/percona-backup-mongodb/point-in-time-recovery.html#incremental-backups

artemgavrilov: On the other hand, it seems that for PG it's different things: https://pgdash.io/blog/postgres-incremental-backup-recovery.html


So yep, you're right. Will add.

artemgavrilov: Done

atymchuk (approved):

fabio-silva (approved):

Dasio (approved):

oter (approved):

BupycHuk (approved):

oter (dismissed):

BupycHuk (dismissed):

BupycHuk (approved):

fabio-silva (approved):

PR #775, opened 2021-08-24
PMM-8673 mongodb backup logs

Dasio:
https://jira.percona.com/browse/PMM-8673

- [ ] https://github.com/percona/pmm-agent/pull/289
- [ ] https://github.com/percona/pmm-managed/pull/822


api/managementpb/backup/backups.proto
```diff
@@ -177,6 +177,27 @@ message RemoveScheduledBackupRequest {
 
 message RemoveScheduledBackupResponse {}
 
+message GetLogsRequest {
+  string artifact_id = 1 [
```
Dasio, line 181: probably this should be optional, and we can later add restore_id also.

api/managementpb/backup/backups.proto
```diff
@@ -177,6 +177,27 @@ message RemoveScheduledBackupRequest {
...
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  uint32 from_chunk = 2;
```
oter, line 186: maybe `chunk_offset`?

api/agentpb/agent.proto
```diff
@@ -526,11 +526,18 @@ message JobProgress {
   // MySQLRestoreBackup contains restore backup job status update.
   message MySQLRestoreBackup {}
 
+  // Logs contains generic logs from job.
+  message Logs {
+    uint32 chunk_id = 1;
+    bytes message = 2;
```
oter, line 532: maybe better string? Since we are going to display them right to the user.

api/managementpb/backup/backups.proto
oter: or even offset

api/managementpb/backup/backups.proto
```diff
@@ -177,6 +177,27 @@ message RemoveScheduledBackupRequest {
...
+
+message GetLogsResponse {
+  repeated LogChunk logs = 1;
+}
+
+// LogChunk represent one chunk of logs.
+message LogChunk {
+  uint32 chunk_id = 1;
+  string message = 2;
+  bool last_chunk = 3;
```
oter, line 198: how you will detect that returned log chunk is the last one? Maybe better to move this field to just GetLogsResponse?

Dasio: pmm-agent, after streaming logs is finished, will send special message that logs are finished and pmm-managed will set this flag for last chunk

api/managementpb/backup/backups.proto
artemgavrilov: Please note that one artifact can have multiple jobs, for example PITR. What we should return in that case? Only last job logs, or all logs for this artifact?

fabio-silva:
@Dasio Wouldn't it be helpful to add a timestamp to the log chunk?

api/managementpb/backup/backups.proto
Dasio: I wasn't aware of this.
Maybe just last job? Not sure

api/managementpb/backup/backups.proto
Dasio: > @Dasio Wouldn't it be helpful to add a timestamp to the log chunk?

What if we don't have timestamp available? (eg. in xtrabackup-mysql). In MongoDB it's inside message so we have this information.

Dasio:
> @Dasio Wouldn't it be helpful to add a timestamp to the log chunk?

What if we don't have timestamp available? (eg. in xtrabackup-mysql). In MongoDB it's inside message so we have this information.

api/managementpb/backup/backups.proto
Dasio: `hasNext` but what if we don't know yet it has next or not. We just know that it maybe it's not end. Then after some time (pmm-agent sends done message), there will be no logs. (In pmm-managed I set flag for last chunk)

fabio-silva:
> > @Dasio Wouldn't it be helpful to add a timestamp to the log chunk?
> 
> What if we don't have timestamp available? (eg. in xtrabackup-mysql). In MongoDB it's inside message so we have this information.

If it's within the message, I guess we could live with that for now. I am assuming the chunks are ordered, right? I.e. the last chunk is the most recent event

Dasio:
Probably I can add timestamp
- pbm - available
- xtrabackup - time when we processed that line

But what will be use case?
Chunks are already ordered.

In case of xtrabackup, timestamp is not part of message.

fabio-silva:
> Probably I can add timestamp
> 
> * pbm - available
> * xtrabackup - time when we processed that line
> 
> But what will be use case?
> Chunks are already ordered.
> 
> In case of xtrabackup, timestamp is not part of message.

If it's easy enough, I think it might be helpful to know when some event happened. We might even consider adding it to the UI, now or later.

fabio-silva:
Another question: is `chunk_id` the chunk order number (0, 1, 2,...), or is it an actual random ID?

Dasio:
ChunkID is sequence so (0,1,2..)
Regarding timestamp. It should be on chunk and it would be time we actually parsed it.
But in case of pbm, we got time for each line.

api/managementpb/backup/backups.proto
Dasio: added `end` as it is inside GetLogsResponse

artemgavrilov (dismissed):

api/managementpb/backup/backups.proto
```diff
@@ -177,6 +177,28 @@ message RemoveScheduledBackupRequest {
...
+
+message GetLogsResponse {
+  repeated LogChunk logs = 1;
+  bool end = 2;
+}
+
+// LogChunk represent one chunk of logs.
+message LogChunk {
+  uint32 chunk_id = 1;
+  string message = 2;
```
artemgavrilov, line 198: Something like `text` or `data` is probably more suitable name

oter (approved):

api/agentpb/agent.proto
```diff
@@ -526,11 +526,19 @@ message JobProgress {
   // MySQLRestoreBackup contains restore backup job status update.
   message MySQLRestoreBackup {}
 
+  // Logs contains generic logs from job.
+  message Logs {
+    uint32 chunk_id = 1;
+    string message = 2;
+    bool done = 3;
+    google.protobuf.Timestamp time = 4;
```
artemgavrilov, line 534: Why do we need a timestamp here? It's time where we collected logs, but how we can use this information?

artemgavrilov: By the way, we have timestamp on `JobProgress` message level.

Dasio: Reflected this comment
https://github.com/percona/pmm/pull/775#issuecomment-919083768
for now, it's not used

artemgavrilov: But do you have a potential use case for that? We know message timestamp and logs have their own timestamp. I can't find a scenario when we need time when we collected logs.

It's easy to add this field later, if there are no use-cases for now then I think it's better to remove it.

api/managementpb/backup/backups.proto
artemgavrilov: Anyway, it's about implementation on managed side. So I resolve this conversation.

oter (approved):

fabio-silva:
Quick question: shouldn't we set a `total_chunks` in the response, after we know it? Otherwise, API consumers can't go to the end of the logs, unless they send `offset=0` and a huge `limit`. If we return the number of chunks, consumers can just send `offset=total_chunks - limit` and get the last `limit` chunks

oter (approved):

Dasio:
> Quick question: shouldn't we set a `total_chunks` in the response, after we know it? Otherwise, API consumers can't go to the end of the logs, unless they send `offset=0` and a huge `limit`. If we return the number of chunks, consumers can just send `offset=total_chunks - limit` and get the last `limit` chunks
> 
> Also, by doing this, `end` becomes redundant: if we get `total_chunks`, we know end happens if `chunk_id = total_chunks - 1`

Usually by "cursor-paging" user can't go to last page. But in this case, we probably can select max chunk id and return it.
Anyway you suggest send `total_chunks` only if it's already end? IDK.
@artemgavrilov @oter any suggestions? Or leave it and improve it later.

fabio-silva:
> > Quick question: shouldn't we set a `total_chunks` in the response, after we know it? Otherwise, API consumers can't go to the end of the logs, unless they send `offset=0` and a huge `limit`. If we return the number of chunks, consumers can just send `offset=total_chunks - limit` and get the last `limit` chunks
> > Also, by doing this, `end` becomes redundant: if we get `total_chunks`, we know end happens if `chunk_id = total_chunks - 1`
> 
> Usually by "cursor-paging" user can't go to last page. But in this case, we probably can select max chunk id and return it.
> Anyway you suggest send `total_chunks` only if it's already end? IDK.
> @artemgavrilov @oter any suggestions? Or leave it and improve it later.

I guess sending `max_chunk_id` works as fine, in my opinion.
I mean always sending it as soon as we know what's the last ID. So, if you call the API while the backup is being done, you don't know the last ID (`max_chunk_id = null`, or inexistent). But after it's done, it should come, in case I want to jump to the end without having to poll for more and more chunks.

artemgavrilov:
> > > Quick question: shouldn't we set a `total_chunks` in the response, after we know it? Otherwise, API consumers can't go to the end of the logs, unless they send `offset=0` and a huge `limit`. If we return the number of chunks, consumers can just send `offset=total_chunks - limit` and get the last `limit` chunks
> > > Also, by doing this, `end` becomes redundant: if we get `total_chunks`, we know end happens if `chunk_id = total_chunks - 1`
> > 
> > 
> > Usually by "cursor-paging" user can't go to last page. But in this case, we probably can select max chunk id and return it.
> > Anyway you suggest send `total_chunks` only if it's already end? IDK.
> > @artemgavrilov @oter any suggestions? Or leave it and improve it later.
> 
> I guess sending `max_chunk_id` works as fine, in my opinion.
> I mean always sending it as soon as we know what's the last ID. So, if you call the API while the backup is being done, you don't know the last ID (`max_chunk_id = null`, or inexistent). But after it's done, it should come, in case I want to jump to the end without having to poll for more and more chunks.

Hmm, sounds good. But if we don't have such use-case right now I propose to implement it later.

artemgavrilov (approved):

fabio-silva:
> > > > Quick question: shouldn't we set a `total_chunks` in the response, after we know it? Otherwise, API consumers can't go to the end of the logs, unless they send `offset=0` and a huge `limit`. If we return the number of chunks, consumers can just send `offset=total_chunks - limit` and get the last `limit` chunks
> > > > Also, by doing this, `end` becomes redundant: if we get `total_chunks`, we know end happens if `chunk_id = total_chunks - 1`
> > > 
> > > 
> > > Usually by "cursor-paging" user can't go to last page. But in this case, we probably can select max chunk id and return it.
> > > Anyway you suggest send `total_chunks` only if it's already end? IDK.
> > > @artemgavrilov @oter any suggestions? Or leave it and improve it later.
> > 
> > 
> > I guess sending `max_chunk_id` works as fine, in my opinion.
> > I mean always sending it as soon as we know what's the last ID. So, if you call the API while the backup is being done, you don't know the last ID (`max_chunk_id = null`, or inexistent). But after it's done, it should come, in case I want to jump to the end without having to poll for more and more chunks.
> 
> Hmm, sounds good. But if we don't have such use-case right now I propose to implement it later.

If we are confident we won't have much logs for now, I guess we can work with only a big limit for now.

fabio-silva (approved):

BupycHuk (dismissed):

oter (approved):

artemgavrilov (approved):

artemgavrilov:
@BupycHuk this feature is ready for merge.

BupycHuk (approved):

PR #777, opened 2021-09-14
PMM-8706: Add specific error codes for incompatible service

oter:
Ticket: https://jira.percona.com/browse/PMM-8706

managed PR: https://github.com/percona/pmm-managed/pull/825
FB: https://github.com/Percona-Lab/pmm-submodules/pull/2016


fabio-silva:
@oter If I get it correctly, when there's an error related to incompatible versions, the error response will contain, for example:
`details: [{ type_url: "ERROR_CODE_INVALID_XTRABACKUP", value: "2" }]`. Is it something like this? Thanks!

Dasio:
So do we want have error codes in some scope like backup?
Firstly I thought about general codes, but this is maybe better

oter:
@fabio-silva 
> @oter If I get it correctly, when there's an error related to incompatible versions, the error response will contain, for example:
> `details: [{ type_url: "ERROR_CODE_INVALID_XTRABACKUP", value: "2" }]`. Is it something like this? Thanks!

It should be something like
`{
    "error": "something happened",
    "code": code_precondition_failed,
    "message": "something happened",
    "details": [
        {
            "@type": "type.googleapis.com/backup.v1beta1.Error",
            "code": "ERROR_CODE_INVALID_XTRABACKUP"
        }
    ]
}`

oter:
@fabio-silva 

> So do we want have error codes in some scope like backup?
> Firstly I thought about general codes, but this is maybe better

I decided to introduce domain based error, so we will not have one enum with aggregation of all errors. In my opinion that approach is better. Any thoughts? @artemgavrilov

api/managementpb/backup/backups.proto
```diff
@@ -195,6 +195,10 @@ message RemoveScheduledBackupResponse {}
 // Backups service handles backup operations to DB.
 service Backups {
   // StartBackup request backup specified service to location.
+  // Could return the Error message in the details containing specific ErrorCode indicating failure reason:
+  // ERROR_CODE_XTRABACKUP_NOT_INSTALLED
+  // ERROR_CODE_INVALID_XTRABACKUP
+  // ERROR_CODE_INCOMPATIBLE_XTRABACKUP
```
artemgavrilov, line 201: Please also `description` option same as we have here: https://github.com/percona/pmm/blob/0f316cc1adf997d39e6afc4aad9e75a809a1c8e3/api/managementpb/checks.proto#L95
to make this information visible is swagger.

api/managementpb/backup/errors.proto
```diff
@@ -0,0 +1,20 @@
...
+  ERROR_CODE_INVALID = 0;
+  ERROR_CODE_XTRABACKUP_NOT_INSTALLED = 1;
+  ERROR_CODE_INVALID_XTRABACKUP = 2;
+  ERROR_CODE_INCOMPATIBLE_XTRABACKUP = 3;
+  ERROR_CODE_INCOMPATIBLE_TARGET_MYSQL = 4;
+}
+
+// Error represents error message returned in the details field in the response.
+message Error {
+  ErrorCode code = 1;
```
artemgavrilov, line 19: Let's rename this field to make it more clear that this is not a gRPC error code.

artemgavrilov (commented):
>  decided to introduce domain based error, so we will not have one enum with aggregation of all errors. In my opinion that approach is better.

Sounds good.

@oter can you please also provide PR/Draft where we can see how these errors can be returned by API?

fabio-silva:
> @fabio-silva
> 
> > @oter If I get it correctly, when there's an error related to incompatible versions, the error response will contain, for example:
> > `details: [{ type_url: "ERROR_CODE_INVALID_XTRABACKUP", value: "2" }]`. Is it something like this? Thanks!
> 
> It should be something like
> 
> ```
> {
>     "error": "something happened",
>     "code": code_precondition_failed,
>     "message": "something happened",
>     "details": [
>         {
>             "@type": "type.googleapis.com/backup.v1beta1.Error",
>             "code": "ERROR_CODE_INVALID_XTRABACKUP"
>         }
>     ]
> }
> ```

Got it !

api/managementpb/backup/errors.proto
oter: not sure, since it is located in the backups package and also has a comment that mentions where this error should be returned.

api/managementpb/backup/backups.proto
oter: added

Dasio (dismissed):

api/managementpb/backup/backups.proto
```diff
@@ -200,6 +201,9 @@ service Backups {
       post: "/v1/management/backup/Backups/Start"
       body: "*"
     };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      description: "Could return the Error message in the details containing specific ErrorCode indicating failure reason:\nERROR_CODE_XTRABACKUP_NOT_INSTALLED - xtrabackup is not installed on the service\nERROR_CODE_INVALID_XTRABACKUP - different versions of xtrabackup and xbcloud\nERROR_CODE_INCOMPATIBLE_XTRABACKUP - xtrabackup is not compatible with MySQL for taking a backup"
```
artemgavrilov, line 205: Why it's written in one line? Can we have it in human-readable format?

api/managementpb/backup/errors.proto
artemgavrilov: IMHO that's can be a little bit confusing:
```
{
    "error": "something happened",
    "code": code_precondition_failed,
    "message": "something happened",
    "details": [
        {
            "@type": "type.googleapis.com/backup.v1beta1.Error",
            "code": "ERROR_CODE_INVALID_XTRABACKUP"
        }
    ]
}
```

artemgavrilov: Ha, I just noticed, why are we returning error text names, not just codes?

api/managementpb/backup/backups.proto
oter: cannot find any info that proto files support multi-line literals :(

api/managementpb/backup/errors.proto
oter: it has a `@type` field indicating from where it came from e.g. from package `backup.v1beta1`

api/managementpb/backup/backups.proto
artemgavrilov: Yep, it seems there is no such option for now: https://github.com/protocolbuffers/protobuf/issues/1297

api/managementpb/backup/errors.proto
oter: > Ha, I just noticed, why are we returning error text names, not just codes?

well looks like it is gonna be just an integer :) because I'm using enum.

artemgavrilov: So on FE, we need to compare strings to understand what error? It looks like have integer code is simpler.

@fabio-silva what do you think?

artemgavrilov: Data race in comments)

oter: @artemgavrilov we had a commenting race condition :D

fabio-silva: > So on FE, we need to compare strings to understand what is the error? It looks like have integer code is simpler.
> 
> 
> 
> @fabio-silva what do you think?

I agree it's simpler the integer way, even though we also have string enums in the FE.

artemgavrilov (dismissed):

artemgavrilov (approved):

maxkondr (approved):

PR #792, opened 2021-10-27
PMM-9060 Add connected_to_platform in settings and endpoint for connecting

jprukner:
This PR adds connected_to_portal into settings so we are able to display connect form if not connected.
Also it replaces SingIn with Connect. It is a breaking change but since SignIn form will be replaced by Connect form we don't need PlatformSignInRequest anymore.


maxkondr (approved):

api/serverpb/json/serverpb.json
```diff
@@ -127,13 +127,13 @@
...
     },
-    "/v1/Platform/SignIn": {
+    "/v1/Platform/Connect": {
       "post": {
         "tags": [
           "Server"
         ],
-        "summary": "PlatformSignIn links that PMM instance to Percona Platform user.",
-        "operationId": "PlatformSignIn",
+        "summary": "PlatformConnect connects a PMM server to Portal organisation. That allows the user to sign in to PMM server with their Percona Account.",
```
atymchuk, line 135:
```suggestion
        "summary": "PlatformConnect connects a PMM server to the organization created on Percona Portal. That allows the user to sign in to the PMM server with their Percona Account.",
```

atymchuk (changes requested):
Sorry, one more correction please )

atymchuk (dismissed):

maxkondr (dismissed):

api/serverpb/server.proto
```diff
@@ -194,6 +194,8 @@ message Settings {
   bool backup_management_enabled = 18;
   // True if Azure Discover is enabled.
   bool azurediscover_enabled = 17;
+  // True if the PMM instance is connected to Portal
+  bool connected_to_portal = 19;
```
palash25, line 198: nitpick: maybe we should call this `connected_to_platform` as everywhere else it mentioned as platform

palash25 (approved):

api/serverpb/server.proto
```diff
@@ -290,18 +292,28 @@ message PlatformSignUpRequest {
 
 message PlatformSignUpResponse {}
 
-message PlatformSignInRequest {
```
artemgavrilov, line 293: That was part of public API and can be used by third-party scripts, etc. I assume we can't just drop this stuff, but we can deprecate it. 

@BupycHuk  correct me if I'm wrong.

BupycHuk: Yeah, that's correct.

BupycHuk: And so, we can revert changes in api.descriptor

jprukner: Ok, to make it clear, you want the singin api not to be removed. So instead of replacing it, we just add a new one. @artemgavrilov @BupycHuk

BupycHuk: yes, and add `[deprecated]` to old ones.

BupycHuk (dismissed):

api/platformpb/json/platformpb.json
```diff
@@ -0,0 +1,190 @@
...
+    "https",
+    "http"
+  ],
+  "swagger": "2.0",
+  "info": {
+    "title": "PMM Platform API",
+    "version": "v0"
+  },
+  "paths": {
+    "/v0/Platform/Connect": {
```
maxkondr, line 18: I don't think we should expose this endpoint. Once user calls it the result will be the following:
- new inventory will be added to Portal organization inventory list
- the SSO params are returned
- but PMM Server doesn't apply SSO Params -> out of sync state of PMM Server and Portal Inventory.

api/platformpb/platform.proto
```diff
@@ -0,0 +1,66 @@
...
+
+message SignUpRequest {
+  // New Percona Platform user's email.
+  string email = 1 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+  // Existing Percona Platform user's password.
+  string password = 2 [deprecated = true];
```
maxkondr, line 18: how it could be existing user if we are going to create it ?

api/platformpb/json/platformpb.json
jprukner: That is not true. This won't return sso params. This calls Platform API, stores SSO and reloads grafana.

jprukner: This call returns empty response and 200 on success, different code and error json on failure.

maxkondr (approved):

artemgavrilov (approved):

api/serverpb/server.proto
atymchuk: @palash25 The platform is not software per se, but the Portal is. We connect PMM to the Portal. By the way, we launch under portal.percona.com.

PR #794, opened 2021-11-02
PMM-7781 Unwind template and rule entities

artemgavrilov:
https://jira.percona.com/browse/PMM-7781

- [ ] https://github.com/percona/pmm-managed/pull/864

FB: https://github.com/Percona-Lab/pmm-submodules/pull/2129


api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  bool disabled = 2;
   // Rule human-readable summary.
-  string summary = 4;
-  // Rule parameters.
-  repeated RuleParam params = 5;
-  // Rule duration.
-  google.protobuf.Duration for = 6;
-  // Rule severity.
-  management.Severity severity = 7;
+  string summary = 3;
```
artemgavrilov, line 70: Memory used by MongoDB connections

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  // Comment.
+  string comment = 4;
```
artemgavrilov, line 72: That field has the same meaning as rule summary, but I propose to rename it to the comment, for example: `My awesome rule for small nodes`

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  // Expression template.
+  string expr_template = 5;
```
artemgavrilov, line 74: Expression as it was specified in template, example: `100 / ((node_memory_MemTotal_bytes)\n/ on(node_name) group_left(service_name) (mongodb_ss_connections{conn_type=\"current\"} * 1024 * 1024))\n> [[ .threshold ]]`

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  // Expression filled with parameters.
+  string expr = 6;
```
artemgavrilov, line 76: Same expression, but with filled placeholders: `100 / ((node_memory_MemTotal_bytes)\n/ on(node_name) group_left(service_name) (mongodb_ss_connections{conn_type=\"current\"} * 1024 * 1024))\n> 20`

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  // Expression parameters definitions.
+  repeated ParamDefinition params_definitions = 7;
```
artemgavrilov, line 78: Definitions of parameters required by expression, contains max, min, default values, param description etc. Equals to `params` field of template entity.

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  // Expression parameters values.
+  repeated ParamValue params_values = 8;
```
artemgavrilov, line 80: Exact param values, requred for params without a default value. For other params may or may not have value.

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  // Default for duration.
+  google.protobuf.Duration default_for = 9;
```
artemgavrilov, line 82: Inherited from template

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  // For duration.
+  google.protobuf.Duration for = 10;
```
artemgavrilov, line 84: Overwrites default_for.

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  // Default severity.
+  management.Severity default_severity = 11;
```
artemgavrilov, line 86: Inherited from template

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  // Severity.
+  management.Severity severity = 12;
```
artemgavrilov, line 88: Overwrites `default_severity`

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
   // Custom labels.
-  map<string, string> custom_labels = 8;
+  map<string, string> custom_labels = 13;
```
artemgavrilov, line 90: Labels that user can specify per rule.

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  // Labels.
+  map<string, string> labels = 14;
```
artemgavrilov, line 92: Labels inherited from template.

api/managementpb/ia/rules.proto
```diff
@@ -62,30 +62,42 @@ message RuleParam {
...
+  // Annotations.
+  map<string, string> annotations = 15;
```
artemgavrilov, line 94: Annotations inherited from template.

api/managementpb/ia/rules.proto
oter: these are completely breaking changes, but probably we can proceed. Should we inform our clients about that?

api/managementpb/ia/rules.proto
oter: BTW let's integrate your nice comments into API :)

oter (approved):
LGTM

api/managementpb/ia/rules.proto
artemgavrilov: That is beta API, we can break it.

atymchuk (approved):
Nice job 👍

api/managementpb/ia/rules.proto
```diff
@@ -54,38 +54,52 @@ message RuleParam {
     // Bool value.
     bool bool = 3;
     // Float value.
-    float float = 4;
+    double float = 4;
```
maxkondr, line 57: why data type change is required?

api/managementpb/ia/templates.proto
```diff
@@ -16,37 +16,37 @@ import "managementpb/severity.proto";
...
 
-// TemplateFloatParam represents float parameter's default value and valid range.
-message TemplateFloatParam {
+// FloatParamDefinition represents float parameter's default value and valid range.
+message FloatParamDefinition {
   // True if default value is set.
   bool has_default = 1;
   // Default value if has_default is true.
-  float default = 2;
+  double default = 2;
```
maxkondr, line 29: the same - why change type?

api/managementpb/ia/rules.proto
artemgavrilov: When we parse templates YAMLs we use `float64`: https://github.com/percona-platform/platform/blob/c6e7618e5d12abb7419b59e073ee0bc75e5eb20b/pkg/alert/parameter.go#L43

Because of types mismatch between API and underlying structs we had such things: https://github.com/percona/pmm-managed/blob/52a4cd82dda43d4161612f9450964e1474be5574/services/management/ia/common.go#L101-L104

oter (approved):

BupycHuk (approved):

fabio-silva (approved):

PR #801, opened 2021-11-30
PMM-7605 Move pmm to the github actions

eoyy:


ritbl (approved):
LGTM

.github/workflows/go.yml
```diff
@@ -0,0 +1,130 @@
...
+          - version: tip
+            may-fail: true
+
+    continue-on-error: ${{ matrix.go.may-fail }}
+
+    runs-on: ubuntu-latest
+
+    env:
+      GOPATH: ${{ github.workspace }}
+      WORKING_DIR: ${{ github.workspace }}/src/github.com/percona/pmm
```
artemgavrilov, line 34: We don't use it in other CI workflows. Why do we need it here? It's used in just one place.

.github/workflows/go.yml
```diff
@@ -0,0 +1,130 @@
+name: build
+
+on:
+  push:
+    branches:
+      - main
+      - release/*
+      # skip non-trunk PMM-XXXX/SAAS-XXXX branch builds, but still build pull requests
+      - '!PMM-[0-9]+'
+      - '!SAAS-[0-9]+'
```
artemgavrilov, line 10: But why do we need negative conditions here? If we will drop these lines only `main` and `release` will trigger build, no?

eoyy: it was migrated from travis.yml 
it ignores (probably) only branches with no description in the name. e.g. the current branch has the description.
Do you think this condition is outdated?

artemgavrilov: I think yes, we are using something like this in all other repos: https://github.com/percona/pmm-managed/blob/58e8a4f6b8f5a1a3b5ba4c1817e757eb388c20bc/.github/workflows/ci.yml#L3-L15

.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,76 @@
...
+          path: ~/.cache/go-build
+          key: ${{ matrix.os }}-go-${{ matrix.go-version }}-build-${{ github.ref }}-${{ hashFiles('**') }}
+          restore-keys: |
+            ${{ matrix.os }}-go-${{ matrix.go-version }}-build-${{ github.ref }}-
+            ${{ matrix.os }}-go-${{ matrix.go-version }}-build-
+      - name: Clean Go modules cache
+        run: go clean -modcache
+
+      - name: Clean Go build cache
+        run: go clean -cache
```
artemgavrilov, line 76: Please add newline

.github/workflows/go.yml
eoyy: Ok, I fixed this.

.github/workflows/go.yml
eoyy: It's needed to run unit tests correctly.

.github/workflows/go.yml
```diff
@@ -0,0 +1,130 @@
...
+          echo "GOROOT=$HOME/gotip" >> $GITHUB_ENV
+          echo "$HOME/gotip/bin" >> $GITHUB_PATH
+
+      - name: Enable Go modules cache
+        uses: percona-platform/cache@v2
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ matrix.os }}-go-${{ matrix.go-version }}-modules-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ matrix.os }}-go-${{ matrix.go-version }}-modules-
```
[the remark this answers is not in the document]
eoyy: Done

artemgavrilov (approved):

artemgavrilov:
Can you please also rename `go.yml` to `ci.yml` as we name this file in other repos?

.github/workflows/go.yml
```diff
@@ -0,0 +1,129 @@
...
+
```
artemgavrilov, line 68: I see that modules cached is not used for some reason and modules are downloaded each time.
Let's add here explicit step to resolve modules
```
      - name: Download Go modules
        run: go mod download
```

.github/workflows/go.yml
```diff
@@ -0,0 +1,132 @@
...
+    runs-on: ubuntu-latest
+    env:
+      SLACK_WEBHOOK: ${{ secrets.SLACK_TOKEN }}
+      SLACK_CHANNEL: "pmm-ci"
+      SLACK_USERNAME: "PR pipelines bot"
+      SLACK_ICON_EMOJI: ":chestnut:"
+      SLACK_COLOR: "#FF0000"
+      SLACK_TITLE: "Finished ${{ github.event.repository.name }} workflow"
+      SLACK_MESSAGE: "Workflow failed: ${{ github.event.inputs.repo || github.repository }}:${{ github.event.inputs.branch || github.head_ref }}"
+      SLACK_FOOTER: "Please check the Actions URL ^"
```
atymchuk, line 127: Can we remove `SLACK_FOOTER: Please check the Actions URL ^`? Looks redundant and just clutters the message.

eoyy:
> Can you please also rename `go.yml` to `ci.yml` as we name this file in other repos?

Done

artemgavrilov (dismissed):

percona-csalguero (approved):

atymchuk (approved):

PR #814, opened 2021-12-27
Readme sync

BupycHuk:


.github/workflows/doc.yml
```diff
@@ -0,0 +1,24 @@
...
+  push:
+    branches:
+      - main
+      - readme-sync
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: readmeio/github-readme-sync@v2
```
denisok, line 12: @atymchuk could you please clone this action? others should be there already.

.github/workflows/doc.yml
```diff
@@ -0,0 +1,24 @@
...
+        with:
+          readme-oas-key: ${{ secrets.README_TOKEN }}
```
denisok, line 14: @atymchuk could you please check if this token is correct.

atymchuk: Not really - the tokens are only known at the time they are posted. If @BupycHuk grabbed it from the readme.io's site it should be OK. This run indirectly confirms it's OK, so should be good to go.

.github/workflows/doc.yml
denisok: @BupycHuk please change all actions to the percona-platform.

denisok (changes requested):

.github/workflows/doc.yml
atymchuk: cloned!

atymchuk (dismissed):

atymchuk:
@denisok all actioned are sourced from /percona-paltform now )

denisok (approved):

denisok (approved):

atymchuk (approved):
:shipit:

PR #818, opened 2022-01-20
add path param to AgentInfo

qwest812:


api/agentlocalpb/agentlocal.proto
```diff
@@ -33,6 +33,7 @@ message AgentInfo {
   // The current listen port of this Agent (exporter or vmagent).
   // Zero for other Agent types, or if unknown or not yet supported.
   uint32 listen_port = 4;
+  string path = 5;
```
BupycHuk, line 36: it probably should not be in AgentInfo, because it's not related to pmm-agent itself.
Just a path doesn't look a good naming
Please add comment.

BupycHuk (changes requested):
Please, add a task number to PR header

denisok:
fixed in https://github.com/percona/pmm/pull/821 ?

BupycHuk:
Fixed in #821 and other PRs in other repositories

PR #832, opened 2022-03-08
PMM-9636 refine failed checks

idoqo:
PMM-9636

Build: SUBMODULES-2371

- [ ] https://github.com/percona/pmm-managed/pull/1012


api/managementpb/pagination.proto
```diff
@@ -1,8 +1,8 @@
 syntax = "proto3";
 
-package ia.v1beta1;
+package management;
 
-option go_package = "api/managementpb/ia;iav1beta1";
+option go_package = "api/managementpb;managementpb";
```
BupycHuk, line 5: 👍🏼

api/managementpb/checks.proto
```diff
@@ -77,8 +110,54 @@ message ChangeSecurityChecksRequest {
...
 service SecurityChecks {
+  // ListFailedServices returns a list of services with failed checks.
+  rpc ListFailedServices(ListFailedServicesRequest) returns (ListFailedServicesResponse) {
+    option (google.api.http) = {
+      post: "/v1/management/SecurityChecks/ListFailedServices"
+      body: "*"
+    };
+  }
+  // GetFailedChecks returns the checks result for a given service.
+  rpc GetFailedChecks(GetFailedChecksRequest) returns (GetFailedChecksResponse) {
```
BupycHuk, line 148: I think this can be replaced by using `GetSecurityCheckResults` with two filters, what do you think?

idoqo: A separate filter to return the failed checks per service, right?

I think the plan is to deprecate and eventually remove `GetSecurityCheckResults` since it's not used (at least by FE).

api/managementpb/checks.proto
```diff
@@ -77,8 +110,66 @@ message ChangeSecurityChecksRequest {
 
 message ChangeSecurityChecksResponse {}
 
+message ListFailedServicesRequest {}
```
PavelKhripkov, line 113: Should not result be paginated?

api/managementpb/checks.proto
```diff
@@ -28,6 +29,38 @@ message SecurityCheckResult {
...
+  string service_name = 6;
+  // ID of the monitored service on which the check ran.
+  string service_id = 7;
+  // Name of the check that failed
+  string check_name = 8;
+  // ID of the check result as stored in AlertManager
+  string alert_id = 9;
+  // Silence status of the check result
+  bool silenced = 10;
+}
```
PavelKhripkov, line 62: We may add `check_type` field (would be something like `MONGODB_REPLSETGETSTATUS`) or split it into `check_type` and `db_family` as requested in the tkt.

PavelKhripkov (changes requested):

api/managementpb/checks.proto
idoqo: Yeah, we figured since this is meant to be a summary (failed services and number of failed checks), it makes sense to display them at once.

api/managementpb/checks.proto
idoqo: Yeah, but I don't think `db_family` is currently supported for checks v1.

api/managementpb/checks.proto
PavelKhripkov: What if someone adds 100 or 300 services? All of them will be shown on single page?

api/managementpb/checks.proto
PavelKhripkov: I think the field may be empty if there is no data to fill it. What do you think?

idoqo: I see, that could work then. but right now it will *always* be empty so I thought maybe we can pend it until we have proper support for it across platform/managed.

api/managementpb/checks.proto
idoqo: Thinking of it now, another approach could be to ignore the cache and query alertmanager directly for the results for this service.

PavelKhripkov (dismissed):

api/managementpb/checks.proto
```diff
@@ -28,6 +29,38 @@ message SecurityCheckResult {
...
+// CheckResultSummary is a summary of check results.
+message CheckResultSummary {
+  string service_name = 1;
+  string service_id = 2;
+  // Number of failed checks for this service that are critical
+  uint32 critical_count = 3;
+  // Number of failed checks for this service that are major
+  uint32 major_count = 4;
+  // Number of failed checks for this service that are trivial
+  uint32 trivial_count = 5;
```
artemgavrilov, line 41: I thought we decided to create counter for each possible severity level, no?

api/managementpb/checks.proto
```diff
@@ -28,6 +29,38 @@ message SecurityCheckResult {
...
+}
+
+// CheckResult represents the check results for a given service.
+message CheckResult {
+  string summary = 1;
+  string description = 2;
+  Severity severity = 3;
+  map<string, string> labels = 4;
```
artemgavrilov, line 49: And where is annotations?

api/managementpb/checks.proto
```diff
@@ -77,8 +110,66 @@ message ChangeSecurityChecksRequest {
...
+      post: "/v1/management/SecurityChecks/ToggleCheckAlert"
+      body: "*"
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "Toggle Check Alert"
+      description: "Silence/Unsilence alerts for a specific check result."
+    };
+  }
   // GetSecurityCheckResults returns Security Thread Tool's latest checks results.
   rpc GetSecurityCheckResults(GetSecurityCheckResultsRequest) returns (GetSecurityCheckResultsResponse) {
```
artemgavrilov, line 174: As I remember we decided to deprecate this endpoint, right?

api/managementpb/checks.proto
idoqo: Yeah,  but that's when frontend is ready to support all the severity levels too.

api/managementpb/checks.proto
idoqo: I think all the fields in the annotations are already in the message directly (summary, name and description). With labels, it helps to send all of it back so that they can be used for filtering, etc

api/managementpb/checks.proto
idoqo: Yeah, you're right

api/managementpb/checks.proto
artemgavrilov: But on the API level we can add all of them now, but fill only these 3.

api/managementpb/checks.proto
```diff
@@ -28,6 +29,50 @@ message SecurityCheckResult {
   string service_name = 6;
 }
 
+// CheckResultSummary is a summary of check results.
+message CheckResultSummary {
+  string service_name = 1;
+  string service_id = 2;
+  // Number of failed checks for this service with unknown severity level.
+  uint32 unknown_count = 3;
```
artemgavrilov, line 37: We can remove that one. `unknown` is used only for invalid states.

artemgavrilov (approved):

artemgavrilov (approved):

PavelKhripkov (approved):

BupycHuk (approved):

PR #854, opened 2022-03-31
Pmm 9724

palash25:
PMM-9724

Build: SUBMODULES-0


api/platformpb/platform.proto
```diff
@@ -158,4 +169,15 @@ service Platform {
       description: "SearchOrganizationEntitlements fetches details of the entitlement's available to the Portal organization that the PMM server is connected to."
     };
   }
+  // ServerInfo returns PMM server ID and name.
+  rpc ServerInfo(ServerInfoRequest) returns (ServerInfoResponse) {
```
palash25, line 173: according to the new design, FE needs to display the pmm server ID and pmm server name after it has been connected to Platform, hence adding this new endpoint

oksana-grishchenko: the design from [Docs](https://confluence.percona.com/display/PMM/Single+Sign-On+-+PMM+Server+several+domains+support)  also shows the 
- Main PMM Server Domain Name 
- Additional PMM Server Names 

we postponed development of the second feature (additional names), but the PMM server still required to have a main domain (currently it is referenced as pmm_server_url). Should this API also provide this url or it will be taken from somewhere else?

palash25: @fabio-silva only asked for ID and name i guess, so for now we will merge and add it later if need be

fabio-silva: these two were out of scope for this version

PR #872, opened 2022-04-20
better user experience and fixes for existing containers

shoffman-percona:
made some changes based on requests in #pmm channel to make it clearer things are happening (no need for spinning bars IMO) but also fixed so that if a pmm-data volume was detected we'd rename it with the same datestamp that the pmm-server would get to allow the user to downgrade easier.

### DIFF
--- a/get-pmm.sh
+++ b/get-pmm.sh
@@ -73,7 +73,7 @@ setup_colors() {
 # Prints message to stderr with new line at the end.
 #######################################
 msg() {
-  echo >&2 -e "${1-}"
+  echo >&2 -ne "${1-}"
 }
 
 #######################################
@@ -223,19 +223,25 @@ run_docker() {
 # If any PMM server instance is run - stop and backup it.
 #######################################
 start_pmm() {
-  msg "Starting PMM server..."
+  msg "Fetching docker image...\n"
   run_docker "pull $repo:$tag 1> /dev/null"
 
+  msg "Checking data volume...\n"
+  pmm_archive="$(date "+%F-%H%M%S")"
   if ! run_docker "inspect pmm-data 1> /dev/null 2> /dev/null"; then
     run_docker "create -v /srv/ --name pmm-data $repo:$tag /bin/true 1> /dev/null"
-    msg "Created PMM Data Volume: pmm-data"
+    msg "Created PMM Data Volume: pmm-data\n"
+  else 
+    msg "\n\t${YELLOW}Existing PMM Data volume found, renaming to pmm-data-$pmm_archive${NOFORMAT}\n"
+    run_docker "rename pmm-data pmm-data-$pmm_archive"
+    run_docker "create -v /srv/ --name pmm-data $repo:$tag /bin/true 1> /dev/null"
+    msg "Created PMM Data Volume: pmm-data\n"
   fi
 
   if run_docker "inspect pmm-server 1> /dev/null 2> /dev/null"; then
-    pmm_archive="pmm-server-$(date "+%F-%H%M%S")"
-    msg "\tExisting PMM Server found, renaming to $pmm_archive\n"
+    msg "\t${YELLOW}Existing PMM Server found, renaming to pmm-server-$pmm_archive${NOFORMAT}\n"
     run_docker 'stop pmm-server' || :
-    run_docker "rename pmm-server $pmm_archive\n"
+    run_docker "rename pmm-server pmm-server-$pmm_archive"
   fi
   run_pmm="run -d -p $port:443 --volumes-from pmm-data --name $container_name --restart always $repo:$tag"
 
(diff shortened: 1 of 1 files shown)


get-pmm.sh
```diff
@@ -223,19 +223,25 @@ run_docker() {
...
+  pmm_archive="$(date "+%F-%H%M%S")"
   if ! run_docker "inspect pmm-data 1> /dev/null 2> /dev/null"; then
     run_docker "create -v /srv/ --name pmm-data $repo:$tag /bin/true 1> /dev/null"
-    msg "Created PMM Data Volume: pmm-data"
+    msg "Created PMM Data Volume: pmm-data\n"
+  else 
+    msg "\n\t${YELLOW}Existing PMM Data volume found, renaming to pmm-data-$pmm_archive${NOFORMAT}\n"
+    run_docker "rename pmm-data pmm-data-$pmm_archive"
+    run_docker "create -v /srv/ --name pmm-data $repo:$tag /bin/true 1> /dev/null"
+    msg "Created PMM Data Volume: pmm-data\n"
```
nikita-b, line 238: You can move these two strings out of the `if` block.

nikita-b (changes requested):

get-pmm.sh
```diff
@@ -260,13 +266,14 @@ show_message() {
     die "${RED}ERROR: cannot detect PMM server address${NOFORMAT}"
```
nikita-b, line 266: Do we need `\n` here?

denisok:
@shoffman-percona maybe also take a look https://jira.percona.com/browse/PMM-9883 .

get-pmm.sh
```diff
@@ -223,19 +223,25 @@ run_docker() {
...
 
+  msg "Checking data volume...\n"
+  pmm_archive="$(date "+%F-%H%M%S")"
   if ! run_docker "inspect pmm-data 1> /dev/null 2> /dev/null"; then
     run_docker "create -v /srv/ --name pmm-data $repo:$tag /bin/true 1> /dev/null"
-    msg "Created PMM Data Volume: pmm-data"
+    msg "Created PMM Data Volume: pmm-data\n"
+  else 
+    msg "\n\t${YELLOW}Existing PMM Data volume found, renaming to pmm-data-$pmm_archive${NOFORMAT}\n"
+    run_docker "rename pmm-data pmm-data-$pmm_archive"
```
BupycHuk, line 236: shouldn't we ask user if he wants to use existing PMM Data volume or create new one? What if user will decide to use the same script to upgrade PMM?

BupycHuk, line 236: I checked this command and looks like it doesn't work for volumes. couldn't find anything in documentation about renamin data volumes. issue in moby project still open https://github.com/moby/moby/issues/31154

get-pmm.sh
```diff
@@ -73,7 +73,7 @@ setup_colors() {
 # Prints message to stderr with new line at the end.
```
BupycHuk, line 73: it doesn't add new line at the end anymore

get-pmm.sh
```diff
@@ -73,7 +73,7 @@ setup_colors() {
...
 #######################################
 msg() {
-  echo >&2 -e "${1-}"
+  echo >&2 -ne "${1-}"
```
BupycHuk, line 76: what was the reason to get rid of new line at the end? I think this change increases risk to break formatting.

BupycHuk (changes requested):

get-pmm.sh
shoffman-percona: @BupycHuk made a good point that people use this for upgrades (as evidenced by the ticket @denisok shared) so I'm rethinking this.  I think as a best practice we *should* be at least taking a backup if we find an existing pmm-data volume and perhaps allow users to skip it if they run in interactive mode.  Thoughts?

get-pmm.sh
shoffman-percona: I didn't think about users doing upgrades with it (which I don't know that they should be...) so we kind of need to think of what the default behavior should be with zero interaction and then allow users to change our decisions with --interactive (which has our defaults set).  I mentioned in an earlier comment that I feel like best practice is to take a backup before we upgrade...there's no going back to a 2.26.0 once the databases in /srv are upgraded to 2.27.0 without a backup...thoughts?

get-pmm.sh
shoffman-percona: I was originally playing with \r to reset cursor to the beginning of the line instead of scrolling but decided against it.  most lines already had newlines so I cleaned up the remaining ones (seems I missed one) but instead I think we'll let the native docker pull text show since it seems dumb to reinvent a spinning progress bar when the docker command already has it.

get-pmm.sh
shoffman-percona: what error are you getting?  I ran this on ubuntu and it successfully renames the volume and the new container references the new volume (the old container would also reference the new volume since they're both looking for pmm-data from the initial docker run).  I'll check on mac shortly but it seems to be working everywhere I'm trying it.  May be moot point if we agree that a proper backup is a smarter way to do upgrades.

get-pmm.sh
shoffman-percona: Yes, missed that one when I changed the default echo behavior.

get-pmm.sh
shoffman-percona: appears to work both on mac and ubuntu?

BupycHuk: tested on my mac
```
nurlan@Nurlans-MacBook-Pro action-slack-notify % docker volume create pmm-data
pmm-data
nurlan@Nurlans-MacBook-Pro action-slack-notify % docker rename pmm-data pmm-data-backup
Error response from daemon: No such container: pmm-data
Error: failed to rename container named pmm-data
nurlan@Nurlans-MacBook-Pro action-slack-notify % 
```

nikita-b: I think an upgrade should be the default option (without pmm-data container rename). If someone just ran the script, it shouldn't delete any data.

nikita-b (changes requested):

BupycHuk:
@shoffman-percona since we are working on the new PMM CLI tool I think we can close this PR, right?

shoffman-percona:
I will close this out and see how PMM CLI takes shape.  I'm concerned we might go backwards though and go from PMM server in a single command to PMM server is 2+ steps (download/install/run CLI command).  If so I'll amend this to be used for first time install only and if already installed exit and point user to PMM CLI but that can be done at a later time.

PR #888, opened 2022-05-07
Hackathon 3

percona-csalguero:
Basic communication workflow between pmm-admin, agent and server


BupycHuk (commented):
@denisok what do you think what is the best place to put this doc? Github wiki, API docs or usual docs?

denisok:
I would think we need one point of truth as you mention. Lets discuss.

[Github Wiki](https://docs.github.com/en/communities/documenting-your-project-with-wikis/about-wikis):
+ provides visual interface to edit markdown
+ easy to use for external contributors
+ push access to the wiki could be enabled for everyone separately
+ has same git interface (edit locally and push)
- lacks review process

API docs and separate pmm doc folder are almost the same in tech:
+ simple markdown
+ PR review process
- no visual editing, full git process

The only difference is that in API docs, special header is needed, which also could have `hidden` values to not publish on ReadMe, but maybe plain markdown also would be ignored (needs testing) also not clear how to work with images for `markdown` folder.

Does it make sense to publish arch in ReadMe ? Maybe, it is more API doc but could be a dev documentation hub.

If we put arch docs in `pmm-doc` repo they will be kinda disjoint from the code and development. But we need some basic doc there as well, and maybe linking here for more detailed documentation. So what choices we have? : 
* submodules and link arch doc from here (we probably want less submodules, not more)
* some Action that just would copy docs from here
* manually copy basic docs if they are changed (should be rare change)
* just link (documentation link) Arch docs in `pmm-doc` to makdown here or ReadMe

I would say that is we can use same folder for API and arch doc, would be cool and go with it. As we have ReadMe, maybe make sense to experiment with it and publish arch doc there and link from `pmm-doc` to ReadMe. So we have only 2 publishing platforms instead of 3 (Wiki, pmm-doc, ReadMe).


CC: @atymchuk

docs/architecture.md
```diff
@@ -0,0 +1,15 @@
+# Architecture
+
+## pmm-admin - server communication workflow.
+
+`pmm-admin` is a command line utility used to send commands to the pmm-server in order to execute actions on clients. Those actions could be to configure the client itself, add or remove monitored instances, etc.
```
atymchuk, line 5:
```suggestion
`pmm-admin` is a command line utility that wraps a number of configuration and inventory endpoints of PMM Server. It it mostly used to configure the client itself, add or remove monitored instances, etc. 
```

docs/architecture.md
```diff
@@ -0,0 +1,15 @@
...
+
+The communication flow is:
+1. pmm-admin gets PMM server URL from the local pmm-agent. 
+2. pmm-admin translates the command into a PMM server message and sends it to the server (pmm-managed API).
+3. pmm-managed process the command and sends the response back to pmm-admin.
+4. pmm-admin intructs the pmm-agent to run an action based on the PMM server response.
+
+### Communication flow diagram.
+
+![communication flow diagram](images/flow.png)
```
atymchuk, line 15: The diagram can we made better. For example, the `get PMM server address from the local agent` is a double-sided arrow right now, but it embeds three separate steps, so IMO it'd better be converted to three arrows.

Another issue is that `Tell pmm-agent to execute the action...` also lacks a response the agent. Than can be added too.

atymchuk (changes requested):

atymchuk:
> I would think we need one point of truth as you mention. Lets discuss.
> 
> [Github Wiki](https://docs.github.com/en/communities/documenting-your-project-with-wikis/about-wikis):
> 
> * provides visual interface to edit markdown
> * easy to use for external contributors
> * push access to the wiki could be enabled for everyone separately
> * has same git interface (edit locally and push)
> 
> * lacks review process
> 
> API docs and separate pmm doc folder are almost the same in tech:
> 
> * simple markdown
> * PR review process
> 
> * no visual editing, full git process
> 
> The only difference is that in API docs, special header is needed, which also could have `hidden` values to not publish on ReadMe, but maybe plain markdown also would be ignored (needs testing) also not clear how to work with images for `markdown` folder.
> 
> Does it make sense to publish arch in ReadMe ? Maybe, it is more API doc but could be a dev documentation hub.
> 
> If we put arch docs in `pmm-doc` repo they will be kinda disjoint from the code and development. But we need some basic doc there as well, and maybe linking here for more detailed documentation. So what choices we have? :
> 
> * submodules and link arch doc from here (we probably want less submodules, not more)
> * some Action that just would copy docs from here
> * manually copy basic docs if they are changed (should be rare change)
> * just link (documentation link) Arch docs in `pmm-doc` to makdown here or ReadMe
> 
> I would say that is we can use same folder for API and arch doc, would be cool and go with it. As we have ReadMe, maybe make sense to experiment with it and publish arch doc there and link from `pmm-doc` to ReadMe. So we have only 2 publishing platforms instead of 3 (Wiki, pmm-doc, ReadMe).
> 
> CC: @atymchuk

I agree: using wiki would be one source too many, so I'd rather focus on readme and pmm-doc. The architecture can be posted to both though.

denisok:
oops, pushed wrong button.

PR #923, opened 2022-06-07
PMM-9913 Added temp cleanup on agent start/stop

tshcherban:
PMM-9913

Build: SUBMODULES-2586


agent/commands/run.go
```diff
@@ -55,14 +58,33 @@ func Run() {
 		config.ConfigureLogger(cfg)
 		l.Debugf("Loaded configuration: %+v", cfg)
 
+		cleanupTmp(cfg.Paths.TempDir, l)
+
 		run(ctx, cfg, configFilepath)
 
+		cleanupTmp(cfg.Paths.TempDir, l)
```
BupycHuk, line 65: don't we remove files on agent stop? I think we should do it on agent stop as well.

tshcherban: its done after `run` completes

tshcherban: added clear on agent stop

YaroslavPodorvanov (approved):
LGTM

PR #927, opened 2022-06-09
PMM-10143 fix agents test

idoqo:
PMM-10143

Build: SUBMODULES-0

### DIFF
--- a/agent/agents/mysql/perfschema/perfschema_test.go
+++ b/agent/agents/mysql/perfschema/perfschema_test.go
@@ -285,20 +285,20 @@ func TestPerfSchema(t *testing.T) {
 
 	case "10.2-mariadb":
 		digests = map[string]string{
-			"SELECT `sleep` (?)":   "e58c348e4947db23b7f3ad30b7ed184a",
-			"SELECT * FROM `city`": "e0f47172152e8750d070a854e607123f",
+			"SELECT `sleep` (?)":   "fe8d67e28d171893e1b33b179394e592",
+			"SELECT * FROM `city`": "7e30fa1763d6d9aa88f359236cedaa78",
 		}
 
 	case "10.3-mariadb":
 		digests = map[string]string{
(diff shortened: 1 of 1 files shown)


agent/agents/mysql/perfschema/perfschema_test.go
```diff
@@ -285,20 +285,20 @@ func TestPerfSchema(t *testing.T) {
 
 	case "10.2-mariadb":
 		digests = map[string]string{
-			"SELECT `sleep` (?)":   "e58c348e4947db23b7f3ad30b7ed184a",
-			"SELECT * FROM `city`": "e0f47172152e8750d070a854e607123f",
+			"SELECT `sleep` (?)":   "fe8d67e28d171893e1b33b179394e592",
+			"SELECT * FROM `city`": "7e30fa1763d6d9aa88f359236cedaa78",
```
idoqo, line 289: I copied these new values from the test results, but I'm not sure how they were generated in the first place, or why they're different for each database distribution.

PavelKhripkov: The only thing I found about algorithm used to produce digest is that it's a MD5 hash. But it seems like it's not just MD5 of an SQL statement, it's something complicated. I don't see any reason to investigate how to get correct hash for different versions of dbs. I believe the previous values were got the same way you've got yours. Mean time we can ask experts in Slack, maybe they're aware of the algorithm.

PavelKhripkov (approved):

agent/agents/mysql/perfschema/perfschema_test.go
BupycHuk: Yeah, it was done the way you've done.

BupycHuk (approved):

agent/agents/mysql/perfschema/perfschema_test.go
idoqo: Right Pavel, I don't think an investigating is super necessary either - it just feels like a black box right now.

PR #944, opened 2022-06-17
Monorepo gh actions

michal-kralik:
PMM-0

Build: SUBMODULES-0


.github/CODEOWNERS
```diff
@@ -1 +1,4 @@
-*	@BupycHuk
+*                           @BupycHuk
+/managed/                   @BupycHuk @artemgavrilov
```
ShashankSinha252, line 2: Is there any benefit in adding multiple entries here at the moment? We have only 2 people responsible more or less for the entire codebase. It would make more sense if we had more people looking at specific parts of the project.

.github/dependabot.yml
```diff
@@ -13,3 +13,20 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
```
ShashankSinha252, line 16: Empty life breaks existing convention in file.

.github/workflows/ci-agent.yml
```diff
@@ -2,6 +2,8 @@ name: pmm-agent CI
 
 on:
   push:
+    paths-ignore:
```
ShashankSinha252, line 5: Would using `paths` instead be a scalable approach?

.github/workflows/ci-managed.yml
```diff
@@ -179,13 +197,13 @@ jobs:
...
       - name: Setup dev-container
-        run: docker exec -i --workdir=/root/go/src/github.com/percona/pmm-managed pmm-managed-server .devcontainer/setup.py
+        run: docker exec -i --workdir=/root/go/src/github.com/percona/pmm/managed pmm-managed-server .devcontainer/setup.py
 
       - name: Run required checks/linters
         env:
           REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          go run .github/check-license.go
+          go run ../.github/check-license.go
```
ShashankSinha252, line 206: Would it be better to use absolute path?

ShashankSinha252 (commented):
LGTM. Minor review comments.

.github/CODEOWNERS
artemgavrilov: Right now we just moved what we had in managed repo. But yes, we will add more codeowners for specific parts here

.github/workflows/ci-agent.yml
artemgavrilov: We need to run top level CI as well. `path-ignore` is the simplest way to do it.

PR #947, opened 2022-06-20
fix issue in get config

denisok:
this is a small fix for kubeconfig generation.

don't know why it has `--` maybe if minikube aliased as `alias kubectl='minikube kubectl'` when it should be `alias kubectl='minikube kubectl --'`.

Anyway it should be fixed, doesn't work in current form.


denisok:
closed in https://github.com/percona/pmm/pull/1000

PR #948, opened 2022-06-20
PMM-7: fix ci

ritbl:


BupycHuk:
What exactly are you fixing here?

ritbl:
@BupycHuk checking if CI works on main

BupycHuk:
CI is failing only on main, I was trying to fix that, but couldn't. It requires deeper investigation to understand how merge-gatekeeper works.

PR #955, opened 2022-06-21
PMM-10145 Handle RDS not having an endpoint

michal-kralik:
- RDS instances in state "Creating" do not have an endpoint yet.
- When discovering such instances we paniced

PMM-10145

Build: SUBMODULES-2580


managed/services/management/rds.go
```diff
@@ -180,6 +180,13 @@ func (s *RDSService) DiscoverRDS(ctx context.Context, req *managementpb.Discover
 			for _, db := range regInstances {
 				l.Debugf("Discovered instance: %+v", db)
 
+				// This happens when the database is in "Creating" state.
+				// At this point there is no endpoint available.
+				if db.Endpoint == nil {
+					l.Infof("Instance %q not ready yet. Please wait until the database is fully created in AWS.", *db.DBInstanceIdentifier)
```
BupycHuk, line 186: let's print it at debug level

michal-kralik: Sure, no problem. Originally I thought this would be helpful to admins to figure out why they don't see the instances they expect, but debug is fine as well.

managed/services/management/rds.go
```diff
@@ -180,6 +180,13 @@ func (s *RDSService) DiscoverRDS(ctx context.Context, req *managementpb.Discover
 			for _, db := range regInstances {
 				l.Debugf("Discovered instance: %+v", db)
 
+				// This happens when the database is in "Creating" state.
+				// At this point there is no endpoint available.
+				if db.Endpoint == nil {
+					l.Debugf("Instance %q not ready yet. Please wait until the database is fully created in AWS.", *db.DBInstanceIdentifier)
+					continue
+				}
+
```
artemgavrilov, line 189: May be we should notify user somehow that instance is present but not ready yet?

michal-kralik: I think that's a good idea. Maybe we can do it in a new ticket just to make sure we don't panic when we encounter db in this state?

artemgavrilov: Yep, can you please create ticket for that?

artemgavrilov (approved):

PR #957, opened 2022-06-21
PM-6715 Add node name to status

michal-kralik:
PMM-6715

Build: SUBMODULES-2587

### DIFF
--- a/admin/agentlocal/agentlocal.go
+++ b/admin/agentlocal/agentlocal.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
 	"time"
 
 	"github.com/AlekSi/pointer"
@@ -64,8 +65,9 @@ var ErrNotConnected = fmt.Errorf("pmm-agent is not connected to PMM Server")
 
 // Status represents pmm-agent status.
 type Status struct {
-	AgentID string `json:"agent_id"`
-	NodeID  string `json:"node_id"`
+	AgentID  string `json:"agent_id"`
+	NodeID   string `json:"node_id"`
+	NodeName string `json:"node_name"`
 
 	ServerURL         string `json:"server_url"`
 	ServerInsecureTLS bool   `json:"server_insecure_tls"`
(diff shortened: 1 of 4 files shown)


BupycHuk (changes requested):

admin/agentlocal/agentlocal.go
```diff
@@ -152,9 +154,12 @@ func GetStatus(requestNetworkInfo NetworkInfo) (*Status, error) {
 		agentVersion = "unknown"
 	}
 
+	hostname, _ := os.Hostname()
```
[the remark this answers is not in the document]
michal-kralik: Thank you for the direction. I changed the implementation. See updated PR description for details.

admin/agentlocal/agentlocal.go
```diff
@@ -64,8 +64,9 @@ var ErrNotConnected = fmt.Errorf("pmm-agent is not connected to PMM Server")
 
 // Status represents pmm-agent status.
 type Status struct {
-	AgentID string `json:"agent_id"`
-	NodeID  string `json:"node_id"`
+	AgentID  string `json:"agent_id"`
```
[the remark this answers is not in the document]
BupycHuk: let's disable this linter or reconfigure it

managed/services/agents/registry.go
```diff
@@ -220,42 +221,51 @@ func (r *Registry) register(stream agentpb.Agent_ConnectServer) (*pmmAgentInfo,
...
+		return nil, err
 	}
 
 	agent.Version = &md.Version
 	if err := q.Update(agent); err != nil {
-		return "", errors.Wrap(err, "failed to update agent")
+		return nil, errors.Wrap(err, "failed to update agent")
+	}
+
+	// Get node name
```
artemgavrilov, line 262: We are getting more than just name here. I thins we can just drop this comment.

.golangci.yml
```diff
@@ -31,6 +31,16 @@ linters-settings:
...
 
+  tagliatelle:
+    # Check the struck tag name case.
+    case:
+      rules:
+        avro: snake
+        bson: camel
+        json: snake
+        yaml: camel
+        xml: camel
```
artemgavrilov, line 42: Right now golangci-lint broken,  @PavelKhripkov  fixed it in this PR: https://github.com/percona/pmm/pull/968
So after fix will be merged you CI will crash because we have mess with json tags, you can find both, snake case and camel case in our codebase.

michal-kralik: I disabled json and yaml for now. We have too many warnings popping up.
We can discuss if and how we want to address it later.

artemgavrilov (approved):

BupycHuk (approved):

PR #963, opened 2022-06-22
PMM-10212 Fixed sha for push events in github ci.

PavelKhripkov:
PMM-10212

Build: SUBMODULES-0

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -114,7 +114,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           interval: 45
           timeout: 1200
-          ref: ${{ github.event.pull_request.head.sha || github.event.push.head.sha }}
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
 
   workflow_success:
     needs: [merge-gatekeeper]
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -106,15 +106,16 @@ jobs:
 
   merge-gatekeeper:
     needs: [check]
+    if: ${{ always() }}
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/common.yml
```diff
@@ -114,7 +114,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           interval: 45
           timeout: 1200
-          ref: ${{ github.event.pull_request.head.sha || github.event.push.head.sha }}
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
```
artemgavrilov, line 117: In fact it seems there is no reason to run gatekeeper on pushes. May be it makes sense to extract it to the separate workflow in run it only for pull requests?

PavelKhripkov: I agree with you. I'd like even to expand the question. Do we really need to run tests against main branch? If I understand correctly, we cannot merge into main until PR tests are green.

BupycHuk: I tried to extract merge-gatekeeper into a separate flow, but it requires additional management to run it once one of the other jobs started.

.github/workflows/common.yml
```diff
@@ -109,12 +109,12 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Run Merge Gatekeeper
+        if: github.event_name == 'pull_request'
```
PavelKhripkov, line 112: The problem here: we still build `percona-platform/merge-gatekeeper@v1.0.2`

.github/workflows/common.yml
PavelKhripkov: Since merge-gatekeeper is our exit point of all checks, and notifications are based on it, we have no options to skip it.

idoqo (approved):

PR #965, opened 2022-06-22
PMM-7462 Compares env var to settings

michal-kralik:
- Settings send all configuration in a request
- This triggered an error if a setting was configured in env var
- This validates if the env var setting did not change

PMM-7462

Build: SUBMODULES-2589


BupycHuk (dismissed):

managed/services/server/server.go
```diff
@@ -550,17 +550,30 @@ func (s *Server) validateChangeSettingsRequest(ctx context.Context, req *serverp
 		return status.Error(codes.FailedPrecondition, "DBaaS is enabled via ENABLE_DBAAS or via deprecated PERCONA_TEST_DBAAS environment variable.")
 	}
 
-	if metricsRes.GetHr().AsDuration() != 0 && s.envSettings.MetricsResolutions.HR != 0 {
+	if metricsRes.GetHr().AsDuration() != 0 &&
+		s.envSettings.MetricsResolutions.HR != 0 &&
+		metricsRes.GetHr().AsDuration() != s.envSettings.MetricsResolutions.HR {
```
artemgavrilov, line 555: It looks like it can be simplified to
```suggestion
	if s.envSettings.MetricsResolutions.HR != 0 && metricsRes.GetHr().AsDuration() != s.envSettings.MetricsResolutions.HR {
```

artemgavrilov (approved):

managed/services/server/server.go
michal-kralik: Yes, that's correct. Logically it is the same condition.
I usually prefer to be more explicit in the conditions because it's easier to understand when you get back to the code later. You don't have to think twice - "and what about this case?" :)

artemgavrilov (approved):

managed/services/server/server.go
BupycHuk: actually, it's not correct.
let's imagine the case that we send a request with HR = 0 and we have env setting = 1.
So in case of new changes, it will return an error, but without these changes, it won't.
Please revert your latest changes.

BupycHuk: at the beginning there was comparison like
`0 != 0 and 1 != 0 && 1 != 0`
now it's like `1 != 0 && 1 != 0`
which is completely different

BupycHuk (changes requested):
please revert a last commit

managed/services/server/server.go
artemgavrilov: > actually, it's not correct. let's imagine the case that we send a request with HR = 0 and we have env setting = 1. So in case of new changes, it will return an error, but without these changes, it won't. Please revert your latest changes.

`HR = 0` means `unset`?

BupycHuk: yes

BupycHuk (approved):

managed/services/server/server.go
artemgavrilov: Ok, got it

artemgavrilov (approved):

managed/services/server/server.go
```diff
@@ -550,17 +550,30 @@ func (s *Server) validateChangeSettingsRequest(ctx context.Context, req *serverp
 		return status.Error(codes.FailedPrecondition, "DBaaS is enabled via ENABLE_DBAAS or via deprecated PERCONA_TEST_DBAAS environment variable.")
 	}
 
-	if metricsRes.GetHr().AsDuration() != 0 && s.envSettings.MetricsResolutions.HR != 0 {
+	if metricsRes.GetHr().AsDuration() != 0 &&
```
ShashankSinha252, line 553: Can we use a separate function to capture conditions here? Same comment for following modifications too.

ShashankSinha252 (changes requested):

managed/services/server/server.go
BupycHuk: good idea

PavelKhripkov (approved):

managed/services/server/server.go
```diff
@@ -908,7 +897,15 @@ func isAgentsStateUpdateNeeded(mr *serverpb.MetricsResolutions) bool {
...
 }
 
-// check interfaces
+// check interfaces.
 var (
 	_ serverpb.ServerServer = (*Server)(nil)
 )
+
+func canUpdateDurationSetting(setting, envSetting time.Duration) bool {
+	if setting != 0 && envSetting != 0 && setting != envSetting {
```
ShashankSinha252, line 906: Let's simplify it using De Morgan's law. It would reduce cognitive load.

managed/services/server/server.go
```diff
@@ -908,7 +897,15 @@ func isAgentsStateUpdateNeeded(mr *serverpb.MetricsResolutions) bool {
 	return true
 }
 
-// check interfaces
+// check interfaces.
 var (
 	_ serverpb.ServerServer = (*Server)(nil)
 )
+
+func canUpdateDurationSetting(setting, envSetting time.Duration) bool {
```
ShashankSinha252, line 905: Great function name!

Can we rename `setting` and `envSetting` to highlight the difference between them. As of now, it's not apparent.

ShashankSinha252 (approved):
LGTM

managed/services/server/server.go
```diff
@@ -895,7 +897,15 @@ func isAgentsStateUpdateNeeded(mr *serverpb.MetricsResolutions) bool {
...
 var (
 	_ serverpb.ServerServer = (*Server)(nil)
 )
+
+func canUpdateDurationSetting(valueToBeSet, valueFromEnvironmentVariable time.Duration) bool {
+	if valueToBeSet == 0 || valueFromEnvironmentVariable == 0 || valueToBeSet == valueFromEnvironmentVariable {
+		return true
+	}
+
+	return false
```
BupycHuk, line 910: it can even be simplified like
```suggestion
	return valueToBeSet == 0 || valueFromEnvironmentVariable == 0 || valueToBeSet == valueFromEnvironmentVariable
```

managed/services/server/server.go
```diff
@@ -895,7 +897,15 @@ func isAgentsStateUpdateNeeded(mr *serverpb.MetricsResolutions) bool {
 	return true
 }
 
-// check interfaces
+// check interfaces.
 var (
 	_ serverpb.ServerServer = (*Server)(nil)
 )
+
+func canUpdateDurationSetting(valueToBeSet, valueFromEnvironmentVariable time.Duration) bool {
```
BupycHuk, line 905: it's better to put this function above check interfaces.

BupycHuk (approved):

managed/services/server/server.go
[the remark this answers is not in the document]
michal-kralik: Thank you for your feedback. Since the PR was already approved I decided to merge. We can address this in https://github.com/percona/pmm/pull/989

Re the OR vs AND - Shashank suggested we switch to OR and I think it's easier to understand as well.
Re the function name - I prefer to use a positive name for functions, purely because if you negate the function you get `not(cannotUpdate)` which I think is harder to comprehend than `not(canUpdate)`

PR #968, opened 2022-06-23
PMM-9089 reformat tip version in ci builds matrix

PavelKhripkov:
Please let me know what Slack settings should I use in this ci

PMM-9089

Build: SUBMODULES-0


.github/workflows/go-tip-checks.yml
```diff
@@ -0,0 +1,119 @@
+name: Go-tip
+
+on:
+  schedule:
+    # Every Sunday at 13:00
+    - cron: "0 13 * * 0"
+
+jobs:
```
artemgavrilov, line 8: I guess we can just trigger all other workflows in this repo, but with `tip` as input. It should be possible.

https://github.com/percona-platform/checked/blob/54e1cde7100f25dd3b5b1fd2d076b9c2bd69ffd2/.github/workflows/ci.yml#L343

.github/workflows/admin.yml
```diff
@@ -7,26 +7,27 @@ on:
...
     name: Test pmm-admin
     runs-on: ubuntu-22.04
-    strategy:
-      fail-fast: false
-      matrix:
-        go-version: [ 1.18.x ]
-        may-fail: [ false ]
-    #        include:
-    #          - go-version: tip
-    #            may-fail: true
```
artemgavrilov, line 27: I think better to keep matrix. Let's try this approach: https://stackoverflow.com/questions/69781005/combine-dynamic-github-workflow-matrix-with-input-values-and-predefined-values

PavelKhripkov: Why do you think we need matrix if we don't use matrix-based workflows? Also there is no any answer at the link you provided. Is there any working example of the approch you described?

artemgavrilov: It will be nice to have matrix because it's easy to add new values during migration period. But yes, it seems that there is no working solution for that.

.github/workflows/common.yml
```diff
@@ -72,25 +72,30 @@ jobs:
 
       - name: Run required checks/linters
         env:
+          # use GITHUB_TOKEN because only it has access to GitHub Checks API
           REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # In case the workflow triggered by pull_request event, we want to redirect linter's output to reviewdog.
+          # The selection made through converting available cases to json and then choosing an element depending on logical value (true=1, false=0).
+          GOLANGCI_LINT_REDIRECT: ${{ fromJson('["", "| bin/reviewdog -f=golangci-lint -reporter=github-pr-review -fail-on-error=true"]')[github.event.pull_request != null] }}
+          GO_CONSISTENT_REDIRECT: ${{ fromJson('["", "| bin/reviewdog -f=go-consistent -reporter=github-pr-review -fail-on-error=true"]')[github.event.pull_request != null] }}
```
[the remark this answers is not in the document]
PavelKhripkov: In case the output always redirected to reviewdog, we cannot see the results of linters for the runs caused by non-PR events. First, reviewdog filters results of linters to show only changes related to PR. Second, reviewdog doesn't get id of PR if workflow triggered by non-PR event, so reviedog reports there is no pull request commit.

idoqo (approved):

.github/workflows/admin.yml
PavelKhripkov: I added this ability.

PavelKhripkov:
After this PR is merged, the main branch won't pass tests until all problems found by required linters are fixed.

.github/workflows/admin.yml
```diff
@@ -7,26 +7,32 @@ on:
...
       matrix:
-        go-version: [ 1.18.x ]
-        may-fail: [ false ]
-    #        include:
-    #          - go-version: tip
-    #            may-fail: true
+        go-version: ${{ fromJson(inputs.go-version || '["1.18.x"]') }}
 
     env:
+      GO_VERSION: ${{ matrix.go-version }}
```
artemgavrilov, line 35: In fact this env variable is not needed anymore, you can use `${{ matrix.go-version }}` directly everywhere

artemgavrilov (approved):
👍

.github/workflows/common.yml
```diff
@@ -70,26 +75,62 @@ jobs:
...
+          if out=$( ${{ env.COMMAND }} ); exit_code=$?; [ $exit_code -ne 0 ]; then
+            if [ $exit_code -gt 1 ] || ${{ github.event.pull_request == null }}; then
+              echo "$out"
+              exit $exit_code
+            else
+              echo "$out" ${{ env.REDIRECT }}
+            fi
+          else
+            echo "$out"
+          fi
```
artemgavrilov, line 100: Later let's replace it with https://github.com/reviewdog/action-golangci-lint

PavelKhripkov: In this case we'd need to take care of versions matching in makefile and ci. Will not that be a problem?

artemgavrilov: Can you please create ticket, let's think about that later. Docs saying that in this action they always using latest release, maybe it will work ok.

BupycHuk (approved):

.github/workflows/admin.yml
```diff
@@ -7,26 +7,32 @@ on:
...
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
-        go-version: [ 1.18.x ]
-        may-fail: [ false ]
-    #        include:
-    #          - go-version: tip
-    #            may-fail: true
+        go-version: ${{ fromJson(inputs.go-version || '["tip"]') }}
```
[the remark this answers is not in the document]
PavelKhripkov: This is only to test pr before merging. I'll reverse these changes.

Makefile.include
```diff
@@ -15,6 +15,9 @@ init:                 ## Install tools.
 	rm -rf bin/*
 	cd tools && go generate -x -tags=tools
 
+	# Install golangci-lint
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ./bin v1.46.2
+
```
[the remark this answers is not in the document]
PavelKhripkov: Devs of this tool advice users to use prebuilt binaries. We already had a problem with deps, and we want to avoid that in future.

gen1us2k (approved):

.github/workflows/common.yml
```diff
@@ -7,23 +7,28 @@ on:
...
+
 jobs:
   check:
     name: Checks
     runs-on: ubuntu-22.04
     strategy:
+      fail-fast: false
       matrix:
-        go-version: [1.18.x]
-        may-fail: [false]
```
BupycHuk, line 19: I think we should change the behavior for standard jobs. it was failing on error, but now it doesn't

.github/workflows/common.yml
PavelKhripkov: https://jira.percona.com/browse/PMM-10445

artemgavrilov (approved):

idoqo (approved):

PR #987, opened 2022-06-28
PMM-7 Make imports tests work with modules

artemgavrilov:
PMM-7

Build: SUBMODULES-0


PavelKhripkov:
I think this is not tests, this is kind of checks, and it requires to be covered with tests itself. Also there are too much responsibilities for one function. Can we split it into smaller ones to simplify understanding? And another point: four levels of nested loops is too much.

artemgavrilov:
> I think this is not tests, this is kind of checks, and it requires to be covered with tests itself. Also there are too much responsibilities for one function. Can we split it into smaller ones to simplify understanding? And another point: four levels of nested loops is too much.

While I'm agree with you, original idea was just make them work in non-GOPATH environments. So I would prefer don't do it now, but review what we can do with these checks later.

BupycHuk:
Can we extract duplicated code into a separate function and just call it everywhere?

BupycHuk (approved):

PR #990, opened 2022-06-29
PMM-10231 Credentials source bugfix

tshcherban:
PMM-10231

Build: SUBMODULES-2609

### DIFF
--- a/admin/commands/management/add_mysql.go
+++ b/admin/commands/management/add_mysql.go
@@ -143,16 +143,35 @@ func (cmd *addMySQLCommand) GetSocket() string {
 	return cmd.Socket
 }
 
+// GetCredentials figures out the credentials to be taken from CLI, credentials-source or defaults-file
 func (cmd *addMySQLCommand) GetCredentials() error {
-	creds, err := commands.ReadFromSource(cmd.CredentialsSource)
-	if err != nil {
-		return fmt.Errorf("%w", err)
+	// passed username has higher priority
+	if cmd.Username != "" {
+		return nil
 	}
 
-	cmd.AgentPassword = creds.AgentPassword
-	cmd.Password = creds.Password
-	cmd.Username = creds.Username
+	// credentials source file has 2nd priority
+	if cmd.CredentialsSource != "" {
+		creds, err := commands.ReadFromSource(cmd.CredentialsSource)
+		if err != nil {
+			return fmt.Errorf("%w", err)
+		}
 
+		cmd.AgentPassword = creds.AgentPassword
+		cmd.Password = creds.Password
+		cmd.Username = creds.Username
+
+		return nil
+	}
+
+	// username not specified, but can be in defaults file
+	// empty return ok here because later username will be read from defaults file
+	if cmd.DefaultsFile != "" {
+		return nil
+	}
+
+	// just use default user if none of above specified
+	cmd.Username = cmd.GetDefaultUsername()
 	return nil
 }
 
@@ -203,8 +222,6 @@ func (cmd *addMySQLCommand) Run() (commands.Result, error) {
 		return nil, err
 	}
 
-	username := defaultsFileUsernameCheck(cmd)
-
 	tablestatsGroupTableLimit := int32(cmd.DisableTablestatsLimit)
 	if cmd.DisableTablestats {
 		if tablestatsGroupTableLimit != 0 {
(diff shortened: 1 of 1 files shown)


admin/commands/management/add_mysql.go
```diff
@@ -143,16 +143,35 @@ func (cmd *addMySQLCommand) GetSocket() string {
 	return cmd.Socket
 }
 
+// GetCredentials figures out the credentials to be taken from CLI, credentials-source or defaults-file
 func (cmd *addMySQLCommand) GetCredentials() error {
-	creds, err := commands.ReadFromSource(cmd.CredentialsSource)
-	if err != nil {
-		return fmt.Errorf("%w", err)
+	// passed username has higher priority
```
ritbl, line 148: nit: it is not getting credential, it is mutating (if needed creds). shall we rename it?

tshcherban: yep, just reused existing method

ritbl (approved):

admin/commands/management/add_mysql.go
ritbl: I understand

admin/commands/management/add_mysql.go
```diff
@@ -143,16 +143,35 @@ func (cmd *addMySQLCommand) GetSocket() string {
 	return cmd.Socket
 }
 
+// GetCredentials figures out the credentials to be taken from CLI, credentials-source or defaults-file
 func (cmd *addMySQLCommand) GetCredentials() error {
-	creds, err := commands.ReadFromSource(cmd.CredentialsSource)
```
cezmunsta, line 147: This used to be:
```go
func (cmd *addMySQLCommand) GetCredentials() error {
       creds, err := commands.ReadFromSource(cmd.CredentialsSource)
       if err != nil {
               return fmt.Errorf("%w", err)
       }

       cmd.AgentPassword = creds.AgentPassword
       cmd.Password = creds.Password
       cmd.Username = creds.Username

       return nil
}
```

Which was called like this:
```go
       if cmd.CredentialsSource != "" {
               if err := cmd.GetCredentials(); err != nil {
                       return nil, fmt.Errorf("failed to retrieve credentials from %s: %w", cmd.CredentialsSource, err)
               }
       }
```

If `--credentials-source` is specified at all, it disregards everything else because its purpose is to mean that a user managing the PMM client has no knowledge and possibly no permission to read the source file.

What happened for it to turn into this?

tshcherban: we can specify credentials via either `credentials-source` or `CLI` or `defaults-file`, so it would be nice to have one point which would decide from where are those read

cezmunsta: It was implemented looking ahead so that a third-party (such as Percona) can deploy PMM client in a customer environment without any knowledge of the database credentials. Looking further ahead, when there is a safe way to sign and register executables to be called by the client that it could be something like a Python script, Go executable, etc that does a live lookup, such as Vault. 

In addition, It is tech-agnostic so that any tech can use the same method.

tshcherban: that part remained intact
it still takes over if `--credentials-source` specified, still throws the same error if file is inaccessible/wrong format/etc
just for brevity I manage all the sources in one method, so its clear the priorities

ritbl: @cezmunsta I see your point, admin should not have access to read creds-file, right?

So, in this case `path` to the file should be passed to pmm-managed, after that to pmm-agent. And agent can read it and pass creds back.

The complication is that pmm-managed is responsible for starting up exporters.

Now we have 2 options.
1. override behavior on pmm-agent, ie. when creds-file is set, it will use creds from this file
2. we need to add extra command to the transport layer (ie. "readConfigFile"), so that pmm-managed can instruct pmm-agent to read the file. now, pmm-managed can construct proper startup params.

IMO approach #2 is a way better, because we will be able to verify if file exist on service add command.

Later we can add functionalify on pmm-agent to monitor file changes, and trigger another event, so that pmm-managed can reload config.
cc @rnovikovP ^^

pkadej (approved):

admin/commands/management/add_mysql.go
cezmunsta: @tshcherban 

> that part remained intact

You have changed one management function, this is in all of the other relevant ones too and it looks like you changed the behaviour for only one.

@ritbl 

> The complication is that pmm-managed is responsible for starting up exporters.

The file is read, the params are set, PMM Server stores those credentials. The credentials are saved, nothing changes. What doesn't happen is the credentials being updated if the source were to return new credentials, but that isn't really an issue and instead just a caveat to its use.

> IMO approach 2 is a way better, because we will be able to verify if file exist on service add command.

That was my original intention, but at the time @BupycHuk didn't see it necessary to store the source and simply change the credentials. As it stands, without the simple update of the credentials being made when the flag is set to use the source, behaviour matches that of a standard password change and the user needs to reconfigure the agent.

BupycHuk: If there was an intention to make it sync with a file on the pmm-agent side it wasn't written anywhere in https://jira.percona.com/browse/PMM-9590. And as far as I know, our PMs didn't have information about that. So if you want to change behavior please create a separate task and let's discuss it with PMs, in that case, I would recommend merging credentials source and default source flags to make it simpler and just supporting both formats on the pmm-agent side.

Discussed with Roma and Denys and they think that it requires a bigger discussion.
Anyway, I don't think that it's in the scope of this task.

PR #993, opened 2022-06-30
PMM-7 Make target to rebuild agent in devcontainer

michal-kralik:
PMM-7

Build: SUBMODULES-2616


Makefile.devcontainer
```diff
@@ -2,22 +2,30 @@
 
 include Makefile.include
 
-release-dev:
+release-dev-managed:
```
artemgavrilov, line 5: Can you please also add comments to all these new targets. We are using them to print output of `make help`

michal-kralik:
Great feedback @artemgavrilov, thanks!

artemgavrilov (approved):

BupycHuk (approved):

BupycHuk:
looks good to me, what if we run a separate pmm-client dev container?

artemgavrilov:
> looks good to me, what if we run a separate pmm-client dev container?

Mmm, but what is the reason to do that?

BupycHuk:
@michal-kralik let's merge this PR

PR #998, opened 2022-07-01
PMM-10095 Automatic service lookup during removal

gen1us2k:
PMM-10095

Build: SUBMODULES-0


managed/services/management/service.go
```diff
@@ -56,6 +56,25 @@ func NewServiceService(db *reform.DB, state agentsStateUpdater, vmdb prometheusS
 
 // RemoveService removes Service with Agents.
 func (s *ServiceService) RemoveService(ctx context.Context, req *managementpb.RemoveServiceRequest) (*managementpb.RemoveServiceResponse, error) {
+	if e := s.db.InTransaction(func(tx *reform.TX) error {
+
+		agent, err := models.FindAgentByID(tx.Querier, req.AgentId)
+		if err != nil {
+			return err
+		}
+		serviceType := serviceTypes[req.ServiceType]
```
gen1us2k, line 65: Do not pay attention to this line. I'll make it panic-free soon. It's just for PoC demo

BupycHuk:
I would recommend keeping this logic on the pmm-admin side and changing the logic this way:
* pmm-admin sends a request to pmm-managed to get services related to the current node
* if we have only one service in the response remove that node

[here](https://github.com/percona/pmm/blob/main/admin/commands/list.go#L127-L142) you can see how to get services for the current node

BupycHuk:
we can extract duplicated code to one place

gen1us2k:
Yep. That makes sense. Thanks

admin/commands/management/remove.go
```diff
@@ -42,9 +45,53 @@ type removeMySQLCommand struct {
...
+		}
+		if len(servicesRes.Payload.Proxysql) == 1 {
+			cmd.ServiceID = servicesRes.Payload.Proxysql[0].ServiceID
+		}
+		if len(servicesRes.Payload.Haproxy) == 1 {
+			cmd.ServiceID = servicesRes.Payload.Haproxy[0].ServiceID
+		}
+		if len(servicesRes.Payload.External) == 1 {
+			cmd.ServiceID = servicesRes.Payload.External[0].ServiceID
+		}
```
BupycHuk, line 92: I think we should make sure that we have only one service, with the current logic we might have 1 MySQL, 1 MongoDB, and 1 PostgreSQL service and it will automatically remove the PostgreSQL service.

gen1us2k: I don't think so because of filtering on request by res.serviceType(). That filters services

BupycHuk: oh, good point 👍🏼

BupycHuk (approved):

admin/commands/management/remove.go
```diff
@@ -42,9 +45,53 @@ type removeMySQLCommand struct {
...
+
+	if cmd.ServiceID == "" && cmd.ServiceName == "" {
+		// PMM-10095: Automatic service lookup during removal
+		//
+		// Get services and remove it automatically once it's only one
+		// service registered
+		servicesRes, err := inventoryClient.Default.Services.ListServices(&services.ListServicesParams{
+			Body: services.ListServicesBody{
+				NodeID:      cmd.NodeID,
+				ServiceType: cmd.serviceType(),
```
gen1us2k, line 68: @BupycHuk returns services filtered by node_id and serviceType. In that case, we are sure that we'll have only one instance per service type.

BupycHuk: Yeah, you are right. Missed that

admin/commands/management/remove.go
artemgavrilov: I think adding switch statement for type can make this part more obvious

admin/commands/management/remove.go
```diff
@@ -42,9 +45,53 @@ type removeMySQLCommand struct {
 	ServiceType string
 	ServiceName string
 	ServiceID   string
+	NodeID      string
```
artemgavrilov, line 48: It seems that flag for that argument is missing

admin/commands/management/remove.go
gen1us2k: Yep. I rewrote the code

admin/commands/management/remove.go
BupycHuk: let's remove it

admin/commands/management/remove.go
```diff
@@ -45,6 +48,42 @@ type removeMySQLCommand struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if cmd.ServiceID == "" && cmd.ServiceName == "" {
+		// Automatic service lookup during removal
+		//
+		// Get services and remove it automatically once it's only one
+		// service registered
+		servicesRes, err := inventoryClient.Default.Services.ListServices(&services.ListServicesParams{
```
artemgavrilov, line 61: Is it possible that node hasn't any services?

artemgavrilov: So we will invoke `RemoveService` with empty parameters in that case? User will get descriptive error in that case?

gen1us2k: `service_id or service_name expected` message returns from the `managed` and right now it returns from the two different cases

1. We have more than one service registered on a node
2. We have no services registered on a node.

I think that I'll edit the returning message to something meaningful

gen1us2k: I just changed the default behavior of `managed` validators. Now it returns more user-friendly errors.

managed/services/management/service.go
```diff
@@ -154,10 +154,10 @@ func (s *ServiceService) checkServiceType(service *models.Service, serviceType i
 
 func (s *ServiceService) validateRequest(request *managementpb.RemoveServiceRequest) error {
 	if request.ServiceName == "" && request.ServiceId == "" {
-		return status.Error(codes.InvalidArgument, "service_id or service_name expected")
+		return status.Error(codes.InvalidArgument, "We can't find service associated with the local node, service_id or service_name expected")
 	}
 	if request.ServiceName != "" && request.ServiceId != "" {
-		return status.Error(codes.InvalidArgument, "service_id or service_name expected; not both")
+		return status.Error(codes.InvalidArgument, "We can't find service associated with the local node, service_id or service_name expected; not both")
```
BupycHuk, line 160: these errors shouldn't be returned on the backend side, because it doesn't make sense for direct API requests. So it's better to wrap the error message returned by pmm-managed on the pmm-admin side.

admin/commands/management/remove.go
```diff
@@ -55,7 +95,7 @@ func (cmd *removeMySQLCommand) Run() (commands.Result, error) {
 	}
 	_, err := client.Default.Service.RemoveService(params)
 	if err != nil {
-		return nil, err
+		return nil, errors.Errorf("We can't find service associated with the local node, %s", err)
```
BupycHuk, line 98: there might be errors not related to finding service(e.g. connection issues, or some internal server error), so this error message will be not relevant. We can check if cmd.ServiceID and cmd.ServiceName are not empty and based on that return an error in lines 85-86.

admin/commands/management/remove.go
```diff
@@ -55,6 +98,10 @@ func (cmd *removeMySQLCommand) Run() (commands.Result, error) {
 	}
 	_, err := client.Default.Service.RemoveService(params)
 	if err != nil {
+		e := commands.GetError(err.(commands.ErrorResponse))
```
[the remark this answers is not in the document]
BupycHuk: Yeah, but first we should make sure that `err` is `commands.ErrorResponse`

admin/commands/management/remove.go
```diff
@@ -55,6 +98,14 @@ func (cmd *removeMySQLCommand) Run() (commands.Result, error) {
 	}
 	_, err := client.Default.Service.RemoveService(params)
 	if err != nil {
+		e, ok := err.(commands.ErrorResponse)
+		if !ok {
+			return nil, err
+		}
+		errResponse := commands.GetError(e)
+		if autoLookup && errResponse.Code < 500 {
```
BupycHuk, line 106: We should know the exact code number for this response, so let's use it.

admin/commands/management/remove.go
```diff
@@ -45,6 +49,45 @@ type removeMySQLCommand struct {
...
+		case len(servicesRes.Payload.Postgresql) == 1:
+			cmd.ServiceID = servicesRes.Payload.Postgresql[0].ServiceID
+		case len(servicesRes.Payload.Proxysql) == 1:
+			cmd.ServiceID = servicesRes.Payload.Proxysql[0].ServiceID
+		case len(servicesRes.Payload.Haproxy) == 1:
+			cmd.ServiceID = servicesRes.Payload.Haproxy[0].ServiceID
+		case len(servicesRes.Payload.External) == 1:
+			cmd.ServiceID = servicesRes.Payload.External[0].ServiceID
+		}
+		autoLookup = true
```
BupycHuk, line 88: the original suggestion was to return the error message here since we already know that we don't pass service id or service name

admin/commands/management/remove.go
gen1us2k: Do we really need this code because we return an error above

BupycHuk: not anymore, we can completely remove this part of the code

BupycHuk (approved):

admin/commands/management/remove.go
```diff
@@ -45,6 +49,45 @@ type removeMySQLCommand struct {
...
+		case len(servicesRes.Payload.Postgresql) == 1:
+			cmd.ServiceID = servicesRes.Payload.Postgresql[0].ServiceID
+		case len(servicesRes.Payload.Proxysql) == 1:
+			cmd.ServiceID = servicesRes.Payload.Proxysql[0].ServiceID
+		case len(servicesRes.Payload.Haproxy) == 1:
+			cmd.ServiceID = servicesRes.Payload.Haproxy[0].ServiceID
+		case len(servicesRes.Payload.External) == 1:
+			cmd.ServiceID = servicesRes.Payload.External[0].ServiceID
+		}
+		if cmd.ServiceID == "" && cmd.ServiceName == "" {
```
michal-kralik, line 86: Let's simplify this by removing `cmd.ServiceName == ""` condition.
It will always be true.

admin/commands/management/remove.go
```diff
@@ -45,6 +49,45 @@ type removeMySQLCommand struct {
...
+			return nil, errors.New("We can't find a service associated with the local node. service_id or service_name is required.")
```
michal-kralik, line 87: I suggest we use "Service ID" and "Service name" instead. The arguments may change in the future and we will forget to update this text.

```suggestion
			return nil, errors.New(`we could not find a service associated with the local node. Please provide "Service ID" or "Service name"`)
```

michal-kralik (changes requested):

michal-kralik:
@gen1us2k looks good, just small comments

michal-kralik (approved):
Looks good. Please create a pmm-submodules build as well to see if everything passed.

gen1us2k:
Yes. Sure

artemgavrilov (approved):

BupycHuk (approved):

PR #999, opened 2022-07-01
PMM-6384 Fixed docker-compose for M1 CPUs

gen1us2k:
PMM-6384

Build: SUBMODULES-0

### DIFF
--- a/agent/docker-compose.yml
+++ b/agent/docker-compose.yml
@@ -31,6 +31,7 @@ services:
   # Otherwise, MySQL in Docker for Mac completely locks during/after slowlog rotation tests.
   mysql:
     image: ${MYSQL_IMAGE:-percona:5.7}
+    platform: linux/amd64
     container_name: pmm-agent_mysql
     command: >
       --sql-mode="ANSI_QUOTES"


BupycHuk (approved):

agent/docker-compose.yml
```diff
@@ -31,6 +31,7 @@ services:
   # Otherwise, MySQL in Docker for Mac completely locks during/after slowlog rotation tests.
   mysql:
     image: ${MYSQL_IMAGE:-percona:5.7}
+    platform: linux/amd64
```
ritbl, line 34: can you make it configurable, so that when MYSQL will support arm64 it can be overriden

artemgavrilov: Once MySQL will support arm64 what is the reason to have this field at all? I guess we will just drop it, no?

artemgavrilov (approved):

agent/docker-compose.yml
ritbl: What I am trying to say, it is always better to avoid hard codding.

ritbl: And the title of the PR misleading, because it is not a fix but rather a workaround.
It can extra details/motivation, ie "On arm64 mac machines, running mysql fails with `no matching manifest for linux/arm64/v8 in the manifest list entries`"

I just checked upstream, and I see that arm64 is supported (ie in mysql:8.0.29-oracle)
image

PR #1003, opened 2022-07-03
PMM-10267: reuse devcontainer and improve docker-compose

ritbl:
PMM-10267

We need to build it only once, we don't need to rely on `perconalab/pmm-server:dev-latest`, we can use the latest stable version instead.

Additionally, it fixes few bugs in current docker compose (ie setting incorrect PATH).

On mac@arm64 reduces `env-up` down to 20-30sec (before it was 1000-1500sec).
On linux@intel (8600k) down to 2-5sec (before it was 90-120sec)


Makefile
```diff
@@ -2,6 +2,22 @@
 
 include Makefile.include
 
+TARGET ?= _bash
+
+env-up-ex:                    ## Start experimental devcontainer.
```
ritbl, line 7: I created new tasks so that we can experiment to a while and decide if we want to use this approach

devcontainer.docker-compose.yml
```diff
@@ -0,0 +1,65 @@
+---
+version: '3.8'
```
pkadej, line 2: The 3.8 compose version was added in docker-compose 1.25.5, ubuntu has only 1.25.0 as stable in his 20.04 repositories. 
I had to change this to 3.7 and everything went well.

pkadej (approved):

devcontainer.docker-compose.yml
ritbl: thanks, will change it 3.7

managed/Makefile
```diff
@@ -33,15 +33,11 @@ clean:                          ## Remove generated files.
...
-	env CGO_ENABLED=0 go build -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/pmm-managed-init ./cmd/pmm-managed-init
-	env CGO_ENABLED=0 go build -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/pmm-managed-starlark ./cmd/pmm-managed-starlark
+	env CGO_ENABLED=0 go build -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/ ./...
 	$(PMM_RELEASE_PATH)/pmm-managed --version
 
 release-dev:
-	go build -race -gcflags="all=-N -l" -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/pmm-managed
-	go build -race -gcflags="all=-N -l" -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/pmm-managed-init ./cmd/pmm-managed-init
-	go build -race -gcflags="all=-N -l" -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/pmm-managed-starlark ./cmd/pmm-managed-starlark
+	go build -race -gcflags="all=-N -l" -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/ ./...
```
ritbl, line 40: Reduces compilation  time by 40-50%

idoqo: This doesn't affect our race tests? (we also run that in CI)

ritbl: Originally, it was compiling into 3 binaries, now it does the same.
Now it does the same, but with one `go build` execution.

ritbl: Yeah, it doesn't really affect compilation speed.

idoqo (approved):
Looks like managed tests are failing after moving the test files. and we should update the import directory accordingly (https://github.com/percona/pmm/blob/101fd85017c811dbf36ef34b55781fd7274a48d2/managed/cmd/pmm-managed/main_test.go#L114)

ritbl:
@idoqo sure, will update tests

tshcherban (approved):

ritbl:
@idoqo looks like these tests are not affected

artemgavrilov:
@ritbl As I remember we discussed a little bit different option. We can prebuild dev-container in CI on latest commit in `main`. And then devs will just pull and use this image.

https://jira.percona.com/browse/PMM-10146

ritbl:
@artemgavrilov this solution it flexible and doesn't require CI in place. Plus, it includes other optimization. And if you need you can update setup script and rebuild container locally.

artemgavrilov:
> we don't need to rely on perconalab/pmm-server:dev-latest, we can use the latest stable version instead.

@ritbl  but why? You create dev branch from `main`. I guess you want your devcontainer to match that state as well, no?

percona-csalguero (approved):

.devcontainer/setup-ex.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/python2
```
ShashankSinha252, line 1: [Python2 has hit EOL](https://www.python.org/doc/sunset-python-2/). Let's switch to Python3

.devcontainer/setup-ex.py
```diff
@@ -0,0 +1,106 @@
...
+
+# See CONTRIBUTING.md.
+
+from __future__ import print_function, unicode_literals
+import multiprocessing, os, subprocess, time
+
+
+GO_VERSION = os.getenv("GO_VERSION")
```
ShashankSinha252, line 9: Let's run this inside a function. We should have a single entry function for the script.

Code scattered across the script outside of functions makes it hard to understand control flow for script.

.devcontainer/setup-ex.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # make install (requires make package)
+    install_packages_p.join()
+    make_init()
+
+    # do basic setup
+    setup()
+
+
+MARKER = "/tmp/devcontainer-setup-done"
```
ShashankSinha252, line 97: Let's wrap this in a main function here. It makes control flow obvious for the script.

Makefile
```diff
@@ -2,6 +2,22 @@
...
+	TARGET="run" make env-ex
+
+env-up-ex-rebuild:            ## Start experimental with rebuild devcontainer.
+	docker-compose -f devcontainer.docker-compose.yml up --build -d
+	TARGET="run" make env-ex
+
+env-down-ex:                  ## Stop experimental devcontainer.
+	docker-compose -f devcontainer.docker-compose.yml down
+
+env-ex:  					  ## Enter devcontainer / or run TARGET
```
ShashankSinha252, line 18: Let's have the comments at same indentation level

ShashankSinha252 (changes requested):
Requested minor changes for Python script

BupycHuk (changes requested):
tests are failing

.devcontainer/setup-ex.py
ritbl: I know... if I am not mistaken we don't have python3 in the base image.

.devcontainer/setup-ex.py
ritbl: it is the same setup.py, I don't want to change it a lot for now.

.devcontainer/setup-ex.py
ritbl: let's improve python setup in next iteration, it is almost identical to setup.py.
I wanted to have this approach co-exist with existing one for a while.

devcontainer.Dockerfile
```diff
@@ -0,0 +1,19 @@
+ARG PMM_VERSION="percona/pmm-server:2.28.0"
```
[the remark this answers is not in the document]
ritbl: Agree, let's have devlatest here

.devcontainer/setup-ex.py
BupycHuk: Yeah, that's because of Centos. @nikita-b, am I right?

nikita-b: Yes, CentOS 7 has only Python2 by default. We can add it, but image will be bigger

BupycHuk (commented):
Mostly looks good to me, but please fix the test.

nikita-b (approved):

.devcontainer/setup-ex.py
ritbl: In existing we do not execute `make release-dev`, because it will be executed as part of container start up.

ritbl:
@BupycHuk I think better to avoid introducing new tasks in makefile, ask people to try new approach, after that cleanup...
What if we try new approach straight away? I don't like to have `setup.py` with `setup-ex.py` and other duplicates.

docker-compose.yml
```diff
@@ -2,14 +2,16 @@
 version: '3.7'
 services:
   pmm-managed-server:
-    image: ${PMM_SERVER_IMAGE:-perconalab/pmm-server:dev-latest}
+    build:
+      context: .
+      args:
+        PMM_VERSION: ${PMM_VERSION:-percona/pmm-server:2.28.0}
```
pkadej, line 8: shouldn't be here `pmm-server:dev-latest`?

ShashankSinha252 (approved):

ritbl:
@BupycHuk agree, updated contrib guide

.github/workflows/managed.yml
```diff
@@ -85,9 +85,6 @@ jobs:
         continue-on-error: true
         run: docker cp ~/go/pkg/mod pmm-managed-server:/root/go/pkg/mod
 
-      - name: Setup dev-container
```
ritbl, line 88: Container is already prepared at this stage

Makefile
```diff
@@ -2,23 +2,28 @@
...
+	docker pull "perconalab/pmm-server:dev-latest"
+
+env-compose-up: env-update-image
 	docker-compose up --detach --renew-anon-volumes --remove-orphans
 
 env-devcontainer:
 	docker exec -it --workdir=/root/go/src/github.com/percona/pmm pmm-managed-server .devcontainer/setup.py
 
-env-down:                                   ## Stop devcontainer.
+env-down:							## Stop devcontainer.
```
BupycHuk, line 20: minor: all comments had the same indents

BupycHuk (approved):

Makefile
```diff
@@ -2,23 +2,28 @@
 
 include Makefile.include
 
-env-up: env-compose-up env-devcontainer     ## Start devcontainer.
+env-up: env-update-image 			## Start devcontainer.
```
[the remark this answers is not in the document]
ritbl: ig is it not changed it will take few seconds, ok, removed

Makefile
ritbl: they are indented 
image

BupycHuk: probably there is combination of tabs and spaces and that's why it's showing differently on your screen and in GitHub

BupycHuk:
Feel free to merge

PR #1016, opened 2022-07-06
PMM-10118 Support multi-license in checks

idoqo:
PMM-10118

This uses https://github.com/apache/skywalking-eyes to re-enable license checks with support for the different licenses in the repo (Apache 2.0 for agent/admin, AGPL 3.0 for other components).


BupycHuk (approved):

.licenserc.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - paths:
+      - "**/managed/**/*.go"
+      - "**/api-tests/**/*.go"
+    paths-ignore:
+      - "**/*_reform.go"
+      - "**/mock_*"
+    license:
+      spdx-id: AGPL-3.0
+      content: |
+        // pmm-managed
```
artemgavrilov, line 44: The only difference between this section and previous is this line. I guess we can merge them and use `Percona Monitoring and Management` in all headers, or omit this part at all.

@atymchuk what do you think?

artemgavrilov (commented):
Great! But I want to clarify if it's possible to unify first row of the license header where we are writing component name

.licenserc.yaml
atymchuk: I prefer omitting that line.

artemgavrilov: @idoqo let's remove this first line from all headers and simplify config for license checker

idoqo: yeah, I'll simplify the license checker config file - it works fine as long as files have the license text in them - so I'll set up a script to remove the license header from the source files later and make that a separate PR.

.licenserc.yaml
```diff
@@ -0,0 +1,68 @@
+header:
+  - paths:
+      - "**/*.go"
+    paths-ignore:
+      - "**/agent/**"
+      - "**/admin/**"
+      - "**/api-tests/**"
```
artemgavrilov, line 7: I looks like we should remove `api-tests` from here, right?

artemgavrilov (approved):

artemgavrilov:
@BupycHuk Time to time `license/cla` check stuck and it's not clear how to re-trigger it.

PR #1020, opened 2022-07-08
PMM-10301 Detailed version of MySQL

rnovikovP:
PMM-10301

Build: SUBMODULES-0

### DIFF
--- a/managed/services/telemetry/config.default.yml
+++ b/managed/services/telemetry/config.default.yml
@@ -57,6 +57,14 @@ telemetry:
       - metric_name: "mysql_version"
         label: "version"
 
+- id: MySQLVersionComment
+    source: VM
+    query: mysql_version_info{service_type="mysql"}
+    summary: "Detailed information about version of monitored MySQL services"
+    data:
+      - metric_name: "mysql_version"
+        label: "version_comment"
+
   - id: MongoDBVersion
     source: VM
     query: mongodb_version_info{service_type="mongodb"}
--- a/managed/services/telemetry/config.default.yml
+++ b/managed/services/telemetry/config.default.yml
@@ -57,7 +57,7 @@ telemetry:
       - metric_name: "mysql_version"
         label: "version"
 
-- id: MySQLVersionComment
+  - id: MySQLVersionComment
     source: VM
     query: mysql_version_info{service_type="mysql"}
     summary: "Detailed information about version of monitored MySQL services"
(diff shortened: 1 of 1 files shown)


managed/services/telemetry/config.default.yml
```diff
@@ -57,6 +57,14 @@ telemetry:
       - metric_name: "mysql_version"
         label: "version"
 
+- id: MySQLVersionComment
```
BupycHuk, line 60:
```suggestion
  - id: MySQLVersionComment
```

managed/services/telemetry/config.default.yml
```diff
@@ -57,6 +57,14 @@ telemetry:
...
+    source: VM
+    query: mysql_version_info{service_type="mysql"}
+    summary: "Detailed information about version of monitored MySQL services"
+    data:
+      - metric_name: "mysql_version"
+        label: "version_comment"
```
BupycHuk, line 66: @ritbl can we include this data in the block above?

BupycHuk (changes requested):
Tests are failing.
@ritbl should we manually update any hashsums or something else?

managed/services/telemetry/config.default.yml
rnovikovP: It was my typo. this need to be a new metric name

managed/services/telemetry/config.default.yml
```diff
@@ -57,6 +57,14 @@ telemetry:
       - metric_name: "mysql_version"
         label: "version"
 
+  - id: MySQLVersionComment
+    source: VM
+    query: mysql_version_info{service_type="mysql"}
+    summary: "Detailed information about version of monitored MySQL services"
+    data:
+      - metric_name: "mysql_version_comment"
+        label: "version_comment"
```
maxkondr, line 66: it can be added to the prev block (MySQLVersion), so it will look like:
```
data:
  - metric_name: "mysql_version"
    label: "version"
  - metric_name: "mysql_version_comment"
    label: "version_comment"
```

In such a case there will be only one query execution to VM instead of 2

maxkondr (changes requested):

ritbl (approved):
LGTM

managed/services/telemetry/config.default.yml
rnovikovP: But, in this case, I will not be able to have separate descriptions for metrics?

maxkondr: you can use the following desc:

“Detailed information about version and comment of monitored MySQL services”

maxkondr (approved):

rnovikovP:
@BupycHuk, @artemgavrilov  🙏

artemgavrilov (approved):

BupycHuk (approved):

PR #1021, opened 2022-07-08
PMM-9921 rewriting pmm-client image entrypoint

nikita-b:
Move this PR from pmm-agent repo: https://github.com/percona/pmm-agent/pull/380

It was in testing and were waiting for RC to merge it. 
PMM-9921


artemgavrilov:
It's not clear from ticket description why we need this new command? Why we can't write `entrypoint.sh` instead of the new command?

nikita-b:
@artemgavrilov It'll be a really big and overcomplicated bash script.

Complex bash scripts are very hard to maintain, hard to handle errors, and hard to improve. If we have a one-page bash script we are good. But if we have more it's usually cheaper to use Python/Go.

We can't use Python in this case because we don't have Python interpretation in the image.

Also, we have only sh (dash: https://man.cx/sh), but not bash in the minimal image.

artemgavrilov (approved):

PR #1025, opened 2022-07-10
PMM-10195: Write unit tests for telemetry

dliakhov:
PMM-10195


ritbl (approved):

managed/services/telemetry/datasources.go
```diff
@@ -30,6 +30,9 @@ import (
 // DataSourceName data source name.
 type DataSourceName string
 
+//go:generate ../../../bin/mockery -name=DataSourceLocator -case=snake -inpkg -testonly
+//go:generate ../../../bin/mockery -name=DataSource -case=snake -inpkg -testonly
+
```
palash25, line 35: nitpick: we usually define a `deps.go` file to contain all the interfaces and mockery directives for a service, is there any reason we are not following this approach here? if no reason, then I think it would be better to move them to a `deps.go` file to maintain consistency across the codebase.

dliakhov: Makes sense

Created deps.go file

managed/services/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,362 @@
...
+		{
+			Key:   "key2",
+			Value: "value2",
+		},
+		{
+			Key:   "key3",
+			Value: "value3",
+		},
+	}
+	expetedReport := &reporter.ReportRequest{
```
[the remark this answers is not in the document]
dliakhov: Fixed

managed/services/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,362 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"gopkg.in/reform.v1"
+	"gopkg.in/reform.v1/dialects/postgresql"
+
+	"github.com/percona/pmm/api/serverpb"
+	"github.com/percona/pmm/managed/models"
+)
+
+func TestService_Run(t *testing.T) {
```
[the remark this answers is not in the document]
dliakhov: Makes sense

Updated

managed/services/telemetry/distribution_util.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/percona/pmm/api/serverpb"
+)
+
+//go:generate ../../../bin/mockery -name=DistributionUtilService -case=snake -inpkg -testonly
+// DistributionUtilService service to get info about OS on which pmm server is running
+type DistributionUtilService interface {
+	getDistributionMethodAndOS(l *logrus.Entry) (serverpb.DistributionMethod, pmmv1.DistributionMethod, string)
+	getLinuxDistribution(procVersion string) string
+}
```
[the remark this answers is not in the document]
dliakhov: Agree

Updated

oksana-grishchenko (approved):

managed/services/telemetry/deps.go
```diff
@@ -0,0 +1,59 @@
...
+// DataSource telemetry data source.
+type DataSource interface {
+	FetchMetrics(ctx context.Context, config Config) ([][]*pmmv1.ServerMetric_Metric, error)
+	Enabled() bool
+}
+
+// LocateTelemetryDataSource retrieves DataSource by name.
+func (s *Service) LocateTelemetryDataSource(name string) (DataSource, error) { //nolint:ireturn
+	return s.dsRegistry.LocateTelemetryDataSource(name)
+}
```
artemgavrilov, line 59: It looks like it should be placed it other file.

managed/services/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,363 @@
...
+				db:         initMockDB(t, now, 1),
+				start:      now,
+				config:     getTestConfig(true, testSourceName, 10*time.Second),
+				dsRegistry: mockDataSourceLocator(t, [][]*pmmv1.ServerMetric_Metric{expectedServerMetrics}, testSourceName, 1),
+				pmmVersion: pmmVersion,
+				dus: func(l *logrus.Entry) distributionUtilService {
+					var dusMock mockDistributionUtilService
+					dusMock.Test(t)
+					dusMock.On("getDistributionMethodAndOS", l).Return(serverpb.DistributionMethod_AMI, pmmv1.DistributionMethod_AMI, "ami").
+						Times(1)
```
artemgavrilov, line 102: To be honest I think there is no need to mock this function at all. We can just make `distributionInfoFilePath` configurable and use in tests tmp files with prepared content . In that case code will be simpler and tests will actually test that we detect distribution method correctly.

managed/services/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,363 @@
...
+			}
+			if mm.Value != gotMetric.Metrics[j].Value {
+				return false
+			}
+		}
+	}
+	return true
+}
+
+func mockDataSourceLocator(t *testing.T, metrics [][]*pmmv1.ServerMetric_Metric, testSourceName string, times int) func() DataSourceLocator {
```
artemgavrilov, line 235: It looks like to grab some real metrics is easier than mock datasources

managed/services/telemetry/telemetry_test.go
dliakhov: It's whant I wanted to do at first, but it's not possible to create this file in OS by user, it returns an error that `/srv/pmm-distribution` file can be read only

managed/services/telemetry/telemetry_test.go
dliakhov: do you mean to hardcode real metrics in the code? or grab from some real datasource in the unit tests these metrics?

managed/services/telemetry/telemetry_test.go
artemgavrilov: You can generate `tmp` files, and overwrite `distributionInfoFilePath`.

managed/services/telemetry/telemetry_test.go
artemgavrilov: Use real datasource.

managed/services/telemetry/distribution_util_test.go
```diff
@@ -0,0 +1,231 @@
...
+		DistributionMethodAMI                       = 3
+		DistributionMethodAZURE                     = 4
+		DistributionMethodDO                        = 5
+
+		PmmV1DistributionMethodDistributionMethodInvalid = 0
+		PmmV1DistributionMethodDOCKER                    = 1
+		PmmV1DistributionMethodOVF                       = 2
+		PmmV1DistributionMethodAMI                       = 3
+		PmmV1DistributionMethodAZURE                     = 4
+		PmmV1DistributionMethodDO                        = 5
```
artemgavrilov, line 56: Why not to use `serverpb.DistributionMethod_*` and `pmmv1.DistributionMethod_*` values directly?

managed/services/telemetry/distribution_util_test.go
```diff
@@ -0,0 +1,231 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			// make sure that each test will run messages in different files
+			suffixFileName := uuid.New().String()
+			tmpDistributionFilePath := fmt.Sprintf("%s%s", tmpDistributionFilePathPrefix, suffixFileName)
+			tmpOsInfoFilePath := fmt.Sprintf("%s%s", tmpOsInfoFilePathPrefix, suffixFileName)
+
+			err := writeToFile(tmpDistributionFilePath, tt.distributionName)
```
artemgavrilov, line 132: It will be nice to cleanup these files in `t.Cleanup`

managed/services/telemetry/distribution_util_test.go
```diff
@@ -0,0 +1,231 @@
...
+	)
+
+	logger := logrus.StandardLogger()
+	logger.SetLevel(logrus.DebugLevel)
+	logEntry := logrus.NewEntry(logger)
+
+	tests := []struct {
+		name             string
+		distributionName string
+		dokcerOsVersion  string
```
artemgavrilov, line 66: dokcer -> docker

managed/services/telemetry/distribution_util_test.go
```diff
@@ -0,0 +1,231 @@
...
+			want: "Amazon",
+		},
+		{
+			name: "should return Microsoft",
+			args: args{procVersion: "Microsoft"},
+			want: "Microsoft",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
artemgavrilov, line 220: Same as above.

managed/services/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,383 @@
...
+			if err := mock.ExpectationsWereMet(); err != nil {
+				t.Errorf("there were unfulfilled expectations for DB mock: %s", err)
+			}
+			defer db.Close() //nolint:errcheck
+		})
+		return reform.NewDB(db, postgresql.Dialect, reform.NewPrintfLogger(l.Printf))
+	}
+}
+
+func initGetSettingsSQLMock(mock sqlmock.Sqlmock, b []byte, now time.Time) {
```
artemgavrilov, line 337: You can just write required values to real DB and read them. Like we do it here: https://github.com/percona/pmm/blob/1fe6469ab403800d2bd339b9ca3b23ca405f0bc2/managed/models/settings_helpers_test.go#L32

managed/services/telemetry/distribution_util_test.go
dliakhov: Better to not use constants from the source code, because in theory they accidentally by mistake could be changed, and in that case we won't find the error.

artemgavrilov: These values are part of API, and we use protobuf descriptors there to detect breaking changes. I think you can skip this kind of checks here. But it's up to you.

managed/services/telemetry/distribution_util_test.go
```diff
@@ -0,0 +1,231 @@
...
+		{
+			name:             "should return Invalid distribution method",
+			distributionName: "invalid",
+			want:             DistributionMethodDistributionMethodInvalid,
+			want1:            PmmV1DistributionMethodDistributionMethodInvalid,
+			want2:            "",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
dliakhov: Got it 

Fixed

managed/services/telemetry/distribution_util_test.go
dliakhov: Fixed as well

managed/services/telemetry/telemetry_test.go
dliakhov: Got it

Updated tests

artemgavrilov:
I wonder why CI passed when tests are failed 🤔

@idoqo  can you please take a look

artemgavrilov (approved):
LGTM. It seems that tests failures are unrelated to these changes.

PR #1032, opened 2022-07-12
PMM-5005: Add field connection_uptime to the pmm-admin status

dliakhov:
[PMM-5005](https://jira.percona.com/browse/PMM-5005)

Build: SUBMODULES-2597

In the scope of this MR were added:

* Displaying status of agent even when agent is not connected to the server. Before it returned and error like this "Failed to get PMM Agent status from local pmm-agent: pmm-agent is not connected to PMM Server.". Now it returns response similar to this:

```
{
 "pmm_agent_status": {
   "agent_id": "/agent_id/d4adc705-d24d-4485-8174-7ca8792880d2",
   "node_id": "",
   "server_url": "https://127.0.0.1:443/",
   "server_insecure_tls": true,
   "server_version": "",
   "agent_version": "2.28.0-896-gaaa90bc2-dirty",
   "agents": [],
   "connection_uptime": "99.00", 
   "connected": false
 },
 "pmm_admin_version": "2.28.0-893-ge3484c6a-dirty"
}
```

where `connected: false` means that there is no connection to server.

* Added to the pmm_agent_status response field 'connection_uptime' which shows percentage of connection time between server and agent.  'connection_uptime' shows percentage uptime during 1 hour.


agent/connectionuptime/service.go
```diff
@@ -0,0 +1,121 @@
...
+	indexLastStatus     int
+	startTime           time.Time
+	lastStatusTimestamp time.Time
+}
+
+// NewService creates new instance of Service
+func NewService(windowPeriod time.Duration) *Service {
+	return &Service{
+		windowPeriodSeconds: int(windowPeriod.Seconds()),
+		connectedStatuses:   make([]bool, int(windowPeriod.Seconds())),
```
ritbl, line 39: Will it grow in size over `windowPeriod.Seconds()`? If not it is I'd suggest using array instead of slice.

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,121 @@
...
+	windowPeriodSeconds int
+	indexLastStatus     int
+	startTime           time.Time
+	lastStatusTimestamp time.Time
+}
+
+// NewService creates new instance of Service
+func NewService(windowPeriod time.Duration) *Service {
+	return &Service{
+		windowPeriodSeconds: int(windowPeriod.Seconds()),
```
ritbl, line 38: Can we have overflow here? I think it is quite possible.

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (s *Service) registerConnectionStatus(timestamp time.Time, connected bool) {
+	if s.startTime.IsZero() {
+		s.startTime = timestamp
+		s.lastStatusTimestamp = timestamp
+		s.connectedStatuses[0] = connected
+		s.indexLastStatus = 0
+
+		return
+	}
```
ritbl, line 59: It looks like some kind of workaround, why timestamp is zero? It should be `current time`, right?

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,121 @@
...
+
+	secondsFromLastEvent := int(timestamp.Unix() - s.lastStatusTimestamp.Unix())
```
ritbl, line 61: overflow?

ritbl (changes requested):
please see my comments

agent/connectionuptime/service.go
dliakhov: Condition `if s.startTime.IsZero() {` checks whether s.startTime is empty or not.

This code will be executed when this method is called first time to initialize some fields

agent/connectionuptime/service.go
dliakhov: Makes sense

Fixed that.

agent/connectionuptime/service.go
dliakhov: agree

changed type of `windowPeriodSeconds` to int64

cezmunsta:
@dliakhov as mentioned in Slack, I am starting to wonder if this approach is overly complex and a simpler approach might even be more useful.

With the current implementation, time is the controlling factor and if I lose 10 seconds then it takes another full window without issues to regain 100. Waiting for 100 is thus problematic, although that isn't to say that it is necessary to wait for 100 and instead use a threshold.

However, consider the following:

- counter starts at zero
- each failed request (connection or otherwise) increments the counter by 1
- when the counter is non-zero, each successful request (connection or otherwise) decrements the counter by 1

In that scenario:

1. OK -> 0
2. OK -> 0
3. NOK -> 1
4. NOK -> 2
5. NOK -> 3
6. OK -> 2
7. OK -> 1
8. OK -> 0
9. OK -> 0
10. NOK -> 1
11. OK -> 0

That would mean that anything non-zero is a sign of an issue, with lower values indicating either the start/recovery/flapping of a problem and higher numbers indicating a more serious problem. Should I wish to wait then it will only take _n_ successful requests instead of a much longer period of time. The field could then be "Connectivity issues" instead of the uptime. The counter could then be easily saved to a state file, or the config, so that it persists across restarts of the agent. 

Why is saving the state relevant? If there are problems, the evidence is lost by restarting the agent. If the agent is terminating itself then that could show 100 when it isn't really true.

dliakhov:
> @dliakhov as mentioned in Slack, I am starting to wonder if this approach is overly complex and a simpler approach might even be more useful.
> 
> With the current implementation, time is the controlling factor and if I lose 10 seconds then it takes another full window without issues to regain 100. Waiting for 100 is thus problematic, although that isn't to say that it is necessary to wait for 100 and instead use a threshold.
> 
> However, consider the following:
> 
>     * counter starts at zero
> 
>     * each failed request (connection or otherwise) increments the counter by 1
> 
>     * when the counter is non-zero, each successful request (connection or otherwise) decrements the counter by 1
> 
> 
> In that scenario:
> 
>     1. OK -> 0
> 
>     2. OK -> 0
> 
>     3. NOK -> 1
> 
>     4. NOK -> 2
> 
>     5. NOK -> 3
> 
>     6. OK -> 2
> 
>     7. OK -> 1
> 
>     8. OK -> 0
> 
>     9. OK -> 0
> 
>     10. NOK -> 1
> 
>     11. OK -> 0
> 
> 
> That would mean that anything non-zero is a sign of an issue, with lower values indicating either the start/recovery/flapping of a problem and higher numbers indicating a more serious problem. Should I wish to wait then it will only take _n_ successful requests instead of a much longer period of time. The field could then be "Connectivity issues" instead of the uptime. The counter could then be easily saved to a state file, or the config, so that it persists across restarts of the agent.
> 
> Why is saving the state relevant? If there are problems, the evidence is lost by restarting the agent. If the agent is terminating itself then that could show 100 when it isn't really true.
> 
> The only issue that I see with a counter is that you can exceed the type's range of values, but it is much more responsive and I am not convinced that using time avoids any of the other issues, such as flapping etc.
> 
> Any thoughts?

@ritbl @BupycHuk what do you think about this? does it make sense to show this counter value in the status information?

or would it make sense to show both these values (connection uptime and counter with number of unsuccessful connections)?

BupycHuk:
The counter doesn't make sense because we use exponential backoff to reconnect to PMM Server. At the same time, we can get a dozen of requests from the server the moment we connected to PMM Server. So the counter will immediately be decreased to 0. 
In case, we want to decrease the counter only when we have a successful connection there might be a case when we lost connection for a few minutes and the counter is increased to 5, a successful connection will decrease it to 4. And we might have this value for days.
If we want to store the current state we easily can marshal data into the state file, so it's not a problem.

dliakhov:
@cezmunsta I guess we can implement this counter in scope of another ticket, because development of this ticket already took a lot of time. 

May you please create a new ticket for implementing displaying counter in the status?

cezmunsta:
> The counter doesn't make sense because we use exponential backoff to reconnect to PMM Server. At the same time, we can get a dozen of requests from the server the moment we connected to PMM Server. So the counter will immediately be decreased to 0.

@BupycHuk no, it does make sense. The counter is for requests, not for connections. A connection is a request and so failed connections or failed requests on active connections change the counter.

I have shown @dliakhov the behaviour of the recovery:


That was before the latest changes, but nonetheless you can see that it became harder and harder to see recovery, as you can request and get the same percentage, as the difference is hidden by the significant figure count.

The counter would not be bound to a window of time, other than since the agent started. Ideally, it would persist across restarts too, although I think that its usefulness is still sufficient without that.

Imagine this:

- PMM Server is upgraded
- All agents lose n seconds of connectivity
- All agents now need to recover to 100
- One agent received a config change
- That one agent is restarted
- That one agent is 100

How do can something detect that the other n agents aren't having a problem? I have to wait 1h
So, if that is automated, something has to run for >= 1 x window for each instance
In that period, other instances could lose connectivity.

Adding a counter alongside the uptime percentage provides the following:
- Ability to determine recovery (uptime <> 100, counter = 0)
- Ability to determine flapping (uptime <> 100, counter = 0 and counter <> 0)
- Ability to determine start of issue (uptime ~ 100 and counter <> 0)

cezmunsta:
> @cezmunsta I guess we can implement this counter in scope of another ticket, because development of this ticket already took a lot of time.
> 
> May you please create a new ticket for implementing displaying counter in the status?

OK :disappointed:

BupycHuk:
@cezmunsta 
As I mentioned above we might have a dozen of requests the moment we connected to PMM Server, so it will decrease the counter to 0 in a second.

we have a field called `connected` for that purpose. So the counter will just duplicate it.
> Ability to determine recovery (uptime <> 100, counter = 0)
can be replaced by (uptime <> 100, connected = true)
> Ability to determine start of issue (uptime ~ 100 and counter <> 0)
can be replaced by (uptime ~ 100 and connected = false)
> Ability to determine flapping (uptime <> 100, counter = 0 and counter <> 0)
it's not clear how the counter might be zero and non-zero at the same time. To determine flapping we can use changes of `connected` value.

BupycHuk:
Do you want to use the status output for metrics? We have another task for internal metrics that will be exposed via `/metrics` endpoint and stored in victoria metrics.
if you compare it with scrape errors then the counter shouldn't decrease on successful connection.

cezmunsta:
> We have another task for internal metrics that will be exposed via `/metrics` endpoint 

Is there a ticket for that? Perhaps that would have the extra information.

> we might have a dozen of requests the moment we connected to PMM Server, so it will decrease the counter to 0 in a second.

So, if a connection is open (connected) and yet there is some issue pushing metrics, what happens to the connection? Does it stay open (thus 100 <> true status as metrics are not being sent). If the connection drops when there is some issue pushing then `connected` would work, but my thought about the counter for requests+connections was that you can do math on it, e.g. uptime - issues == uptime

cezmunsta:
@dliakhov 
> May you please create a new ticket for implementing displaying counter in the status?

I guess that as the status does now report and it is possible to determine at least whether or not the agent has a successful connection then perhaps we can see how well it works on its own. If metrics will be exposed another way then those could be used for closer inspection, in other words perhaps the request issue counter isn't needed.

BupycHuk:
>Is there a ticket for that? Perhaps that would have the extra information.

https://jira.percona.com/browse/PMM-4087

>So, if a connection is open (connected) and yet there is some issue pushing metrics, what happens to the connection? Does it stay open (thus 100 <> true status as metrics are not being sent). If the connection drops when there is some issue pushing then connected would work, but my thought about the counter for requests+connections was that you can do math on it, e.g. uptime - issues == uptime

Yes, it stays open. Collecting metrics about failed requests is definitely a separate task and should be implemented within the scope of that task.

admin/agentlocal/agentlocal.go
```diff
@@ -115,10 +120,6 @@ func GetStatus(requestNetworkInfo NetworkInfo) (*Status, error) {
 		return nil, ErrNotSetUp
 	}
 
-	if p.RunsOnNodeID == "" {
```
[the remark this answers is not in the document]
dliakhov: when agent can't connect to server then `RunsOnNodeID` is empty. But some other information is still available like agent id, for example. 

And instead of returning error message in this case we can return status information, but with some empty fields. Here is an example of such case: 

```{
 "pmm_agent_status": {
   "agent_id": "/agent_id/d4adc705-d24d-4485-8174-7ca8792880d2",
   "node_id": "",
   "server_url": "https://127.0.0.1:443/",
   "server_insecure_tls": true,
   "server_version": "",
   "agent_version": "2.28.0-896-gaaa90bc2-dirty",
   "agents": [],
   "connection_uptime": "99.00", 
   "connected": false
 },
 "pmm_admin_version": "2.28.0-893-ge3484c6a-dirty"
}```

File: agent/config/config.go [hunk 6]
```diff
@@ -193,6 +196,10 @@ func get(args []string, l *logrus.Entry) (cfg *Config, configFileF string, err e
 		if cfg.Ports.Max == 0 {
 			cfg.Ports.Max = 51999
 		}
+		if cfg.WindowConnectedTime == 0 {
+			cfg.WindowConnectedTime = time.Hour
```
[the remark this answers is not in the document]
dliakhov: yes

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,128 @@
...
+func (s *Service) getStartIndex(size int64) int64 {
+	startElement := s.indexLastStatus - size
+	if startElement < 0 {
+		startElement = s.windowPeriodSeconds + startElement
+	}
+	return startElement
+}
+
+func (s *Service) getNumOfConnectedSeconds(startIndex int64, totalNumOfSeconds int64) int {
+	endIndex := startIndex + totalNumOfSeconds
```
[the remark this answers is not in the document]
dliakhov: Not sure that it's possible.

`startIndex` in this case it cannot  be higher then size of bitset. And `totalNumOfSeconds` is calclualted as a difference in seconds between previous call of `registerEvent` and `toTime`. This difference should be huge in order overflow happened.

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (s *Service) calculateConnectionUpTime(toTime time.Time) float32 {
+	totalNumOfSeconds := s.getTotalNumberOfSeconds(toTime)
+	startIndex := s.getStartIndex(totalNumOfSeconds)
+	connectedSeconds := s.getNumOfConnectedSeconds(startIndex, totalNumOfSeconds)
+
+	return float32(connectedSeconds) / float32(totalNumOfSeconds) * 100
+}
+
+func (s *Service) getTotalNumberOfSeconds(toTime time.Time) int64 {
```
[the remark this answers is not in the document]
dliakhov: I'm not sure that when program is running it's possible. 

By design instead of `toTime` we should have `time.Now()`, but in this case it's hard to write good unit tests for this. And I didn't find a good way to mock creation time in golang, I decided to pass time as an argument.

In theory it's possible to have such behavior, but it would mean that service is not used correctly.

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	secondsFromLastEvent := timestamp.Unix() - s.lastStatusTimestamp.Unix()
+	for i := s.indexLastStatus + 1; i < (s.indexLastStatus + secondsFromLastEvent); i++ {
+		// set the same status to elements of previous connection status
+		s.bits.SetBit(&s.bits, int(i%s.windowPeriodSeconds), s.bits.Bit(int(s.indexLastStatus)))
+	}
+
+	s.indexLastStatus = (s.indexLastStatus + secondsFromLastEvent) % s.windowPeriodSeconds
+	s.bits.SetBit(&s.bits, int(s.indexLastStatus), getBit(connected))
```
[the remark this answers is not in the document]
dliakhov: How can we handle that correctly if `SetBit()` method accepts only int?


In theory it's possible that s.indexLastStatus will be bigger then maximum of int32, but secondsFromLastEvent is calculated as difference between two Unix timestamps and practically I'm not sure that it will exceed 2^31 value.

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,128 @@
...
+	s.lastStatusTimestamp = timestamp
+}
+
+func getBit(b bool) uint {
```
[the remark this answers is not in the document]
dliakhov: Renamed

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,128 @@
...
+	return float32(connectedSeconds) / float32(totalNumOfSeconds) * 100
+}
+
+func (s *Service) getTotalNumberOfSeconds(toTime time.Time) int64 {
+	totalNumOfSeconds := s.windowPeriodSeconds
+	diffInSecondsBetweenStartTimeAndToTime := toTime.Unix() - s.startTime.Unix()
+	if diffInSecondsBetweenStartTimeAndToTime < s.windowPeriodSeconds {
+		totalNumOfSeconds = diffInSecondsBetweenStartTimeAndToTime
+	}
+	return totalNumOfSeconds
```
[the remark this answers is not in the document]
dliakhov: Makes sense

Changed name of that variable

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,128 @@
...
+	"math/big"
+	"sync"
+	"time"
+)
+
+// Service calculates connection uptime between agent and server
+type Service struct {
+	mx sync.Mutex
+
+	bits big.Int
```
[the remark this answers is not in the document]
dliakhov: Renamed

agent/config/config.go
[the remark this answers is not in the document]
dliakhov: I removed this constant variable from there.

In this file aren't used constnats for default variables, so I think better to be consistent with that

agent/agentlocal/agent_local.go
```diff
@@ -151,12 +152,13 @@ func (s *Server) Status(ctx context.Context, req *agentlocalpb.StatusRequest) (*
...
-		AgentsInfo:     agentsInfo,
-		ConfigFilepath: s.configFilepath,
-		AgentVersion:   version.Version,
+		AgentId:          s.cfg.ID,
+		RunsOnNodeId:     md.AgentRunsOnNodeID,
+		ServerInfo:       serverInfo,
+		AgentsInfo:       agentsInfo,
+		ConfigFilepath:   s.configFilepath,
+		AgentVersion:     version.Version,
+		ConnectionUptime: float32(math.Round(float64(upTime)*100) / 100),
```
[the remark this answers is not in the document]
dliakhov: Moved

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,128 @@
...
+func NewService(windowPeriod time.Duration) *Service {
+	return &Service{
+		windowPeriodSeconds: int64(windowPeriod.Seconds()),
+	}
+}
+
+// RegisterConnectionStatus adds new connection status
+func (s *Service) RegisterConnectionStatus(timestamp time.Time, connected bool) {
+	s.mx.Lock()
+	defer s.mx.Unlock()
```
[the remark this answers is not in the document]
dliakhov: Sorry missed that point, 

Refactored

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,128 @@
...
+		s.startTime = timestamp
+		s.lastStatusTimestamp = timestamp
+		s.bits.SetBit(&s.bits, 0, getBit(connected))
+		s.indexLastStatus = 0
+
+		return
+	}
+
+	secondsFromLastEvent := timestamp.Unix() - s.lastStatusTimestamp.Unix()
+	for i := s.indexLastStatus + 1; i < (s.indexLastStatus + secondsFromLastEvent); i++ {
```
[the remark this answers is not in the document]
dliakhov: Refactored

agent/connectionuptime/service.go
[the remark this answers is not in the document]
dliakhov: I just wanted to say that `s.indexLastStatus` can't exceed the size of bitset which is used for storing connections statuses (`s.windowPeriodSeconds`)

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,156 @@
...
+import (
+	"math/big"
+	"time"
+)
+
+const channelBufferSize = 10
+
+// Service calculates connection uptime between agent and server
+type Service struct {
+	uptimeSeconds big.Int
```
BupycHuk, line 27: What is the reason to use big.Int instead of a slice of times?
the current implementation in most cases uses more CPU and more memory and at the same time harder to understand. The previous implementation could give us more opportunities to improve debug logic by including a whole array of connection and disconnection logic to logs in the future.

ritbl: >the current implementation in most cases uses more CPU and more memory and at the same time harder to understand

Implementation with bitmap (big.Int) will be very efficient.

BupycHuk: For the one-day time range, it uses 86400 bits which is 10800 bytes. Isn't it too much?
Looping through this range to get the uptime value also doesn't look efficient.

BupycHuk: I said slices of times, not booleans.

dliakhov: Okay, but slice of `time.Time` even will consume much more memory then slice of booleans, because `time.Time` struct has two variables with size of 8 bytes and one pointer. 

I guess from the perspective of memory consumption bit set will be much better.

what do you think?

dliakhov: No, I don't

I already changed the implementation of calculation connection uptime.

Now here will be stored practicallly a ring array and which element of it it means a second. And value means whether it's connected or not. So if specified window time is hour then we should allocate 3600 elements.

Every time when we register a new event, then we set the number of elements which is equal to difference between current timesampt and timestamp of previouse call in seconds.

And for calculation connection up time we just divide the number of elements which makred as `connected` (has 1 value in this bitset) to whole size of bitset (or number of passed time in seconds if it is less then the window time).

cezmunsta: > Do you expect that your system will connect-disconnect more often than each minute?

That is entirely possible

dliakhov: > That is entirely possible, if the requests are frequent enough anyway.

Okay, and it can be handled by this implementation as well.

agent/client/client.go
```diff
@@ -74,12 +75,14 @@ type Client struct {
 	rw      sync.RWMutex
 	md      *agentpb.ServerConnectMetadata
 	channel *channel.Channel
+
+	cs *connectionuptime.Service
```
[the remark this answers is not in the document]
dliakhov: updated

agent/connectionuptime/service.go
BupycHuk: @cezmunsta That sounds like another issue, barely related to this one. which might happen because of too many requests.
Could you please explain how it can bring to a chain of very often successful and unsuccessful connections? Like connected-disconnected-connected-disconnected-connected-disconnected

agent/config/config.go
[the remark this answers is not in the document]
ritbl: can you move this var to main code in used it here and in test?

ritbl (approved):

cezmunsta:
@dliakhov it seems like the the [build branch](https://github.com/Percona-Lab/pmm-submodules/pull/2597) needs a nudge :)

agent/commands/run.go
```diff
@@ -59,8 +61,13 @@ func Run() {
 		l.Debugf("Loaded configuration: %+v", cfg)
 
 		cleanupTmp(cfg.Paths.TempDir, l)
+		if connectionUptimeService == nil {
+			logrus.Infof("Window check connection time is %.2f hour(s)", cfg.WindowConnectedTime.Hours())
+			connectionUptimeService = connectionuptime.NewService(cfg.WindowConnectedTime)
+			connectionUptimeService.RunCleanupGoroutine(ctx)
+		}
```
BupycHuk, line 68: let's implement window period updating logic.

BupycHuk (approved):

agent/commands/run.go
dliakhov: Implemented that

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,176 @@
...
+		if event.Connected {
+			if i+1 >= len(c.events) {
+				connectedTimeMs += toTime.Sub(event.Timestamp).Milliseconds()
+			} else {
+				connectedTimeMs += c.events[i+1].Timestamp.Sub(event.Timestamp).Milliseconds()
+			}
+		}
+	}
+
+	totalTime := toTime.Sub(c.events[0].Timestamp).Milliseconds()
```
YaroslavPodorvanov, line 174: panic!!!

YaroslavPodorvanov: fix in https://github.com/percona/pmm/pull/1074

cezmunsta:
@dliakhov are we now starting at zero instead of 100 for the uptime?

dliakhov:
> @dliakhov are we now starting at zero instead of 100 for the uptime?

@cezmunsta what do you mean?

cezmunsta:
> @cezmunsta what do you mean?

@dliakhov I managed to get this state, but can't seem to reproduce it:
```sh
$ /home/percona/tmp/pmm-client/bin/pmm-agent setup --server-address 127.0.0.1:28443 --server-username REDACTED --server-password REDACTED --paths-base ${HOME}/tmp/pmm --server-insecure-tls $(hostname -I | cut -f1 -d' ') generic test                           
INFO[2022-07-28T07:32:48.811+01:00] Using /home/percona/tmp/pmm/exporters/node_exporter  component=setup                                                                                                           
INFO[2022-07-28T07:32:48.811+01:00] Using /home/percona/tmp/pmm/exporters/mysqld_exporter  component=setup                                                                                                         
INFO[2022-07-28T07:32:48.811+01:00] Using /home/percona/tmp/pmm/exporters/mongodb_exporter  component=setup                                                                                                        
INFO[2022-07-28T07:32:48.811+01:00] Using /home/percona/tmp/pmm/exporters/postgres_exporter  component=setup                                                                                                       
INFO[2022-07-28T07:32:48.811+01:00] Using /home/percona/tmp/pmm/exporters/proxysql_exporter  component=setup                                                                                                       
INFO[2022-07-28T07:32:48.811+01:00] Using /home/percona/tmp/pmm/exporters/rds_exporter  component=setup                                                                                                            
INFO[2022-07-28T07:32:48.811+01:00] Using /home/percona/tmp/pmm/exporters/azure_exporter  component=setup                                                                                                          
INFO[2022-07-28T07:32:48.811+01:00] Using /home/percona/tmp/pmm/exporters/vmagent  component=setup                                                                                                                 
Checking local pmm-agent status...                                                                                                                                                                                 
pmm-agent is running.                                                                                                                                                                                              
Registering pmm-agent on PMM Server...                                                                                                                                                                             
Registered.                                                                                                                                                                                                        
Configuration file /home/percona/.config/pmm/pmm-agent-test.yaml updated.                                                                                                                                          
Reloading pmm-agent configuration...                                                                     
Configuration reloaded.                                                                                                                                                                                            
Checking local pmm-agent status...                                                                       
pmm-agent is running.                                                                                    
                                                                                            
$ /home/percona/tmp/pmm-client/bin/pmm-admin status --json 2>&1 | jq                                                                                                                                              
{                                                                                                        
  "pmm_agent_status": {                                                                                  
    "agent_id": "/agent_id/442d3954-0764-4a6e-9073-0f4d6cddbef5",          
    "node_id": "/node_id/4185d05c-8295-4f88-9c3e-6be41c006921",            
    "node_name": "test",                                                                         
    "server_url": "https://127.0.0.1:28443/",       
    "server_insecure_tls": true,                                                                         
    "server_version": "2.30.0-HEAD-bcf8a701",       
    "agent_version": "2.30.0-HEAD-bcf8a701",                                                             
    "agents": [                                                                                          
      {                                        
        "agent_id": "/agent_id/e0e5cc2b-993a-4532-86c1-e95e4fd31716",           
        "agent_type": "VM_AGENT",                                                                        
        "status": "RUNNING",           
        "listen_port": 42000                                                                             
      },                                                                                                 
      {                                      
        "agent_id": "/agent_id/f15eeabd-e7df-4603-9ce7-1253ae76c4cc",         
        "agent_type": "NODE_EXPORTER",   
        "status": "RUNNING",                                                                             
        "listen_port": 42001                                                                             
      }                                                                                                  
    ],                                                                                                                                                                                                             
    "connected": true,                                                                                   
    "server_clock_drift": 70557,                   
    "server_latency": 198878,                                                                                                                                                                                      
    "connection_uptime": 3.75                                                                            
  },                                                                                                     
  "pmm_admin_version": "2.30.0-HEAD-bcf8a701"                                                                                                                                                                      
}   
```

When trying to reproduce it then I seem to get the 100 as expected. 

The following output can still appear though:
```sh
$ /home/percona/tmp/pmm-client/bin/pmm-admin status --json 2>&1 | jq
"Failed to get PMM Agent status from local pmm-agent: pmm-agent is running, but not set up.\nPlease run `pmm-admin config` with --server-url flag."
```
I guess that is OK to leave as "pmm-agent is running, but not set up" can be used

cezmunsta (approved):

YaroslavPodorvanov (approved):

pmm.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
dliakhov: got it

removed

BupycHuk (approved):

BupycHuk (approved):

YaroslavPodorvanov (approved):

ritbl (approved):
LGTM

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,168 @@
...
+// Here is example how it works.
+// When we have such set of events in connection set `f1 s1 f2`
+// where f1 - first event of failed connection
+//       s1 - first event of successful connection
+//       f2 - second event of failed connection
+//
+// method will return result using next formula `time_between(s1, f2)/time_between(f1, now)*100`
+// where time_between(s1, f2) - connection up time
+//       time_between(f1, now) - total time betweeen first event (f1) and current moment
+func (c *Service) GetConnectedUpTimeSince(toTime time.Time) float32 {
```
michal-kralik, line 142: Should we rename this to `GetConnectedUpTimeUntil` since we provide a `toTime`?

admin/agentlocal/agentlocal.go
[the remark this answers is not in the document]
michal-kralik: Before we stopped execution here.
What happens in case agent is not connected and we run a command such as `pmm-admin list` which relies on `NodeID` not being empty?

michal-kralik (changes requested):

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,168 @@
...
+// set period of time (by default it's 24 hours).
+// Method will calculate connected time as interval between connected and disconneced events
+//
+// Here is example how it works.
+// When we have such set of events in connection set `f1 s1 f2`
+// where f1 - first event of failed connection
+//       s1 - first event of successful connection
+//       f2 - second event of failed connection
+//
+// method will return result using next formula `time_between(s1, f2)/time_between(f1, now)*100`
```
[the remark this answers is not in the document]
dliakhov: This formula is correct. First version of implementation this was used that formula which is in the ticket, but it was decided to change algorithm because previous one ` num_of_successfull_requests_to_server/num_of_total_requests_to_server*100` is not accurate.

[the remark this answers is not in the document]
michal-kralik: sounds good, thank you. I suggest we update the ticket as well.

agent/config/config.go
```diff
@@ -368,6 +375,8 @@ func Application(cfg *Config) (*kingpin.Application, *string) {
 		Envar("PMM_AGENT_PORTS_MIN").Uint16Var(&cfg.Ports.Min)
 	app.Flag("ports-max", "Maximal allowed port number for listening sockets [PMM_AGENT_PORTS_MAX]").
 		Envar("PMM_AGENT_PORTS_MAX").Uint16Var(&cfg.Ports.Max)
+	app.Flag("window-connected-time", "Window time for which we tracke the status of connection between agent and server").
```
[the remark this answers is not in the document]
dliakhov: default value is specified here when specified default values for other configuration options: https://github.com/percona/pmm/pull/1032/files#diff-1308de62937b63bdfe5da80b0bf5216c4916d12ad7a40aee4843649c1ca9c02fR200

Agree, that better use .Default() method here, but we here is used another approach to set it.

It was agreed to change default period to 1 hour.

admin/agentlocal/agentlocal.go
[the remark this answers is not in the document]
dliakhov: @michal-kralik in this case I can see that `pmm-admin list` command locally returns error `Post "https://127.0.0.1:443/v1/inventory/Services/List": dial tcp 127.0.0.1:443: connect: connection refused`, when agent lost connection to server.

agent/connectionuptime/service.go
dliakhov: got it

renamed

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,168 @@
...
+// In the connection event set we store only when connection status was changed
+// (was it connected or not) in the next format:
+// {<timestamp, is_connected>, <timestamp, is_connected>, ...}
+//
+// For example:
+// {<'2022-01-01 15:00:00', true>, <'2022-01-01 15:20:00', false>, <'2022-01-01 15:20:10', true>}
+//
+// GetConnectionUpTime returns the percentage of connection uptime during
+// set period of time (by default it's 24 hours).
+// Method will calculate connected time as interval between connected and disconneced events
```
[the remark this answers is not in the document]
dliakhov: fixed

agent/connectionuptime/service.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	// Move first elements which are already expired to the start of the slice
+	// in order to not lose information about previous state of connection.
+	// The latest expired element in the slice will be the first one to calculate
+	// uptime correctly during set up window time
+	expired := time.Now().Add(-c.windowPeriod)
+	for i := len(c.events) - 1; i >= 0; i-- {
+		if c.events[i].Timestamp.Before(expired) {
+			c.events[i].Timestamp = expired.Add(time.Second)
```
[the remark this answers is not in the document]
dliakhov: yeah, it's not needed

removed

admin/agentlocal/agentlocal.go
```diff
@@ -116,10 +121,6 @@ func GetStatus(requestNetworkInfo NetworkInfo) (*Status, error) {
...
-		return nil, ErrNotConnected
```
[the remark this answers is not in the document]
dliakhov: removed

agent/connectionuptime/service.go
```diff
@@ -86,7 +86,7 @@ func (c *Service) deleteOldEvents() {
 	expired := time.Now().Add(-c.windowPeriod)
 	for i := len(c.events) - 1; i >= 0; i-- {
 		if c.events[i].Timestamp.Before(expired) {
-			c.events[i].Timestamp = expired.Add(time.Second)
+			c.events[i].Timestamp = expired
```
michal-kralik, line 89: Is there a reason we need to rewrite the event's timestamp to `expired`?

dliakhov: yes, there is

Imagine case when we have the next set of events
```[{time: 00:00, connected: true}, {time: 12:30, connected: false}, {time: 12:40, connected: true}]```

and you want to get a connection uptime during last hour from 12:40. In this case first event will be out of the period window (1 hour). 

In case if we only remove expired events then we lose information about connection between 00:00 and 12:30. And instead of removing we update (shift) the last expired timestamp to the `time.Now() - windowPeriod` time. 

And in our case events for calculation uptime will look like this: ```[{time: 11:40, connected: true}, {time: 12:30, connected: false}, {time: 12:40, connected: true}]```

Comment (michal-kralik) [reply agent/connectionuptime/service.go:89]: Thank you, that's clear now.
In these cases I usually recommend working with the window period instead of changing a timestamp. Changing the timestamp may have other unexpected consequences. The "truth" has changed.
When someone updates this part of code, they need to be aware the timestamp can change.

Imagine this situation:
1. `[{time: 00:00, connected: true}, {time: 12:30, connected: false}, {time: 12:40, connected: true}]`
2. Remove expired -> `[{time: 11:40, connected: true}, {time: 12:30, connected: false}, {time: 12:40, connected: true}]`
3. Set window period to 10 minutes
4. At 12:40 you would expect the uptime to be 0% whereas it would be >0% because we still have the 11:40 entry.

When calculating uptime, if we limited the total time to the window period instead, we could avoid this issue.
There could be more cases where this could potentially cause problems, if we relied on changing the timestamp.

I may be missing something, but I think working with the window period is less error-prone.

**Review by michal-kralik — CHANGES_REQUESTED** [review]
As discussed over slack, let's address the missing ErrNotConnected showing cryptic errors for other commands

File: agent/connectionuptime/service.go (hunk 27 above)
Comment (dliakhov) [reply agent/connectionuptime/service.go:89]: > When calculating uptime, if we limited the total time to the window period instead, we could avoid this issue.

But window period can be changed only by user who starts pmm-agent. It's defined in the configuration.

Let's take a look on our example but without changing window period:
1. We have next events [{time: 00:00, connected: true}, {time: 12:30, connected: false}, {time: 12:40, connected: true}]
1. Remove expired event (which happened more than 1 hour before)-> [{time: 12:30, connected: false}, {time: 12:40, connected: true}

In this case first event was removed and we lost info what connections status was between 11:40 and 12:30.

Comment (michal-kralik) [reply agent/connectionuptime/service.go:89]: Based on the code the window period can be [changed here](https://github.com/percona/pmm/blob/0601387fb8f62e82275507b0e15bdce7d9e127bc/agent/commands/run.go#L77). 

Not sure if it can be really changed during runtime though.
In my example I don't recommend removing events, rather limiting the uptime calculation to the window period.
This means limiting the earliest time we account for in the [uptime calculation](https://github.com/percona/pmm/blob/0601387fb8f62e82275507b0e15bdce7d9e127bc/agent/connectionuptime/service.go#L156-L166).

If you had the following events
`[{time: 00:00, connected: true}, {time: 12:30, connected: false}, {time: 12:40, connected: true}]`
The current time is 12:40 and period is 1h.
The first 00:00 event would be considered only in the period 11:40-12:30.
This way having expired events does not cause issues with calculated uptime and there's no need to touch the timestamp.

**Review by michal-kralik — CHANGES_REQUESTED** [review]
The error handling works like a charm now 👍
Let's finish the discussion around window period, expired events and timestamp changes and this is ready to go.

File: agent/connectionuptime/service.go (hunk 27 above)
Comment (dliakhov) [reply agent/connectionuptime/service.go:89]: But without removing events at one point of time user will throw error that it's out of memory, we can forever add elements to slice without removing elements. Otherwise it will be memory leak here.

>If you had the following events
[{time: 00:00, connected: true}, {time: 12:30, connected: false}, {time: 12:40, connected: true}]
The current time is 12:40 and period is 1h.
The first 00:00 event would be considered only in the period 11:40-12:30.

But to calculate this we would need to know as well from which time we can calculate connection uptime (in our case it's 11:40). I don't see much difference between what you suggest and what's already implemented. Only in my implementation we update timestamp of last expired event from which we can start to calculate connection uptime. But I agree that it's hard to understand there.

In this case instead of updating existing event's timestamp, we can remove it and set a new event with updated timestamp. wdyt about this?

Comment (dliakhov) [reply agent/connectionuptime/service.go]: >In this case instead of updating existing event's timestamp, we can remove it and set a new event with updated timestamp to make code more readable. wdyt about this?

@michal-kralik I pushed these changes, please take a look on it

Comment (michal-kralik) [reply agent/connectionuptime/service.go]: I agree we need to clear the expired events because of memory.
I suggest we change the way [uptime is calculated](https://github.com/percona/pmm/blob/1258affe0cb100984eeb9fc70ce3a39c2f3b7ff7/agent/connectionuptime/service.go#L159-L167).

Here's a pseudo-code I would suggest:

```go
expired := now - windowPeriod
for i, event := range c.events {
	if event.Connected {
		from := max(expired, event.timestamp)
		to := toTime
		if i+1 < len(c.events) {
			to = max(expired, c.events[i+1])
		}

		if from >= to {
			continue
		}

		connectedTime += to - from
	}
}
```

This way the calculation properly takes into account the window period only. And we can keep the timestamps intact.
Let me know if you think this makes sense.

michal-kralik (changes requested):
See my comment re logic of the uptime calculation. Happy to jump on a quick call if needed.

agent/connectionuptime/service.go
dliakhov: Hey Michal,

I got your point, it makes sense.
I changed implementation, please take a look on it

agent/connectionuptime/service_test.go
```diff
@@ -267,3 +267,68 @@ func compareFloatWithTolerance(a, b float32) bool {
...
+	now := time.Now()
+	tests := []struct {
+		name             string
+		setOfConnections map[time.Time]bool
+		expectedUpTime   float32
+		windowPeriod     time.Duration
+		toTime           time.Time
+	}{
+		{
+			name: "should return 50% uptime when window period is 10s, and 100% uptime when window period is 5s",
```
michal-kralik, line 281: It does not seem we're testing the second part of the test "and 100% uptime when window period is 5s".
The same applies to the following test.

Shouldn't the uptime be 80% with 5s window period?

agent/connectionuptime/service_test.go
```diff
@@ -267,3 +267,68 @@ func compareFloatWithTolerance(a, b float32) bool {
...
+				now.Add(-3 * time.Second): true,
+				now.Add(-2 * time.Second): true,
+				now.Add(-1 * time.Second): true,
+			},
+			expectedUpTime: 50,
+			windowPeriod:   10 * time.Second,
+			toTime:         now,
+		},
+		{
+			name: "should return 100% uptime when window period is 5s, and 100% uptime when window period is 10s",
```
michal-kralik, line 297: Same as above - we're not testing the 2nd part of the test name.

michal-kralik (changes requested):
Looks solid, thanks for the updates.
Last two comments re tests. I think we're missing two test cases.

agent/connectionuptime/service_test.go
dliakhov: my bad, it's copy-paste issues

renamed test

agent/connectionuptime/service_test.go
dliakhov: fixed as well

michal-kralik (changes requested):
Do we have a test to check if window period is applied properly if there are events outside the window period?
[Such as this](https://github.com/percona/pmm/pull/1032/commits/5e54187679bb59c6af37bcc8867ad616d7715f8f#diff-85baa20f141cdbf836a3b41f987d3b7fd281bd17caf2c7aba485bac50bba5528R281) but with 5 second period?

dliakhov:
> Do we have a test to check if window period is applied properly if there are events outside the window period? [Such as this](https://github.com/percona/pmm/pull/1032/commits/5e54187679bb59c6af37bcc8867ad616d7715f8f#diff-85baa20f141cdbf836a3b41f987d3b7fd281bd17caf2c7aba485bac50bba5528R281) but with 5 second period?

@michal-kralik yes, it makes sense to add them.

Added.

michal-kralik (approved):
Looks great, thank you

PR #1039, opened 2022-07-15
Check client/server versions compatibility in pmm-admin

idoqo:
PMM-10101

Build: **[SUBMODULES-2640](https://github.com/Percona-Lab/pmm-submodules/pull/2640)**

Following the comment in the linked JIRA ticket, This adds a version compatibility warning to `pmm-admin` when the client and server warnings are not in sync.

The checks here are based on the recommendations described in the [PMM upgrade documentation](https://docs.percona.com/percona-monitoring-and-management/how-to/upgrade.html).

Here's a sample output from using an older PMM server image (2.26) and the `pmm-admin` build from this branch:


admin/main_test.go
```diff
@@ -53,3 +53,54 @@ func TestVersionJson(t *testing.T) {
...
+			serverVersion: "1.19.0",
+			clientVersion: "2.28.0",
+			error:         true,
+		},
+		{
+			name:          "server version ahead of client",
+			serverVersion: "2.30.0",
+			clientVersion: "2.28.0",
+			error:         true,
+		},
```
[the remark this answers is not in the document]
idoqo: yeah, that situation is fine, but still generally advisable to keep the versions close.

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.11
 	github.com/ClickHouse/clickhouse-go/v2 v2.2.0
 	github.com/DATA-DOG/go-sqlmock v1.5.0
+	github.com/Masterminds/semver v1.5.0
```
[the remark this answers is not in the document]
idoqo: no actually, replaced semver with the `version` package now anyway - since it's what we use in other places.

BupycHuk:
Hi @idoqo, could you create a FB, please?
I think this change might break some E2E tests.

BupycHuk:
@idoqo any update on this?

admin/cli/flags/global.go
```diff
@@ -34,6 +34,7 @@ type GlobalFlags struct {
 	PMMAgentListenPort      uint32      `default:"${defaultListenPort}" help:"Set listen port of pmm-agent"`
 	JSON                    jsonFlag    `help:"Enable JSON output"`
 	Version                 versionFlag `short:"v" help:"Show application version"`
+	DisableVersionCheck     bool        `name:"disable-version-check" help:"Disables pmm-server/client compatibility checks"`
```
michal-kralik, line 37: Let's remove the `name` here. By default it's the same as specified.

admin/cmd/bootstrap.go
```diff
@@ -211,3 +218,46 @@ func processFinalError(err error, isJSON bool) {
...
+			b, jErr := json.Marshal(err.Error())
+			if jErr != nil {
+				logrus.Infof("Error: %#v.", err)
+				logrus.Panicf("Failed to marshal error to JSON.\n%s.\nPlease report this bug.", jErr)
+			}
+			fmt.Printf("%s\n", b) //nolint:forbidigo
+		} else if errors.Is(err, errMismatchedVersion) {
+			fmt.Println(err) //nolint:forbidigo
+		} else {
+			logrus.Infof("Failed to check version compatibility: %s.", err.Error())
```
michal-kralik, line 240: Let's do `Errorf` here instead.

admin/cmd/bootstrap.go
```diff
@@ -187,6 +190,10 @@ func finishBootstrap(globalFlags *flags.GlobalFlags) {
 
 	agentlocal.SetTransport(ctx, globalFlags.EnableDebug || globalFlags.EnableTrace, globalFlags.PMMAgentListenPort)
 
+	if !globalFlags.DisableVersionCheck {
+		checkVersionCompatibility(bool(globalFlags.JSON))
```
michal-kralik, line 194: Would it make sense to run this in a go routine not to block the bootstrap?

admin/cmd/bootstrap.go
```diff
@@ -187,6 +190,10 @@ func finishBootstrap(globalFlags *flags.GlobalFlags) {
 
 	agentlocal.SetTransport(ctx, globalFlags.EnableDebug || globalFlags.EnableTrace, globalFlags.PMMAgentListenPort)
 
+	if !globalFlags.DisableVersionCheck {
```
michal-kralik, line 193: We need to make sure this runs only for the `pmm-admin` command, not `pmm` or `pmm-server-upgrade`

admin/cmd/bootstrap.go
```diff
@@ -211,3 +218,46 @@ func processFinalError(err error, isJSON bool) {
...
+			if jErr != nil {
+				logrus.Infof("Error: %#v.", err)
+				logrus.Panicf("Failed to marshal error to JSON.\n%s.\nPlease report this bug.", jErr)
+			}
+			fmt.Printf("%s\n", b) //nolint:forbidigo
+		} else if errors.Is(err, errMismatchedVersion) {
+			fmt.Println(err) //nolint:forbidigo
+		} else {
+			logrus.Infof("Failed to check version compatibility: %s.", err.Error())
+		}
```
michal-kralik, line 241: Can we use `logrus` instead of `fmt` here?
It appears it shall be fully initialized by this point, including JSON formatting.

michal-kralik (changes requested):

admin/cmd/bootstrap.go
idoqo: I think even with a goroutine, we'll still have to wait for it to return before proceeding to run the invoked command, so I'm not sure it offers an advantage here.

michal-kralik: I think we don't have to wait for the output. The only outcome of the check seems to be a message in the log which can come at any point without slowing down the boot process.
Is there maybe anything I missed?

PR #1041, opened 2022-07-17
PMM-7 allow env-up-rebuild to use custom docker image

idoqo:
PMM-7

At the moment, `make env-up-rebuild` continues to pull the latest docker image and makes it impossible to use a custom docker image (say a feature build or an older image).

This removes the extra make target and allows docker-compose to continue using any custom image passed to it (using the `PMM_VERSION` env variable).

Build: SUBMODULES-0


Makefile
```diff
@@ -5,7 +5,7 @@ include Makefile.include
 env-up: 							## Start devcontainer.
 	docker-compose up -d
 
-env-up-rebuild: env-update-image	## Start devcontainer with rebuild.
+env-up-rebuild:						## Rebuild and start devcontainer
```
BupycHuk, line 8: I think it's better to change the behavior of env-update-image to pull the latest image based on some environment variable and it should be the same as the one used in docker-compose.yml.
P.S. PMM_VERSION doesn't look like a good name for this environment variable :)

idoqo: I think it used to be `PMM_SERVER_IMAGE`, will update it to use that then.

BupycHuk (approved):

ritbl (approved):
looks good! thanks @idoqo

artemgavrilov (approved):

BupycHuk (approved):

docker-compose.yml
```diff
@@ -2,12 +2,11 @@
...
-#    build:
-#      context: .
-#      args:
-#        PMM_VERSION: ${PMM_VERSION:-perconalab/pmm-server:dev-latest}
-#      dockerfile: devcontainer.Dockerfile
+    build:
+      context: .
+      args:
+        PMM_SERVER_IMAGE: ${PMM_SERVER_IMAGE:-perconalab/pmm-server:dev-container}
+      dockerfile: devcontainer.Dockerfile
```
idoqo, line 9: This doesn't have any effect on prebuilt devcontainer (since it already contains all the tools we install), but we need if $PMM_SERVER_IMAGE is set to maybe a feature build.

PR #1050, opened 2022-07-18
PMM-10218 Broken Pprof file returned on concurrent pprof request.

pkadej:
[PMM-10218](https://jira.percona.com/browse/PMM-10218) Broken Pprof file returned on concurrent pprof request

Build: SUBMODULES-2661


ritbl:
@pkadej you can lock resource, in case it is locked you can return  something like `429` (Too Many Requests) or other 4xx.

ritbl (approved):
LGTM

managed/services/supervisord/logs.go
```diff
@@ -217,38 +217,45 @@ func (l *Logs) files(ctx context.Context, pprofConfig *PprofConfig) []fileConten
 
 	// add pprof
 	if pprofConfig != nil {
+		filesSync := &sync.Mutex{}
```
pkadej, line 220: This one is not related to PR's goal. 
I found race problem (with appending to files array) and decided to fix it.

artemgavrilov, line 220:
```suggestion
		var filesSync sync.Mutex
```

artemgavrilov (approved):

BupycHuk (approved):

PR #1053, opened 2022-07-20
PMM-7: use prebuild deccontainer by default and fix CI

ritbl:


Makefile
```diff
@@ -9,7 +9,7 @@ env-up-rebuild: env-update-image	## Start devcontainer with rebuild.
 	docker-compose up --build -d
 
 env-update-image:					## Pull latest dev image
-	docker pull "perconalab/pmm-server:dev-latest"
+	docker-compose pull
```
artemgavrilov, line 12: It can be like `docker compose pull`, without dash

docker-compose.yml
```diff
@@ -2,11 +2,12 @@
...
-      context: .
-      args:
-        PMM_VERSION: ${PMM_VERSION:-perconalab/pmm-server:dev-latest}
-      dockerfile: devcontainer.Dockerfile
+    image: perconalab/pmm-server:dev-container
+#    build:
+#      context: .
+#      args:
+#        PMM_VERSION: ${PMM_VERSION:-perconalab/pmm-server:dev-latest}
+#      dockerfile: devcontainer.Dockerfile
```
artemgavrilov, line 10: Why not to remove it?

artemgavrilov (approved):

BupycHuk (approved):

Makefile
ritbl: We can, it is compose v2 format, I don't think we need to push for it. It will cause issue for people using linux.

docker-compose.yml
ritbl: It is useful then you need to build container locally.

managed/Makefile
```diff
@@ -33,11 +33,11 @@ clean:                          ## Remove generated files.
 	find . -name mock_*.go -delete
 
 release:                        ## Build pmm-managed release binaries.
-	env CGO_ENABLED=0 go build -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/ ./...
+	env CGO_ENABLED=0 go build -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/ ./cmd/...
```
ritbl, line 36: fixes issue in linux:
--- FAIL: TestStarlarkSandbox (60.02s)
    main_test.go:107: 
        	Error Trace:	/root/go/src/github.com/percona/pmm/managed/cmd/pmm-managed-starlark/main_test.go:107
        	Error:      	Received unexpected error:
        	            	exit status 2
        	Test:       	TestStarlarkSandbox
FAIL
FAIL	github.com/percona/pmm/managed/cmd/pmm-managed-starlark	60.095s

ritbl: plus, it should be faster

docker-compose.yml
```diff
@@ -55,15 +57,13 @@ services:
     volumes:
       - ./:/root/go/src/github.com/percona/pmm
       - ./Makefile.devcontainer:/root/go/src/github.com/percona/pmm/Makefile:ro # change Makefile in devcontainer
-      - pmm-bin:/root/go/src/github.com/percona/pmm/bin
```
ritbl, line 58: it will result in issues with access on linux, and you will have linux binaries in mac, so it is better to use `/root/go/bin` instead

managed/models/node_helpers_test.go
```diff
@@ -130,7 +130,7 @@ func TestNodeHelpers(t *testing.T) {
 			})
 			assert.NoError(t, err)
 
-			structs, err := q.SelectAllFrom(models.NodeTable, "WHERE machine_id = $1 ORDER BY node_id", machineID)
+			structs, err := q.SelectAllFrom(models.NodeTable, "WHERE machine_id = $1 ORDER BY node_id DESC", machineID)
```
ritbl, line 133: to fix the test

.github/workflows/managed.yml
```diff
@@ -23,6 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        use-cache: [true]
```
ritbl, line 26: by default don't use cache, as of know it is faster and CI doesn't fail due to low disk space

.github/workflows/managed.yml
```diff
@@ -23,6 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        use-cache: [false]
```
ritbl, line 26: by default don't use cache, as of know it is faster and CI doesn't fail due to low disk space

idoqo: I'm not sure the cache is the problem. Issue is We mount the cache as a volume, and the copy it to and fro with docker cp thus doubling the size. In this [PR](https://github.com/percona/pmm/pull/1049) we removed the copy and it seems to be fine.

ritbl: It is a problem
image

image

ritbl: We at least need to run on machine with more disk space. Plus caching should be revisited, because it's slows things down

idoqo: Yeah, the screenshots were from an older run, but we can revisit caching later, I guess.

idoqo (approved):

BupycHuk (approved):

.github/workflows/managed.yml
```diff
@@ -54,7 +55,11 @@ jobs:
           fetch-depth: 1
           path: ${{ env.WORKING_DIR }}
 
+      - name: Space available on machine
+        run: df -h
```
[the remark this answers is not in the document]
artemgavrilov: Yep, lets remove debug steps

.github/workflows/managed.yml
```diff
@@ -72,19 +78,25 @@ jobs:
...
         continue-on-error: true
         run: docker cp ~/.cache/go-build pmm-managed-server:/root/.cache/go-build
 
       - name: Restore Go modules cache
+        if: matrix.use-cache
         continue-on-error: true
         run: docker cp ~/go/pkg/mod pmm-managed-server:/root/go/pkg/mod
 
+      - name: Space available on machine
+        run: df -h
```
artemgavrilov, line 98: Same

docker-compose.yml
```diff
@@ -2,16 +2,18 @@
...
+#      context: .
+#      args:
+#        PMM_VERSION: ${PMM_VERSION:-perconalab/pmm-server:dev-latest}
+#      dockerfile: devcontainer.Dockerfile
     container_name: pmm-managed-server
     hostname: pmm-managed-server
     networks:
       - ${NETWORK:-default}
     environment:
+      - PMM_RELEASE_PATH=/root/go/bin
```
artemgavrilov, line 16: I guess we can remove it, `PMM_RELEASE_PATH` has a default value defined in makefile

.github/workflows/managed.yml
artemgavrilov: Please create ticket for that

artemgavrilov (approved):

docker-compose.yml
BupycHuk: I think the comment below explains why we need to override PMM_RELEASE_PATH here

.github/workflows/managed.yml
ritbl: @artemgavrilov sure

.github/workflows/managed.yml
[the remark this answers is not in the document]
ritbl: I don't see any harm, I think it is useful.
But I don't mind removing it as well.

docker-compose.yml
ritbl: @BupycHuk yes, you are correct. It is a nasty bug I was able to reproduce after a while on Linux, on mac all works.

artemgavrilov: Then I didn't get it. Makefile defines it's own `PMM_RELEASE_PATH` and all actions we perform with makefile targets. What it the reason to overwrite it?

PR #1062, opened 2022-07-23
PMM-10365: retry agent tests

ritbl:
PMM-10365

### DIFF
--- a/.github/workflows/agent.yml
+++ b/.github/workflows/agent.yml
@@ -111,13 +111,28 @@ jobs:
         run: make env-up
 
       - name: Test race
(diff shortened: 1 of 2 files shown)


artemgavrilov (changes requested):
I think it's really wrong way to solve the problem with unstable tests. It's hiding the root cause and consumes CI time.

ritbl:
@artemgavrilov as for now, it just an automation for what we repeat over and over. Yes, it needs to be fixes in long run.
But for now, it is good enough to not waste our time, repeating and waiting CI to finish.

PR #1078, opened 2022-07-28
PMM-10018 set CPU: 500m and RAM: 500M for Haproxy by default

gen1us2k:
PMM-10018

Build: SUBMODULES-0


BupycHuk (approved):

managed/services/management/dbaas/pxc_cluster_service.go
```diff
@@ -35,12 +35,14 @@ import (
...
-	proxyDefaultCpuM        = 1000
-	proxyDefaultMemoryBytes = 2000000000
+	pxcDefaultClusterSize      = 3
+	pxcDefaultCPUM             = 1000
+	pxcDefaultMemoryBytes      = 2000000000
+	pxcDefaultDiskSize         = 25000000000
+	proxyDefaultCPUM           = 500
+	proxyDefaultMemoryBytes    = 500000000
+	proxySQLDefaultCPUM        = 1000
+	proxySQLDefaultMemoryBytes = 2000000000
```
artemgavrilov, line 45: To be honest it's not obvious from the first sight what is the difference between these variables. DBasS always use haproxy? If yes, maybe, we can name variables like `haproxyDefaultCPUM` and `haproxyDefaultMemoryBytes`?

artemgavrilov (approved):

managed/services/management/dbaas/pxc_cluster_service.go
gen1us2k: After last discussion with Max Bubenick we decided to use same defaults for proxysql and for haproxy. I reverted changes

PR #1082, opened 2022-07-29
PMM-10119: automated list of gathered telemetry

arriven:
PMM-10119

Build: SUBMODULES-2688


dliakhov (approved):

managed/services/telemetry/telemetry.go
```diff
@@ -303,3 +303,12 @@ func (s *Service) Format(report *reporter.ReportRequest) string {
 
 	return builder.String()
 }
+
+// GetSummaries returns the list of gathered telemetry
+func (s *Service) GetSummaries() []string {
```
BupycHuk, line 308: if telemetry is disabled should we return an empty array?

dliakhov: Not sure about this. 
This method is used to get summaries of datapoints to display them on UI, in the Advanced Settings page.

arriven: the UI just uses this to show a tooltip on what telemetry would be collected if enabled, so we need to be able to show it all the time

api/serverpb/server.proto
```diff
@@ -197,6 +197,8 @@ message Settings {
   bool azurediscover_enabled = 17;
   // True if the PMM instance is connected to Platform
   bool connected_to_platform = 19;
+  // Includes list of collected telemetry
+  repeated string telemetry_summaries = 20;
```
artemgavrilov, line 201: Maybe it makes sense to create endpoint for that data, not sure. 🤔

artemgavrilov (approved):

BupycHuk (approved):

YaroslavPodorvanov (approved):

api/serverpb/server.proto
dliakhov: I guess better to use it here, because these summaries are used only on the Advanced settings page, and it's simpler to handle them on UI together with other data on this page.

I don't see any sense to move it to another endpoint.

PR #1110, opened 2022-08-01
PMM-10231 Service Params Source

pkadej:
[PMM-10231](https://jira.percona.com/browse/PMM-10231)

Build: SUBMODULES-2719


agent/credentialssource/credentials_source.go
```diff
@@ -0,0 +1,196 @@
...
+	credentialsJSONFile, err := parseJsonFile(filePath)
+	if err == nil {
+		return credentialsJSONFile, nil
+	}
+
+	if serviceType == inventorypb.ServiceType_MYSQL_SERVICE {
+		return parseMySQLDefaultsFile(filePath)
+	}
+
+	return nil, errors.Errorf("unrecognized file type %s", filePath)
```
BupycHuk, line 102: Maybe it's better to wrap `err` returned by `parseJsonFile`

agent/credentialssource/credentials_source.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func parseJsonFile(filePath string) (*credentialsSource, error) {
+	creds := credentialsJSON{"", "", ""}
+
+	f, err := os.Lstat(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("%w", err)
```
BupycHuk, line 110: let's use `errors.Wrapf` in all cases below with `%w`

agent/credentialssource/credentials_source.go
```diff
@@ -0,0 +1,196 @@
...
+
+	if serviceType == inventorypb.ServiceType_MYSQL_SERVICE {
+		return parseMySQLDefaultsFile(filePath)
+	}
+
+	return nil, errors.Errorf("unrecognized file type %s", filePath)
+}
+
+func parseJsonFile(filePath string) (*credentialsSource, error) {
+	creds := credentialsJSON{"", "", ""}
```
BupycHuk, line 106: no reason to initialize empty strings.

agent/credentialssource/credentials_source.go
```diff
@@ -0,0 +1,196 @@
...
+
+	f, err := os.Lstat(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("%w", err)
+	}
+
+	if f.Mode()&0o111 != 0 {
+		return nil, fmt.Errorf("%w: %s", errors.New("file execution is not supported"), filePath)
+	}
```
BupycHuk, line 115: I still think that we don't need to check the file permissions.

agent/credentialssource/credentials_source.go
```diff
@@ -0,0 +1,196 @@
...
+	parsedData := &credentialsSource{
+		username:      creds.Username,
+		password:      creds.Password,
+		agentPassword: creds.AgentPassword,
+	}
+
+	err = validateResults(parsedData)
+	if err != nil {
+		return nil, err
+	}
```
BupycHuk, line 136: if we validate data in both functions can we do it in https://github.com/percona/pmm/pull/1110/files#diff-3f486b304421a573b79690cddeb7fe149e9d4159999e8ad5bad717bbd4d85bc9R60 once we get the result?

api-tests/management/mysql_test.go
```diff
@@ -595,6 +595,7 @@ func TestAddMySQL(t *testing.T) {
 		assert.Nil(t, addMySQLOK)
 	})
 
+	// according to credentials-source parameter, passing empty username is allowed
```
BupycHuk, line 598: it should be allowed only if `credentials source` is passed. In other cases behaviour should be the same.

api-tests/management/mysql_test.go
```diff
@@ -613,11 +614,16 @@ func TestAddMySQL(t *testing.T) {
 				Address:     "10.10.10.10",
 				Port:        3306,
 				PMMAgentID:  pmmAgentID,
+
+				SkipConnectionCheck: true,
```
BupycHuk, line 618: let's revert this change

api-tests/management/mysql_test.go
BupycHuk: the same in all cases below

managed/services/agents/credentials_source_agent_invoker.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/percona/pmm/api/agentpb"
+	"github.com/percona/pmm/api/inventorypb"
+	"github.com/percona/pmm/managed/models"
+	"github.com/percona/pmm/managed/utils/logger"
+)
+
+// CredentialsSourceAgentInvoker requests from agent to parse credentials-source passed in request.
+type CredentialsSourceAgentInvoker struct {
```
BupycHuk, line 31: IMHO, it's already in `agents` directory, so `AgentInvoker` suffix is unnecessary. Maybe `CredentialsSourceLoader` or checker?

managed/services/agents/credentials_source_agent_invoker.go
```diff
@@ -0,0 +1,123 @@
...
+
+// NewCredentialsSourceAgentInvoker creates new CredentialsSourceAgentInvoker request.
+func NewCredentialsSourceAgentInvoker(r *Registry) *CredentialsSourceAgentInvoker {
+	return &CredentialsSourceAgentInvoker{
+		r: r,
+	}
+}
+
+// InvokeAgent sends request (with file path) to pmm-agent to parse credentials-source file.
+func (p *CredentialsSourceAgentInvoker) InvokeAgent(ctx context.Context, pmmAgentID, filePath string, serviceType models.ServiceType) (*models.CredentialsSourceParsingResult, error) {
```
BupycHuk, line 43: let's rename it to `Credentials`, `GetCredentials` or `LoadCredentials`

managed/services/agents/credentials_source_agent_invoker.go
```diff
@@ -0,0 +1,123 @@
...
+		Username:      parserResponse.Username,
+		Password:      parserResponse.Password,
+		AgentPassword: parserResponse.AgentPassword,
+		Host:          parserResponse.Host,
+		Port:          parserResponse.Port,
+		Socket:        parserResponse.Socket,
+	}, nil
+}
+
+func createRequest(configPath string, serviceType models.ServiceType) (*agentpb.ParseCredentialsSourceRequest, error) {
```
BupycHuk, line 88: as far as I see here only service type is different here, so we could use the same mechanism like we use in https://github.com/percona/pmm/blob/4c0aaa7cea4444ae37414ce2c43b21fd1c4772b1/managed/services/management/service.go#L31 to convert from `model` to `inventorypb`

managed/services/management/external.go
```diff
@@ -68,6 +71,18 @@ func (e *ExternalService) AddExternal(ctx context.Context, req *managementpb.Add
 		if req.AddNode != nil && runsOnNodeId == "" {
 			runsOnNodeId = nodeID
 		}
+		if req.CredentialsSource != "" {
```
BupycHuk, line 74: let's add additional checks for the field we removed validation

BupycHuk: it should be implemented for 2 cases:
1. credentials source is not passed
2. credentials source is passed, but fields aren't presented in the source.

BupycHuk (changes requested):
Let's store credentials source in database for the future usage.

agent/credentialssource/credentials_source.go
pkadej: I just moved whole logic/method from pmm-admin: https://github.com/percona/pmm/blob/main/admin/commands/base.go#L102
Removing it.

managed/services/management/mongodb.go
```diff
@@ -49,6 +54,15 @@ func (s *MongoDBService) Add(ctx context.Context, req *managementpb.AddMongoDBRe
...
 	if e := s.db.InTransaction(func(tx *reform.TX) error {
+		if req.CredentialsSource != "" {
+			result, err := s.csai.InvokeAgent(ctx, req.PmmAgentId, req.CredentialsSource, models.MongoDBServiceType)
+			if err != nil {
+				return status.Error(codes.FailedPrecondition, fmt.Sprintf("Credentials Source file error: %s.", err))
+			}
+
+			s.applyCredentialsSource(req, result)
+		}
+
```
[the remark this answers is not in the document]
pkadej: Fixed.

managed/services/agents/credentials_source_agent_invoker.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	start := time.Now()
+	defer func() {
+		if dur := time.Since(start); dur > 5*time.Second {
+			l.Warnf("Invoking agent took %s.", dur)
+		}
+	}()
```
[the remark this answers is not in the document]
pkadej: Changed.

agent/credentialssource/credentials_source.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func parseCredentialsSourceFile(filePath string, serviceType inventorypb.ServiceType) (*credentialsSource, error) {
+	if len(filePath) == 0 {
+		return nil, errors.New("configPath for parseCredentialsSourceFile is empty")
+	}
+
+	filePath, err := expandPath(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("fail to normalize path: %w", err)
```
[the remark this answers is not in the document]
pkadej: Changed.

agent/credentialssource/credentials_source.go
```diff
@@ -0,0 +1,196 @@
+// Copyright 2019 Percona LLC
```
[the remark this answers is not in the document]
pkadej: Done

agent/credentialssource/credentials_source.go
```diff
@@ -0,0 +1,196 @@
...
+	res.AgentPassword = parsedData.agentPassword
+	res.Host = parsedData.host
+	res.Port = parsedData.port
+	res.Socket = parsedData.socket
+
+	return &res
+}
+
+func parseCredentialsSourceFile(filePath string, serviceType inventorypb.ServiceType) (*credentialsSource, error) {
+	if len(filePath) == 0 {
```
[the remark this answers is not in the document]
pkadej: Changed.

agent/credentialssource/credentials_source.go
pkadej: Changed.

agent/credentialssource/credentials_source.go
pkadej: Changed.

agent/credentialssource/credentials_source.go
pkadej: @cezmunsta It shouldn't be removed right? Can we decide if should it be keeped or not?

managed/services/agents/credentials_source_agent_invoker.go
pkadej: Done.

managed/services/agents/credentials_source_agent_invoker.go
pkadej: Done.

managed/services/management/external.go
pkadej: 1. If credentials source is not passed, pmm-admin sets & sends default username.
2. Added condition: 
check if source-credentials is passed and if no username were found throws the same message as in case before removing validation (it allow us to keep backward compatibility without test/api test changes).

api-tests/management/mysql_test.go
pkadej: Changed.

api-tests/management/mysql_test.go
pkadej: Done.

managed/services/agents/credentials_source_agent_invoker.go
pkadej: Changed to `CredentialsSourceLoader`

managed/services/agents/credentials_source_loader.go
```diff
@@ -0,0 +1,106 @@
...
+		Password:      parserResponse.Password,
+		AgentPassword: parserResponse.AgentPassword,
+		Host:          parserResponse.Host,
+		Port:          parserResponse.Port,
+		Socket:        parserResponse.Socket,
+	}, nil
+}
+
+func createRequest(configPath string, serviceType models.ServiceType) (*agentpb.ParseCredentialsSourceRequest, error) {
+	inventorypbServiceType, serviceTypeExist := serviceTypes[serviceType]
```
BupycHuk, line 97: it's a common practice in go to use `ok` as a second argument to check if the key exists
```suggestion
	inventorypbServiceType, ok := serviceTypes[serviceType]
```

managed/services/management/external.go
BupycHuk: API can be used directly, so we shouldn't rely on the `pmm-admin` logic.

pkadej: Got it.
I changed in every place where the field "required" was removed from API:
- proxysql
- postgresql
- mysql

BupycHuk (approved):

pkadej:
I had to regenerate descriptor.bin, because unused "ParseDefaultsFile" fields in agent API  was renamed.
```
bin/buf breaking --against descriptor.bin api
api/agentpb/agent.proto:1:1:Previously present message "ParseDefaultsFileRequest" was deleted from file.
api/agentpb/agent.proto:1:1:Previously present message "ParseDefaultsFileResponse" was deleted from file.
api/agentpb/agent.proto:694:5:Field "20" with name "parse_credentials_source" on message "AgentMessage" changed option "json_name" from "parseDefaultsFile" to "parseCredentialsSource".
api/agentpb/agent.proto:694:5:Field "20" on message "AgentMessage" changed type from "agent.ParseDefaultsFileResponse" to "agent.ParseCredentialsSourceResponse".
api/agentpb/agent.proto:694:36:Field "20" on message "AgentMessage" changed name from "parse_defaults_file" to "parse_credentials_source".
api/agentpb/agent.proto:726:5:Field "18" with name "parse_credentials_source" on message "ServerMessage" changed option "json_name" from "parseDefaultsFile" to "parseCredentialsSource".
api/agentpb/agent.proto:726:5:Field "18" on message "ServerMessage" changed type from "agent.ParseDefaultsFileRequest" to "agent.ParseCredentialsSourceRequest".
api/agentpb/agent.proto:726:35:Field "18" on message "ServerMessage" changed name from "parse_defaults_file" to "parse_credentials_source".
```

admin/commands/management/add_proxysql.go
```diff
@@ -112,31 +106,36 @@ func (cmd *AddProxySQLCommand) RunCmd() (commands.Result, error) {
...
+	if cmd.CredentialsSource == "" {
+		if cmd.Username == "" {
+			cmd.Username = cmd.GetDefaultUsername()
+		}
+
+		if cmd.Password == "" {
+			cmd.Password = cmd.GetDefaultPassword()
+		}
+	}
+
```
michal-kralik, line 118: Is there a specific reason we replace an empty password or username if not provided?
I could imagine if both are empty, we apply defaults, but password could be an empty string, right?

admin/commands/management/add_mysql.go
```diff
@@ -89,14 +89,15 @@ func (res *addMySQLResult) TablestatStatus() string {
...
-	AgentPassword string `help:"Custom password for /metrics endpoint"`
+	ServiceName       string `name:"name" arg:"" default:"${hostname}-mysql" help:"Service name (autodetected default: ${hostname}-mysql)"`
+	Address           string `arg:"" optional:"" help:"MySQL address and port (default: 127.0.0.1:3306)"`
+	Socket            string `help:"Path to MySQL socket"`
+	NodeID            string `help:"Node ID (default is autodetected)"`
+	PMMAgentID        string `help:"The pmm-agent identifier which runs this instance (default is autodetected)"`
+	Username          string `help:"MySQL username"`
+	Password          string `help:"MySQL password"`
+	AgentPassword     string `help:"Custom password for /metrics endpoint"`
+	CredentialsSource string `help:"Credentials provider"`
```
michal-kralik, line 100: Can we please improve the help text here and in other instances of `CredentialsSource`?
It's not clear what value is expected here.

michal-kralik (changes requested):

admin/commands/management/add_proxysql.go
pkadej: Default credentials for both username and password were "admin" (line 52 - 53), so if `--credentials-source` parameter wasn't passed and `--password` also, pmm-admin should apply 'admin' as default.

michal-kralik: That's correct. There was one case though which we don't support now.

```
pmm-admin add --username=root --password=""
```

This allows you to configure an empty password while the new solution does support this.
That is, of course, unless we did not discover a bug in Kong where the command would use a default value instead :)

pkadej: Got your point, thank you.

If we run pmm-admin from main branch: 
```
pmm-admin add proxysql --username=root --password=""
```
pmm-server will get username=root and password="" in request.
For now pmm-admin allows sending empty password even if default value for proxysql is `admin`.

In this PR even if we pass or not empty parmeter `--password=""` the pmm-server will get default `admin` password in request from admin. 






That's interesting case.
Should pmm-admin support passing empty parameters/passwords even if defaults values has been declared?
CC @BupycHuk

pkadej: > Yeah, it should support it, another question is can we differentiate the flag which wasn't passed from the flag which is empty for reason?

Unfortunately, I don't have idea how we can distinguish them :disappointed:

michal-kralik: It appears we shall be able to distinguish this through [Value.Set](https://pkg.go.dev/github.com/alecthomas/kong#Value) attribute.

pkadej: Fixed. 
image

admin/commands/management/add_mysql.go
pkadej: Done.

admin/commands/management/add.go
```diff
@@ -97,3 +97,27 @@ func processGlobalAddFlagsWithSocket(cmd connectionGetter, opts AddCommonFlags)
...
+func (u username) AfterApply() error {
+	usernameParameterSpecified = true
+	return nil
+}
+
+// AfterApply Change passwordParameterSpecified flag if argument was passed
+func (p password) AfterApply() error {
+	passwordParameterSpecified = true
+	return nil
+}
```
BupycHuk, line 123: seems like these new types can be easily replaced by just pointers https://github.com/alecthomas/kong#pointers

BupycHuk (changes requested):

admin/commands/management/add.go
pkadej: This is going to be a quite long comment.

When decided to use pointers https://github.com/alecthomas/kong#pointers, pmm-admin compiled but failed to launch:
`panic: AddProxySQLCommand.Username: unsupported field type *string, perhaps missing a cmd:"" tag?`
since --username/--password are **flags** adding cmd:"" will make parameter as a **subcommand**.


I did some research and found simillar problem with pointers in subcommands: https://github.com/alecthomas/kong/issues/196. Issue was solved in latest version which pmm-admin uses, so I jumped into PRs source https://github.com/alecthomas/kong/commit/43616c6a6d5989dd9b3512e80f00ed3337b8b842#diff-679d4747ec9379321e0f2e758c26439f99115a71b4618af9713a65f65d160fb7R1286
The problem is that the fix is only for subcommands, not for flags.

I've replaced

> Cmd *cmd \`cmd:""\`

with
> Cmd   *string \`help:"PostgreSQL username"\`

and the result was the same, as for pmm-admin.


It looks like Kong's **inner commands** have problems with **pointers for parameters**.

In next step i've tried to add `cmd:""` just to see how the app will behave.
Unfortunately, it doesn't work as expected. Passed empty parameter `--username` and skipped `--password` wasn't nill but equal "" in runtime. 
So pointers won't solve this issue :( 



Should we stick to the current solution?

BupycHuk: Yes let's stick on current solution with some TODO that ot should be fixed once kong will start supporting it. And can we create an issues in kongs repository, please.

pkadej: Just checked that main branch with kong source code and it works fine. 
then find out that this code https://github.com/alecthomas/kong/pull/296 hasn't been released yet (merged 20th Sept).

There is no need to create new issue. Will add TODO that it should be fixed after kong version bumping.
Thanks!

BupycHuk: Great, thank you

BupycHuk:
Feel free to merge without waiting for my approval, just make sure that it was QA-ed

admin/commands/management/add.go
michal-kralik: @pkadej pointers don't work because they have not been release yet and are not part of v0.6.1
With the master branch they work as expected.

I'd opt for using the pointers. We can either wait until a new release is available or we can pin the module to a specific commit and upgrade as soon as the new version is out.
Either way, I tried pinging the maintainers to see when a new release is due https://github.com/alecthomas/kong/pull/296#issuecomment-1290048303

michal-kralik: v0.7.0 has just been released. We shall upgrade and use pointers.

pkadej: Done. Updated, retested and pushed. Works. Thanks :)

admin/commands/management/add_external.go
```diff
@@ -57,7 +57,7 @@ type AddExternalCommand struct {
 	RunsOnNodeID        string            `name:"agent-node-id" help:"Node ID where agent runs (default is autodetected)"`
 	Username            string            `help:"External username"`
 	Password            string            `help:"External password"`
-	CredentialsSource   string            `type:"existingfile" help:"Credentials provider"`
+	ServiceParamsSource string            `help:"Path to file with service parameters"`
```
michal-kralik, line 60: Let's please update the help here so it's clear what is expected.
With the current help it's not clear what the file shall contain.

PR #926, opened 2022-06-08
PMM-5680 store agents logs

qwest812:
[PMM-5680](https://jira.percona.com/browse/PMM-5680)

Build: SUBMODULES-2630 ~~SUBMODULES-2425~~

PR #1113, opened 2022-08-02
PMM-5680 skip with dynamic IDs now

YaroslavPodorvanov:
PMM-5680

Build: SUBMODULES-0

- [ ] https://github.com/percona/pmm/pull/926


artemgavrilov (approved):
As a quickfix it's ok. But we should check that expected files are present. We can use regexp for that, for example.

BupycHuk (approved):

PR #1114, opened 2022-08-02
PMM-7 Fix linter warnings

artemgavrilov:
PMM-7

Build: SUBMODULES-0


managed/services/management/azure_database.go
```diff
@@ -21,7 +21,7 @@ import (
 	"fmt"
 
 	"github.com/AlekSi/pointer"
-	"github.com/Azure/azure-sdk-for-go/services/resourcegraph/mgmt/2019-04-01/resourcegraph"
+	"github.com/Azure/azure-sdk-for-go/services/resourcegraph/mgmt/2019-04-01/resourcegraph" //nolint:staticcheck
```
artemgavrilov, line 24: That package is deprecated, here is a ticket to update it: https://github.com/percona/pmm/issues/1250

agent/agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -241,7 +241,7 @@ func (m *PGStatMonitorQAN) Run(ctx context.Context) {
 	if current, _, err := m.monitorCache.getStatMonitorExtended(ctx, m.q, normalizedQuery); err == nil {
 		m.monitorCache.refresh(current)
 		m.l.Debugf("Got %d initial stat monitor.", len(current))
-		running = true
+		running = true //nolint:ineffassign //TODO: @Nurlan please check this warning, it seems that we may have a bug here.
```
artemgavrilov, line 244: @BupycHuk please take a look

agent/agents/postgres/pgstatmonitor/pgstatmonitor_test.go
```diff
@@ -181,7 +181,7 @@ func TestPGStatMonitorSchema(t *testing.T) {
 	var mPlansCallsCnt, mPlansTimeCnt float32
 	pgsmVersion, _, err := getPGMonitorVersion(db.Querier)
 	assert.NoError(t, err)
-	switch pgsmVersion {
+	switch pgsmVersion { //nolint:exhaustive //TODO:@Nurlan please check if this was good warning and we should add more cases
```
artemgavrilov, line 184: @BupycHuk and that too

agent/agents/postgres/pgstatmonitor/pgstatmonitor.go
BupycHuk: looks like it's in a right place, but the next one which overwrites it is incorrect.

BupycHuk: I think the line 255 should be
```
running = running && m.checkDefaultWaitTime(waitTime)
```

agent/agents/postgres/pgstatmonitor/pgstatmonitor_test.go
BupycHuk: Looks ok for me.
@JiriCtvrtka PTAL

BupycHuk (approved):

agent/agents/postgres/pgstatmonitor/pgstatmonitor.go
artemgavrilov: I applied suggested fix, but whole this chunk looks strange, at first we may send `running` state, and just after that switch it back to `waiting`. It looks like we can at first check all conditions, and then send status.

BupycHuk: agree

JiriCtvrtka: Good idea. Thanks.

@BupycHuk is it possible to drop older versions and have compatibility only with PGSM 1.0 and higher? In next PMM versions. Code will be shorter and cleaner then and because of problems in older  PGSM versions (like wrong metrics etc) noone should use it anyway.

agent/agents/postgres/pgstatmonitor/pgstatmonitor_test.go
JiriCtvrtka: Since PMM now already supports also PGSM 1.1 and 2.0 I added these cases.

JiriCtvrtka (approved):

michal-kralik (approved):

PR #1117, opened 2022-08-02
PMM-7 Fix codecov upload for managed

artemgavrilov:
PMM-7

https://community.codecov.com/t/cant-make-ignore-option-work-for-directory-with-strict-permissions/3814


.github/workflows/admin.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         run: make test-cover
 
       - name: Upload coverage results
-        uses: percona-platform/codecov-action@v3
+        uses: codecov/codecov-action@v3
```
artemgavrilov, line 79: @atymchuk please add original codecov action to allow list of actions

BupycHuk (approved):

oksana-grishchenko (approved):

.github/workflows/admin.yml
atymchuk: I checked - it's there, so we should be good to go

atymchuk (approved):

PR #1119, opened 2022-08-03
PMM-10455: batch telemetry sending

ritbl:
PMM-10455


managed/services/telemetry/telemetry.go
```diff
@@ -129,12 +132,7 @@ func (s *Service) Run(ctx context.Context) {
 		s.l.Debugf("\nTelemetry captured:\n%s\n", s.Format(report))
 
 		if s.config.Reporting.Send {
-			err := s.send(ctx, report)
-			if err != nil {
-				s.l.Debugf("Telemetry info not sent, due to error: %s.", err)
-				return
-			}
-			s.l.Debug("Telemetry info sent.")
+			s.sendCh <- report
```
ritbl, line 135: do not block timer ticking

managed/services/telemetry/telemetry.go
```diff
@@ -160,7 +160,47 @@ func (s *Service) DistributionMethod() serverpb.DistributionMethod {
 	return s.sDistributionMethod
 }
 
-func (s *Service) prepareReport(ctx context.Context) *reporter.ReportRequest {
+func (s *Service) processSendCh(ctx context.Context) {
+	var inflightReports []*pmmv1.ServerMetric
```
ritbl, line 164: buffer where we hold not sent reports

managed/services/telemetry/telemetry.go
```diff
@@ -160,7 +160,47 @@ func (s *Service) DistributionMethod() serverpb.DistributionMethod {
...
+	var sendCtx context.Context
+	var cancel context.CancelFunc
+
+	for {
+		select {
+		case report, ok := <-s.sendCh:
+			if ok {
+				inflightReports = append(inflightReports, report)
+				if sendCtx != nil {
```
ritbl, line 173: if previous request is in flight (e.g. in retry loop) cancel it

managed/services/telemetry/telemetry.go
```diff
@@ -160,7 +160,47 @@ func (s *Service) DistributionMethod() serverpb.DistributionMethod {
...
+					cancel()
+				}
+				sendCtx, cancel = context.WithCancel(context.Background())
+
+				go func(ctx context.Context, reports *[]*pmmv1.ServerMetric) {
+					reportsCopy := make([]*pmmv1.ServerMetric, len(*reports))
+					copy(reportsCopy, *reports)
```
ritbl, line 180: send copy of slice, to avoid potential concurrent modification

managed/services/telemetry/telemetry.go
```diff
@@ -160,7 +160,47 @@ func (s *Service) DistributionMethod() serverpb.DistributionMethod {
...
+func (s *Service) processSendCh(ctx context.Context) {
+	var inflightReports []*pmmv1.ServerMetric
+	var sendCtx context.Context
+	var cancel context.CancelFunc
+
+	for {
+		select {
+		case report, ok := <-s.sendCh:
+			if ok {
+				inflightReports = append(inflightReports, report)
```
artemgavrilov, line 172: So `inflightReports` can grow ethereally if we have some problem and some report always can't be sent?

ritbl: it will grow if it will be not possible to send, even after years of not sending it should now consume more than few megs of ram

ritbl: after successful send, it will be cleaned up

managed/services/telemetry/telemetry.go
BupycHuk: I think we should copy in the main goroutine avoid concurrency conflicts

managed/services/telemetry/telemetry.go
BupycHuk: Could you point to the logic where we clean up it? I can't find in the code where we clean up this slice.

BupycHuk: oh, I see, we should make sure that we won't have a race condition because of access to the same array from a multiple places.

BupycHuk (changes requested):

managed/services/telemetry/telemetry.go
artemgavrilov: May be it's not for now, but I suggest to drop old telemetry data if queue size is exceeded.

managed/services/telemetry/telemetry.go
```diff
@@ -160,7 +160,47 @@ func (s *Service) DistributionMethod() serverpb.DistributionMethod {
...
+				if sendCtx != nil {
+					cancel()
+				}
+				sendCtx, cancel = context.WithCancel(context.Background())
```
maxkondr, line 176: shouldn't it inherit context passed to `processSendCh`?

managed/services/telemetry/telemetry.go
```diff
@@ -160,7 +160,47 @@ func (s *Service) DistributionMethod() serverpb.DistributionMethod {
...
+					reportsCopy := make([]*pmmv1.ServerMetric, len(*reports))
+					copy(reportsCopy, *reports)
+					err := s.send(ctx, &reporter.ReportRequest{
+						Metrics: reportsCopy,
+					})
+					if err != nil {
+						s.l.Debugf("Telemetry info not sent, due to error: %s.", err)
+						return
+					}
+					*reports = nil
```
maxkondr, line 188: here it is possible that the initial `inflightReports` form line 172 contains more records that have been copied in line 180 -> data loss

maxkondr (changes requested):

managed/services/telemetry/telemetry.go
ritbl: >Could you point to the logic where we clean up it? I can't find in the code where we clean up this slice.

L188

>we should make sure that we won't have a race condition because of access to the same array from a multiple places.

it is being copied before it is used, to avoid race

managed/services/telemetry/telemetry.go
ritbl: yeah, make sense. thanks

managed/services/telemetry/telemetry.go
ritbl: yeah, copy probably not an atomic. thanks

managed/services/telemetry/telemetry.go
ritbl: agree

managed/services/telemetry/telemetry.go
```diff
@@ -160,7 +161,60 @@ func (s *Service) DistributionMethod() serverpb.DistributionMethod {
...
+					})
+					if err != nil {
+						s.l.Debugf("Telemetry info not sent, due to error: %s.", err)
+						return
+					}
+
+					cleanup()
+
+					s.l.Debug("Telemetry info sent.")
+				}(sendCtx, reportsCopy, cleanup)
```
BupycHuk, line 206: do we really need to pass these arguments, they are all available in a function and don't save us from a race condition.

managed/services/telemetry/telemetry.go
BupycHuk: but at line 188 we were setting nil value. that's what I meant. I see that you fixed it.

BupycHuk (approved):

BupycHuk:
Please fix linters

managed/services/telemetry/telemetry.go
ritbl: I know it can be a closure, but I think it is better to explicitly specify what is used, and it is personal preference for cases where we have concurrency. Please let me know if you think it is better to use varibles directly.

dliakhov (approved):

managed/services/telemetry/telemetry.go
```diff
@@ -160,7 +161,60 @@ func (s *Service) DistributionMethod() serverpb.DistributionMethod {
...
+				go func(ctx context.Context, reports []*pmmv1.ServerMetric, cleanup func()) {
+					err := s.send(ctx, &reporter.ReportRequest{
+						Metrics: reports,
+					})
+					if err != nil {
+						s.l.Debugf("Telemetry info not sent, due to error: %s.", err)
+						return
+					}
+
+					cleanup()
```
maxkondr, line 203: Here is a risk of loosing data in `inflightReports`. by the time when this function is called it is possible that `inflightReports` already contains more records that have been copied in L191.

maxkondr (changes requested):

managed/services/telemetry/telemetry.go
ritbl: ok, fixed

managed/services/telemetry/telemetry.go
```diff
@@ -160,7 +161,61 @@ func (s *Service) DistributionMethod() serverpb.DistributionMethod {
...
+				inflightReports = append(inflightReports, report)
+				sendCtx, cancel = context.WithCancel(ctx)
+
+				reportsCopy := make([]*pmmv1.ServerMetric, len(inflightReports))
+				copy(reportsCopy, inflightReports)
+				reportsSync.Unlock()
+
+				go func(ctx context.Context, reports []*pmmv1.ServerMetric, cleanup func()) {
+					reportsSync.Lock()
+					defer reportsSync.Unlock()
```
maxkondr, line 194: now there is no sense in moving this functionality into a separate go routine since the mutex is locked and L184 will wait while the sending is finished -> usual sync sending

ritbl: Don't agree, request will be canceled in other goroutine, this part is out of critical section

maxkondr (approved):

PR #1125, opened 2022-08-04
PMM-9946 Automatically add k8s cluster

gen1us2k:
* Check if PMM is inside k8s cluster
* Get config and convert it to kubeconfig
* Trigger register cluster call

PMM-0

Build: SUBMODULES-0


gen1us2k:
There's no need for review at this moment. I'm just trying to create PoC of this automation

managed/services/dbaas/dbaas_client.go
```diff
@@ -54,6 +55,30 @@ func NewClient(dbaasControllerAPIAddress string) *Client {
...
+	}
+	c.k8sClient = client
+	return nil
+}
+func (c *Client) GetKubeConfig() (string, error) {
+
+	if err := c.initializeInClusterK8sClient(); err != nil {
+		return "", err
+	}
+	if err := c.Connect(context.Background()); err != nil {
```
gen1us2k, line 71: @BupycHuk I don't like connecting to dbaas-controller here. It kind of breaks logic of enabling/disabling dbaas feature. Any suggestions here?

managed/services/dbaas/kubernetes_client.go
```diff
@@ -0,0 +1,140 @@
+package dbaas
+
+import (
+	"context"
+	"encoding/json"
+
```
gen1us2k, line 6: Tests are missing here. I'll fix it soon

managed/services/dbaas/dbaas_client.go
BupycHuk: in the case of disabled dbaas, we should return an empty value by default.

managed/services/management/dbaas/dbaas_initializer.go
```diff
@@ -0,0 +1,251 @@
...
+// Enable enables dbaas feature and connects to dbaas-controller
+func (in *Initializer) Enable(ctx context.Context) error {
+	in.m.Lock()
+	defer in.m.Unlock()
+	if in.enabled {
+		return nil
+	}
+	timeoutCtx, cancel := context.WithTimeout(ctx, 20*time.Second)
+	err := in.dbaasClient.Connect(timeoutCtx)
+	cancel()
```
oksana-grishchenko, line 90: shouldn't it be 
```
defer cancel()
``` 
here?

managed/services/management/dbaas/dbaas_initializer.go
```diff
@@ -0,0 +1,251 @@
...
+		req := &dbaasv1beta1.RegisterKubernetesClusterRequest{
+			KubernetesClusterName: defaultClusterName,
+			KubeAuth: &dbaasv1beta1.KubeAuth{
+				Kubeconfig: kubeConfig.Kubeconfig,
+			},
+		}
+		_, err = in.RegisterCluster(ctx, req)
+		if err != nil {
+			return err
+		}
```
oksana-grishchenko, line 127: Suggestion: add logs before all `return`s to simplify troubleshooting

managed/services/management/dbaas/dbaas_initializer.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	pmmVersion, err := goversion.NewVersion(pmmversion.PMMVersion)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	pxcOperatorVersion, psmdbOperatorVersion, err := in.versionService.LatestOperatorVersion(ctx, pmmVersion.Core().String())
+	if err != nil {
+		return nil, err
+	}
```
oksana-grishchenko, line 192: Suggestion: add logs before all `return`s to simplify troubleshooting

oksana-grishchenko (approved):
Posted some comments, everything else LGTM

managed/services/management/dbaas/dbaas_initializer.go
```diff
@@ -0,0 +1,251 @@
...
+	timeoutCtx, cancel := context.WithTimeout(ctx, 20*time.Second)
+	err := in.dbaasClient.Connect(timeoutCtx)
+	cancel()
+	if err != nil {
+		return err
+	}
+	ctx, in.cancel = context.WithCancel(ctx)
+
+	in.enabled = true
+	kubeConfig, err := in.dbaasClient.GetKubeConfig(ctx, &dbaascontrollerv1beta1.GetKubeconfigRequest{})
```
BupycHuk, line 97: can we extract everything below this line to a separate method?

BupycHuk: And should we do it each time?

managed/services/management/dbaas/dbaas_initializer.go
```diff
@@ -0,0 +1,251 @@
...
+	err := in.dbaasClient.Disconnect()
+	if err != nil {
+		return err
+	}
+	in.enabled = false
+	return nil
+}
+
+// RegisterCluster registers k8s cluster and installs all required operators
+func (in *Initializer) RegisterCluster(ctx context.Context, req *dbaasv1beta1.RegisterKubernetesClusterRequest) (*dbaasv1beta1.RegisterKubernetesClusterResponse, error) {
```
BupycHuk, line 155: TBH I don't like the idea that we moved this method here. It doesn't feel like it belongs here.

BupycHuk (changes requested):

managed/services/management/dbaas/dbaas_initializer.go
gen1us2k: Yep. Thanks

managed/services/management/dbaas/dbaas_initializer.go
gen1us2k: What do you suggest?

managed/services/management/dbaas/dbaas_initializer.go
gen1us2k: I moved it to a separate function. 

> And should we do it each time?

Yes, but we do not do anything else once PMM has registered cluster.

managed/services/management/dbaas/dbaas_initializer.go
BupycHuk: 1. The best option would be to have layered architecture, so we wouldn't call a gRPC server. But it makes the code inconsistent with what we have in the project now.
2. Copy and paste?
3. Calling kubernetes_server from the initializer.

BupycHuk (approved):

managed/services/management/dbaas/dbaas_initializer.go
```diff
@@ -0,0 +1,151 @@
...
+			if err != nil {
+				in.l.Errorf("failed finding cluster: %v", err)
+				return nil
+			}
+			if cluster != nil {
+				return errClusterExists
+			}
+			return nil
+		})
+		if err != nil {
```
BupycHuk, line 111: err can be only `errClusterExists`

managed/services/management/dbaas/dbaas_initializer.go
```diff
@@ -0,0 +1,151 @@
...
+				Kubeconfig: kubeConfig.Kubeconfig,
+			},
+		}
+		_, err = in.kubernetesServer.RegisterKubernetesCluster(ctx, req)
+		if err != nil {
+			return err
+		}
+		in.l.Info("Cluster is successfully initialized")
+	} else {
+		in.l.Errorf("failed getting kubeconfig inside cluster: %v", err)
```
BupycHuk, line 130: what response do we get if we run PMM outside of the cluster?

PR #1126, opened 2022-08-04
PMM-10383 Exclude managed from pmm-client build

artemgavrilov:
PMM-10383

Build: SUBMODULES-2686


atymchuk (approved):

build/scripts/build-client-binary
```diff
@@ -27,8 +27,10 @@ gobuild_component() {
         echo skip build
         return
     fi
-    if [ ${component} = 'pmm-admin' -o ${component} = 'pmm-agent' ]; then
-        build_command="cd /tmp/go/src/${extract_path} && make release"
+    if [ ${component} = 'pmm-admin' ]; then
+        build_command="cd /tmp/go/src/${extract_path}/admin && make release"
+    elif [ ${component} = 'pmm-agent' ]; then
+        build_command="cd /tmp/go/src/${extract_path}/agent && make release"
```
BupycHuk, line 33: looks like we are calling gobuild_component for these components incorrectly here https://github.com/percona/pmm/pull/1126/files#diff-d73461fa7f39050da15a5425872480d2f0e377753a7a1a502b715c42f8f79d4dR110-R111
We don't use component_path at all and I think that it could be used the same way as for percona-toolkit.
Also, we can override extract_path there.

artemgavrilov:
> @artemgavrilov Do we need a FB to be sure of the result?
> 
> btw, codecov's reporting is super weird ))

FB is ready. I already checked pmm-client tarball.

build/scripts/build-client-binary
artemgavrilov: It seems that you added wrong link, in points to your comment

BupycHuk: line 110-111 in this file where we call `gobuild_component`.
https://github.com/percona/pmm/blob/d74fa2b31add84840df0cddc313d91549c6f135d/build/scripts/build-client-binary#L110-L111

artemgavrilov: Done

artemgavrilov:
> btw, codecov's reporting is super weird ))

@atymchuk I thinks that's because we run same tests on different DB versions, and for each version we have some specific logic.

BupycHuk (approved):

PR #1127, opened 2022-08-04
PMM-9682 large scale support

BupycHuk:
PMM-9682

Build: SUBMODULES-2689


michal-kralik (approved):
`/admin` folder approved

agent/agentlocal/deps.go
```diff
@@ -23,8 +23,8 @@ import (
 	"github.com/percona/pmm/api/agentpb"
 )
 
-//go:generate ../../bin/mockery -name=client -case=snake -inpkg -testonly
-//go:generate ../../bin/mockery -name=supervisor -case=snake -inpkg -testonly
+//go:generate mockery -name=client -case=snake -inpkg -testonly
+//go:generate mockery -name=supervisor -case=snake -inpkg -testonly
```
artemgavrilov, line 27: Instead of `mockery` in our tools folder here we will use mockery from that `PATH`. It may be a completely different version. I propose to keep exact path to tools binaries.

Makefile.include
```diff
@@ -1,6 +1,8 @@
 .PHONY: default init release gen test clean all
 default: help
 
+PMM_RELEASE_PATH ?= ./bin
```
artemgavrilov, line 4: Why do you prefer variable for that? Do we need to overwrite it somewhere?

docker-compose.yml
```diff
@@ -56,6 +58,7 @@ services:
       - ${PMM_PORT_CH_HTTP:-11123}:8123
     volumes:
       - ./:/root/go/src/github.com/percona/pmm
+      - ../../percona-platform/grafana/public:/usr/share/grafana/public
```
artemgavrilov, line 61: That may be a problem for other devs. Do we need it, or that left after debugging?

managed/Makefile
```diff
@@ -15,6 +15,7 @@ PMM_RELEASE_VERSION ?= $(shell git describe --always --dirty | cut -b2-)
 PMM_RELEASE_TIMESTAMP ?= $(shell date '+%s')
 PMM_RELEASE_FULLCOMMIT ?= $(shell git rev-parse HEAD)
 PMM_RELEASE_BRANCH ?= $(shell git describe --always --contains --all)
+export PATH := $(PMM_RELEASE_PATH):$(PATH)
```
artemgavrilov, line 18: Ok, here is the problem. Actually we had such export everywhere some time ago, and then we dropped it. Trick is that your targets may work even if you forgot to invoke `make init`. Some outdated tool will be taken from your PATH.

agent/agentlocal/deps.go
BupycHuk: We have conflicts on macs because of that. If we run `make init` locally and inside dev container they generate tools for different architectures.

Makefile.include
BupycHuk: Yes, as I mentioned above we are not able to use even tools outside of dev container on systems with arm architecture.

docker-compose.yml
BupycHuk: Yeah, it's committed by mistake

agent/agentlocal/deps.go
artemgavrilov: I remember we fixed this problem

Ah, ok, that was broken again in that PR: https://github.com/percona/pmm/pull/1003/files

Line 57 of `docker-compose.yml`

artemgavrilov: So let's just revert this change, this fake volume will let us to have different set of tools inside container and on the host machine

managed/Makefile
BupycHuk: That's possible, I'll check if it's possible to use absolute path in generate comments.

agent/agentlocal/deps.go
BupycHuk: having `/root/go/src/github.com/percona/pmm/bin` in volumes doesn't fix it for local runs, it just rewrites the directory.

artemgavrilov: Mmm, what do you mean? Local run will use local `bin`, container run will use container's `bin`

managed/services/management/dbaas/db_cluster_service_test.go
```diff
@@ -114,85 +114,103 @@ func TestDBClusterService(t *testing.T) {
...
-					Operation: &controllerv1beta1.RunningOperation{
-						TotalSteps:    int32(10),
-						FinishedSteps: int32(10),
-					},
-				},
-			},
-		}
-		dbaasClient.On("ListPSMDBClusters", ctx, mock.Anything).Return(&mockPSMDBResp, nil)
+		s := NewDBClusterService(db, dbaasClient, grafanaClient, versionService, nil)
+		//mockPXCResp := controllerv1beta1.ListPXCClustersResponse{
```
JiriCtvrtka, line 122: @gen1us2k  Can I ask what is the reason of it? I am just wondering.

JiriCtvrtka (approved):

agent/agentlocal/deps.go
BupycHuk: it works if you have long running containers, but if you re-create it each time. it rewrites local directory on setup.

artemgavrilov: So problem is that you start new containers frequently, and you don't want to install tools each time?

BupycHuk: Yes

PR #1135, opened 2022-08-08
PMM-7 Fix gatekeeper

artemgavrilov:
PMM-7

Build: SUBMODULES-0

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -153,7 +153,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Run Merge Gatekeeper
-        uses: upsidr/merge-gatekeeper@v1
+        uses: upsidr/merge-gatekeeper@v1.1.0
         with:
           self: Merge Gatekeeper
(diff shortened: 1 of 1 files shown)


.github/workflows/common.yml
```diff
@@ -153,8 +153,9 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Run Merge Gatekeeper
-        uses: percona-platform/merge-gatekeeper@v1.0.2
+        uses: upsidr/merge-gatekeeper@v1
```
artemgavrilov, line 156: @atymchuk  Please add this action to allow-list. Also, do we want to use major versions like `@v1`, or exact like `@v1.1.0`?

ritbl (approved):

.github/workflows/common.yml
atymchuk: It depends. For a tool like this with only 20-ish starts on github, I would apply zero trust. Therefore, I'd suggest we pin it to the exact version.

atymchuk (approved):

BupycHuk (approved):

PR #1140, opened 2022-08-10
PMM-10495 Avoid swagger validation when browsing /swagger

cezmunsta:
By default, Swagger UI attempts to validate specs against swagger.io's
online validator. You can use this parameter to set a different
validator URL, for example for locally deployed validators
(Validator Badge). Setting it to either none, 127.0.0.1 or
localhost will disable validation.

https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/#network

* Set validatorUrl to an empty string

[PMM-10495](https://jira.percona.com/browse/PMM-10495)

Build: [SUBMODULES-2701](https://github.com/Percona-Lab/pmm-submodules/pull/2701)


atymchuk:
Ceri, this is quite hard to review. Is there any way for us to "rebuild from source"? I know we are just grabbing the files from swagger's ui repo, but I was wondering how you actually made those changes to compiles files?

cezmunsta:
> Ceri, this is quite hard to review. Is there any way for us to "rebuild from source"? I know we are just grabbing the files from swagger's ui repo, but I was wondering how you actually made those changes to compiled files?

@atymchuk they aren't compiled, just minified from what I can tell.
It was simply applying the workaround from the ticket:
```
sed -i 's;validatorUrl:"https://validator.swagger.io/validator";validatorUrl:"";g' *
```

atymchuk:
> How are the files normally updated? I didn't notice a `make` command to pull in files, so I presume that it is manual, in which case perhaps `Makefile.include` needs something extra

I believe they are copied from [here](https://registry.npmjs.org/swagger-ui-dist/-/swagger-ui-dist-4.13.2.tgz)

BupycHuk (changes requested):
I would recommend just set `validationUrl: 'none'` here https://github.com/percona/pmm/blob/main/api/swagger/swagger-initializer.js#L6 to disable validation. it's much easier to maintain than custom build.

cezmunsta:
@BupycHuk that file is also part of the bundle when downloaded, so really it only reduces the number of files that need to be edited. I have added `download_swagger` and `update_swagger`, so `make update_swagger` can be used to pull in the latest version and patch `swagger-initializer.js` as part of the process

BupycHuk:
@cezmunsta, yeah, but it's much easier to see differences in the configuration file than in obfuscated files

BupycHuk (approved):

Makefile.include
```diff
@@ -110,6 +112,20 @@ gen-alertmanager:     # Generate Alertmanager client.
 clean_swagger:
 	find api -name '*.swagger.json' -print -delete
 
+LATEST_SWAGGER = $(shell curl --silent --head https://github.com/swagger-api/swagger-ui/releases/latest | grep -E '^[lL]ocation:' | grep -oE 'v[0-9.]+[[:space:]]*' | tr -d ' ')
```
atymchuk, line 115: Dude, you are a bash ninja :)

atymchuk (approved):

Makefile.include
atymchuk: 🚢

atymchuk: Let's wait until it's tested, OK?

Makefile.include
```diff
@@ -110,6 +112,20 @@ gen-alertmanager:     # Generate Alertmanager client.
...
+
+download_swagger:
+ifeq ($(shell test "${SWAGGER_UI_VERSION}" = "latest" && echo 1 || echo 0), 1)
+	curl --output /tmp/swagger.tar.gz -sSfL "https://github.com/swagger-api/swagger-ui/archive/refs/tags/$(LATEST_SWAGGER).tar.gz"
+else
+	curl --output /tmp/swagger.tar.gz -sSfL "https://github.com/swagger-api/swagger-ui/archive/refs/tags/${SWAGGER_UI_VERSION}.tar.gz"
+endif
+
+update_swagger: download_swagger
```
artemgavrilov, line 124: Please add comment to that target, we are using them for `help` target.

Makefile.include
artemgavrilov: Minor: this variable can be moved under first `if` in `download` swagger. It's used only there.

artemgavrilov (approved):

Makefile.include
```diff
@@ -1,3 +1,6 @@
+# vim: ts=8:sw=8:ft=make:noai:noet
```
cezmunsta, line 1: Let me know if there is any reason to remove the modeline, it makes life easier for me with my vim settings

Makefile.include
```diff
@@ -110,6 +113,20 @@ gen-alertmanager:     # Generate Alertmanager client.
 clean_swagger:
 	find api -name '*.swagger.json' -print -delete
 
+
+ifeq ($(shell test "${SWAGGER_UI_VERSION}" = "latest" && echo 1 || echo 0), 1)
+get_swagger_version:
+override SWAGGER_UI_VERSION = $(shell curl --silent --head https://github.com/swagger-api/swagger-ui/releases/latest | grep -E '^[lL]ocation' | sed 's;^.*/;;')
```
cezmunsta, line 119: @artemgavrilov I presume that you wanted to avoid calls when not updating swagger, so refactored it slightly to override the version instead and moved the download into `update-swagger`

Makefile.include
```diff
@@ -110,6 +113,20 @@ gen-alertmanager:     # Generate Alertmanager client.
...
+else
+get_swagger_version:
+endif
+
+update-swagger: get_swagger_version               ## Update Swagger UI in api/swagger, use `SWAGGER_UI_VERSION=xxx make update-swagger` to choose a version other than latest.
```
cezmunsta, line 124: Renamed this to match the format of the other recipes that are show via help, where a hypen is used instead of underscores

artemgavrilov (approved):

BupycHuk (approved):

PR #1159, opened 2022-08-20
PMM-10513 redact user creds in server url

atymchuk:
PMM-10513

Build: SUBMODULES-2723


build/packages/rpm/server/SPECS/victoriametrics.spec
```diff
@@ -1,3 +1,5 @@
+%undefine _missing_build_ids_terminate_build
```
michal-kralik, line 1: Is this change related to the PR?

michal-kralik (approved):
All looks good. Just one comment. I'm not sure if the `rpm` change is related to the PR.

build/packages/rpm/server/SPECS/victoriametrics.spec
atymchuk: Yes and no. If you look at the logs, the pipeline was failing without this. It's the last component that does not have this statement, all others do already.

BupycHuk (approved):

PR #1165, opened 2022-08-25
PMM-3396: Support for unix sockets in agent and admin

michal-kralik:
PMM-3396

Build: SUBMODULES-0


admin/agentlocal/agentlocal.go
```diff
@@ -31,21 +33,57 @@ import (
...
+
+		t := tr.Clone()
+		t.DialContext = func(ctx context.Context, _, _ string) (net.Conn, error) {
+			dialer := net.Dialer{}
+			return dialer.DialContext(ctx, "unix", socket)
+		}
+		cl := &http.Client{
+			Transport: t,
+		}
+		transport := httptransport.NewWithClient(GetHostname("", 0, socket), "/", []string{"http"}, cl)
```
gen1us2k, line 76: Why we can't use `t` as transport here?

admin/agentlocal/agentlocal.go
```diff
@@ -31,21 +33,57 @@ import (
...
+// DefaultClient is http client configured either for socket or tcp connection to local pmm-agent.
+// It shall be used to talk to the local pmm-agent.
+var DefaultClient = http.DefaultClient
+
 // SetTransport configures transport for accessing local pmm-agent API.
-func SetTransport(ctx context.Context, debug bool, port uint32) {
+func SetTransport(ctx context.Context, debug bool, port uint32, socket string) {
 	// use JSON APIs over HTTP/1.1
-	transport := httptransport.New(fmt.Sprintf("%s:%d", Localhost, port), "/", []string{"http"})
+	transport := configureTransport(port, socket)
```
gen1us2k, line 43: I assume that this code can be simplified

```
transport := httptransport.New(fmt.Sprintf("%s:%d", Localhost, port), "/", []string{"http"}) // default fallback transport

if socket != "" {
		transport =  &http.Transport{
			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
                                 var dialer net.Dialer
				return dialer.DialContext(ctx, "unix", socket)
			},
		},
}
```

admin/agentlocal/agentlocal.go
michal-kralik: `transport` is of type `*httptransport.Runtime` while `t` is `*http.Transport`

admin/agentlocal/agentlocal.go
michal-kralik: `httptransport.New` returns `*httptransport.Runtime` while you would try to assign `*http.Transport` later. These are incompatible types I believe.

admin/main.go
```diff
@@ -138,3 +150,77 @@ func main() {
...
+	if sock := <-socketChan; sock != "" {
+		logrus.Debugf("Found socket %s", sock)
+		return sock, 0
+	}
+
+	if port := <-portChan; port != 0 {
+		logrus.Debugf("Found port %d", port)
+		return "", port
+	}
+
```
oksana-grishchenko, line 223: It looks like this code will wait a value from the `socketChan` until it appears and never will go down and check the `portChan`. Is it intended? The [simplified test](https://go.dev/play/p/n55giKb85g1) to play around

agent/agentlocal/agent_local.go
```diff
@@ -340,7 +341,35 @@ var (
...
 )
 
-// addData add data to zip file
+var errSocketOrPortRequired = errors.New("socketOrPortRequired")
+
+// getListener returns a net.Listener on socket or tcp based on configuration.
+func (s *Server) getListener(l *logrus.Entry) (net.Listener, error) {
+	if s.cfg.ListenSocket != "" {
+		l.Infof("Starting local API server on unix:%s", s.cfg.ListenSocket)
+		l, err := net.Listen("unix", s.cfg.ListenSocket)
```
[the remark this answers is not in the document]
michal-kralik: Good catch. And go did not complain about it either :)

admin/main.go
michal-kralik: In your example that's true. In this case though, `socketChan` can receive an empty string, in which case `portChan` message is processed.

oksana-grishchenko: Oh, I see, you are right)

oksana-grishchenko (approved):
LGTM. Great that you keep the PR clean and resolve all the conversations that don't need the reviewer's attention anymore, it's a pleasure to come back and review it again :)

agent/config/config.go
```diff
@@ -312,6 +217,147 @@ func get(args []string, l *logrus.Entry) (cfg *Config, configFileF string, err e
...
+	if b, _ := strconv.ParseBool(os.Getenv("PMM_TRACE")); b {
+		cfg.Trace = true
+	}
+}
+
+var (
+	defaultListenSocketJSON         = "/usr/local/percona/pmm2/pmm-agent.sock"
+	defaultListenSocketGRPCFilename = "pmm-agent-grpc.sock"
+	// DefaultListenSocketGRPC defines path to a socket on which GRPC server is listening.
+	DefaultListenSocketGRPC = "/usr/local/percona/pmm2/" + defaultListenSocketGRPCFilename
```
BupycHuk, line 255: we have `cfg.Paths.PathsBase`, let's put sockets in the same directory or into subdirectory of `cfg.Paths.PathsBase` like we do with `exporters`

agent/config/config.go
```diff
@@ -312,6 +217,147 @@ func get(args []string, l *logrus.Entry) (cfg *Config, configFileF string, err e
...
+	}
+	if !filepath.IsAbs(cfg.Paths.PTPGSummary) {
+		cfg.Paths.PTPGSummary = filepath.Join(cfg.Paths.PathsBase, cfg.Paths.PTPGSummary)
+	}
+	if !filepath.IsAbs(cfg.Paths.PTMongoDBSummary) {
+		cfg.Paths.PTMongoDBSummary = filepath.Join(cfg.Paths.PathsBase, cfg.Paths.PTMongoDBSummary)
+	}
+	if !filepath.IsAbs(cfg.Paths.PTMySQLSummary) {
+		cfg.Paths.PTMySQLSummary = filepath.Join(cfg.Paths.PathsBase, cfg.Paths.PTMySQLSummary)
+	}
```
BupycHuk, line 316: since we are refactoring it, can we update paths in a loop, like for exporters?

agent/config/config.go
```diff
@@ -312,6 +217,147 @@ func get(args []string, l *logrus.Entry) (cfg *Config, configFileF string, err e
...
+	DefaultListenSocketGRPC = "/usr/local/percona/pmm2/" + defaultListenSocketGRPCFilename
+)
+
+func configureNetworkDefaults(cfg *Config) {
+	if cfg.ListenSocket == "" {
+		configureListenDefaults(cfg)
+	}
+	if cfg.ListenSocketGRPC == "" && cfg.ListenSocket != "" {
+		cfg.ListenSocketGRPC = path.Dir(cfg.ListenSocket) + "/" + defaultListenSocketGRPCFilename
+	}
```
BupycHuk, line 264: shouldn't these lines be part of configureListenDefaults?

agent/config/config.go
```diff
@@ -312,6 +217,147 @@ func get(args []string, l *logrus.Entry) (cfg *Config, configFileF string, err e
...
+		if cfg.ListenSocketGRPC == "" {
+			cfg.ListenSocketGRPC = DefaultListenSocketGRPC
+		}
+	}
+	if cfg.ListenAddress == "" && cfg.ListenPort != 0 {
+		cfg.ListenAddress = "127.0.0.1"
+	}
+	if cfg.ListenAddress != "" && cfg.ListenPort == 0 {
+		cfg.ListenPort = 7777
+	}
```
BupycHuk, line 285: can we use a `switch case` here?

agent/agentlocal/agent_local.go
```diff
@@ -192,7 +195,7 @@ func (s *Server) Reload(ctx context.Context, req *agentlocalpb.ReloadRequest) (*
 // runGRPCServer runs gRPC server until context is canceled, then gracefully stops it.
 func (s *Server) runGRPCServer(ctx context.Context, listener net.Listener) {
 	l := s.l.WithField("component", "local-server/gRPC")
-	l.Debugf("Starting gRPC server on http://%s/ ...", listener.Addr().String())
+	l.Debugf("Starting gRPC server on unix:%s ...", listener.Addr().String())
```
BupycHuk, line 198: existing pmm-agents should keep working via `http://`

BupycHuk: only new pmm-agents should use sockets

BupycHuk: that can be checked by the existence of the port configuration in `pmm-agent.yaml` file

agent/config/config_test.go
```diff
@@ -444,6 +444,179 @@ func TestGet(t *testing.T) {
...
+			"--id=agent-id",
+			"--listen-socket=/tmp/pmm-agent.sock",
+			"--server-address=127.0.0.1",
+		}, logrus.WithField("test", t.Name()))
+		require.NoError(t, err)
+
+		expected := &Config{
+			ID:               "agent-id",
+			ListenSocket:     "/tmp/pmm-agent.sock",
+			ListenSocketGRPC: "/tmp/pmm-agent-grpc.sock",
```
BupycHuk, line 459: do we really need two separate files?

BupycHuk (changes requested):

agent/agentlocal/agent_local.go
michal-kralik: I understand the grpc server is used internally only.
Originally it listened on a random port.

Since it's for internal purpose only, I believe we don't need to keep this backwards compatible.
Any thoughts?

agent/config/config_test.go
michal-kralik: I may be wrong but I could not find a way for the grpc gateway to talk directly to grpc server without exposing a socket or a port. For this reason we have two socket files.

admin/commands/summary.go
```diff
@@ -317,14 +346,14 @@ func addPprofData(ctx context.Context, zipW *zip.Writer, skipServer bool, global
 		var wg sync.WaitGroup
 		ch := make(chan pprofData, len(sources))
 
-		for dir, urlPrefix := range sources {
+		for dir, source := range sources {
 			wg.Add(1)
 
-			go func(url, name string) {
+			go func(url, name string, source sourceConfig) {
```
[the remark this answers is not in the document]
michal-kralik: Updated. Is there any specific reason you prefer source being first?

agent/config/config.go
michal-kralik: Sure, good catch

agent/config/config.go
michal-kralik: The `configureListenDefaults` are triggered only in case `ListenSocket` is empty.
I moved it to a separate function and added one more check. We had a bug there.

agent/config/config.go
michal-kralik: Wasn't aware of this. Great catch.

admin/main.go
```diff
@@ -138,3 +150,79 @@ func main() {
...
+			err := conn.Close()
+			if err != nil {
+				logrus.Debugf("Socket close error: %#v", err)
+			}
+
+			socket = socketPath
+		}
+
+		socketChan <- socket
+	}()
```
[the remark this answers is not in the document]
BupycHuk: we can extract duplicated part to a separate part and call it `checkConnection(protocol, url) error` and keep only different parts.

agent/agentlocal/agent_local.go
BupycHuk: if pmm-admin doesn't connect to pmm-agent via gRPC maybe it's better to completely remove support of gRPC?

BupycHuk (approved):

gen1us2k (approved):

PR #1168, opened 2022-08-25
PMM-10403 Define user APIs

ShashankSinha252:
PMM-10403 : Add user related information in PMM

Build: SUBMODULES-2735


managed/models/user_flags_helpers.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+type UpdateUserParams struct {
+	UserID int
+	Tour   bool
+}
+
+func CreateUser(q *reform.Querier, params *CreateUserParams) (*UserDetails, error) {
+	// TODO : Confirm if check is required
+	if params.UserID < 0 {
```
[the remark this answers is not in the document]
ShashankSinha252: For **admin** account in default setup, user ID is **1**. It doesn't look like **0** can be a user ID.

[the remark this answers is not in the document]
BupycHuk: let's than check if the id less or equal then 0

managed/models/user_flags_helpers.go
```diff
@@ -22,24 +22,26 @@ import (
...
 func CreateUser(q *reform.Querier, params *CreateUserParams) (*UserDetails, error) {
-	// TODO : Confirm if check is required
 	if params.UserID < 0 {
 		return nil, status.Errorf(codes.InvalidArgument, "Invalid user ID")
 	}
 
 	// Check user ID is unique
 	row := &UserDetails{ID: params.UserID}
-	if err := q.Reload(row); err != reform.ErrNoRows {
+	if err := q.Reload(row); errors.Is(err, reform.ErrNoRows) {
```
BupycHuk, line 44: something looks wrong here.
```suggestion
	if err := q.Reload(row); !errors.Is(err, reform.ErrNoRows) {
```

managed/services/grafana/client.go
```diff
@@ -182,19 +182,27 @@ func (r role) String() string {
 }
 
 // getUserID returns user ID from Grafana for given user
-func (c *Client) getUserID(ctx context.Context, authHeaders http.Header) (int, error) {
+func (c *Client) GetUserID(ctx context.Context) (int, error) {
+	userID := -1
+
+	authHeaders, err := c.authHeadersFromContext(ctx)
+	if err != nil {
+		return userID, err
```
BupycHuk, line 190: if the user ID can't be 0, then let's return 0 instead of -1
```suggestion
		return 0, err
```

managed/services/grafana/client.go
```diff
@@ -182,19 +182,27 @@ func (r role) String() string {
...
 	if err != nil {
-		return -1, err
+		return userID, err
 	}
 
-	if val, ok := m["id"].(int); ok {
-		return val, nil
-	} else {
-		return -1, errors.New("Unable to find User ID")
+	val, ok := m["id"].(float64)
```
BupycHuk, line 200:
```suggestion
	userID, ok := m["id"].(float64)
```

managed/services/user/user.go
```diff
@@ -32,25 +32,34 @@ import (
 type Service struct {
 	db *reform.DB
 	l  *logrus.Entry
+	c  clientInterface
 
 	userpb.UnimplementedUserServer
 }
 
+type clientInterface interface {
```
BupycHuk, line 40: maybe just `grafanaClient` like in other places?

managed/services/user/user.go
```diff
@@ -85,8 +94,10 @@ func (s *Service) UpdateUser(ctx context.Context, req *userpb.UserUpdateRequest)
...
 
-	// TODO : Get ID from Grafana
-	userID := 32
+	userID, err := s.c.GetUserID(ctx)
+	if err != nil {
+		return nil, err
+	}
 
 	userInfo := &models.UserDetails{}
 	e := s.db.InTransaction(func(tx *reform.TX) error {
```
BupycHuk, line 103: let's merge these two transactions

BupycHuk (changes requested):

managed/services/user/user.go
oksana-grishchenko: In addition to the @BupycHuk suggestion - as far as I remember, even if the SELECT and then the UPDATE queries being called inside a single transaction it doesn't guarantee that the selected rows not to be changed between those calls. The SELECT FOR UPDATE does.

BupycHuk: @artemgavrilov have we fixed that?

managed/services/grafana/client.go
```diff
@@ -183,26 +183,26 @@ func (r role) String() string {
 
 // getUserID returns user ID from Grafana for given user
 func (c *Client) GetUserID(ctx context.Context) (int, error) {
-	userID := -1
+	userID := 0.0
```
BupycHuk, line 186: I think this line can be removed.

managed/services/grafana/client.go
```diff
@@ -183,26 +183,26 @@ func (r role) String() string {
...
 	var m map[string]interface{}
 	err = c.do(ctx, "GET", "/api/user", "", authHeaders, nil, &m)
 
 	if err != nil {
-		return userID, err
+		return 0, err
 	}
 
-	val, ok := m["id"].(float64)
+	userID, ok := m["id"].(float64)
```
BupycHuk, line 200: interesting case, why do we cast to float instead of int?

ShashankSinha252: We don't dictate how it gets **unmarshaled**. I noticed it gets read as **float64** instead of an **int**.

managed/services/grafana/client.go
ShashankSinha252: Done

managed/services/user/user.go
ShashankSinha252: Done

managed/models/user_flags_helpers.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func UpdateUser(q *reform.Querier, params *UpdateUserParams) (*UserDetails, error) {
+	// TODO : Confirm if check is required
+	if params.UserID < 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "Invalid user ID")
+	}
+	if !params.Tour {
+		return nil, status.Errorf(codes.InvalidArgument, "Cannot unset tour flag")
+	}
```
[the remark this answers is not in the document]
ShashankSinha252: Done

managed/services/user/user.go
artemgavrilov: @BupycHuk Well, technically it's not a bug or something that should be fixed. We have a ticket about increasing default transaction isolation level https://jira.percona.com/browse/PMM-8433 up to `serializable`. After that there will be no need to use `select for update`.

However in this particular case we can use simple `select` because update didn't use anything from selected data. `select for update` needed when you have logic like this:

```
SELECT money FROM user WHERE id = 1 FOR UPDATE;
var newMoney = money + 100
UPDATE user SET money = newMoney WHERE id = 1;
```

Moreover I don't know why we need `select` here at all, we can just use `update`. DB will retrun error if user is not exists.

managed/services/user/user.go
```diff
@@ -0,0 +1,129 @@
...
+// GetUser creates a new user
+func (s *Service) GetUser(ctx context.Context, req *userpb.UserDetailsRequest) (*userpb.UserDetailsResponse, error) {
+	// TODO : Get user ID from Grafana
+	userID := 32
+
+	userInfo := &models.UserDetails{}
+	e := s.db.InTransaction(func(tx *reform.TX) error {
+		var err error
+		userInfo, err = models.FindUser(tx.Querier, userID)
+		if err != nil {
```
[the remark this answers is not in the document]
ShashankSinha252: Done

managed/services/user/user.go
ShashankSinha252: Merged transactions. Skipping SELECT FOR usage for now. Let me know if it's compulsory.

oksana-grishchenko (approved):
LGTM. Please check the latest linter comments, it may worth to fix

managed/services/grafana/client.go
BupycHuk: we can dictate by using structs instead of map :) But ok

BupycHuk (approved):

PR #1171, opened 2022-08-26
PMM-9599: Make key=value CLI help consistent

michal-kralik:
PMM-9599

Build: SUBMODULES-2737


admin/commands/inventory/add_service_haproxy.go
```diff
@@ -49,7 +49,7 @@ type AddServiceHAProxyCommand struct {
 	Environment    string            `placeholder:"prod" help:"Environment name like 'production' or 'qa'"`
 	Cluster        string            `placeholder:"east-cluster" help:"Cluster name"`
 	ReplicationSet string            `placeholder:"rs1" help:"Replication set name"`
-	CustomLabels   map[string]string `help:"Custom user-assigned labels. Example: region=east,app=app1"`
+	CustomLabels   map[string]string `help:"Custom user-assigned labels"`
```
idoqo, line 52: Looks like we just removed the examples completely. Is that the idea?

michal-kralik: That's correct.
The new framework includes them in the help such as 
```
--custom-labels=KEY=VALUE;...    Custom user-assigned labels
```

idoqo (approved):

PR #1181, opened 2022-08-29
PMM-9599: Custom labels example for config

michal-kralik:
PMM-9599

Build: SUBMODULES-2740

### DIFF
--- a/admin/commands/inventory/add_agent_external_exporter.go
+++ b/admin/commands/inventory/add_agent_external_exporter.go
@@ -56,7 +56,7 @@ type AddAgentExternalExporterCommand struct {
 	Scheme       string            `help:"Scheme to generate URI to exporter metrics endpoints (http, https)"`
 	MetricsPath  string            `help:"Path under which metrics are exposed, used to generate URI"`
 	ListenPort   int64             `required:"" placeholder:"port" help:"Listen port for scraping metrics"`
-	CustomLabels map[string]string `help:"Custom user-assigned labels"`
+	CustomLabels map[string]string `mapsep:"," help:"Custom user-assigned labels"`
 	PushMetrics  bool              `help:"Enables push metrics model flow, it will be sent to the server by an agent"`
 }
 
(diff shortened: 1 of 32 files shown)


oksana-grishchenko (approved):

admin/commands/inventory/add_agent_external_exporter.go
```diff
@@ -56,7 +56,7 @@ type AddAgentExternalExporterCommand struct {
 	Scheme       string            `help:"Scheme to generate URI to exporter metrics endpoints (http, https)"`
 	MetricsPath  string            `help:"Path under which metrics are exposed, used to generate URI"`
 	ListenPort   int64             `required:"" placeholder:"port" help:"Listen port for scraping metrics"`
-	CustomLabels map[string]string `help:"Custom user-assigned labels"`
+	CustomLabels map[string]string `mapsep:"," help:"Custom user-assigned labels"`
```
PavelKhripkov, line 59: Why do we use `string` type in one case and `map` in others?

PavelKhripkov (approved):

admin/commands/inventory/add_agent_external_exporter.go
michal-kralik: That's a great catch. Looking for feedback here.
The `string` type is in https://github.com/percona/pmm/blob/5783c71fb1f73df4565fcf4bfccd295deaeb8a03/agent/config/config.go#L130

The reason is we pass the label as-is, as a `string`, to the agent.
The alternative is to switch it to `map` and then parse it back into a `string` but we would have to account for special characters, etc. I opted for keeping it as a `string` instead as it seems it's less error prone. The downside is there's not the default `map` validation applied.

PavelKhripkov: What if we choose only one type for all cases and experience the same problems in all cases?)

michal-kralik: If we choose `map` we will have the problem only in one case - the config.
I don't think it would make sense to choose `string` because the library supports it out of the box.

I feel it's safer to just have `string` with config. We're not making the situation worse, compared to the current state.

PR #1193, opened 2022-09-06
PMM-10044 Fix enormous numbers in QAN.

BupycHuk:
PMM-10044

Build: SUBMODULES-2754


michal-kralik (approved):

PavelKhripkov:
This PR doesn't fix the bug, so do we really need these changes? I left some thoughts in the tkt.

BupycHuk:
@PavelKhripkov 10000 is a default limit value for perf schema tables, so it should fix the problem for default installations.

PavelKhripkov:
@BupycHuk Aha, I see, we rely on the default values for now.

PR #1196, opened 2022-09-07
Run `buf mod update` to update buf.lock

stefanvanburen:
We recently made a change to our managed [buf.build/googleapis/googleapis](https://buf.build/googleapis/googleapis) repository before certifying BSR as v1. You'll need to run `buf mod update` to update your buf.lock. This PR runs that for you, which fixes some of the [recent failures you've seen in your API action](https://github.com/percona/pmm/runs/8229834047?check_suite_focus=true#step:8:203).

[googleapis](https://github.com/googleapis/googleapis) contains over 3800 files, mostly relating to Google's core APIs. This causes numerous issues for most users, such as timeouts when installing packages with huge swaths of unused code to pull in just a few files. Not only is this a lot of code over the network and on disk, but it can cause issues in editors that try to parse all of that code!

However, there are only ~30 files used by 99.999% of developers, and these files are the most common dependency in the Protobuf ecosystem. For example:

- google.type, which defines useful messages such as DateTime and Money.-
- google.rpc, for interacting with gRPC.
- google.api, for defining APIs, such as with grpc-gateway.

We considered these commonly used packages and included a subset of them into the new, slim googleapis repository.

We *hate* breaking anyone, even in an alpha, and our apologies that it's causing any difficulties here. The issues with googleapis were serious enough that we felt that we had to do this in this extraordinary circumstance. We're coming out of beta extremely soon, and once we do, we take our responsibility not to break customers (free users or otherwise) very seriously, we appreciate you riding the rough road for the moment!


BupycHuk (approved):

artemgavrilov (approved):

artemgavrilov:
@stefanvanburen Thank you!

PR #1215, opened 2022-09-15
PMM-10662: Add datapoints for MySQL

dliakhov:
PMM-10662: Add datapoints for MySQL

Build: https://github.com/Percona-Lab/pmm-submodules/pull/2777


ritbl (approved):

managed/services/telemetry/config.default.yml
```diff
@@ -60,6 +60,38 @@ telemetry:
...
+      - metric_name: "mysql_global_status_queries_running"
+        value: 1
+
+  - id: MySQLSpaceUsed
+    source: VM
+    query: sum(count(mysql_info_schema_table_version{engine="InnoDB"}) by (schema,table) * sum (mysql_info_schema_table_size{component=~"data_length|index_length"}) by (schema,table))
+    summary: "Data usage of MySQL"
+    data:
+      - metric_name: "mysql_space_used"
+        value: 1
```
artemgavrilov, line 93: Shouldn't we mention that it's InnoDB specific metrics is it's name?

BupycHuk (approved):

artemgavrilov (approved):

managed/services/telemetry/config.default.yml
dliakhov: makes sense

updated

PR #1216, opened 2022-09-15
PMM-8792 Configurable query length for MySQL.

JiriCtvrtka:
PMM-8792

Build: https://github.com/Percona-Lab/pmm-submodules/pull/2765

Test failing due to merge changes from main. There are some diffs.


agent/utils/truncate/query.go
```diff
@@ -15,15 +15,34 @@
...
 	// copy MySQL behavior
-	return string(runes[:maxQueryLength-4]) + " ...", true
+	return string(runes[:queryLength-4]) + " ...", true
+}
+
+// GetDefaultQueryLength returns default decimal value for query length.
+func GetDefaultQueryLength() int {
+	var mx sync.RWMutex
+	mx.RLock()
+	defer mx.RUnlock()
```
BupycHuk, line 45: it seems like we don't need it

managed/models/database_test.go
```diff
@@ -384,7 +384,7 @@ func TestDatabaseMigrations(t *testing.T) {
 		}, settings.MetricsResolutions)
 	})
 	t.Run("stats_collections field migration: string to string array", func(t *testing.T) {
-		sqlDB := testdb.Open(t, models.SkipFixtures, pointer.ToInt(57))
+		sqlDB := testdb.Open(t, models.SkipFixtures, pointer.ToInt(67))
```
BupycHuk, line 387: value here should be exact number

managed/models/database_test.go
```diff
@@ -393,29 +393,34 @@ func TestDatabaseMigrations(t *testing.T) {
...
 
-		agent, err := models.FindAgentByID(db.Querier, "id")
+		agent, err := models.FindAgentByID(db.Querier, createdAgent.AgentID)
 		require.NoError(t, err)
-		require.Equal(t, "id", agent.AgentID)
+		require.Equal(t, createdAgent.AgentID, agent.AgentID)
 		require.Equal(t, []string{"db.col1", "db.col2", "db.col3"}, agent.MongoDBOptions.StatsCollections)
 	})
 	t.Run("stats_collections field migration: string array to string array", func(t *testing.T) {
-		sqlDB := testdb.Open(t, models.SkipFixtures, pointer.ToInt(63))
```
BupycHuk, line 418: it supposed to be 58 if I remember correctly

BupycHuk (changes requested):

managed/models/database_test.go
JiriCtvrtka: Thanks. Test should be fixed now.

JiriCtvrtka:
Test failing due to merge changes from main. There are some diffs.

michal-kralik (changes requested):

admin/commands/inventory/add_agent_qan_mysql_perfschema_agent.go
```diff
@@ -60,6 +60,7 @@ type AddAgentQANMySQLPerfSchemaAgentCommand struct {
 	Password             string            `help:"MySQL password for scraping metrics"`
 	CustomLabels         map[string]string `mapsep:"," help:"Custom user-assigned labels"`
 	SkipConnectionCheck  bool              `help:"Skip connection check"`
+	QueryLength          int32             `placeholder:"NUMBER" help:"Define how many chars can a query have until it's truncated (default: server-defined; -1: no limit)"`
```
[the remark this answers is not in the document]
JiriCtvrtka: @michal-kralik Changed for PMM Admin and API.

michal-kralik (approved):

percona-csalguero (changes requested):

agent/utils/truncate/query.go
```diff
@@ -15,15 +15,28 @@
...
-func Query(q string) (query string, truncated bool) {
 	runes := []rune(q)
-	if len(runes) <= maxQueryLength {
+	if len(runes) <= queryLength {
 		return string(runes), false
 	}
 
 	// copy MySQL behavior
-	return string(runes[:maxQueryLength-4]) + " ...", true
+	return string(runes[:queryLength-4]) + " ...", true
```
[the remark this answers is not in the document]
JiriCtvrtka: Right, good point. Changed, thanks.

percona-csalguero (approved):

agent/agents/mysql/perfschema/perfschema.go
```diff
@@ -85,6 +85,7 @@ type PerfSchema struct {
 	q                    *reform.Querier
 	dbCloser             io.Closer
 	agentID              string
+	queryLength          int
```
artemgavrilov, line 88: Same question as for PG

agent/agents/supervisor/supervisor.go
```diff
@@ -461,6 +461,7 @@ func (s *Supervisor) startBuiltin(agentID string, builtinAgent *agentpb.SetState
 		params := &perfschema.Params{
 			DSN:                  dsn,
 			AgentID:              agentID,
+			QueryLength:          int(builtinAgent.QueryLength),
```
artemgavrilov, line 464: Same, what is the reason to use `int`?

api/agentpb/agent.proto
```diff
@@ -67,6 +67,8 @@ message SetStateRequest {
   message BuiltinAgent {
     inventory.AgentType type = 1;
     string dsn = 2;
+    // Limit query length in QAN (default: server-defined; -1: no limit).
+    int32 query_length = 8;
```
artemgavrilov, line 71: `max_query_lenght`?

agent/agents/mysql/perfschema/perfschema.go
JiriCtvrtka: Renamed to maxQueryLength on all places (will do same for postgres PR), because its better naming.

api/agentpb/agent.proto
JiriCtvrtka: Renamed to maxQueryLength on all places (will do same for postgres PR), because its better naming.

agent/agents/supervisor/supervisor.go
JiriCtvrtka: @artemgavrilov 

1) It was actually unspecified integer already before.
2) Postgres query max length is 32 bits, but for example MySQL is almost infinitely. Thats why I picked int to have same data type for all agents types. Anyway because I am limiting it on API layer to 32 bits anyway maybe its probably not necessary.

Its up to you about point 2. Just let me know.

artemgavrilov: I guess we can use `int32` everywhere just to eliminate converting and simplify code.

agent/utils/truncate/query.go
```diff
@@ -15,10 +15,19 @@
 // Package truncate privides strings truncation utilities.
 package truncate
 
-var maxQueryLength = 2048
+var defaultMaxQueryLength = 2048
+
+// Query truncate query to specific length of chars, if needed. -1: No limit, 0: Default (2048).
+// Also truncate all invalid UTF-8 chars.
+func Query(q string, maxQueryLength int) (query string, truncated bool) {
```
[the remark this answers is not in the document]
artemgavrilov: Yes, better to not use named return where it's possible

artemgavrilov (approved):

agent/utils/truncate/query.go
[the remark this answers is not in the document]
JiriCtvrtka: Agree, but we have a lot of places in PMM where namedreturn is used and we dont see this warnings because nobody touched it.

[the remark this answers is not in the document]
JiriCtvrtka: Removed atleast from touched place.

managed/models/database_test.go
BupycHuk: it's supposed to be 58, not 68 :)

BupycHuk (approved):

managed/models/database_test.go
artemgavrilov: @BupycHuk There is the problem with this test, it uses some old migrations, but our models are always represent last migration. So we should rewrite inserts here from reform to plain SQL, or drop this test at all. To be honest I don't see any value in testing one year old migration over and over again.

PR #1223, opened 2022-09-16
Add issues templates

artemgavrilov:
Here is how it looks like: 
Bug form: https://github.com/percona/pmm/blob/issue-tempates/.github/ISSUE_TEMPLATE/bug.yml
Features form: https://github.com/percona/pmm/blob/issue-tempates/.github/ISSUE_TEMPLATE/feature.yml


BupycHuk (approved):

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,58 @@
+name: Bug Report
+description: File a bug report
+labels: ["bug", "triage"]
```
artemgavrilov, line 3: `triage` lable means that `issue needs a confirmation` or something like this. I took it from github examples, we can decide later how we will work with this.

denisok (approved):

PR #1234, opened 2022-09-20
PMM-10467 Alert rules migration script

artemgavrilov:
PMM-10467

Build: SUBMODULES-0


migration.py
```diff
@@ -0,0 +1,108 @@
...
+import json
+
+import requests
+
+
+def prepare_labels(rule):
+    if "labels" not in rule:
+        return
+
+    labels = rule["labels"]
```
gen1us2k, line 13: It can produce an exception. Better to use 

```
labels = rule.get('labels', None)
if not labels :
  labels.update{}...
  
```

migration.py
```diff
@@ -0,0 +1,108 @@
...
+    labels.update({
+        "percona_alerting": "1",
+        "severity": rule["severity"].lstrip("SEVERITY_").lower(),
+        "template_name": rule["template_name"]
```
gen1us2k, line 17: Same thing here. Script can just fail with `KeyError` exception

migration.py
```diff
@@ -0,0 +1,108 @@
...
+    return annotations
+
+
+def prepare_expression(rule):
+    expr = rule["expr"]
+
+    if "filters" not in rule:
+        return expr
+
+    for f in rule["filters"]:
```
gen1us2k, line 35:
```
for f in rule.get('filters', []):
  key = f.get('key', "")
  value = f.get('key', '')
  expr = f"label_match({expr}, "{key}", "{value}")
```
This is more pythonic way and it has some fallbacks to the empty data type without any exceptions

migration.py
```diff
@@ -0,0 +1,108 @@
...
+                    help="folder in Grafana where to put migrated alert rules (default: %(default)s)")
+parser.add_argument("-g", "--group", default="migrated",
+                    help="alert group name for migrated alert rules (default: %(default)s)")
+
+config = vars(parser.parse_args())
+auth = (config["user"], config["password"])
+
+# Create alert group for migrated rules
+groupURL = '{server_url}/graph/api/ruler/grafana/api/v1/rules/{folder}/{group}'.format(**config)
+groupReq = requests.get(groupURL, auth=auth, verify=config["insecure"])
```
gen1us2k, line 82: requests.get can produce exceptions that should be handled (e.g. call sys.exit here with some message). The best way is to use try/except here

migration.py
```diff
@@ -0,0 +1,108 @@
...
+alertRulesGroup = json.loads(groupReq.text)
+if "interval" not in alertRulesGroup:
+    alertRulesGroup["interval"] = "1m"
+
+# Get Metrics datasource UID
+datasourceURL = '{server_url}/graph/api/datasources/1'.format(**config)
+datasourceReq = requests.get(datasourceURL, auth=auth, verify=config["insecure"])
+datasource = json.loads(datasourceReq.text)
```
gen1us2k, line 90: requests library has `datasourceReq.json()` method that returns parsed json

migration.py
```diff
@@ -0,0 +1,108 @@
...
+datasourceUID = datasource["uid"]
+
+# Get existing Integrated Alerting alert rules
+iaRulesURL = '{server_url}/v1/management/ia/Rules/List'.format(**config)
+iaRulesReq = requests.post(iaRulesURL, auth=auth, verify=config["insecure"])
```
gen1us2k, line 95: What about checking statuscode here?

BupycHuk (approved):

ia_migration.py
```diff
@@ -0,0 +1,111 @@
...
+parser.add_argument("-f", "--folder", default="Experimental",
+                    help="folder in Grafana where to put migrated alert rules (default: %(default)s)")
+parser.add_argument("-g", "--group", default="migrated",
+                    help="alert group name for migrated alert rules (default: %(default)s)")
+
+config = vars(parser.parse_args())
+auth = (config["user"], config["password"])
+
+# Create alert group for migrated rules
+groupURL = '{server_url}/graph/api/ruler/grafana/api/v1/rules/{folder}/{group}'.format(**config)
```
oksana-grishchenko, line 83: Does the parser allow to handle `server-url` argument as `server_url`?

oksana-grishchenko (approved):
LGTM considering the humble experience I have in Python

PR #1270, opened 2022-09-29
Fix IA migration script

artemgavrilov:
PMM-0

Build: SUBMODULES-0


oksana-grishchenko (approved):

PavelKhripkov:
This will wrap `expr` recursively. Is it what we want? https://github.com/percona/pmm/blob/4076a3f08f2ecbc1a668e71d906c7aeb0f332b92/ia_migration.py#L43

artemgavrilov:
> This will wrap `expr` recursively. Is it what we want?
> 
> https://github.com/percona/pmm/blob/4076a3f08f2ecbc1a668e71d906c7aeb0f332b92/ia_migration.py#L43

Yes,  we are using VM filter function here, and it can handle only one KV pair at once. So if we have multiple filters we should wrap original expression multiple times

https://docs.victoriametrics.com/MetricsQL.html#label_match

PavelKhripkov (approved):

PavelKhripkov:
@artemgavrilov do we need license header in the migration script?

artemgavrilov:
> @artemgavrilov do we need license header in the migration script?

I don't know, but added)

BupycHuk (approved):

gen1us2k (approved):

PR #1281, opened 2022-10-04
PMM-10814: List active service types

michal-kralik:
PMM-10814

Build: SUBMODULES-2813


BupycHuk (approved):

BupycHuk:
@michal-kralik Can we write a few API tests?

managed/models/service_helpers.go
```diff
@@ -123,6 +124,33 @@ func FindServices(q *reform.Querier, filters ServiceFilters) ([]*Service, error)
...
+	rows, err := q.Query(query)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	defer func() {
+		if err := rows.Close(); err != nil {
+			logrus.Debug(err)
+		}
+	}()
```
artemgavrilov, line 139: Here we will overwrite error returned on line 145

managed/models/service_helpers.go
```diff
@@ -123,6 +124,33 @@ func FindServices(q *reform.Querier, filters ServiceFilters) ([]*Service, error)
...
+
+	res := []ServiceType{}
```
artemgavrilov, line 141: https://dave.cheney.net/2014/05/24/on-declaring-variables
```suggestion
	var res []ServiceType
```

managed/models/service_helpers.go
```diff
@@ -123,6 +124,33 @@ func FindServices(q *reform.Querier, filters ServiceFilters) ([]*Service, error)
...
+	for rows.Next() {
+		var serviceType ServiceType
+		if err := rows.Scan(&serviceType); err != nil {
```
artemgavrilov, line 144:
```suggestion
		if err = rows.Scan(&serviceType); err != nil {
```

managed/models/service_helpers_test.go
```diff
@@ -266,6 +266,15 @@ func TestServiceHelpers(t *testing.T) {
 		}})
 	})
 
+	t.Run("FindActiveServiceTypes", func(t *testing.T) {
+		q, teardown := setup(t)
+		defer teardown(t)
+
+		types, err := models.FindActiveServiceTypes(q)
+		assert.NoError(t, err)
+		assert.Equal(t, 6, len(types))
```
artemgavrilov:
```suggestion
		assert.Len(t, types, 6)
```

managed/services/inventory/grpc/services_server.go
```diff
@@ -88,6 +88,25 @@ func (s *servicesServer) ListServices(ctx context.Context, req *inventorypb.List
...
+	types, err := s.s.ListActiveServiceTypes(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	res := &inventorypb.ListActiveServiceTypesResponse{
+		ServiceTypes: make([]inventorypb.ServiceType, 0, len(types)),
+	}
+
+	res.ServiceTypes = append(res.ServiceTypes, types...)
```
artemgavrilov, line 105: It seems we can just do it like this:
```suggestion
	res := &inventorypb.ListActiveServiceTypesResponse{
		ServiceTypes: types,
	}
```

managed/services/inventory/services.go
```diff
@@ -76,6 +76,40 @@ func (ss *ServicesService) List(ctx context.Context, filters models.ServiceFilte
...
+func (ss *ServicesService) ListActiveServiceTypes(ctx context.Context) ([]inventorypb.ServiceType, error) {
+	var types []models.ServiceType
+	err := ss.db.InTransaction(func(tx *reform.TX) error {
+		var err error
+		types, err = models.FindActiveServiceTypes(tx.Querier)
+		return err
+	})
+	if err != nil {
+		return nil, err
+	}
```
artemgavrilov, line 91: Let's keep it consistent with other parts
```suggestion
	e := ss.db.InTransaction(func(tx *reform.TX) error {
		var err error
		types, err = models.FindActiveServiceTypes(tx.Querier)
		return err
	})
	if e != nil {
		return nil, e
	}
```

managed/services/inventory/services.go
```diff
@@ -76,6 +76,40 @@ func (ss *ServicesService) List(ctx context.Context, filters models.ServiceFilte
...
+			res = append(res, inventorypb.ServiceType_MONGODB_SERVICE) //nolint:nosnakecase
+		case models.PostgreSQLServiceType:
+			res = append(res, inventorypb.ServiceType_POSTGRESQL_SERVICE) //nolint:nosnakecase
+		case models.ProxySQLServiceType:
+			res = append(res, inventorypb.ServiceType_PROXYSQL_SERVICE) //nolint:nosnakecase
+		case models.HAProxyServiceType:
+			res = append(res, inventorypb.ServiceType_HAPROXY_SERVICE) //nolint:nosnakecase
+		case models.ExternalServiceType:
+			res = append(res, inventorypb.ServiceType_EXTERNAL_SERVICE) //nolint:nosnakecase
+		}
```
artemgavrilov, line 108: It seems that we can just remove this switch-case and just add every `t` to result

managed/models/service_helpers.go
artemgavrilov: Ah, no, I'm wrong.

artemgavrilov: BTW can we use different name for error variable here, to make it more obvious?

michal-kralik: Sure

managed/services/inventory/grpc/services_server.go
michal-kralik: Great catch

managed/services/inventory/services.go
michal-kralik: `t` is of type `string` and `res` requires `int`. It appears to me there needs to be a way to map the values to each other.
How would you suggest we use just `t`? Maybe I'm missing something.

artemgavrilov: Ah, yes, my bad. Nevermind.

artemgavrilov (approved):

PR #1285, opened 2022-10-06
PMM-10464  Telemetry data from Grafana

pkadej:
[PMM-10464](https://jira.percona.com/browse/PMM-10464) Telemetry data from Grafana

Build: [SUBMODULES-2914](https://github.com/Percona-Lab/pmm-submodules/pull/2914)


managed/services/telemetry/deps.go
```diff
@@ -47,6 +47,8 @@ type DataSourceLocator interface {
 
 // DataSource telemetry data source.
 type DataSource interface {
+	PreFetch(ctx context.Context) error
 	FetchMetrics(ctx context.Context, config Config) ([][]*pmmv1.ServerMetric_Metric, error)
+	PostFetch(ctx context.Context) error
```
ritbl, line 52: nit: I'd call it `AfterFetch`, because `PostFetch` is a bit less clear

managed/services/telemetry/telemetry.go
```diff
@@ -208,17 +208,22 @@ func (s *Service) processSendCh(ctx context.Context) {
...
 	var totalTime time.Duration
-	for _, telemetry := range s.config.telemetry {
-		// locate DS
-		ds, err := s.LocateTelemetryDataSource(telemetry.Source)
+
+	for _, dataSource := range dataSources {
+		err := dataSource.PreFetch(ctx)
 		if err != nil {
-			s.l.Debugf("failed to lookup telemetry datasource for [%s]:[%s]", telemetry.Source, telemetry.ID)
+			s.l.Debug("PreFetch failed", err)
```
ritbl, line 218: why it is not an Error?

managed/services/telemetry/datasource_grafana_sqlitedb.go
```diff
@@ -0,0 +1,124 @@
...
+		return err
+	}
+
+	if !sourceFileStat.Mode().IsRegular() {
+		return errors.Wrapf(err, "%s is not a regular file", d.config.DBFile)
+	}
+
+	source, err := os.Open(d.config.DBFile)
+	if err != nil {
+		d.log.Error(err)
```
ritbl, line 72: This is not a very good practice, it can lead to logging 2 times. I think it would be better to delegate handling to a caller.

ritbl: I understand that you are trying to have "critical" and "non-critical" errors this way. But it is not obvious.

managed/services/telemetry/datasource_grafana_sqlitedb.go
```diff
@@ -0,0 +1,124 @@
...
+		return err
+	}
+	defer func() {
+		if err := source.Close(); err != nil {
+			d.log.Errorf("Error closing file. %s", err)
+		}
+	}()
+
+	nBytes, err := io.Copy(tempFile, source)
```
ritbl, line 87: how about logging how many bytes were read? how big this file can become?

managed/services/telemetry/datasource_grafana_sqlitedb.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil || nBytes == 0 {
```
ritbl, line 88: I understand that `nBytes == 0` should not happen, but what if source file is empty?

I think we should state "nBytes == 0 && sourcefileLen > 0"

managed/services/telemetry/datasource_grafana_sqlitedb.go
```diff
@@ -0,0 +1,124 @@
...
+func NewDataSourceGrafanaSqliteDB(config DSGrafanaSqliteDB, l *logrus.Entry) DataSource {
+	return &dsGrafanaSelect{
+		log:      l,
+		config:   config,
+		db:       nil,
+		tempFile: "",
+	}
+}
+
+func (d *dsGrafanaSelect) PreFetch(ctx context.Context) error {
```
ritbl, line 59: We are copying and opening DB, what if we don't have config that is using this DataSource?

How about lazily openning? (on first attempt to use this DS)

We need `finalization` block though, to be called in the end.

managed/services/telemetry/datasource_grafana_sqlitedb.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (d *dsGrafanaSelect) PostFetch(ctx context.Context) error {
+	err := d.db.Close()
+	if err != nil {
+		return err
+	}
+
+	err = os.Remove(d.tempFile)
+	if err != nil {
+		d.log.Errorf("Error removing file. %s", err)
```
ritbl, line 120: wrap error instead?

managed/services/telemetry/telemetry.go
```diff
@@ -237,11 +242,34 @@ func (s *Service) prepareReport(ctx context.Context) *pmmv1.ServerMetric {
 			telemetryMetric.Metrics = append(telemetryMetric.Metrics, each...)
 		}
 	}
+
+	for _, dataSource := range dataSources {
+		err := dataSource.PostFetch(ctx)
+		if err != nil {
+			s.l.Debug("PostFetch failed", err)
```
ritbl, line 249: I think it is better to log with Error + specify where it is occuring (telemetry.id)

managed/services/telemetry/telemetry.go
```diff
@@ -208,17 +208,22 @@ func (s *Service) processSendCh(ctx context.Context) {
...
 	telemetryMetric, _ := s.makeMetric(ctx)
-
 	var totalTime time.Duration
-	for _, telemetry := range s.config.telemetry {
-		// locate DS
-		ds, err := s.LocateTelemetryDataSource(telemetry.Source)
+
+	for _, dataSource := range dataSources {
+		err := dataSource.PreFetch(ctx)
 		if err != nil {
```
ritbl, line 217: if `init` for a DS is failing should be consider that this DS should not be used for metrics fetching?

managed/services/telemetry/datasource_grafana_sqlitedb.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (d *dsGrafanaSelect) FetchMetrics(ctx context.Context, config Config) ([][]*pmmv1.ServerMetric_Metric, error) {
+	if d.db == nil {
+		return nil, errors.Errorf("temporary grafana database is not initialized: %s", d.config.DBFile)
+	}
+	return fetchMetricsFromDB(ctx, d.log, d.config.Timeout, d.db, config)
+}
+
+func (d *dsGrafanaSelect) PostFetch(ctx context.Context) error {
+	err := d.db.Close()
```
ritbl, line 113: what if `Init` is not called? db will be nil causing a panic

managed/services/telemetry/telemetry.go
```diff
@@ -237,11 +242,34 @@ func (s *Service) prepareReport(ctx context.Context) *pmmv1.ServerMetric {
 			telemetryMetric.Metrics = append(telemetryMetric.Metrics, each...)
 		}
 	}
+
+	for _, dataSource := range dataSources {
+		err := dataSource.PostFetch(ctx)
```
ritbl, line 247: should we cleanup is `init` has failed?

ritbl (commented):
please see my comments

managed/services/telemetry/datasource_grafana_sqlitedb.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+	}()
+
+	nBytes, err := io.Copy(tempFile, source)
+	if err != nil || nBytes == 0 {
+		d.log.Error(err)
+		return errors.Wrapf(err, "cannot create copy of database file %s", d.config.DBFile)
+	}
+
+	db, err := sql.Open("sqlite3", tempFile.Name())
```
dliakhov, line 93: Is it possible to open new connection for existing sqllite db file without copying it?

managed/services/telemetry/telemetry.go
```diff
@@ -208,17 +208,22 @@ func (s *Service) processSendCh(ctx context.Context) {
 }
 
 func (s *Service) prepareReport(ctx context.Context) *pmmv1.ServerMetric {
+	dataSources := s.locateDataSources(s.config.telemetry)
```
dliakhov, line 211: I guess it's not needed every time init this dataSources map, because config file cannot be changed in the runtime. 

I guess you can move initailization of this variable to Run() or New() method.

pkadej: Done.

managed/services/telemetry/datasource_grafana_sqlitedb.go
pkadej: We need to copy database to avoid its corruption.

managed/services/telemetry/telemetry.go
pkadej: Fixed.

managed/services/telemetry/telemetry.go
pkadej: Fixed.

managed/services/telemetry/telemetry.go
pkadej: Fixed.

managed/services/telemetry/datasource_grafana_sqlitedb.go
pkadej: Fixed.

managed/services/telemetry/datasource_grafana_sqlitedb.go
pkadej: Fixed for both cases.

managed/services/telemetry/datasource_grafana_sqlitedb.go
pkadej: Done

managed/services/telemetry/datasource_grafana_sqlitedb.go
pkadej: Fixed.

managed/services/telemetry/deps.go
pkadej: Changed.

managed/services/telemetry/deps.go
```diff
@@ -47,6 +47,8 @@ type DataSourceLocator interface {
 
 // DataSource telemetry data source.
 type DataSource interface {
+	PreFetch(ctx context.Context) error
```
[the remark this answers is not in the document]
pkadej: Changed.

managed/services/telemetry/telemetry.go
```diff
@@ -128,7 +128,8 @@ func (s *Service) Run(ctx context.Context) {
 			return
 		}
 
-		report := s.prepareReport(ctx)
+		dataSources := s.locateDataSources(s.config.telemetry)
```
dliakhov, line 131: could you please move code `dataSources := s.locateDataSources(s.config.telemetry)` outside of doSend function? I guess it's not needed to init `dataSources` every time when doSend function is called because config file cannot be changed in the runtime.

pkadej: Done.

managed/services/telemetry/datasource_grafana_sqlitedb.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+
+	defer func() {
+		if err := source.Close(); err != nil {
+			d.log.Errorf("Error closing file. %s", err)
+		}
+	}()
+
+	nBytes, err := io.Copy(tempFile, source)
```
gen1us2k, line 90: Why do we need to copy a database file?

pkadej: Because SQLite database file is can be corrupted.
The problem with file based sqlite3 database is the concurrency. The database can be easily corrupted if two processes are using the same file.  Official site says that db should be threadsafe, but many users complained about their corrupted databases.

gen1us2k: AFAIK SQLite handles concurrent reads gracefully and has problems with concurrent writes, leading to lock contention. Anyway, looks good.

gen1us2k (approved):

managed/services/telemetry/datasource_grafana_sqlitedb.go
pkadej: Yes, but we have no full control on sql driver behavior. Thank you.

artemgavrilov: We still should consider that copy of DB may be corrupted: https://www.sqlite.org/howtocorrupt.html

managed/services/telemetry/datasource_grafana_sqlitedb.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// Events, errors and driver for for grafana sqlite database.
+	pmmv1 "github.com/percona-platform/saas/gen/telemetry/events/pmm"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	_ "modernc.org/sqlite"
+)
+
+type dsGrafanaSelect struct {
+	log      *logrus.Entry
```
artemgavrilov, line 33: We name it just `l` everywhere.

managed/services/telemetry/datasource_grafana_sqlitedb.go
artemgavrilov: Does this question has an answer?

managed/services/telemetry/datasource_grafana_sqlitedb.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (d *dsGrafanaSelect) Dispose(ctx context.Context) error {
+	err := d.db.Close()
+	if err != nil {
+		return err
+	}
+
+	err = os.Remove(d.tempFile)
+	if err != nil {
+		return errors.Wrapf(err, "Error removing file.")
```
artemgavrilov, line 122: https://github.com/golang/go/wiki/CodeReviewComments#error-strings
```suggestion
		return errors.Wrapf(err, "failed to remove sqlite database file")
```

managed/services/telemetry/datasource_grafana_sqlitedb_test.go
```diff
@@ -0,0 +1,97 @@
...
+func TestDatasource(t *testing.T) {
+	t.Parallel()
+	logger := logrus.StandardLogger()
+	logger.SetLevel(logrus.DebugLevel)
+	logEntry := logrus.NewEntry(logger)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*100)
+	t.Cleanup(func() {
+		cancel()
+	})
```
artemgavrilov, line 38: Why do you  need context with timeout in tests?

managed/services/telemetry/datasource_grafana_sqlitedb_test.go
```diff
@@ -0,0 +1,97 @@
...
+				MetricName: "total_users_in_database",
+				Column:     "total",
+			},
+		},
+	}
+
+	t.Run("get metrics from db", func(t *testing.T) {
+		t.Parallel()
+		databaseFile, err := filepath.Abs("../../testdata/telemetry/grafana_sqlite.db")
+		assert.NoError(t, err)
```
artemgavrilov, line 56: It has no sense to continue this test if you got an error here.
```suggestion
		require.NoError(t, err)
```

managed/services/telemetry/datasource_grafana_sqlitedb_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+		conf := &DSGrafanaSqliteDB{
+			Enabled: true,
+			Timeout: time.Second * 10,
+			DBFile:  databaseFile,
+		}
+		grafanaDB := NewDataSourceGrafanaSqliteDB(*conf, logEntry)
+
+		err = grafanaDB.Init(ctx)
+		assert.NoError(t, err)
```
artemgavrilov, line 66:
```suggestion
		require.NoError(t, err)
```

managed/services/telemetry/datasource_grafana_sqlitedb_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+		metrics, err := grafanaDB.FetchMetrics(ctx, *config)
+		assert.NoError(t, err)
```
artemgavrilov, line 69:
```suggestion
		require.NoError(t, err)
```

managed/services/telemetry/datasource_grafana_sqlitedb_test.go
```diff
@@ -0,0 +1,97 @@
...
+		assert.Equal(t, len(metrics), 1)
+
+		err = grafanaDB.Dispose(ctx)
+		assert.NoError(t, err)
```
artemgavrilov, line 73:
```suggestion
		require.NoError(t, err)
```

managed/services/telemetry/telemetry.go
```diff
@@ -208,17 +210,29 @@ func (s *Service) processSendCh(ctx context.Context) {
...
-	for _, telemetry := range s.config.telemetry {
-		// locate DS
-		ds, err := s.LocateTelemetryDataSource(telemetry.Source)
+
+	// initialize datasources
+	for sourceName, dataSource := range s.dataSourcesMap {
+		err := dataSource.Init(ctx)
 		if err != nil {
-			s.l.Debugf("failed to lookup telemetry datasource for [%s]:[%s]", telemetry.Source, telemetry.ID)
+			s.l.Error("Telemetry datasource init failed", err)
```
artemgavrilov, line 221: Better to use `warn` level for cases when work can be continued.
```suggestion
			s.l.Warnf("Telemetry datasource %s init failed: %v", sourceName, err)
```

managed/services/telemetry/telemetry.go
```diff
@@ -248,13 +262,37 @@ func (s *Service) prepareReport(ctx context.Context) *pmmv1.ServerMetric {
 
 		telemetryMetric.Metrics = append(telemetryMetric.Metrics, metrics...)
 	}
+
+	// datasources disposal
+	for sourceName, dataSource := range initializedDataSources {
+		err := dataSource.Dispose(ctx)
+		if err != nil {
+			s.l.Debugf("Dispose of %s datasource failed. %v", sourceName, err)
```
artemgavrilov, line 270:
```suggestion
			s.l.Debugf("Dispose of %s datasource failed: %v", sourceName, err)
```

artemgavrilov (commented):
My biggest question why are we doing this now, when in the next release we plan to migrate Grafana from sqlite to PG?

https://jira.percona.com/browse/PMM-4466

managed/services/config/pmm-managed.yaml
```diff
@@ -16,6 +16,10 @@ services:
         separate_credentials:
           username: pmm
           password: pmm
+      GRAFANADB_SELECT:
+        enabled: true
+        timeout: 5s
+        db_file: /srv/grafana/grafana.db
```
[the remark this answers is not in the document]
pkadej: I think moving this out somewhere else is not a part of this ticket.
All pmm-server's telemetry datasources are defined here. I suggest making a task for cleaning this up.

managed/services/telemetry/datasource_grafana_sqlitedb.go
pkadej: Changed.

managed/services/telemetry/telemetry.go
pkadej: Changed.

managed/services/telemetry/telemetry.go
pkadej: Changed.

managed/services/telemetry/datasource_grafana_sqlitedb_test.go
pkadej: Changed.

managed/services/telemetry/datasource_grafana_sqlitedb_test.go
pkadej: Changed.

managed/services/telemetry/datasource_grafana_sqlitedb_test.go
pkadej: Changed.

managed/services/telemetry/datasource_grafana_sqlitedb_test.go
pkadej: Changed.

managed/services/telemetry/datasource_grafana_sqlitedb_test.go
pkadej: Changed.

managed/services/telemetry/datasource_grafana_sqlitedb.go
pkadej: Changed.

managed/services/telemetry/datasource_grafana_sqlitedb.go
pkadej: I had missed this one. Yes.
Now, code runs calls Dispose (ex. Postfetch) only on successfully initialized datasource.

https://github.com/percona/pmm/blob/PMM-10464_grafana-telemetry/managed/services/telemetry/telemetry.go#L224
and
https://github.com/percona/pmm/blob/PMM-10464_grafana-telemetry/managed/services/telemetry/telemetry.go#L267

pkadej:
> My biggest question why are we doing this now, when in the next release we plan to migrate Grafana from sqlite to PG?
> 
> https://jira.percona.com/browse/PMM-4466

Since migration is a big and complex task which haven't ended successfully in the past, @rnovikovP agreed for such solution.

managed/services/telemetry/datasource_grafana_sqlitedb.go
pkadej: Still - better to have corrupted copy of database and having not sent telemetry than corrupt the original one :)

managed/services/telemetry/datasource_grafana_sqlitedb.go
```diff
@@ -0,0 +1,126 @@
...
+	return d.config.Enabled
+}
+
+// NewDataSourceGrafanaSqliteDB makes new data source for grafana sqlite database metrics.
+func NewDataSourceGrafanaSqliteDB(config DSGrafanaSqliteDB, l *logrus.Entry) DataSource {
+	return &dsGrafanaSelect{
+		l:        l,
+		config:   config,
+		db:       nil,
+		tempFile: "",
```
artemgavrilov, line 55: You can omit these fields, it doesn't add any value.

PR #1291, opened 2022-10-11
PMM-7: describe managed update process

michal-kralik:
PMM-7

### DIFF
--- a/docs/managed/update-process.md
+++ b/docs/managed/update-process.md
@@ -1,0 +1,11 @@
+# Update of managed
+
+Update of `managed` is triggered by sending a [StartUpdate](https://github.com/percona/pmm/blob/6761010b8b30042936c58c022752f6b57581afee/api/serverpb/server.proto#L325) message.
+This performs the following actions:
+1. Runs [pmm-update](https://github.com/percona/pmm-update) command to initiate an update
+2. `pmm-update` first updates itself to the latest version and restarts
+3. `pmm-update` then runs a set of Ansible tasks to properly update all components
+
+## Notes
+- `pmm-update` does not handle rollbacks in case of errors
+- `pmm-update` assumes we have root


docs/managed/update-process.md
```diff
@@ -0,0 +1,11 @@
+# Update of managed
+
+Update of `managed` is triggered by sending a [StartUpdate](https://github.com/percona/pmm/blob/6761010b8b30042936c58c022752f6b57581afee/api/serverpb/server.proto#L325) message.
```
atymchuk, line 3: This endpoint is not just updating `managed`, but the whole PMM Server.

atymchuk (changes requested):

docs/managed/update-process.md
michal-kralik: Great catch

palash25 (approved):

BupycHuk (approved):

docs/managed/update-process.md
```diff
@@ -0,0 +1,11 @@
+# Update of PMM Server
+
+Update of PMM Server which includes `managed` and other compontents is triggered by sending a [StartUpdate](https://github.com/percona/pmm/blob/6761010b8b30042936c58c022752f6b57581afee/api/serverpb/server.proto#L325) message.
```
atymchuk, line 3:
```suggestion
Update of PMM Server which includes `managed` and other components is triggered by sending a [StartUpdate](https://github.com/percona/pmm/blob/6761010b8b30042936c58c022752f6b57581afee/api/serverpb/server.proto#L325) message.
```

docs/managed/update-process.md
```diff
@@ -0,0 +1,11 @@
...
+This performs the following actions:
+1. Runs [pmm-update](https://github.com/percona/pmm-update) command to initiate an update
+2. `pmm-update` first updates itself to the latest version and restarts
+3. `pmm-update` then runs a set of Ansible tasks to properly update all components
```
atymchuk, line 7:
```suggestion
3. `pmm-update` then runs a set of Ansible tasks to update all other components
```

docs/managed/update-process.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Notes
+- `pmm-update` does not handle rollbacks in case of errors
+- `pmm-update` assumes we have root
```
atymchuk, line 11:
```suggestion
- `pmm-update` runs with root privileges
```

atymchuk (changes requested):

docs/managed/update-process.md
michal-kralik: I suggest we use the following instead

"`pmm-update` requires root priveleges to run"

atymchuk: Not bad, but it sounds like the user needs to additionally configure something. In docker, for example, it is already configured as root.

artemgavrilov (approved):

atymchuk (approved):

docs/managed/update-process.md
```diff
@@ -0,0 +1,11 @@
+# Update of PMM Server
+
+Update of PMM Server which includes `managed` and other components is triggered by sending a [StartUpdate](https://github.com/percona/pmm/blob/6761010b8b30042936c58c022752f6b57581afee/api/serverpb/server.proto#L325) message.
```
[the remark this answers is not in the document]
artemgavrilov: @michal-kralik ^ Let's use link to main branch instead of particular commit

PR #1293, opened 2022-10-12
PMM-7: properly handle exec.ExitError

michal-kralik:
PMM-7

Build: SUBMODULES-2824


artemgavrilov (approved):

admin/cli/cli.go
```diff
@@ -138,8 +138,10 @@ func printResponse(opts *flags.GlobalFlags, res commands.Result, err error) erro
 		os.Exit(1)
 
 	case *exec.ExitError: // from config command that execs `pmm-agent setup`
-		printExitError(opts, res, err)
-		os.Exit(err.ExitCode())
+		if res != nil {
+			printExitError(opts, res, err)
+			os.Exit(err.ExitCode())
```
BupycHuk, line 143: shouldn't this be outside of the `if` clause?

michal-kralik: No, I believe this is correct.
Today when a command returns an `exec.ExitError` and there's no `res`, we panic.
I believe this is because up until now the `exec.ExitError` occurred in only one case and was handled in a special way.

From design perspective this is not correct and we shall be able to handle this type of error in a standard way.
With this change the `exec.ExitError`, not handled in the special way of receiving `res`, is handled as any other error returned by commands.

palash25 (approved):

admin/cli/cli.go
BupycHuk: I meant only `os.Exit`.
But anyway why we may have `*exec.ExitError` with no `res`?.

michal-kralik: The `exec.ExitError` can be returned as part of running commands on the machine such as `apt update`.
If apt is locked, the command fails, it's propagated up and handled in this function.

PR #1312, opened 2022-10-18
PMM-4087 pmm agent metrics collection for itself

rishat-ishbulatov:
[PMM-4087](https://jira.percona.com/browse/PMM-4087)

Build: SUBMODULES-0


ghost (changes requested):
@rishat-ishbulatov could you please elaborate on the changes in `agents/supervisor/supervisor.go`?

I'd like to better understand the design choice of combining supervisor with prometheus stats.
Would it make more sense to keep the stats separately?

rishat-ishbulatov:
> @rishat-ishbulatov could you please elaborate on the changes in `agents/supervisor/supervisor.go`?
> 
> I'd like to better understand the design choice of combining supervisor with prometheus stats. Would it make more sense to keep the stats separately?

agent/agents/supervisor/supervisor.go
```diff
@@ -214,6 +244,9 @@ func (s *Supervisor) SetState(state *agentpb.SetStateRequest) {
 }
 
 func (s *Supervisor) storeLastStatus(agentID string, status inventorypb.AgentStatus) {
+	s.agentStatuses.WithLabelValues(agentID).Set(float64(status))
+	s.agentStatusChangeTotal.WithLabelValues(agentID).Inc()
```
michal-kralik, line 248: Is there any reason we update the values before acquiring a lock?

agent/agents/supervisor/supervisor.go
```diff
@@ -663,6 +697,9 @@ func (s *Supervisor) stopAll() {
 
 // Describe implements prometheus.Collector.
 func (s *Supervisor) Describe(ch chan<- *prometheus.Desc) {
+	s.agentStatuses.Describe(ch)
+	s.agentStatusChangeTotal.Describe(ch)
+	s.agentQANBucketLength.Describe(ch)
```
michal-kralik, line 702: Is there any reason we update the values before acquiring a lock?

agent/agents/supervisor/supervisor.go
```diff
@@ -673,6 +710,9 @@ func (s *Supervisor) Describe(ch chan<- *prometheus.Desc) {
 
 // Collect implement prometheus.Collector.
 func (s *Supervisor) Collect(ch chan<- prometheus.Metric) {
+	s.agentStatuses.Collect(ch)
+	s.agentStatusChangeTotal.Collect(ch)
+	s.agentQANBucketLength.Collect(ch)
```
michal-kralik, line 715: Is there any reason we update the values before acquiring a lock?

ghost (changes requested):
I'd opt for separating metrics from the actual logic to make the code more readable.
I understand `channel` has been designed in a similar fashion before. Either way, this could be a good time for a refactor.

Please also address linting issues.

agent/agents/supervisor/supervisor.go
rishat-ishbulatov: Hi, Michal. Fixed a place with mutex. According to linter errors, everything seems to be fixed, maybe with the exception of cyclomatic complexity, but there is just one switch. Regarding the refactoring proposal. No problem. Is it to be done as a service or... ? Maybe in the project there is an example of what you would like to see? I suppose this will also need to be done for the channel?

rishat-ishbulatov:
Hi, Michal. Fixed a place with mutex. According to linter errors, everything seems to be fixed, maybe with the exception of cyclomatic complexity, but there is just one switch. Regarding the refactoring proposal. No problem. Is it to be done as a service or... ? Maybe in the project there is an example of what you would like to see? I suppose this will also need to be done for the channel?

michal-kralik:
> Hi, Michal. Fixed a place with mutex. According to linter errors, everything seems to be fixed, maybe with the exception of cyclomatic complexity, but there is just one switch. Regarding the refactoring proposal. No problem. Is it to be done as a service or... ? Maybe in the project there is an example of what you would like to see? I suppose this will also need to be done for the channel?

I would suggest we create a `metrics` package in `agent` and `client` and store this information there.
I'd be looking for a solution where the metrics logic would be separated from the application logic so the code is easier to understand and maintain.

Re linting - try running the linter locally such as `bin/golangci-lint run -c=.golangci.yml --new-from-rev=main`
This prints a couple issues.

agent/agents/supervisor/supervisor.go
```diff
@@ -533,6 +565,7 @@ func (s *Supervisor) startBuiltin(agentID string, builtinAgent *agentpb.SetState
 				}
 			}
 			if change.MetricsBucket != nil {
+				s.agentQANBucketLength.WithLabelValues(agentID).Set(float64(len(change.MetricsBucket)))
```
BupycHuk, line 568: should it be set? is there an option to increase the value?

agent/client/channel/channel.go
```diff
@@ -202,9 +247,13 @@ func (c *Channel) send(msg *agentpb.AgentMessage) {
...
 	if err != nil {
 		c.close(errors.Wrap(err, "failed to send message"))
+		c.mFailed.Inc()
 		return
 	}
-	c.mSend.Inc()
+
+	messageType := string(getMessageTypeFromAgentMessage(msg))
+	c.mSend.WithLabelValues(messageType).Inc()
+	c.mSizeSend.WithLabelValues(messageType).Set(float64(proto.Size(msg)))
```
BupycHuk, line 256: let's use `Add` instead of `Set`

agent/client/channel/channel.go
```diff
@@ -233,78 +281,95 @@ func (c *Channel) runReceiver() {
...
 
+		messageType := Unknown
 		switch p := msg.Payload.(type) {
 		// requests
 		case *agentpb.ServerMessage_Ping:
 			c.requests <- &ServerRequest{
 				ID:      msg.Id,
 				Payload: p.Ping,
 			}
+			messageType = Ping
```
BupycHuk, line 292: can we use `getMessageTypeFromAgentMessage`?

agent/client/client.go
```diff
@@ -77,6 +77,9 @@ type Client struct {
 
 	cus      *connectionuptime.Service
 	logStore *tailog.Store
+
+	connectionUptime          prometheus.Gauge
+	connectionEstablishedTime time.Time
```
BupycHuk, line 82: can we get these metrics from`connectionuptime.Service`?

agent/client/client.go
```diff
@@ -819,9 +829,23 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
...
+func (c *Client) registerConnectionStatus(connected bool) {
+	switch {
+	case !c.connectionEstablishedTime.IsZero() && connected:
+		c.connectionUptime.Set(time.Since(c.connectionEstablishedTime).Seconds())
+	case !connected:
+		c.connectionEstablishedTime = time.Time{}
+		c.connectionUptime.Set(0)
+	case connected:
+		c.connectionEstablishedTime = time.Now()
+		c.connectionUptime.Set(0)
```
BupycHuk, line 845: why does it 0?

BupycHuk (changes requested):

PR #1313, opened 2022-10-18
PMM-10867: Upgrade client via pmm cli

michal-kralik:
PMM-10867

Build: SUBMODULES-2834


admin/commands/pmm/client/upgrade.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+type upgradeResult struct{}
+
+// Result is a command run result.
+func (res *upgradeResult) Result() {}
+
+// String stringifies command result.
+func (res *upgradeResult) String() string {
+	return "ok"
```
BupycHuk, line 59: I think message shouldn't be just `ok` :)

admin/pkg/common/distribution.go
```diff
@@ -0,0 +1,231 @@
+// Copyright 2019 Percona LLC
```
BupycHuk, line 1: shouldn't this file be in the client package?

admin/commands/pmm/client/upgrade.go
```diff
@@ -0,0 +1,200 @@
...
+
+const versionLatest = "latest"
+
+type packageManager int
+
+const (
+	dnf packageManager = iota
+	yum
+	apt
+)
```
BupycHuk, line 182: let's put all constants to the top

admin/pkg/common/distribution.go
michal-kralik: Great catch.
Originally I planned to reuse this file for server as well but I just realized most of it is just client. Moved.

admin/commands/pmm/client/upgrade.go
michal-kralik: Agreed. I suggest we keep it like this for now though.
The reason is the UI still needs a lot of polishing. The goal of this PR is to have a functional prototype ready, review it and address the UI in the upcoming PRs.
With this approach I expect the PRs shall be easier to review, not having thousands of lines of changes :)

gen1us2k (approved):
Looks good to me and I have nothing to add except Nurlan's comment

idoqo (approved):

admin/pkg/client/client.go
```diff
@@ -0,0 +1,222 @@
...
+	// Docker represents Docker installation of PMM Agent or Server.
+	Docker DistributionType = "docker"
+	// PackageManager represents installation of PMM Agent or Server via a package manager.
+	PackageManager DistributionType = "package-manager"
+	// Tarball represents installation of PMM Agent or Server via a tarball.
+	Tarball DistributionType = "tarball"
+)
+
+// ErrLatestVersionNotFound is returned when we cannot determine what the latest version is.
+var ErrLatestVersionNotFound = fmt.Errorf("LatestVersionNotFound")
```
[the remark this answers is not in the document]
michal-kralik: Trying to be consistent with the rest of the code.
We discussed this particular topic in https://github.com/percona/pmm/pull/1207#pullrequestreview-1127154047
Feel free to share your thoughts though.

PR #1314, opened 2022-10-18
PMM-10900: custom image and option to run pmm with external clickhouse

ritbl:
[PMM-10900](https://jira.percona.com/browse/PMM-10900)

Fully backward-compatible way to allow developers to override used Devcontainer and option to use external ClickHouse DB


Makefile
```diff
@@ -2,28 +2,40 @@
 
 include Makefile.include
 
+ifeq ($(PROFILES),)
+PROFILES := 'pmm'
```
ritbl, line 6: by default run as it runs now

docker-compose.yml
```diff
@@ -19,15 +21,86 @@ services:
...
+  # PMM with external ClickHouse DB
+  ch:
+    profiles:
+      - pmm-ch
+    image: ${CH_IMAGE:-clickhouse/clickhouse-server:22.6.9.11-alpine}
+    platform: linux/amd64
+    hostname: ${CH_HOSTNAME:-ch}
+    ports:
+      - ${CH_PORT:-9000}:9000
+  pmm-managed-server-ch:
```
ritbl, line 78: `ch` and `pmm-managed-server-ch` will be used when develop runs container the following way:

```
PROFILES=pmm-ch make env-up
```

docker-compose.yml
```diff
@@ -19,15 +21,86 @@ services:
...
+      - ${PMM_PORT_CH_HTTP:-11123}:8123
+    volumes:
+      - ./:/root/go/src/github.com/percona/pmm
+      - ./Makefile.devcontainer:/root/go/src/github.com/percona/pmm/Makefile:ro # change Makefile in devcontainer
+      # caching
+      - go-modules:/root/go/pkg/mod
+      - root-cache:/root/.cache
+
+  # PMM with external ClickHouse DB
+  ch:
```
ritbl, line 70: use external ClickHouse (force AMD64), as DBaaS fails (in arm64 builds, amd64 works fine) with one that is running in container.

managed/Makefile
```diff
@@ -36,8 +36,13 @@ release:                        ## Build pmm-managed release binaries.
 	env CGO_ENABLED=0 go build -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/ ./cmd/...
 	$(PMM_RELEASE_PATH)/pmm-managed --version
 
+ARCH=$(shell uname -m)
 release-dev:
-	go build -race -gcflags="all=-N -l" -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/ ./cmd/...
+	@if [ $(ARCH) = "aarch64" ]; then\
+	    go build -gcflags="all=-N -l" -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/ ./cmd/...
```
ritbl, line 42: to ARM do not run with -race


`-race` causes crash building pmm-managed binaries


```
# github.com/percona/pmm/managed/cmd/pmm-managed
reflect.methodValueCall: nosplit stack over 792 byte limit
reflect.methodValueCall<0>
    grows 448 bytes, calls reflect.moveMakeFuncArgPtrs<1>
        grows 288 bytes, calls internal/abi.(*IntArgRegBitmap).Get<1>
            grows 48 bytes, calls runtime.racefuncenter<1>
                grows 0 bytes, calls racefuncenter<121>
                    grows 16 bytes, calls racecall<121>
                        grows 0 bytes, calls indirect
                            grows 0 bytes, calls runtime.morestack<0>
                            8 bytes over limit
```

Makefile
```diff
@@ -2,28 +2,40 @@
...
+endif
+
 env-up: 							## Start devcontainer.
+	COMPOSE_PROFILES=$(PROFILES) \
```
BupycHuk, line 10: do we really need to set it for each item?
can we extract it outside of items?

.gitignore
```diff
@@ -27,3 +27,5 @@ pmm-agent-dev.yaml
 
 # ViM temporary files
 *.sw[o,p]
+
+.env
```
ritbl, line 31: docker compose uses `.env` by default, it should not be committed

BupycHuk (commented):
will it work if user runs just `docker-compose up`?

docker-compose.yml
```diff
@@ -2,7 +2,9 @@
 version: '3.7'
 services:
   pmm-managed-server:
```
ritbl, line 4: this section was not changed, only `profiles` section is added

Makefile
ritbl: I need to pass default profile (`pmm`), for some reason it does'n work with default
image

image

BupycHuk (approved):

tshcherban (approved):

PR #1321, opened 2022-10-19
PMM-10029 Fix gRPC message size error during upgrade.

BupycHuk:
PMM-10029

Build: SUBMODULES-2837


managed/services/supervisord/supervisord.go
```diff
@@ -372,6 +374,9 @@ func (s *Service) UpdateLog(offset uint32) ([]string, uint32, error) {
 	for {
 		line, err := reader.ReadString('\n')
 		if err == nil {
+			if newOffset > s.gRPCMessageMaxSize {
+				return lines, newOffset, errors.WithStack(err)
```
artemgavrilov, line 378: `err` here is `nil`

artemgavrilov: BTW, we just return error an that's it? How it solves the problem?

denisok: @artemgavrilov where do we return an error? I think idea to return part of the log that fits grpc message size?
But it probably should be after this line: `lines = append` as otherwise `lines`is empty.

Looks like log api has pagination (request log by offset) and this forces this pagination.

denisok (approved):

denisok (changes requested):
wrong status, I mean to request changes

managed/services/supervisord/supervisord.go
BupycHuk: @artemgavrilov , @denisok fixed

denisok: but `lines` have nothing ?

BupycHuk: it does, it contains everything we collect during previous loops. it can be empty only in one case: if the length of one line is higher than max limit, which is almost impossible 
if we put this if after `lines = append` the size of lines message might become bigger than max limit.

BupycHuk: and there was error that I was comparing newOffset with max limit instead of diff between newOffset and offset

denisok (approved):

managed/services/supervisord/supervisord.go
denisok: yep, I see it now

JiriCtvrtka:
@artemgavrilov Everything should be fixed now. PTAL

idoqo (approved):

denisok:
@gen1us2k @artemgavrilov please review

managed/services/supervisord/supervisord.go
```diff
@@ -373,6 +375,9 @@ func (s *Service) UpdateLog(offset uint32) ([]string, uint32, error) {
 		line, err := reader.ReadString('\n')
 		if err == nil {
 			newOffset += uint32(len(line))
+			if newOffset-offset > s.gRPCMessageMaxSize {
```
artemgavrilov, line 378: What gonna happen in we have line that is longer than `gRPCMessageMaxSize` ?

gen1us2k (approved):

managed/services/supervisord/supervisord.go
denisok: @BupycHuk could you please check this comment?

PR #1327, opened 2022-10-21
PMM-10677: Implement parsing value of metric to json

dliakhov:
PMM-10677: Implement parsing value of metric to json

Explanation how it should work you can find here https://confluence.percona.com/pages/viewpage.action?pageId=138486089. See there info about field `data_json`.


managed/services/telemetry/config.default.yml
```diff
@@ -49,6 +49,20 @@ telemetry:
       - metric_name: "pmm_server_updates_disabled"
         column: "updates_disabled"
 
+  - id: PMMAgentRandomInfo
+    source: PMMDB_SELECT
+    query: agent_type, status from agents
+    summary: "PMMAgentRandomInfo"
+    json_format:
```
ritbl, line 56: nit: it would be good to follow naming schema. e.g. `data_json`

dliakhov: changed to data_json

managed/services/telemetry/config.default.yml
```diff
@@ -49,6 +49,7 @@ telemetry:
       - metric_name: "pmm_server_updates_disabled"
         column: "updates_disabled"
 
+
```
ritbl, line 52: nit: no need for additional space

dliakhov: Removed this line

PR #1340, opened 2022-10-24
Remove server local storage from code

PavelKhripkov:
PMM-0

Build: SUBMODULES-0


agent/runner/jobs/backup_location.go
```diff
@@ -19,8 +19,8 @@ type BackupLocationType string
 
 // BackupLocation types. Same as in managed/models/location_model.go.
 const (
-	S3BackupLocationType        BackupLocationType = "s3"
-	PMMClientBackupLocationType BackupLocationType = "pmm-client"
+	S3BackupLocationType         BackupLocationType = "s3"
+	FilesystemBackupLocationType BackupLocationType = "pmm-client"
```
artemgavrilov, line 23: It seems that we can also rename value here and in `managed/models/location_model.go`

idoqo, line 23:
```suggestion
	FilesystemBackupLocationType BackupLocationType = "filesystem"
```

idoqo (approved):

agent/runner/jobs/backup_location.go
PavelKhripkov: Double) Looks like I really have to change it)))

PavelKhripkov: Double) Looks like I really have to change it)))

PR #1342, opened 2022-10-25
PMM-10466: UI events

ritbl:
[PMM-10466](https://jira.percona.com/browse/PMM-10466)

Related Grafana changes:
 - https://github.com/percona-platform/grafana/pull/541

FB: 
 - https://github.com/Percona-Lab/pmm-submodules/pull/2847


managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,259 @@
...
+	pmmv1 "github.com/percona-platform/saas/gen/telemetry/events/pmm"
+	"github.com/percona/pmm/managed/services/telemetry"
+	"github.com/sirupsen/logrus"
+	"sort"
+	"time"
+
+	uievents "github.com/percona/pmm/api/uieventspb"
+)
+
+const cleanupAfterHours = 30
```
ritbl, line 32: I know that is not ideal and can cause reporting telemetry "from yesterday". But this can happen only if send interval is modified or there is network issue that last 24+h. This should be a very rare case.

managed/services/telemetry/extension.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"context"
+
+	pmmv1 "github.com/percona-platform/saas/gen/telemetry/events/pmm"
+)
+
+// Extension provides dynamic extension point for Telemetry.
+type Extension interface {
+	FetchMetrics(ctx context.Context, report *Config) ([]*pmmv1.ServerMetric_Metric, error)
+}
```
BupycHuk, line 12: do we really need a separate interface for extensions? it looks similar to `DataSource`

BupycHuk: btw, why do we use pointer for config here and value in `DataSource`

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,262 @@
...
+func (s *Service) FetchMetrics(ctx context.Context, report *telemetry.Config) ([]*pmmv1.ServerMetric_Metric, error) {
+	var metrics []*pmmv1.ServerMetric_Metric
+
+	metrics = s.processDashboardStat(metrics)
+	metrics = s.processComponentsStat(metrics)
+
+	return metrics, nil
+}
+
+func (s *Service) processDashboardStat(metrics []*pmmv1.ServerMetric_Metric) []*pmmv1.ServerMetric_Metric {
```
BupycHuk, line 101: Seems like `metrics` argument isn't required here and in `processComponentsStat`

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,262 @@
...
+	pmmv1 "github.com/percona-platform/saas/gen/telemetry/events/pmm"
+	"github.com/sirupsen/logrus"
+
+	uievents "github.com/percona/pmm/api/uieventspb"
+	"github.com/percona/pmm/managed/services/telemetry"
+)
+
+const (
+	cleanupAfterHours    = 30
+	cleanupCheckInterval = 1 * time.Minute
```
BupycHuk, line 35: if we need to clean up every 30 hours, no reason to check it every 1 minute. we can use a ticker with 30 hours interval

BupycHuk: in that case, we don't need extra checks in `cleanupIfNeeded`

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,262 @@
...
+	return metrics, nil
+}
+
+func (s *Service) processDashboardStat(metrics []*pmmv1.ServerMetric_Metric) []*pmmv1.ServerMetric_Metric {
+	type Stat struct {
+		TopDashboards         []string `json:"top_dashboards"`
+		SlowDashboardsP95_1s  []string `json:"slow_dashboards_p95_1s"`
+		SlowDashboardsP95_5s  []string `json:"slow_dashboards_p95_5s"`
+		SlowDashboardsP95_10s []string `json:"slow_dashboards_p95_10s"`
+	}
```
BupycHuk, line 107: can we extract this struct outside of the method?

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,262 @@
...
+	cleanupAfterHours    = 30
+	cleanupCheckInterval = 1 * time.Minute
+)
+
+// Service provides facility for storing UI events.
+type Service struct {
+	l               *logrus.Entry
+	lastCleanup     time.Time
+	dashboardUsage  map[string]*DashboardUsageStat
+	componentsUsage map[string]*ComponentsUsageStat
```
BupycHuk, line 43: these two variables are a possible cause of race conditions.

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,262 @@
...
+		SlowDashboardsP95_1s  []string `json:"slow_dashboards_p95_1s"`
+		SlowDashboardsP95_5s  []string `json:"slow_dashboards_p95_5s"`
+		SlowDashboardsP95_10s []string `json:"slow_dashboards_p95_10s"`
+	}
+
+	if len(s.dashboardUsage) > 0 {
+		dashboardStat := &Stat{
+			SlowDashboardsP95_1s:  []string{},
+			SlowDashboardsP95_5s:  []string{},
+			SlowDashboardsP95_10s: []string{},
```
BupycHuk, line 113: just a question, why do we set value for only 3 of 4 fields?

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+
+		marshal, err := json.Marshal(dashboardStat)
+		if err != nil {
+			s.l.Error("failed to marshal to JSON")
+		}
+		metrics = append(metrics, &pmmv1.ServerMetric_Metric{
+			Key:   "ui-events-dashboards",
+			Value: string(marshal),
+		})
```
BupycHuk, line 156: should we append a metric even if the `err` above isn't `nil`

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	return metrics
+}
+
+func (s *Service) processComponentsStat(metrics []*pmmv1.ServerMetric_Metric) []*pmmv1.ServerMetric_Metric {
+	type Stat struct {
+		SlowComponentsP95_1s  []string `json:"slow_components_p95_1s"`
+		SlowComponentsP95_5s  []string `json:"slow_components_p95_5s"`
+		SlowComponentsP95_10s []string `json:"slow_components_p95_10s"`
+	}
```
BupycHuk, line 166: let's extract this struct as well

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+	}
+
+	return &uievents.StoreResponse{}, nil
+}
+
+func (s *Service) loadTimeHistogram() *hdrhistogram.Histogram {
+	lowersValue := int64(0)
+	highestValue := int64(1_000 * 60 * 10) // 10 min
+	numberOfSignificantValueDigits := 5
```
BupycHuk, line 252: seems like these values can be constants

BupycHuk: in that case using `loadTimeHistogram` can be replaced by calling `hdrhistogram.New(lowersValue, highestValue, numberOfSignificantValueDigits)`

BupycHuk (changes requested):

managed/services/telemetry/extension.go
ritbl: They are similar, but the concept if different. DataSources are predefined in the architecture document.
Idea with extensions is more broad, in the future, if there will be a use case we can have extensions that modify configuration. Or other aspects of telemetry processing.
In future, it can be used as pre/post processor together with DataSource. That is why I didn't want to mix these concepts in the first place.

managed/services/uievents/uievents.go
ritbl: From a performance standpoint, it is nothing. The idea here was to be able to modify cleanupAfterHours if needed or move it to config.

managed/services/uievents/uievents.go
ritbl: There is an extremely low chance of one of dashboardUsage or componentsUsage to be old and other to be new. I think it is similar to guessing UUID, but I will put it under MX

managed/services/uievents/uievents.go
ritbl: it is a matter of style, it is similar to appender

managed/services/uievents/uievents.go
ritbl: The point here is visibility, this type makes sense only in this function.
You think it has performance implications?
There is none:

```

import (
	"testing"
)

type StatOutside struct {
	TopDashboards         []string `json:"top_dashboards"`
	SlowDashboardsP95_1s  []string `json:"slow_dashboards_p95_1s"`
	SlowDashboardsP95_5s  []string `json:"slow_dashboards_p95_5s"`
	SlowDashboardsP95_10s []string `json:"slow_dashboards_p95_10s"`
}

func Benchmark(b *testing.B) {
	b.Run("outside", func(b *testing.B) {
		items := make([]StatOutside, 0, 100)
		for i := 0; i < 100; i++ {
			items = append(items, StatOutside{})

		}
	})
	b.Run("inside", func(b *testing.B) {
		type StatInside struct {
			TopDashboards         []string `json:"top_dashboards"`
			SlowDashboardsP95_1s  []string `json:"slow_dashboards_p95_1s"`
			SlowDashboardsP95_5s  []string `json:"slow_dashboards_p95_5s"`
			SlowDashboardsP95_10s []string `json:"slow_dashboards_p95_10s"`
		}

		items := make([]StatInside, 0, 100)
		for i := 0; i < 100; i++ {
			items = append(items, StatInside{})

		}
	})
}
```

running with `go test -bench=. -benchtime=10000000x -count=10`

```
goos: darwin
goarch: arm64
pkg: github.com/percona/pmm/agent/agents/mysql/slowlog/parser
Benchmark/outside-10            10000000                 0.0005083 ns/op
Benchmark/outside-10            10000000                 0.0004250 ns/op
Benchmark/outside-10            10000000                 0.0005834 ns/op
Benchmark/outside-10            10000000                 0.0004292 ns/op
Benchmark/outside-10            10000000                 0.0003334 ns/op
Benchmark/outside-10            10000000                 0.0003917 ns/op
Benchmark/outside-10            10000000                 0.0006208 ns/op
Benchmark/outside-10            10000000                 0.0005125 ns/op
Benchmark/outside-10            10000000                 0.0002625 ns/op
Benchmark/outside-10            10000000                 0.0003083 ns/op
Benchmark/inside-10             10000000                 0.0004750 ns/op
Benchmark/inside-10             10000000                 0.0002750 ns/op
Benchmark/inside-10             10000000                 0.0005667 ns/op
Benchmark/inside-10             10000000                 0.0003541 ns/op
Benchmark/inside-10             10000000                 0.0002542 ns/op
Benchmark/inside-10             10000000                 0.0004916 ns/op
Benchmark/inside-10             10000000                 0.0002750 ns/op
Benchmark/inside-10             10000000                 0.0002625 ns/op
Benchmark/inside-10             10000000                 0.0002542 ns/op
Benchmark/inside-10             10000000                 0.0003375 ns/op
```

managed/services/uievents/uievents.go
ritbl: sorry, didn't get the question

managed/services/uievents/uievents.go
ritbl: good question, I think we need to skip it, there will be nil most likely

managed/cmd/pmm-managed/main.go
```diff
@@ -311,6 +314,12 @@ func runGRPCServer(ctx context.Context, deps *gRPCServerDeps, features gRPCServe
 		l.Fatalf("Failed to register platform service: %s", err.Error())
 	}
 
+	if err == nil {
+		uieventspb.RegisterUIEventsServer(gRPCServer, deps.uieventsService)
+	} else {
+		l.Fatalf("Failed to register UI events service: %s", err.Error())
+	}
```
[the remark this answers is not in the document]
BupycHuk: if we call l.Fatalf it stops the process `https://pkg.go.dev/github.com/sirupsen/logrus#Fatalf`, so we can put it after `if err != nil {}` block

managed/services/uievents/uievents.go
ritbl: it just a factory method, and good practice, we should try to minimize scope of variables/constants. that is why extracting makes a little sense.

managed/services/uievents/uievents.go
BupycHuk: this struct has 4 fields, but here we fill only 3 of them

managed/services/uievents/uievents.go
BupycHuk: there might be multiple calls from the frontend and multiple goroutines might try to store data for them which will raise a panic.

managed/cmd/pmm-managed/main.go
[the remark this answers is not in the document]
ritbl: sure, good catch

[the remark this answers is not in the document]
ritbl: by the way, it works the same way as platformService
```
	if err == nil {
		platformpb.RegisterPlatformServer(gRPCServer, platformService)
	} else {
		l.Fatalf("Failed to register platform service: %s", err.Error())
	}
```

actually constructor for `platform Service` never returns error, I will fix the code there

BupycHuk:
@ritbl please fix CI checks

managed/cmd/pmm-managed/main.go
```diff
@@ -805,7 +810,17 @@ func main() {
...
-	telemetry, err := telemetry.NewService(db, platformClient, version.Version, cfg.Config.Services.Telemetry)
+	uieventsService, err := uievents.New()
+	if err != nil {
+		l.Fatalf("Could not create UI Events Service: %s", err)
+	}
+	uieventsService.ScheduleCleanup(ctx)
+
+	telemetryExtensions := map[telemetry.ExtensionType]telemetry.Extension{
+		telemetry.UIEventsExtension: uieventsService,
+	}
```
artemgavrilov, line 821: Why it's made this way? Why do we need it pluggable?

go.mod
```diff
@@ -162,6 +162,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
+require github.com/HdrHistogram/hdrhistogram-go v1.1.2
+
```
artemgavrilov, line 166: I wonder why it's in the separate section?

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,292 @@
+// Copyright (C) 2017 Percona LLC
```
artemgavrilov, line 1:
```suggestion
// Copyright (C) 2023 Percona LLC
```

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,292 @@
...
+)
+
+// Service provides facility for storing UI events.
+type Service struct {
+	l               *logrus.Entry
+	dashboardUsage  map[string]*DashboardUsageStat
+	componentsUsage map[string]*ComponentsUsageStat
+	userFlowEvents  []*uievents.UserFlowEvent
+
+	stateLock sync.Mutex
```
artemgavrilov, line 48: Let's use mutex hat approach: https://dmitri.shuralyov.com/idiomatic-go/entries/2

And we usually using naming like`stateM` or just `m` if service has only one mutex.

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,292 @@
...
+
+type ComponentsUsageStat struct {
+	uid      string
+	useCount int32
+	loadTime *hdrhistogram.Histogram
+}
+
+// New returns platform Service.
+func New() (*Service, error) {
+	l := logrus.WithField("component", "platform")
```
artemgavrilov, line 68: I guess it should be:
```suggestion
	l := logrus.WithField("component", "uievents")
```

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+	}()
+}
+
+func (s *Service) FetchMetrics(ctx context.Context, config telemetry.Config) ([]*pmmv1.ServerMetric_Metric, error) {
+	s.stateLock.Lock()
+	dashboardUsage := s.dashboardUsage
+	componentsUsage := s.componentsUsage
+	userFlowEvents := s.userFlowEvents
+	s.stateLock.Unlock()
```
artemgavrilov, line 101: That won't help from data race. You need to make deep copy of these maps, of warp `processDashboardMetrics` functions with mutex.

managed/services/uievents/uievents.go
```diff
@@ -0,0 +1,292 @@
...
+	for _, event := range request.DashboardUsage {
+		stat, ok := s.dashboardUsage[event.Uid]
+		if !ok {
+			stat = &DashboardUsageStat{
+				title:    event.Title,
+				uid:      event.Uid,
+				useCount: 0,
+				loadTime: hdrhistogram.New(lowersValue, highestValue, numberOfSignificantValueDigits),
+			}
+			s.dashboardUsage[event.Uid] = stat
```
artemgavrilov, line 255: Not protected by mutex

artemgavrilov (changes requested):

managed/services/uievents/uievents.go
artemgavrilov: Also it seems it makes sense to use RWMutex

managed/services/uievents/uievents.go
ritbl: there is not race in store, but I'll add

managed/services/uievents/uievents.go
ritbl: why? we are not modifying items, we are just adding other items into collections

ritbl: okay, let's do that

managed/cmd/pmm-managed/main.go
ritbl: yes, it should be configurable

artemgavrilov: But why?

managed/services/uievents/uievents.go
artemgavrilov: This test was failing with datarace before latest changes:
```
func Test(t *testing.T) {
	t.Parallel()

	s, err := New()
	require.NoError(t, err)

	t.Run("A", func(t *testing.T) {
		t.Parallel()
		_, err := s.Store(context.Background(), &uievents.StoreRequest{
			Notifications: []*uievents.NotificationEvent{
				{
					Title:          "test",
					Text:           "test",
					Location:       "test",
					LocationParams: "test",
				},
			},
			DashboardUsage: []*uievents.DashboardUsageEvent{
				{
					Uid:            "test",
					Title:          "test",
					Tags:           nil,
					LoadTime:       0,
					Location:       "test",
					LocationParams: "test",
				},
			},
		})
		require.NoError(t, err)
	})

	t.Run("B", func(t *testing.T) {
		t.Parallel()
		_, err := s.FetchMetrics(context.Background(), telemetry.Config{})
		require.NoError(t, err)
	})
}
```

managed/services/telemetry/telemetry.go
```diff
@@ -225,6 +228,22 @@ func (s *Service) prepareReport(ctx context.Context) *pmmv1.ServerMetric {
...
+			if ok {
+				metrics, err := extension.FetchMetrics(ctx, telemetry)
+				if err != nil {
+					s.l.Debugf("failed while calling extension [%s]:%s", telemetry.Extension, err)
+					continue
+				}
+				telemetryMetric.Metrics = append(telemetryMetric.Metrics, metrics...)
+			} else {
+				s.l.Errorf("telemetry extension [%s] is not supported", telemetry.Extension)
+			}
```
JiriCtvrtka, line 242:
```suggestion
			if !ok {
				s.l.Errorf("telemetry extension [%s] is not supported", telemetry.Extension)
				continue
			}

			metrics, err := extension.FetchMetrics(ctx, telemetry)
			if err != nil {
				s.l.Debugf("failed while calling extension [%s]:%s", telemetry.Extension, err)
			}
			telemetryMetric.Metrics = append(telemetryMetric.Metrics, metrics...)

```

JiriCtvrtka: Its little bit less nested, but its up to you.

artemgavrilov (approved):

JiriCtvrtka:
Just minor comment, please take a look why main test failing.

managed/cmd/pmm-managed/main.go
ritbl: It should be configurable so that it is explicit what is being collected. Plus we need to be able to remove collection is needed.

go.mod
ritbl: good question, moving

managed/services/telemetry/telemetry.go
ritbl: but 1 more continue, and inderection. IMO it is personal style preference.

ritbl: discussed offline, code is updated

BupycHuk (approved):

PR #1344, opened 2022-10-25
PMM-10944 Restart exporters after restore

artemgavrilov:
PMM-10944

Build: SUBMODULES-2851

### DIFF
--- a/agent/agents/supervisor/supervisor.go
+++ b/agent/agents/supervisor/supervisor.go
@@ -224,7 +224,6 @@ func (s *Supervisor) RestartAgents() {
 
 		if err := s.startProcess(id, agent.requestedState, agent.listenPort); err != nil {
 			s.l.Errorf("Failed to restart Agent: %s.", err)
-			// TODO report that error to server
 		}
 	}
 
@@ -234,7 +233,6 @@ func (s *Supervisor) RestartAgents() {
 
 		if err := s.startBuiltin(id, agent.requestedState); err != nil {
 			s.l.Errorf("Failed to restart Agent: %s.", err)
-			// TODO report that error to server
 		}
 	}
 }
--- a/agent/agents/supervisor/supervisor.go
+++ b/agent/agents/supervisor/supervisor.go
@@ -213,6 +213,32 @@ func (s *Supervisor) SetState(state *agentpb.SetStateRequest) {
 	s.setBuiltinAgents(state.BuiltinAgents)
 }
 
+// RestartAgents restarts all existing agents.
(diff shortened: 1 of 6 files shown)


agent/runner/jobs/deps.go
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package jobs
+
+type restarter interface {
```
PavelKhripkov, line 17:
```suggestion
type agentsRestarter interface {
```

agent/runner/jobs/mongodb_restore_job.go
```diff
@@ -106,6 +116,7 @@ func (j *MongoDBRestoreJob) Run(ctx context.Context, send Send) error {
 		return errors.WithStack(err)
 	}
 
+	defer j.restarter.RestartAgents()
```
PavelKhripkov, line 119: Don't we need to restart agents only if restoring has been done successfully?

agent/agents/supervisor/supervisor.go
```diff
@@ -213,6 +213,32 @@ func (s *Supervisor) SetState(state *agentpb.SetStateRequest) {
...
+	for id, agent := range s.builtinAgents {
+		agent.cancel()
+		<-agent.done
+
+		if err := s.startBuiltin(id, agent.requestedState); err != nil {
+			s.l.Errorf("Failed to restart Agent: %s.", err)
+			// TODO report that error to server
+		}
+	}
+}
```
PavelKhripkov, line 240: Is it possible to use existing functions `setAgentProcesses` and `setBuiltinAgents`? If those functions are too big it's better to split them into smaller ones. Otherwise we're duplicating the code and the TODOs.

artemgavrilov: These functions restart agents only when state was changed, while we need to restart agents without any change.

agent/runner/jobs/mongodb_restore_job.go
artemgavrilov: I guess unsuccessful restore may cause problems as well. Better to restart.

agent/runner/jobs/deps.go
artemgavrilov: renamed

PR #1350, opened 2022-10-26
PMM-10586: basic alerts tracking

ritbl:
[PMM-10586](https://jira.percona.com/browse/PMM-10586)

FB: https://github.com/Percona-Lab/pmm-submodules/pull/2854


rnovikovP (approved):

managed/services/telemetry/config.default.yml
```diff
@@ -8,6 +8,22 @@ telemetry:
       - metric_name: "pmm_server_ia_enabled"
         column: "ia_enabled"
 
+  - id: PMMServerAlertsFire
+    source: VM
+    query: max_over_time(grafana_alerting_alerts{state="alerting"}[24h])
+    summary: "Max Number of Alerts Firing over last 24h"
```
michal-kralik, line 14:
```suggestion
    summary: "Maximum number of active concurrent alerts in the last 24 hours"
```

managed/services/telemetry/config.default.yml
```diff
@@ -8,6 +8,22 @@ telemetry:
...
+    data:
+      - metric_name: "pmm_server_alerts_firing_max_24h"
```
michal-kralik, line 16:
```suggestion
      - metric_name: "pmm_server_max_active_concurrent_alerts_24h"
```

michal-kralik (approved):
Just minor suggestions to the wording.

managed/services/telemetry/config.default.yml
ritbl: I think metric name it better to leave as it is, it is not intended to be "correct English", but rather the same prefix for similar metrics.
e.g. pmm_server_alerts_firing_max_24h, pmm_server_alerts_firing_max_12h

PR #1341, opened 2022-10-24
PMM-10928: install vmgateway via rpm

michal-kralik:
PMM-10928

Build: SUBMODULES-2845

PR #1353, opened 2022-10-27
PMM-10966: add vmgateway

michal-kralik:
PMM-10966

Related PRs:
- https://github.com/percona/pmm-update/pull/252
- https://github.com/percona/pmm/pull/1341

Build: SUBMODULES-0


managed/services/grafana/auth_server.go
```diff
@@ -108,6 +108,9 @@ const authenticationErrorCode = 401
 // cacheInvalidationPeriod is and period when cache for grafana response should be invalidated.
 const cacheInvalidationPeriod = 3 * time.Second
 
+// Base64 encoded jwt token providing full access via vmgateway with long expiration.
+const jwtFullAccess = "eyJ0eXAiOiJKV1QiLCJhbGciOiJQUzI1NiIsImtpZCI6IjI1OTMzN2RiLTc0MTItNDVkYS1hZDg2LWI2M2M5Nzc5NjU4OCJ9.eyJpYXQiOjE2NjE3NjM3MzUsIm5iZiI6MTY2MTc2MzczNSwiZXhwIjo0ODE1MzYzNzQwLCJqdGkiOiJ2eTU5VzRFMXJQVjk0OGVyazNCOTYiLCJ2bV9hY2Nlc3MiOnsidGVuYW5kX2lkIjp7fX19.yckDKjbFrnnBNlxN5Cjlk8fSbgKc2KzToJVbfQw_rOqSgdl_lNe7TDEtQ7NG2BITJW9rxbN4vdAYY-7gQ2loEz5Ev9YIl1QnzM553Eyw6HcES0JGjU2b9Pn7LXDlAigORoeX_3BryG7LXMUv7Cz2zfjrfaYuStBX1Jcv13BayGDeVpUAJ4s1IPlDwMpHvxqvGH4OjsxugdVIyfumhEvNpAx6kdy8SUABSsmh5Hb2yuIhyW0LKiSMgoCxegjShUFwiark7WkUb7APrO4yYRUV6DYnaYvP8HmV1ItD-wvl_2NDHe9xnUJXx0zMrszl36V5H15_5IDJdAccCAO7CaZr7g" //nolint:lll
```
gen1us2k, line 112: Can we make it configurable and pass it via args/env vars? It's usually not a bad practice to hardcode tokens

managed/services/grafana/auth_server.go
```diff
@@ -214,6 +217,9 @@ func (s *AuthServer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 			l.Warnf("%s", err)
 		}
 	}
+
+	// Adds a header indicating full access to all metrics when used with vmgateway.
+	rw.Header().Set("X-Percona-Token", "Bearer "+jwtFullAccess)
```
gen1us2k, line 222: I assume that we need to have a middleware/proxy to remove this header for the end user and add it only for a vmgateway. In that case, adding vmgateway makes no sense in terms of security.

michal-kralik:
Andrew's feedback discussed in a call.

managed/services/grafana/auth_server.go
artemgavrilov, line 112: Shouldn't it be configurable? Why we hardcoding this in code?

michal-kralik: The goal is to test the `vmgateway` component. For this it's configured basically as a transparent proxy.
The jwt you're referring to is required for `vmgateway` to work and it basically says, do not limit access to any metrics and expire in 100 years. It includes no secrets.

For this reason I think it's fine if it's hardcoded. It's going to be removed in the next version when we start limiting access to metrics.

artemgavrilov: Aha, got it. Thanks!

artemgavrilov (approved):

managed/services/supervisord/supervisord.go
```diff
@@ -675,6 +675,34 @@ stdout_logfile_backups = 3
...
+priority = 7
+command =
+    /usr/sbin/vmgateway
+      -eula
+      -enable.auth
+      -envflag.enable
+      -envflag.prefix=VMG_
+      -http.disableResponseCompression
+      -write.url=http://127.0.0.1:9090/prometheus
+      -read.url=http://127.0.0.1:9090/prometheus
```
artemgavrilov, line 689: Also it seems that we should check nginx configuration to be sure that VM is not accessible directly anymore

michal-kralik: Here's the related PR for this change https://github.com/percona/pmm-update/pull/252
As researched by Maxim, it shall be sufficient to pass `/prometheus/api/v1` through `vmgateway`.

PR #1358, opened 2022-10-28
PMM-10976: API for managing service labels

michal-kralik:
PMM-10976

Build: SUBMODULES-2874


managed/services/management/service.go
```diff
@@ -144,6 +146,96 @@ func (s *ServiceService) RemoveService(ctx context.Context, req *managementpb.Re
 	return &managementpb.RemoveServiceResponse{}, nil
 }
 
+// AddCustomLabels adds or replaces (if key exists) custom labels for a service.
+func (s *ServiceService) AddCustomLabels(ctx context.Context, req *managementpb.AddCustomLabelsRequest) (*managementpb.AddCustomLabelsResponse, error) {
+	if req.ServiceId == "" {
+		return nil, status.Error(codes.InvalidArgument, "service_id is required")
+	}
```
oksana-grishchenko, line 153: This check can be moved to the protobuf level using the `string_not_empty` field validator

managed/services/management/service.go
```diff
@@ -144,6 +146,96 @@ func (s *ServiceService) RemoveService(ctx context.Context, req *managementpb.Re
...
+		if err != nil {
+			return err
+		}
+
+		labels, err := service.GetCustomLabels()
+		if err != nil {
+			return err
+		}
+		if labels == nil {
+			labels = make(map[string]string)
```
oksana-grishchenko, line 166: Is it worth it to predefine capacity? how many labels are usually being added?

managed/services/management/service_test.go
```diff
@@ -256,4 +256,134 @@ func TestServiceService(t *testing.T) {
...
+			response, err := s.AddCustomLabels(ctx, &managementpb.AddCustomLabelsRequest{})
+			assert.Nil(t, response)
+			tests.AssertGRPCError(t, status.New(codes.InvalidArgument, "service_id is required"), err)
+		})
+
+		t.Run("Add a label", func(t *testing.T) {
+			ctx, s, teardown, vmdb := setup(t)
+			defer teardown(t)
+
+			vmdb.Mock.On("RequestConfigurationUpdate").Once().Return()
```
oksana-grishchenko, line 274: The `Once()` call makes sense together with
```
mock.AssertExpectationsForObjects(t, vmdb.Mock)
```
which checks that all mocked methods are actually being called and called the exact amount of times. Otherwise tests with mocks may end up with a lot of dead code and build an incorrect picture of how the function really works.

oksana-grishchenko (approved):

managed/services/management/service_test.go
michal-kralik: I may be missing something, but this shall be checked during `teardown` in https://github.com/percona/pmm/blob/65291e6c78f03d8b9bbe436a64db3d436b6acc68/managed/services/management/service_test.go#L60

Is the approach you're suggesting different in some ways to the `teardown` approach?

managed/services/management/service.go
michal-kralik: Great catch. We could even do better and assign the labels directly, without iterating through the map.

managed/services/management/service.go
michal-kralik: I didn't see the validators being used very often.
Is the expected design to use validators instead of custom validation in the service?

managed/services/management/service_test.go
oksana-grishchenko: That's what I've missed. Then it's totally fine, I don't think there is any sufficient difference.

managed/services/management/service.go
oksana-grishchenko: Not sure about the company-wide conventions, but it's definitely being used in the portal APIs. 
I think it makes the codebase cleaner, however it's up to you to decide what option to use.

michal-kralik: Sounds great. Thank you

michal-kralik:
@oksana-grishchenko I just moved the functionality from `management` to `inventory` as recommended by Nurlan.
Other than this there's no other change.

managed/services/inventory/services.go
```diff
@@ -346,3 +346,85 @@ func (ss *ServicesService) Remove(ctx context.Context, id string, force bool) er
 
 	return nil
 }
+
+// AddCustomLabels adds or replaces (if key exists) custom labels for a service.
+func (ss *ServicesService) AddCustomLabels(ctx context.Context, req *inventorypb.AddCustomLabelsRequest) (*inventorypb.AddCustomLabelsResponse, error) {
+	err := ss.db.InTransactionContext(ctx, nil, func(tx *reform.TX) error {
```
artemgavrilov, line 352: To eliminate shadowing:
```suggestion
	errTx := ss.db.InTransactionContext(ctx, nil, func(tx *reform.TX) error {
```

managed/services/inventory/services.go
```diff
@@ -346,3 +346,85 @@ func (ss *ServicesService) Remove(ctx context.Context, id string, force bool) er
...
+	}
+
+	ss.vmdb.RequestConfigurationUpdate()
+
+	return &inventorypb.AddCustomLabelsResponse{}, nil
+}
+
+// RemoveCustomLabels removes custom labels from a service.
+func (ss *ServicesService) RemoveCustomLabels(ctx context.Context, req *inventorypb.RemoveCustomLabelsRequest) (*inventorypb.RemoveCustomLabelsResponse, error) {
+	err := ss.db.InTransactionContext(ctx, nil, func(tx *reform.TX) error {
```
artemgavrilov, line 393: Same

artemgavrilov (approved):

PR #1370, opened 2022-11-01
PMM-10948 go-grpc-prometheus use fork with new label "source"

YaroslavPodorvanov:
PMM-10948

Build: SUBMODULES-2865

- [] https://github.com/Percona-Lab/go-grpc-prometheus/pull/2

https://github.com/percona/pmm/pull/1398 better


go.mod
```diff
@@ -10,6 +10,8 @@ replace github.com/pganalyze/pg_query_go v1.0.3 => github.com/Percona-Lab/pg_que
 
 replace golang.org/x/crypto v0.0.0-20220722155217-630584e8d5aa => github.com/percona-lab/crypto v0.0.0-20220811043533-d164de3c7f08
 
+replace github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0 => github.com/percona-lab/go-grpc-prometheus v0.0.0-20221101194635-957e112f35fe
```
artemgavrilov, line 13: Why don't we consider creating another interceptor with metric that we need? Do we need all this information provided by `go-grpc-prometheus` **together** with the source info? I mean forking of some project should be the last option, is there any chance to eliminate that?

Also why we are using `replace` directive, can't we just change all imports in our code to our fork?

YaroslavPodorvanov: > Why don't we consider creating another interceptor with metric that we need?
Yes, we can do it and it will look like ~
```text
grpc_server_handled_external_total{grpc_code="OK",grpc_method="PingList",grpc_service="mwitkow.testproto.TestService",grpc_type="server_stream"}
```
But for me look better with new label and we discuss this solution with @rnovikovP and @ritbl 

> Do we need all this information provided by `go-grpc-prometheus` **together** with the source info?
In first step no, because we need to know "is our API integrated to another systems" (from issue description _What endpoints used directly by end users (e.g. CI integration / automation)_)
But in future we want to know why external API calls are failed

> I mean forking of some project should be the last option, is there any chance to eliminate that?
With new label no

> Also why we are using `replace` directive, can't we just change all imports in our code to our fork?
If we want direct use then we need rename inside repository https://github.com/Percona-Lab/go-grpc-prometheus all imports and after that `sync code`  with main branch will be much harder

Probably repository https://github.com/grpc-ecosystem/go-grpc-prometheus will not supported according info from repository README.md , but we use `replace` for case when `sync code` needed

YaroslavPodorvanov:
Draft after discussions with @ritbl and @artemgavrilov to try use https://github.com/grpc-ecosystem/go-grpc-middleware/tree/v2

YaroslavPodorvanov:
https://github.com/percona/pmm/pull/1398 better

PR #1374, opened 2022-11-04
PMM-10807 Reorder function calls in PMM Disconnect and minor refactor

palash25:
PMM-10807

Build: [SUBMODULES-2931](https://github.com/Percona-Lab/pmm-submodules/pull/2931)

This PR: reorders the function calls in Disconnect endpoint such that the request to disconnect from Portal is made before deleting the records from PMM DB. This prevents situations where a disconnect request that would fail midway (in case of admin-admin credentials user)  and lead to a dirty state in the DB.


managed/services/platform/platform.go
```diff
@@ -130,44 +137,28 @@ func (s *Service) Disconnect(ctx context.Context, req *platformpb.DisconnectRequ
...
 		if errors.Is(err, grafana.ErrFailedToGetToken) {
-			return nil, status.Error(codes.FailedPrecondition, "Failed to get access token. Please sign in using your Percona Account.")
+			return nil, errGrafanaAccessTokenFailed
 		}
 		s.l.Errorf("Disconnect to Platform request failed: %s", err)
 		return nil, err
 	}
 
 	err = s.client.Disconnect(ctx, userAccessToken, settings.PMMServerID)
-	needRecover := err != nil && !req.Force
```
palash25, line 152: @oksana-grishchenko can we deprecate this flag now? or do we still need it?

oksana-grishchenko: Looks like we don't need it anymore, this flag was introduced to prevent inserting the sso data back to the DB after a failed disconnect. 
Do you know if there was any specific reason why the functions calls were in that order initially? It looks like there is some story behind that order.

oksana-grishchenko (approved):
With that change the `force` flag is not needed anymore in the API, and there will be also some useless code on the FE side that uses that flag. Not sure what we should do about it.

palash25:
moving this to draft till i confirm that it doesn't violate with the spec

managed/services/platform/platform.go
palash25: @oksana-grishchenko i checked the spec https://confluence.percona.com/display/PMM/Single+Sign-On+-+PMM+Server+integration+with+Portal+and+Okta  and it mentions the exact same flow for Disconnect that this PR implements it seems it was implemented the wrong way initially.

oksana-grishchenko (approved):
Great job - the problem is solved and the API users don't have to change anything on their side

BupycHuk (approved):

PR #1375, opened 2022-11-04
PMM-7 Single linter config, fix linters

oksana-grishchenko:
PMM-7 Single linter config

### Part 1

The idea behind this change is to migrate from a white-list linter config to the black-list linter config, which means enabling all linters by default and disable only certain of them. 

Since the repo contains tons of linter errors, here is the plan:
1. Enable all linters by default (done)
2. Disable all linters that fail currently (done) 
3. Process the disabled linters one by one - decide if we should keep one disabled or we should fix the related errors. It's ok to add //nolint directives when it's too invasive to fix, at least the linter will work for the new code.  (done for several linters, the rest is TBD in another PRs)

Build: SUBMODULES-0


agent/connectionchecker/connection_checker.go
```diff
@@ -130,7 +130,9 @@ func (cc *ConnectionChecker) checkMySQLConnection(ctx context.Context, dsn strin
 
 	if err = cc.sqlPing(ctx, db); err != nil {
 		if errors.As(err, &x509.HostnameError{}) {
-			res.Error = errors.Wrap(err, "mysql ssl certificate is misconfigured, make sure the certificate includes the requested hostname/IP in CN or subjectAltName fields").Error()
+			res.Error = errors.Wrap(err,
```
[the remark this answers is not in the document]
oksana-grishchenko: updated

oksana-grishchenko (commented):
The `Managed / Tests (false, 1.19.x) (pull_request) Failing after 8m` seems to be unrelated, please let me know if it's not

michal-kralik:
> The `Managed / Tests (false, 1.19.x) (pull_request) Failing after 8m` seems to be unrelated, please let me know if it's not

Let's please make sure the failed test is not because of these changes. Other PRs have the test passing. It could be because of this.

admin/commands/list_test.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//nolint:lll
```
artemgavrilov, line 15: I guess we can disable this linter for tests globally

admin/helpers/helpers.go
```diff
@@ -59,8 +59,9 @@ func IsHAProxySupported() (bool, error) {
 	if err != nil {
 		return false, err
 	}
+	errLessThanMinVersion := fmt.Errorf("haproxy is not supported in this version, please update your pmm-server to %s or higher", minVersion) //nolint:goerr113
```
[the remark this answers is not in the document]
artemgavrilov: Why it was changed at all?

agent/versioner/versioner.go
```diff
@@ -90,7 +90,7 @@ func (v *Versioner) binaryVersion(
 	defer cancel()
 
 	if _, err := v.ef.LookPath(binaryName); err != nil {
-		if err.(*exec.Error).Err == exec.ErrNotFound {
+		if errors.Is(err.(*exec.Error).Err, exec.ErrNotFound) {
```
artemgavrilov, line 93: Can it be just `errors.Is(err, exec.ErrNotFound)` ?

api-tests/init.go
```diff
@@ -76,7 +76,7 @@ var (
 )
 
 // ErrFromNginx is an error type for nginx HTML response.
-type ErrFromNginx string
+type ErrFromNginx string //nolint:errname
```
artemgavrilov, line 79: I think we can rename it to `NginxError` instead

managed/models/common_helpers.go
```diff
@@ -18,7 +18,7 @@ package models
 import "fmt"
 
 // ErrInvalidArgument returned when some passed argument is invalid.
-type ErrInvalidArgument struct {
+type ErrInvalidArgument struct { //nolint:errname
```
artemgavrilov, line 21: Let's rename

managed/services/backup/backup_service.go
```diff
@@ -158,13 +158,15 @@ func (s *Service) PerformBackup(ctx context.Context, params PerformBackupParams)
 			}
 		} else {
 			if artifact, err = models.UpdateArtifact(tx.Querier, artifact.ID, models.UpdateArtifactParams{
-				Status: models.BackupStatusPointer(models.PendingBackupStatus),
+				Status:     models.BackupStatusPointer(models.PendingBackupStatus),
+				ServiceID:  nil,
+				ScheduleID: nil,
```
artemgavrilov, line 163: Let's revert these two lines. This linter already disabled.

managed/services/grafana/client.go
```diff
@@ -42,8 +42,14 @@ import (
...
 
-// ErrFailedToGetToken means it failed to get user's token. Most likely due to the fact user is not logged in using Percona Account.
-var ErrFailedToGetToken = errors.New("failed to get token")
+var (
+	// ErrFailedToGetToken means it failed to get user's token. Most likely due to the fact user is not logged in using Percona Account.
+	ErrFailedToGetToken = errors.New("failed to get token")
+
+	errGettingHeaders        = fmt.Errorf("cannot get headers from metadata")
+	errGrafanaAPIUnavailable = fmt.Errorf("cannot reach Grafana API")
+	errGrafanaErrors         = fmt.Errorf("grafana is running with errors")
```
artemgavrilov, line 51: Can we revert this? There is no any value in these variables.

artemgavrilov (commented):
👍

agent/versioner/versioner.go
oksana-grishchenko: Since exec.Error is a custom Error with its own logic inside, I would keep the original check to avoid unexpected side effects

admin/commands/list_test.go
oksana-grishchenko: A good idea, thanks, updated.

admin/helpers/helpers.go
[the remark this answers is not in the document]
oksana-grishchenko: Reverted. It was a failed attempt to satisfy linter requirements, I agree this change is useless.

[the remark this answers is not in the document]
michal-kralik: Let's disable the `goerr113` linter in this case.
It adds extra work and we don't benefit from the linter, it does not make sense to keep it enabled.

artemgavrilov (approved):

percona-csalguero (approved):
Thanks for this.

PR #1388, opened 2022-11-09
PMM-10743 Extend metrics for Advisor

0leksii:
PMM-10743

Build: SUBMODULES-2881


dliakhov (approved):

managed/services/checks/checks.go
```diff
@@ -150,36 +151,29 @@ func New(db *reform.DB, platformClient *platform.Client, agentsRegistry agentsRe
 			Namespace: prometheusNamespace,
 			Subsystem: prometheusSubsystem,
 			Name:      "scripts_executed_total",
-			Help:      "Counter of check scripts executed per service type",
-		}, []string{"service_type"}),
+			Help:      "Counter of check scripts executed per check name",
```
ritbl, line 154: `per check name`, is it correct?

managed/services/checks/checks.go
```diff
@@ -150,36 +151,29 @@ func New(db *reform.DB, platformClient *platform.Client, agentsRegistry agentsRe
...
+		}, []string{"service_type", "check_type", "check_name"}),
 
 		mAlertsGenerated: prom.NewCounterVec(prom.CounterOpts{
 			Namespace: prometheusNamespace,
 			Subsystem: prometheusSubsystem,
 			Name:      "alerts_generated_total",
-			Help:      "Counter of alerts generated per service type per check type",
-		}, []string{"service_type", "check_type"}),
+			Help:      "Counter of alerts generated per check name",
```
ritbl, line 161: is it correct? I think it is still used as before, we added another label.

managed/services/checks/checks.go
```diff
@@ -150,36 +151,29 @@ func New(db *reform.DB, platformClient *platform.Client, agentsRegistry agentsRe
...
+		}, []string{"service_type", "check_type", "check_name"}),
+
+		mChecksDownloaded: prom.NewCounterVec(prom.CounterOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "checks_downloaded_total",
+			Help:      "Counter of checks downloaded per check name",
```
ritbl, line 168: please check if it is correct

managed/services/checks/checks.go
```diff
@@ -150,36 +151,29 @@ func New(db *reform.DB, platformClient *platform.Client, agentsRegistry agentsRe
...
+		}, []string{"service_type", "check_type", "check_name"}),
 	}
 
 	if d, _ := strconv.ParseBool(os.Getenv(envDisableStartDelay)); d {
 		l.Warn("Start delay disabled.")
 		s.startDelay = 0
 	}
 
-	s.mScriptsExecuted.WithLabelValues(string(models.MySQLServiceType))
```
ritbl, line 169: why it was removed?

managed/services/checks/checks.go
```diff
@@ -1519,12 +1515,46 @@ func (s *Service) UpdateIntervals(rare, standard, frequent time.Duration) {
...
+// initializeServiceMetrics initializes metrics for serviceType.
+func (s *Service) initializeServiceMetrics(serviceType models.ServiceType, checks map[string]check.Check) {
+	for _, c := range checks {
+		s.mAlertsGenerated.WithLabelValues(string(serviceType), string(c.Type), c.Name)
+		s.mScriptsExecuted.WithLabelValues(string(serviceType), string(c.Type), c.Name)
+		s.mChecksDownloaded.WithLabelValues(string(serviceType), string(c.Type), c.Name)
+	}
+}
+
+// incChecksDownload increases check download counters.
```
ritbl, line 1545: increases -> increments

managed/services/checks/checks.go
```diff
@@ -1519,12 +1515,46 @@ func (s *Service) UpdateIntervals(rare, standard, frequent time.Duration) {
...
+func (s *Service) incChecksDownload() {
+	mySQLChecks, postgreSQLChecks, mongoDBChecks := s.groupChecksByDB(s.checks)
+	s.incServiceCheckDownloadMetrics(models.MySQLServiceType, mySQLChecks)
+	s.incServiceCheckDownloadMetrics(models.PostgreSQLServiceType, postgreSQLChecks)
+	s.incServiceCheckDownloadMetrics(models.MongoDBServiceType, mongoDBChecks)
+}
+
+// incServiceCheckDownloadMetrics increases check download counters for serviceType.
```
ritbl, line 1553: don't think we need this comment, I think it is obvious and it is private

managed/services/checks/checks.go
```diff
@@ -1519,12 +1515,46 @@ func (s *Service) UpdateIntervals(rare, standard, frequent time.Duration) {
...
 }
 
 // Collect implements prom.Collector.
 func (s *Service) Collect(ch chan<- prom.Metric) {
 	s.mScriptsExecuted.Collect(ch)
 	s.mAlertsGenerated.Collect(ch)
+	s.mChecksDownloaded.Collect(ch)
+}
+
+// initializeMetrics initializes all metrics for checks.
```
ritbl, line 1528: don't think we need this comment, I think it is obvious and it is private

managed/services/checks/checks.go
```diff
@@ -1519,12 +1515,46 @@ func (s *Service) UpdateIntervals(rare, standard, frequent time.Duration) {
...
+func (s *Service) initializeMetrics() {
+	mySQLChecks, postgreSQLChecks, mongoDBChecks := s.groupChecksByDB(s.checks)
+	s.initializeServiceMetrics(models.MySQLServiceType, mySQLChecks)
+	s.initializeServiceMetrics(models.PostgreSQLServiceType, postgreSQLChecks)
+	s.initializeServiceMetrics(models.MongoDBServiceType, mongoDBChecks)
+}
+
+// initializeServiceMetrics initializes metrics for serviceType.
```
ritbl, line 1536: don't think we need this comment, I think it is obvious and it is private

managed/services/checks/checks.go
```diff
@@ -1378,6 +1373,7 @@ func (s *Service) CollectChecks(ctx context.Context) {
 			s.l.Errorf("Failed to download checks: %s.", err)
 			return // keep previously downloaded checks
 		}
+		defer s.incChecksDownload()
```
ritbl, line 1376: please add comment describing why it is executed with `defer`

ritbl (approved):

managed/services/checks/checks.go
```diff
@@ -189,6 +183,7 @@ func (s *Service) Run(ctx context.Context) {
 	defer s.l.Info("Done.")
 
 	s.CollectChecks(ctx)
+	s.initializeMetrics()
```
artemgavrilov, line 186: I wonder do we need to initialize them at all? Set of labels become quite dynamic, it depends what was downloaded from Protal. So maybe it makes sense just to init them in place?

artemgavrilov (approved):

managed/services/checks/checks.go
0leksii: On one hand yes - we can leave the init till the point when we actually need that label, on the other hand having no labels until checks are executed/failed is a bit weird too. So, I would keep it with the current logic: all labels are initialized at startup.

dliakhov (approved):

managed/services/checks/checks.go
```diff
@@ -1378,6 +1373,8 @@ func (s *Service) CollectChecks(ctx context.Context) {
 			s.l.Errorf("Failed to download checks: %s.", err)
 			return // keep previously downloaded checks
 		}
+		// defer it ro run after updateChecks
```
artemgavrilov, line 1376:
```suggestion
		// defer it to run after updateChecks
```

managed/services/checks/checks.go
artemgavrilov: I don't have strong opinion on that. Maybe @ritbl can suggest what is better for telemetry needs?

managed/services/telemetry/config.default.yml
```diff
@@ -531,3 +531,43 @@ telemetry:
...
+        label: "check_name"
+      - metric_name: "count"
+        value: 1
+
+  - id: AdvisorChecksFailedCount
+    source: VM
+    query: pmm_managed_checks_alerts_generated_total
+    transform:
+      type: JSON
+      metric: "advisor_checks_failed"
```
artemgavrilov, line 567: That sounds more like check failed itself, not that it detected some problem and fired an alert. I think it makes sense to name it more obvious.

artemgavrilov (approved):
LGTM with comments

managed/services/telemetry/config.default.yml
rnovikovP, line 567:
```suggestion
      metric: "advisor_checks_generated_fails"
```
?

rnovikovP: @artemgavrilov ^ WDYT?

0leksii: Updated.

artemgavrilov: Yep, that's better. BTW original metric name(`checks_alerts_generated`) was good enough as for me 😄

managed/services/checks/checks.go
ritbl: These two solutions are effectively equivalent. In both cases we have initialization with `0` value.
This one looks a bit cleaner, though.

ritbl: @artemgavrilov, @0leksii and I discussed offline. It looks like we might not initialize all the possible metrics, hence it is better to avoid initialization to not confuse users and not produce wrong charts.
Thank you!

managed/services/telemetry/config.default.yml
```diff
@@ -566,3 +566,43 @@ telemetry:
     data:
       - metric_name: "k8s_clusters_count"
         column: "count"
+
+  # Advisor
+  - id: AdvisorChecksDownloadedCount
+    source: VM
+    query: pmm_managed_checks_checks_downloaded_total
```
ritbl, line 573: pmm_managed_checks_checks_downloaded_total -> pmm_server_advisors_downloaded_total ?

and others the same way

managed/services/telemetry/config.default.yml
```diff
@@ -566,3 +566,43 @@ telemetry:
...
+        label: "check_name"
+      - metric_name: "count"
+        value: 1
+
+  - id: AdvisorChecksExecutedCount
+    source: VM
+    query: pmm_managed_checks_scripts_executed_total
+    transform:
+      type: JSON
+      metric: "advisor_checks_executed"
```
ritbl, line 589: pmm_server_advisors_checks_executed

managed/services/telemetry/config.default.yml
```diff
@@ -566,3 +566,43 @@ telemetry:
...
       - metric_name: "k8s_clusters_count"
         column: "count"
+
+  # Advisor
+  - id: AdvisorChecksDownloadedCount
+    source: VM
+    query: pmm_managed_checks_checks_downloaded_total
+    transform:
+      type: JSON
+      metric: "advisor_checks_downloaded"
```
ritbl, line 576: advisor_checks_downloaded -> pmm_server_advisor_checks_downloaded

managed/services/telemetry/config.default.yml
```diff
@@ -566,3 +566,43 @@ telemetry:
...
+    summary: "Advisor - number of downloaded checks per name"
+    data:
+      - metric_name: "check_name"
```
ritbl, line 579: check_name -> `1`

managed/services/telemetry/config.default.yml
```diff
@@ -566,3 +566,43 @@ telemetry:
...
+        label: "check_name"
+      - metric_name: "count"
```
ritbl, line 581: count -> `2`

managed/services/telemetry/config.default.yml
```diff
@@ -566,3 +566,43 @@ telemetry:
...
+    source: VM
+    query: pmm_managed_checks_scripts_executed_total
+    transform:
+      type: JSON
+      metric: "advisor_checks_executed"
+    summary: "Advisor - number of executed checks per name"
+    data:
+      - metric_name: "check_name"
+        label: "check_name"
+      - metric_name: "count"
```
ritbl, line 594: 1, 2 as names

managed/services/telemetry/config.default.yml
```diff
@@ -566,3 +566,43 @@ telemetry:
...
+    query: pmm_managed_checks_alerts_generated_total
+    transform:
+      type: JSON
+      metric: "advisor_checks_generated_fails"
+    summary: "Advisor - number of failed checks per name"
+    data:
+      - metric_name: "check_name"
+        label: "check_name"
+      - metric_name: "count"
+        value: 1
```
ritbl, line 608: 1,2 as names

managed/services/telemetry/config.default.yml
```diff
@@ -566,3 +566,43 @@ telemetry:
...
+
+  - id: AdvisorChecksFailedCount
+    source: VM
+    query: pmm_managed_checks_alerts_generated_total
+    transform:
+      type: JSON
+      metric: "advisor_checks_generated_fails"
```
ritbl, line 602: advisor_checks_generated_fails -> pmm_server_advisor_checks_generated_fails

managed/services/telemetry/config.default.yml
0leksii: Done

managed/services/telemetry/config.default.yml
0leksii: Done

managed/services/telemetry/config.default.yml
0leksii: Done

managed/services/telemetry/config.default.yml
0leksii: Done

managed/services/telemetry/config.default.yml
0leksii: Done

managed/services/telemetry/config.default.yml
0leksii: Done

PR #1394, opened 2022-11-11
PMM-10875 Added alerting tour flag

ShashankSinha252:
PMM-10875 : Add flag to track alerting tour completion for user.

Build: SUBMODULES-2888


ShashankSinha252:
Reviewers, please note that files to review are following:

- api/userpb/user.proto
- managed/models/database.go
- managed/models/user_flags_helpers.go
- managed/models/user_flags_model.go
- managed/models/user_flags_model_reform.go
- managed/services/user/user.go

Rest are all autogenerated files.

managed/models/user_flags_helpers.go
```diff
@@ -29,8 +29,9 @@ type CreateUserParams struct {
 
 // UpdateUserParams has parameters to update existing user.
 type UpdateUserParams struct {
-	UserID int
-	Tour   bool
+	UserID       int
+	Tour         bool
+	AlertingTour bool
```
oksana-grishchenko, line 34: One concern - will the business-logic still work correctly in all cases with the default `false` value? 
I mean, what if I need to change only the `Tour` and therefore I will create the params like this:
```
UpdateUserParams {UserID: "someID", Tour: true}
```
so I just skip the `AlertingTool` field and it becomes a default value (although the real `AlertingTool` value might be `true`), so the `AlertingTool` field got unintendedly updated to `false`

If this is a possible case it could be avoided by using *bool type that allows to distinguish empty from false.

ShashankSinha252: Well noticed @oksana-grishchenko. The current logic for update allows only to set flags as `true`. With default value of `false`, this means that API allows only to set flags, never unset them.

oksana-grishchenko (approved):

percona-csalguero (approved):

PR #1404, opened 2022-11-15
PMM-7: fix docs pipeline with empty grep

michal-kralik:
PMM-7

### DIFF
--- a/.github/workflows/release-doc.yml
+++ b/.github/workflows/release-doc.yml
@@ -17,7 +17,7 @@ jobs:
         env:
           release_tag: ${{ github.ref }}
       - name: Convert mkdocs
-        shell: bash --noprofile --norc -e {0}
+        shell: bash --noprofile --norc -ex {0}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release-doc.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         env:
           release_tag: ${{ github.ref }}
       - name: Convert mkdocs
-        shell: bash
+        shell: bash --noprofile --norc -e {0}
```
atymchuk, line 20:
```suggestion
        shell: bash --noprofile --norc -ex {0}
```
Maybe adding -x (aka `xtrace`) will help troubleshoot?

michal-kralik: The `-o pipefail` made the pipeline fail on `grep` if the pattern was not found

atymchuk (approved):

artemgavrilov:
@michal-kralik But what was the reason of failure?

michal-kralik:
> @michal-kralik But what was the reason of failure?

The flag `-o pipefail` makes the pipeline fail with exit code of the first command which failed. 
By default pipeline returns the exit code of the last command.

In our case `grep` exits with code `1` because it did not find any match.

See https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425#set--o-pipefail for reference

PR #1411, opened 2022-11-16
PMM-11069 scheduled backup identical names allowed fix

PavelKhripkov:
PMM-11069

Build: SUBMODULES-2905


managed/models/scheduled_tasks_helpers.go
```diff
@@ -266,3 +300,32 @@ func checkUniqueScheduledTaskID(q *reform.Querier, id string) error {
...
+	return "", nil
+}
+
+func nameIsFree(q *reform.Querier, name string) error {
+	tasks, err := FindScheduledTasks(q, ScheduledTasksFilter{Name: name})
+	if err != nil {
+		return err
+	}
+	if len(tasks) != 0 {
+		return ErrAlreadyExists
```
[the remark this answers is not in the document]
PavelKhripkov: I wanted to check if it's possible for our data model but then just forgot. Thanks for the reminder!

managed/models/scheduled_tasks_helpers.go
```diff
@@ -39,7 +42,7 @@ func FindScheduledTaskByID(q *reform.Querier, id string) (*ScheduledTask, error)
 	case nil:
 		return res, nil
 	case reform.ErrNoRows:
-		return nil, status.Errorf(codes.NotFound, "ScheduledTask with ID %q not found.", id)
+		return nil, errors.Wrapf(ErrNotFound, "couldn't get scheduled task with ID %q", id)
```
artemgavrilov, line 45: That was common approach for all such helpers. I guess we should refactor it everywhere then.

managed/models/scheduled_tasks_helpers.go
```diff
@@ -266,3 +296,31 @@ func checkUniqueScheduledTaskID(q *reform.Querier, id string) error {
...
+
+func isNameAlreadyExistsErr(err error, name string) bool {
+	if err, ok := err.(*pq.Error); ok {
+		// We check like this because there is no structured info in the returned error.
+		if err.Code == pgUniqueConstraintViolationCode && strings.Contains(err.Detail, name) {
+			return true
+		}
+	}
+	return false
+}
```
artemgavrilov, line 326: I guess because of that we have `checkUnique*` functions. While this may work too(i'm not sure about "contains" part), isn't it better to use same approach as everywhere in code base?

managed/models/scheduled_tasks_helpers.go
PavelKhripkov: Yeap, we should. API errors should be used only on API level otherwise we have a mess. But it should be done in separate PR I think.

artemgavrilov (approved):

idoqo (approved):

PR #1415, opened 2022-11-17
PMM-7 Fix go-consistent warnings

artemgavrilov:
PMM-7

### DIFF
--- a/qan-api2/db.go
+++ b/qan-api2/db.go
@@ -133,8 +133,7 @@ func DropOldPartition(db *sqlx.DB, days uint) {
 	err := db.Select(
 		&partitions,
 		query,
-		days,
-	)
+		days)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


qan-api2/db.go
```diff
@@ -133,8 +133,7 @@ func DropOldPartition(db *sqlx.DB, days uint) {
 	err := db.Select(
 		&partitions,
 		query,
-		days,
-	)
+		days)
```
michal-kralik, line 136: I suggest we get rid of this rule. I think it's easier to navigate the codebase with bracket on new line.

artemgavrilov: If you interested in this change you can create poll for BE devs and if majority will prefer bracket on newline then you can rewrite it in all codebase 😁

recharte: I agree :+1:

artemgavrilov: We have democracy here, go on)

recharte (approved):

PR #1422, opened 2022-11-18
PMM-7: fix slow tests on Ubuntu

michal-kralik:
PMM-7

Build: SUBMODULES-2903

### DIFF
--- a/.devcontainer/setup.py
+++ b/.devcontainer/setup.py
@@ -91,6 +91,9 @@ def setup():
     run_commands([
         # allow connecting from any host, needed to connect from host to PG running in docker
         "sed -i -e \"s/#listen_addresses = \'localhost\'/listen_addresses = \'*\'/\" /srv/postgres14/postgresql.conf",
+        # Turns fsync off. Create database operations with fsync on are very slow on Ubuntu.
+        # Having fsync off in dev environment is fine.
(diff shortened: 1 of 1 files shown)


recharte (approved):
:raised_hands:

.devcontainer/setup.py
```diff
@@ -91,6 +91,9 @@ def setup():
     run_commands([
         # allow connecting from any host, needed to connect from host to PG running in docker
         "sed -i -e \"s/#listen_addresses = \'localhost\'/listen_addresses = \'*\'/\" /srv/postgres14/postgresql.conf",
+        # Turns fsync off. Create database operations with fsync on are very slow on Ubuntu.
+        # Having fsync off in dev environment is fine.
+        "sed -i -e \"s/#fsync = on/fsync = off/\" /srv/postgres14/postgresql.conf",
```
artemgavrilov, line 96: I wonder why this problem is Ubuntu specific? 🤔

artemgavrilov (approved):

.devcontainer/setup.py
michal-kralik: Could be filesystem specific, mine is `ext4` and maybe mount configuration, `lvm` config. Who knows :)

PR #1424, opened 2022-11-18
PMM-10938 restart after physical restore

idoqo:
PMM-10938

Build: SUBMODULES-2907


managed/services/backup/backup_service.go
```diff
@@ -288,14 +288,17 @@ func (s *Service) RestoreBackup(ctx context.Context, serviceID, artifactID strin
 			jobType = models.MySQLRestoreBackupJob
 			jobData = &models.JobData{
 				MySQLRestoreBackup: &models.MySQLRestoreBackupJobData{
+					ServiceID: serviceID,
```
PavelKhripkov, line 291: Since you found this bug I think we should add some tests that now it works as expected.

managed/models/service_helpers_test.go
```diff
@@ -401,6 +401,45 @@ func TestServiceHelpers(t *testing.T) {
...
+		})
+		require.NoError(t, err)
+
+		services, err := models.FindServices(q, models.ServiceFilters{
+			ServiceType:    pointerToServiceType(models.MongoDBServiceType),
+			ReplicationSet: "rs0",
+		})
+		assert.NoError(t, err)
+		assert.NotNil(t, services)
+		assert.Len(t, services, 2)
```
PavelKhripkov, line 441: Can we also test the content of the returned values?

api/agentpb/agent.proto
```diff
@@ -326,6 +326,17 @@ message StartActionRequest {
...
   }
+
+  // RestartMongoDBServiceParams describes an action request to restart components in a MongoDB node
+  message RestartMongoDBServiceParams {
+    // Service is a service that can be restarted from PMM server
+    enum Service {
+      SERVICE_NAME_INVALID = 0;
+      MONGOD = 1;
+      PBM_AGENT = 2;
+    }
```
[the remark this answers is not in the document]
idoqo: Alright, maybe something like "unit"? since that is what systemctl calls them.

api/agentpb/agent.proto
```diff
@@ -326,6 +326,17 @@ message StartActionRequest {
...
+    Service service = 1;
```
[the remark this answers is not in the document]
idoqo: For the current use case, we can't restart pbm-agent and mongod at once because pbm-agent wants all mongod instances to be available otherwise, it will fail (so an array won't work)

agent/runner/runner.go
```diff
@@ -237,6 +237,7 @@ func (r *Runner) handleAction(ctx context.Context, action actions.Action) {
 		var errMsg string
 		if err != nil {
 			errMsg = err.Error()
+			l.Infof("Action produced output: %+v", string(output))
 			l.Warnf("Action terminated with error: %+v", err)
```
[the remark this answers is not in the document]
idoqo: Agree about the level and will update that, though yeah I think whichever order is fine.

agent/client/client.go
```diff
@@ -501,6 +501,17 @@ func (c *Client) handleStartActionRequest(p *agentpb.StartActionRequest) error {
...
 		action = actions.NewProcessAction(p.ActionId, timeout, c.cfg.Paths.PTMongoDBSummary, argListFromMongoDBParams(params.PtMongodbSummaryParams))
+	case *agentpb.StartActionRequest_RestartMongodbServiceParams:
+		var service string
+		switch params.RestartMongodbServiceParams.Service {
+		case agentpb.StartActionRequest_RestartMongoDBServiceParams_MONGOD:
+			service = "mongod"
+		case agentpb.StartActionRequest_RestartMongoDBServiceParams_PBM_AGENT:
+			service = "pbm-agent"
+		default:
+			return errors.Errorf("unknown service '%s' specified in mongod restart request", params.RestartMongodbServiceParams.Service)
```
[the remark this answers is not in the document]
idoqo: we return errors from here to the job result handler (and for now it's only logged). ideally, we should include such errors in the job's logs (in our database), but we currently don't have logs for restore, I'll try to see if we can start doing that from here.

agent/client/client.go
```diff
@@ -501,6 +501,17 @@ func (c *Client) handleStartActionRequest(p *agentpb.StartActionRequest) error {
...
+		}
+		action = actions.NewProcessAction(p.ActionId, timeout, "systemctl", []string{"restart", service})
```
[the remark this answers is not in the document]
idoqo: i don't know the best approach here really, but `NewProcessAction` was written to take CLI commands as arguments (https://github.com/percona/pmm/blob/192c07b65ed0a0ddda18da4df7bd380e2dbbe8cf/agent/runner/actions/process_action.go#L38)

managed/services/agents/jobs.go
```diff
@@ -575,7 +582,92 @@ func (s *JobsService) StartMongoDBRestoreBackupJob(
...
+	// pbmRestarts is a list of PMM agent IDs on which we successfully restarted pbm-agent
+	pbmAgentRestarts := make(map[string]struct{})
+
+	for _, pmmAgent := range rsAgents {
+		if err = s.restartMongoComponent(pmmAgent.AgentID, agentpb.StartActionRequest_RestartMongoDBServiceParams_MONGOD); err != nil {
+			return err
+		}
+		mongoRestarts[pmmAgent.AgentID] = struct{}{}
+	}
+	s.l.Infof("restarted mongod on %d out of %d services", len(mongoRestarts), len(rsMembers))
```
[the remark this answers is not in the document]
idoqo: yeah, I just thought it's still okay to log info about success operations

agent/runner/runner.go
```diff
@@ -238,6 +238,7 @@ func (r *Runner) handleAction(ctx context.Context, action actions.Action) {
 		if err != nil {
 			errMsg = err.Error()
 			l.Warnf("Action terminated with error: %+v", err)
+			l.Warnf("Action produced output: %s", string(output))
```
artemgavrilov, line 241: I think `Debug` will fit better here

managed/services/backup/backup_service.go
```diff
@@ -288,14 +288,17 @@ func (s *Service) RestoreBackup(ctx context.Context, serviceID, artifactID strin
...
 					RestoreID: restoreID,
 				},
 			}
 		case models.MongoDBServiceType:
 			jobType = models.MongoDBRestoreBackupJob
 			jobData = &models.JobData{
 				MongoDBRestoreBackup: &models.MongoDBRestoreBackupJobData{
+					ServiceID: serviceID,
 					RestoreID: restoreID,
+					DataModel: params.DataModel,
```
artemgavrilov, line 301: We have this information in artifact. Do we really need it here?

managed/services/agents/jobs.go
```diff
@@ -227,7 +227,42 @@ func (s *JobsService) handleJobResult(_ context.Context, l *logrus.Entry, result
...
+				logParams = models.CreateJobLogParams{
+					JobID:     job.ID,
+					ChunkID:   logChunkID,
+					LastChunk: true,
+					Data:      "restore successfully completed.",
+				}
+			}
+			if err := createJobLog(t.Querier, logParams.JobID, logParams.Data, logParams.ChunkID, logParams.LastChunk); err != nil {
+				s.l.WithError(err).Errorf("failed to create log for job %s [chunk: %d]", job.ID, logChunkID)
+			}
```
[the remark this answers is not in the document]
idoqo: In that case we can read `logChunkID`  from the DB (i.e last chunk that was saved) instead of starting from 0

managed/services/agents/jobs.go
```diff
@@ -617,3 +732,15 @@ func convertDataModel(model models.DataModel) (backuppb.DataModel, error) {
 		return 0, errors.Errorf("unknown data model: %s", model)
 	}
 }
+
+func createJobLog(querier *reform.Querier, jobID, data string, chunkID int, lastChunk bool) error {
```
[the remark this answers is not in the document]
idoqo: I tried that but the code looked way too messy.

managed/services/backup/backup_service.go
idoqo: I added it to make sure we only restart for physical backups. The alternative will be to use restore ID to get artifact ID, then use artifact ID to get the data model - and this felt simpler.

api/agentpb/agent.proto
```diff
@@ -328,7 +328,7 @@ message StartActionRequest {
   }
 
   // RestartMongoDBServiceParams describes an action request to restart components in a MongoDB node
-  message RestartMongoDBServiceParams {
+  message RestartServiceParams {
```
PavelKhripkov, line 331:
```suggestion
  message RestartSystemServiceParams {
```

managed/models/service_helpers_test.go
```diff
@@ -402,45 +402,43 @@ func TestServiceHelpers(t *testing.T) {
...
 
 		services, err := models.FindServices(q, models.ServiceFilters{
-			ServiceType:    pointerToServiceType(models.MongoDBServiceType),
-			ReplicationSet: "rs0",
+			ServiceType: pointerToServiceType(models.MongoDBServiceType),
+			Cluster:     "cluster0",
 		})
 		assert.NoError(t, err)
 		assert.NotNil(t, services)
 		assert.Len(t, services, 2)
```
PavelKhripkov, line 440: We don't need this line any more.

agent/client/client.go
```diff
@@ -504,12 +504,12 @@ func (c *Client) handleStartActionRequest(p *agentpb.StartActionRequest) error {
...
 		switch params.RestartMongodbServiceParams.SystemService {
-		case agentpb.StartActionRequest_RestartMongoDBServiceParams_MONGOD:
+		case agentpb.StartActionRequest_RestartServiceParams_MONGOD:
 			service = "mongod"
-		case agentpb.StartActionRequest_RestartMongoDBServiceParams_PBM_AGENT:
+		case agentpb.StartActionRequest_RestartServiceParams_PBM_AGENT:
 			service = "pbm-agent"
 		default:
-			return errors.Errorf("unknown service '%s' specified in mongod restart request", params.RestartMongodbServiceParams.SystemService)
+			return errors.Errorf("invalid service '%s' specified in mongod restart request", params.RestartMongodbServiceParams.SystemService)
```
PavelKhripkov, line 512: We're still losing this error)

managed/services/backup/backup_service.go
idoqo: we aren't using it anywhere, for now, so I've reverted it to keep the changes within scope.

agent/client/client_test.go
```diff
@@ -217,19 +217,47 @@ func TestUnexpectedActionType(t *testing.T) {
...
+					},
+				},
+				expectedCode: codes.InvalidArgument,
+			},
+		}
+
+		for _, tc := range cases {
+			t.Run(tc.name, func(t *testing.T) {
+				err = stream.Send(&agentpb.ServerMessage{Id: tc.id, Payload: tc.payload})
+				assert.NoError(t, err)
```
PavelKhripkov, line 253:
```suggestion
				require.NoError(t, err)
```

agent/client/client_test.go
```diff
@@ -217,19 +217,47 @@ func TestUnexpectedActionType(t *testing.T) {
...
 
-		msg, err = stream.Recv()
-		assert.NoError(t, err)
-		assert.Equal(t, int32(codes.Unimplemented), msg.GetStatus().GetCode())
-		assert.NoError(t, err)
+				msg, err = stream.Recv()
+				assert.NoError(t, err)
+				assert.Equal(t, int32(tc.expectedCode), msg.GetStatus().GetCode())
+				assert.NoError(t, err)
```
PavelKhripkov, line 258:
```suggestion
				require.NoError(t, err)
				assert.Equal(t, int32(tc.expectedCode), msg.GetStatus().GetCode())
```

PavelKhripkov (approved):

managed/services/agents/jobs.go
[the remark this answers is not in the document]
artemgavrilov: Yep, that's wrong. Chunk ID counter defined on the agent side, and can be incremented only there. Moreover logs chunks can arrive on PMM server after job result. In your implementation you will have chunk ID collision because you adding something there on server side.

managed/services/agents/jobs.go
```diff
@@ -235,7 +235,49 @@ func (s *JobsService) handleJobResult(_ context.Context, l *logrus.Entry, result
...
+				if err := s.runMongoPostRestore(ctx, job.Data.MongoDBRestoreBackup.ServiceID); err != nil {
+					s.l.WithError(err).Error("failed to restart components after restore from a physical backup")
+					if logErr := createJobLog(t.Querier, job.ID, err.Error(), logChunkID, true); logErr != nil {
+						s.l.WithError(logErr).Errorf("failed to create log for job %s [chunk: %d]", job.ID, logChunkID)
+					}
+					_, err = models.ChangeRestoreHistoryItem(
+						t.Querier,
+						job.Data.MongoDBRestoreBackup.RestoreID,
+						models.ChangeRestoreHistoryItemParams{
+							Status: models.ErrorRestoreStatus,
```
artemgavrilov, line 270: Missed `FinishedAt`

managed/services/agents/jobs.go
```diff
@@ -586,7 +623,92 @@ func (s *JobsService) StartMongoDBRestoreBackupJob(
...
+	serviceType := models.MongoDBServiceType
+	clusterMembers, err := models.FindServices(
+		s.db.Querier,
+		models.ServiceFilters{
+			ServiceType: &serviceType,
+			Cluster:     service.Cluster,
+		})
+	if err != nil {
+		return err
+	}
```
artemgavrilov, line 644: That should be wrapped in transaction

managed/services/agents/jobs.go
```diff
@@ -586,7 +623,92 @@ func (s *JobsService) StartMongoDBRestoreBackupJob(
...
+		s.l.Debugf("found service: %s in replica set: %s", service.ServiceName, service.ReplicationSet)
+		serviceAgents, err := models.FindPMMAgentsForService(s.db.Querier, service.ServiceID)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get pmm agent for replica set member: %s", service.ServiceID)
+		}
+		if len(serviceAgents) == 0 {
+			return errors.Errorf("cannot find pmm agent for service %s", service.ServiceID)
+		}
+		clusterAgents = append(clusterAgents, serviceAgents[0])
+	}
```
artemgavrilov, line 657: This too

managed/services/agents/jobs.go
```diff
@@ -586,7 +623,92 @@ func (s *JobsService) StartMongoDBRestoreBackupJob(
...
+		if err = s.restartMongoSystemService(pmmAgent.AgentID, agentpb.StartActionRequest_RestartSystemServiceParams_PBM_AGENT); err != nil {
+			return err
+		}
+		pbmAgentRestarts[pmmAgent.AgentID] = struct{}{}
+	}
+	s.l.Infof("successfully restarted pbm-agent on all %d services", len(pbmAgentRestarts))
+	return nil
+}
+
+func (s *JobsService) restartMongoSystemService(agentID string, service agentpb.StartActionRequest_RestartSystemServiceParams_SystemService) error {
```
artemgavrilov, line 684:
```suggestion
func (s *JobsService) restartSystemService(agentID string, service agentpb.StartActionRequest_RestartSystemServiceParams_SystemService) error {
```

managed/services/agents/jobs.go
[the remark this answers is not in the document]
idoqo: update on this: we eventually agreed to keep it this way since we don't stream restore logs yet, and re-work logs later.

PR #1425, opened 2022-11-18
PMM-11080 initial CLI tests for PMM binary

yurkovychv:
PMM-11080

Build: SUBMODULES-0


cli-tests/README.md
```diff
@@ -0,0 +1,40 @@
...
+* Open the _**cli**_ folder in console
+* Install Node.js 12+ version and make sure npx is included
+* Install project dependencies: `npm ci`
+* Install "playwright": `npx playwright install`
+* Install test module if required: `npm install -D @playwright/test`
+
+this is it! tests are good to go on specified PMM Server and/or Client.
+
+### Getting PMM server and client ready
+  * **Setup environment for CLI tests:**
```
michal-kralik, line 16: What does this item include?

michal-kralik (commented):
Looks solid. I recommend placing this in a directory such as `tests` or similar.
I expect we'll extend the test suite in the future to cover more, not just cli, in which case the folder `cli-tests` is misleading.

Can we also extend this with an example how to configure an environment from the tests?
When this runs in CI, what does the command `pmm-admin add mongodb` connect to?

michal-kralik:
> > Looks solid. I recommend placing this in a directory such as `tests` or similar. I expect we'll extend the test suite in the future to cover more, not just cli, in which case the folder `cli-tests` is misleading.
> > Can we also extend this with an example how to configure an environment from the tests? When this runs in CI, what does the command `pmm-admin add mongodb` connect to?
> 
> @michal-kralik About separate `tests` folder. I followed the strategy you have right now. You already have `api-tests` folder. So there is not sense to create `tests` folder and put just cli tests there. About mongodb related tests. Those were placed here as a POC. They are not being executed yet. So lets just ignore those.

@BupycHuk @artemgavrilov any preference here?

JiriCtvrtka (approved):

cli-tests/tests/pmm-cli/pmm.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { test, expect } from '@playwright/test';
+import * as cli from '@helpers/cliHelper';
+import PMMRestClient from '@support/types/request';
+
+test.describe('PMM binary tests @pmm-cli', async () => {
+  test('--version', async ({}) => {
+    const output = await cli.exec('pmm --version');
+    await output.assertSuccess();
```
BupycHuk, line 8: should we check if the output is correct?

cli-tests/tests/pmm-cli/pmm.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  test('server docker install with --admin-password flag', async ({ }) => {
```
BupycHuk, line 11: do we have the step to shut down the container?

cli-tests/README.md
```diff
@@ -0,0 +1,39 @@
...
+* Install project dependencies: `npm ci`
+* Install "playwright": `npx playwright install`
+* Install test module if required: `npm install -D @playwright/test`
+
+this is it! tests are good to go on specified PMM Server and/or Client.
+
+### Getting PMM server and client ready
+  * **Setup PMM Client**
+    * Download and unzip client tarball
+    * Put binaries into `/usr/bin` directory and grant corresponding permissions
```
BupycHuk, line 18: since it's already in the `pmm` repository maybe we can give information on how to build it from sources?

cli-tests/README.md
```diff
@@ -0,0 +1,39 @@
...
+  * **Setup PMM Server**
+
+      _coming soon_
+      
+    * **Run CLI tests upon remote PMM server:**  
+      _coming soon_
```
BupycHuk, line 24: as far as I see it doesn't work against remote PMM Server yet, so let's remove this line to not have confusion.

cli-tests/tests/mongodb.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  });
+
+  test('run pmm-admin add mongodb --help to check host @mongo', async ({}) => {
+    const output = await cli.exec('pmm-admin add mongodb --help');
+    /*
+    echo "$output"
+        [ "$status" -eq 0 ]
+    echo "${output}" | grep "host"
+*/
+    await test.step('Verify "metrics-mode="auto"" is present', async () => {
```
BupycHuk, line 42: looks like step name is incorrect here

cli-tests/tests/mongodb.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    const output = await cli.exec('pmm-admin add mongodb --help');
+    /*
+    echo "$output"
+        [ "$status" -eq 0 ]
+    echo "${output}" | grep "port"
+*/
+    await test.step('Verify "port" is present', async () => {
+      await output.assertSuccess();
+      await expect(output.stdout).toContain('Usage: pmm-admin add mongodb [<name> [<address>]]');
+      await expect(output.stdout).toContain('--socket=STRING');
```
BupycHuk, line 59: looks like we should check port here

cli-tests/tests/mongodb.spec.ts
```diff
@@ -0,0 +1,102 @@
+import { test, expect } from '@playwright/test';
+// import cli = require('@helpers/cliHelper'); //optional way to import with local name
+import * as cli from '@helpers/cliHelper';
+
+test.describe('Spec file for MongoDB CLI tests ', async () => {
+  test('pmm-admin mongodb --help check for socket @cli @mongo', async ({}) => {
```
BupycHuk, line 6: can we combine all these tests related to `--help` output?

BupycHuk (commented):
just curious if we could use Go instead of JS?

michal-kralik (approved):

artemgavrilov (approved):

cli-tests/tests/pmm-cli/pmm.spec.ts
```diff
@@ -3,6 +3,10 @@ import * as cli from '@helpers/cliHelper';
 import PMMRestClient from '@support/types/request';
 
 test.describe('PMM binary tests @pmm-cli', async () => {
+  test.afterEach(async ({}) => {
+    await cli.exec('docker rm -f $(docker ps -a -q)')
```
BupycHuk, line 7: does this remove all docker containers? I think we should remove only ones that we started

michal-kralik: I just learned the hard way. All my containers are gone :)

PR #1432, opened 2022-11-21
PMM-10981 Add datapoint for MySQL clusters

0leksii:
Add datapoint to indicate how many clusters/group replications with the amount of nodes there are set up in PMM.

PMM-10981

Build: SUBMODULES-0

### DIFF
--- a/managed/services/telemetry/config.default.yml
+++ b/managed/services/telemetry/config.default.yml
@@ -544,3 +544,17 @@ telemetry:
     data:
       - metric_name: "k8s_clusters_count"
         column: "count"
+
+# MySQL Clusters
+  - id: MySQLClustersCount
+    source: PMMDB_SELECT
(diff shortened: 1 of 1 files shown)


managed/services/telemetry/config.default.yml
```diff
@@ -544,3 +544,17 @@ telemetry:
     data:
       - metric_name: "k8s_clusters_count"
         column: "count"
+
+# MySQL Clusters
+  - id: MySQLClustersCount
+    source: PMMDB_SELECT
+    query: aggregated_nodes.nodes_count as nodes_count, COUNT(*) as clusters_count FROM (SELECT cluster, replication_set, COUNT(*) as nodes_count FROM services WHERE service_type = 'mysql' GROUP BY cluster, replication_set) as aggregated_nodes GROUP BY aggregated_nodes.nodes_count
```
dliakhov, line 551: Does it work for PXC clusters and Group Replication, right? Should we distinguish whether it's PXC cluster or Group Replication?

0leksii: Discussed verbally with @rnovikovP  - will try to rework to fetch data from VM as there we have more consistent data than in PG replication names/ cluster names

0leksii:
Will need to extend the exporter to get the real number of nodes from the group replication rather then depend on non-mandatory replication-set value.

PR #1434, opened 2022-11-23
PMM-10891 PG Explain with placeholders.

JiriCtvrtka:
PMM-10891
Build: https://github.com/Percona-Lab/pmm-submodules/pull/2921

**Contains also some changes/improvements for MySQL implementation of this.**


agent/agents/postgres/pgstatmonitor/stat_monitor_cache.go
```diff
@@ -151,9 +152,22 @@ func (ssc *statMonitorCache) getStatMonitorExtended(ctx context.Context, q *refo
...
 			fingerprint := c.Query
 			example := ""
+
+			var explainFingerprint string
+			var placeholdersCount uint32
+			var errParsing error
 			if !normalizedQuery {
 				example = c.Query
 				fingerprint, err = ssc.generateFingerprint(c.Query)
+				explainFingerprint, placeholdersCount, errParsing = queryparser.PostgreSQL(c.Query)
```
BupycHuk, line 162: fingerprint is already a normalized query, so we can just call `queryparser.PostgreSQLNormalized` with it as argument.

agent/agents/postgres/pgstatmonitor/stat_monitor_cache.go
```diff
@@ -151,9 +152,22 @@ func (ssc *statMonitorCache) getStatMonitorExtended(ctx context.Context, q *refo
...
+			} else {
+				explainFingerprint, placeholdersCount, errParsing = queryparser.PostgreSQLNormalized(c.Query)
```
BupycHuk, line 164: this could be moved into the `else` below

agent/agents/postgres/pgstatstatements/pgstatstatements.go
```diff
@@ -341,6 +342,16 @@ func makeBuckets(current, prev statementsMap, l *logrus.Entry) []*agentpb.Metric
 			Postgresql: &agentpb.MetricsBucket_PostgreSQL{},
 		}
 
+		if currentPSS.Query != "" {
+			explainFingerprint, placeholdersCount, err := queryparser.PostgreSQL(currentPSS.Query)
```
BupycHuk, line 346: we can get these values from currentPSS without parsing it again

BupycHuk (commented):
don't we have tests for pg_stat_monitor, why didn't they fail because of new fields?

agent/agents/postgres/pgstatmonitor/stat_monitor_cache.go
```diff
@@ -177,6 +183,12 @@ func (ssc *statMonitorCache) getStatMonitorExtended(ctx context.Context, q *refo
 				if isTruncated {
 					c.IsQueryTruncated = isTruncated
 				}
+				explainFingerprint, placeholdersCount, errParsing = queryparser.PostgreSQLNormalized(fingerprint)
+			}
+
+			if errParsing == nil {
+				c.ExplainFingerprint = explainFingerprint
```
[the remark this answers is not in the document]
JiriCtvrtka: No. I changed it and now we dont save explain fingerprint for PG since both fingerprints were same in this case.

api/managementpb/actions.proto
```diff
@@ -55,13 +55,16 @@ message StartMySQLExplainActionRequest {
...
-  // Query ID of query. 
-  string query_id = 6;
+  // Deprecated: should not be used.
+  string query = 3 [deprecated = true];
+  // Query ID of query. Required. 
+  string query_id = 6 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
```
BupycHuk, line 65: let's keep it optional for a few versions, we have to support backward compatibility with old pmm-agents

JiriCtvrtka: Added back query for MySQL explain actions. Just marked "deprecated" by comment. Query or query ID required. Because if query ID should be optional we need also query.

agent/queryparser/parser.go
```diff
@@ -15,26 +15,55 @@
...
+	return PostgreSQLNormalized(query)
+}
+
+// PostgreSQLNormalized parse query, which is already normalized and return fingeprint and placeholders.
+func PostgreSQLNormalized(q string) (string, uint32, error) {
+	// PG 10 and above has $ as a placeholders.
+	r, err := regexp.Compile(`[\$]{1}\d`)
+	if err != nil {
+		return "", 0, errors.Wrap(err, "cannot get placeholders count")
+	}
```
rishat-ishbulatov, line 61: Do we actually need to compile every time?

agent/runner/actions/mysql_explain_action.go
```diff
@@ -127,7 +125,12 @@ func (a *mysqlExplainAction) Run(ctx context.Context) ([]byte, error) {
 func (a *mysqlExplainAction) sealed() {}
 
 func (a *mysqlExplainAction) explainDefault(ctx context.Context, tx *sql.Tx) ([]byte, error) {
-	rows, err := tx.QueryContext(ctx, fmt.Sprintf("EXPLAIN /* pmm-agent */ %s", a.query))
+	inter := []any{}
+	for _, p := range a.params.Values {
```
percona-csalguero, line 129: Maybe you can use `copy` here

agent/queryparser/parser.go
JiriCtvrtka: @rishat-ishbulatov Since its used in actions, which could be ran multiple times, its better to keep it inside. Otherwise it will be global variable and then we need mutex there. Or another solution is on your mind?

percona-csalguero (approved):

agent/queryparser/parser.go
rishat-ishbulatov: @JiriCtvrtka https://stackoverflow.com/questions/46746988/is-regexp-matchstring-thread-safe

rishat-ishbulatov (approved):

agent/queryparser/parser.go
JiriCtvrtka: @rishat-ishbulatov Good point with thread safe regexp. I did small refactor as we discussed on the Slack.

BupycHuk (approved):

agent/queryparser/parser.go
rishat-ishbulatov: looks good )

PR #1435, opened 2022-11-23
PMM-11133: fix broken custom labels api

michal-kralik:
PMM-11133

Build: SUBMODULES-0


managed/services/inventory/grpc/services_server.go
```diff
@@ -268,3 +268,13 @@ func (s *servicesServer) RemoveService(ctx context.Context, req *inventorypb.Rem
 
 	return &inventorypb.RemoveServiceResponse{}, nil
 }
+
+// AddCustomLabels adds or replaces (if key exists) custom labels for a service.
+func (s *servicesServer) AddCustomLabels(ctx context.Context, req *inventorypb.AddCustomLabelsRequest) (*inventorypb.AddCustomLabelsResponse, error) {
```
artemgavrilov, line 273: We has similar `grpc` package in managemet before, but we removed it. I guess we can do the same here if you have time for that.

artemgavrilov (approved):

oksana-grishchenko (approved):

PR #1438, opened 2022-11-24
PMM-11136: vmproxy initial version

michal-kralik:
PMM-11136

Build: SUBMODULES-2916


atymchuk (changes requested):

vmproxy/main.go
```diff
@@ -0,0 +1,56 @@
...
+			Compact:             true,
+			NoExpandSubcommands: true,
+		}),
+	)
+
+	proxy.StartProxy(proxy.Config{
+		HeaderName:    opts.HeaderName,
+		ListenAddress: net.JoinHostPort(opts.ListenAddress, strconv.Itoa(opts.ListenPort)),
+		TargetURL:     opts.TargetURL,
+	})
```
maxkondr, line 55: what about program exit code?

vmproxy/pkg/proxy/proxy.go
```diff
@@ -0,0 +1,121 @@
...
+	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		if failOnInvalidHeader(rw, req, cfg.HeaderName) {
+			return
+		}
+
+		rProxy.ServeHTTP(rw, req)
+	})
+}
+
+func failOnInvalidHeader(rw http.ResponseWriter, req *http.Request, headerName string) bool {
```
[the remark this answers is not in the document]
maxkondr: `vmgateway` responses with `403 Access denied` in such cases.
I think we should behave the same. Otherwise it may be a security breach and user will get access to metrics he is not allowed to read

vmproxy/README.md
```diff
@@ -0,0 +1,36 @@
+# VMProxy
+
+VMProxy is a stateless reverse proxy which proxies requests to VictoriaMetrics and
+optionally adds `extra_filters` query based on the provided configuration.
```
maxkondr, line 4: which modes does it support now?
- read
- write
- both
?

vmproxy/pkg/proxy/proxy.go
```diff
@@ -0,0 +1,121 @@
...
+	if filters := req.Header.Get(headerName); filters != "" {
+		if _, err := parseFilters(filters); err != nil {
+			rw.Header().Set("Content-Type", "text/plain; charset=utf-8")
+			rw.WriteHeader(http.StatusPreconditionFailed)
+			io.WriteString(rw, "Failed to parse headers")
+			return true
+		}
+	}
+
+	return false
```
maxkondr, line 73: I think it should fail in case header is absent in the incoming request

vmproxy/pkg/proxy/proxy.go
```diff
@@ -0,0 +1,121 @@
...
+		rProxy.ServeHTTP(rw, req)
+	})
+}
+
+func failOnInvalidHeader(rw http.ResponseWriter, req *http.Request, headerName string) bool {
+	if filters := req.Header.Get(headerName); filters != "" {
+		if _, err := parseFilters(filters); err != nil {
+			rw.Header().Set("Content-Type", "text/plain; charset=utf-8")
+			rw.WriteHeader(http.StatusPreconditionFailed)
+			io.WriteString(rw, "Failed to parse headers")
```
maxkondr, line 68: I think it should print header name as well

.github/workflows/vmproxy.yml
```diff
@@ -0,0 +1,94 @@
+name: VMProxy
+
+on:
+  push:
+    branches:
+      - main
+      - release-*
```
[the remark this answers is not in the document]
michal-kralik: I kept the `release-*` there for consistency with other actions where it's listed as well.
Do you think we shall remove it from all the workflows?

vmproxy/main.go
michal-kralik: It's zero by default.

vmproxy/pkg/proxy/proxy_test.go
```diff
@@ -0,0 +1,122 @@
...
+				rec := httptest.NewRecorder()
+				req := httptest.NewRequest(http.MethodGet, targetURL, nil)
+				req.Header.Set(headerName, tc.headerContent)
+
+				handler.ServeHTTP(rec, req)
+				require.Equal(t, rec.Result().StatusCode, tc.expectedStatus)
+			})
+		}
+	})
+}
```
maxkondr, line 122: this test checks the response from `vmproxy`. I think we should check the outgoing request as well (after modification) that `vmproxy` forwards to VM.

Because the main purpose of this component - adjust request

vmproxy/pkg/proxy/proxy_test.go
```diff
@@ -0,0 +1,122 @@
...
+		t.Parallel()
+
+		type testParams struct {
+			expectedFilters []string
+			expectedStatus  int
+			headerContent   string
+			name            string
+		}
+
+		testCases := []testParams{
```
maxkondr, line 80: btw, the real cases will be in format `[{<key>="<value>"}]`
```
[{region="west"}]
[{region="west", project="app"}]

[{region="west", project="app"}, {region="east", project="shop"}]
```

vmproxy/pkg/proxy/proxy.go
```diff
@@ -0,0 +1,121 @@
...
+		req.URL.Host = target.Host
+
+		// Replace extra filters if present
+		if filters := req.Header.Get(headerName); filters != "" {
+			q := req.URL.Query()
+			q.Del("extra_filters")
+
+			if parsed, _ := parseFilters(filters); parsed != nil {
+				for _, f := range parsed {
+					q.Add("extra_filters", f)
```
maxkondr, line 89: accordingly to https://github.com/VictoriaMetrics/VictoriaMetrics/#prometheus-querying-api-enhancements it shall be 
```suggestion
					q.Add("extra_filters[]", f)
```

maxkondr: and the value shall be like:
```
{env=~"prod|staging",user="xyz"}
```

but not just array of strings

maxkondr: and one more - need to understand how several filters with condition "OR" between them shall be added to query

maxkondr (changes requested):

vmproxy/pkg/proxy/proxy.go
michal-kralik: The scope of this tool is to restrict access to metrics as defined in the incoming request.
In case the incoming request does not define any restrictions, the is no filtering applied.

`managed` is out of scope of this tool.
We can discuss further if this decision affects Access Control in any way. I believe it's fine and shall respect all requirements.

vmproxy/main.go
```diff
@@ -0,0 +1,56 @@
...
+type flags struct {
+	TargetURL     *url.URL `default:"http://127.0.0.1:9090" help:"Target URL where to proxy requests"`
+	ListenPort    int      `default:"1280" help:"Listen port for proxy"`
+	ListenAddress string   `default:"127.0.0.1" help:"Listen address for proxy"`
+	HeaderName    string   `default:"X-Proxy-Filter" help:"Header name to read filter configuration from. The content of the header shall be a base64 encoded JSON array with strings. Each string is a filter. Multiple filters are joined with a logical OR."` //nolint:lll
+}
+
+func main() {
+	var opts flags
+	kong.Parse(
```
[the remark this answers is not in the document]
michal-kralik: Yes, it is.

vmproxy/pkg/proxy/proxy.go
```diff
@@ -0,0 +1,121 @@
...
+	HeaderName string
+	// Address the proxy is listening on
+	ListenAddress string
+	// Target URL to forward requests to
+	TargetURL *url.URL
+}
+
+// StartProxy starts proxy which adds extra filters based on configuration.
+func StartProxy(cfg Config) {
+	logrus.Infof("Starting to proxy at http://%s to %s", cfg.ListenAddress, cfg.TargetURL.String())
```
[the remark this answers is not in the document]
michal-kralik: There are no debug levels supported at the moment.

vmproxy/pkg/proxy/proxy_test.go
michal-kralik: That's fine. These are unit tests and vmproxy does not enforce any format.

vmproxy/pkg/proxy/proxy.go
michal-kralik: vmproxy does not validate the format of the filters. That is left up to VictoriaMetrics. It just passes the provided filters.

michal-kralik: > and one more - need to understand how several filters with condition "OR" between them shall be added to query

By sending the query parameter multiple times.

vmproxy/pkg/proxy/proxy.go
michal-kralik: > another point to think over - should `vmproxy` accept all incoming requests or only requests from `managed` (how to determine that request is sent by `managed`) ?

This is out of scope of `vmproxy`. It needs to be managed on a different level.

maxkondr: Ok, we can start with it now. Later, depending on future requirements, we can improve this component.

maxkondr (approved):

recharte (changes requested):
Good job @michal-kralik just a small typo in the README

vmproxy/README.md
```diff
@@ -0,0 +1,36 @@
...
+- `region=EU`
+
+Multiple filters are joined with logical OR - such as `env=QA OR region=EU`
+
+If the header is not present, no extra filters are applied.
+
+## Contributing notes
+
+### Building
+`make build`
```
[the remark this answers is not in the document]
michal-kralik: Nice catch

vmproxy/pkg/proxy/proxy.go
```diff
@@ -0,0 +1,121 @@
...
+		req.URL.Scheme = target.Scheme
+		req.URL.Host = target.Host
+
+		// Replace extra filters if present
+		if filters := req.Header.Get(headerName); filters != "" {
+			q := req.URL.Query()
+			q.Del("extra_filters")
+
+			if parsed, _ := parseFilters(filters); parsed != nil {
+				for _, f := range parsed {
```
maxkondr, line 88: One more: If filter is passed via X-Proxy-Filter, `extra_filters` from original request shall be removed removed. 
So that the outgoing request  to VictoriaMetrics will contain only filters defined by user's role

michal-kralik: I tried adding a test case for this and it failed.
Thanks to this I figured you were right about the parameter name - it shall include the brackets `extra_filters[]`.

I was under the impression the brackets are automatically added by the struct and they weren't.
Fixed now.

vmproxy/pkg/proxy/proxy.go
```diff
@@ -0,0 +1,121 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+// Package proxy provides http reverse proxy functionality
+package proxy
```
artemgavrilov, line 17: We are not using this `pkg` naming, I propose to put `proxy` package right in the root.

artemgavrilov (commented):
Why we are not embedding this into managed? Why do we need it a separate process?

.github/workflows/vmproxy.yml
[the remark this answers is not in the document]
atymchuk: Ok, let's remove then?

michal-kralik:
> Why we are not embedding this into managed? Why do we need it a separate process?

The idea is this can be a separate component released to the community as an alternative to `vmgateway`.
For this reason we did not want to tightly couple it with `managed`

artemgavrilov (approved):

recharte (approved):

PR #1449, opened 2022-11-28
PMM-11128: nodes info v2

ritbl:
[PMM-11128](https://jira.percona.com/browse/PMM-10541)

Build: 
- https://github.com/Percona-Lab/pmm-submodules/pull/2920


managed/services/telemetry/transform.go
```diff
@@ -41,7 +41,8 @@ func transformToJSON(config *Config, metrics []*pmmv1.ServerMetric_Metric) ([]*p
 		return nil, errors.Errorf("invalid metrics config")
 	}
 
-	firstMetric := config.Data[0].MetricName
+	// consider first metric is the beginning of the object window
+	firstMetric := metrics[0].Key
```
ritbl, line 45: small bug, that was found while testing

0leksii (approved):

managed/services/telemetry/transform.go
gen1us2k: There can be panic once there are no metrics returned

gen1us2k (approved):

managed/services/telemetry/transform.go
ritbl: if metrics cannot be collected, I think, we should not panic.

without this change if would with "return nil, errors.Errorf("invalid metrics sequence: no match with first metric")", which will be logged.

atymchuk (approved):

atymchuk:
@ritbl Looks like we need to fix the API test `Test_transformToJSON`

managed/services/telemetry/transform.go
ritbl: @gen1us2k I see what you do mean, I will fix that potential case and write test for it

docker-compose.yml
```diff
@@ -21,23 +21,24 @@ services:
...
     extra_hosts:
       - host.docker.internal:host-gateway
-    #      - portal.localhost:${PORTAL_HOST:-host-gateway}
-    #      - check.localhost:${PORTAL_CHECK_HOST:-host-gateway}
-    #      - pmm.localhost:${PORTAL_PMM_HOST:-host-gateway}
-    #      - check-dev.percona.com:${PORTAL_PMM_HOST:-host-gateway}
+    #  - portal.localhost:${PORTAL_HOST:-host-gateway}
+    #  - check.localhost:${PORTAL_CHECK_HOST:-host-gateway}
+    #  - pmm.localhost:${PORTAL_PMM_HOST:-host-gateway}
+    #  - check-dev.percona.com:${PORTAL_PMM_HOST:-host-gateway}
```
pkadej, line 41: I think we can unstage whitespaces/comments formatting.

pkadej (approved):

pkadej (approved):

managed/services/telemetry/transform.go
gen1us2k: I didn't see `if metrics == nil` somewhere above the code.

PR #1470, opened 2022-12-01
PMM-10325 vmagent auth to grafana service token

rishat-ishbulatov:
[PMM-10325](https://jira.percona.com/browse/PMM-10325)

[Build](https://github.com/Percona-Lab/pmm-submodules/pull/2928)


api/inventorypb/agents.proto
```diff
@@ -34,6 +34,10 @@ enum AgentType {
 message PMMAgent {
   // Unique randomly generated instance identifier.
   string agent_id = 1;
+  // Login represent login for vmagent auth config.
```
BupycHuk, line 37: let's move these fields into register node response and leave only one field: token

agent/commands/setup.go
```diff
@@ -158,6 +158,8 @@ func register(cfg *config.Config, l *logrus.Entry) {
 
 	fmt.Printf("Registered.\n")
 	cfg.ID = agentID
+	cfg.Server.Username = "api_key"
+	cfg.Server.Password = token
```
BupycHuk, line 162: let's check that the token is not empty in this response. it might happen when PMM Server version is older then PMM client

managed/services/management/node.go
```diff
@@ -121,9 +129,16 @@ func (s *NodeService) Register(_ context.Context, req *managementpb.RegisterNode
 			CreateNodeExporter(tx.Querier, pmmAgent.AgentID, nil, isPushMode(req.MetricsMode), req.DisableCollectors,
 				pointer.ToStringOrNil(req.AgentPassword), "")
 		return err
-	}); e != nil {
+	})
+	if e != nil {
 		return nil, e
 	}
 
+	apiKeyName := fmt.Sprintf("pmm-vmagent-%s-%d", req.NodeName, rand.Int63())
```
BupycHuk, line 137:
```suggestion
	apiKeyName := fmt.Sprintf("pmm-agent-%s-%d", req.NodeName, rand.Int63())
```
it can be used for pmm-admin commands too

managed/services/management/node_test.go
```diff
@@ -35,6 +35,12 @@ import (
 	"github.com/percona/pmm/managed/utils/tests"
 )
 
+type dummyProvider struct{}
+
+func (d *dummyProvider) CreateAdminAPIKey(context.Context, string) (int64, string, error) {
+	return 0, "", nil
+}
```
BupycHuk, line 42: we have mocks for that, let's use them

BupycHuk (approved):

agent/commands/clients.go
```diff
@@ -200,9 +200,14 @@ func serverRegister(cfgSetup *config.Setup) (string, error) {
 		Context: context.Background(),
 	})
 	if err != nil {
-		return "", err
+		return "", "", err
 	}
-	return res.Payload.PMMAgent.AgentID, nil
+
+	if res.Payload.Token == "" {
+		return "", "", errors.New("server responded with empty token (please consider upgrade server version).")
```
michal-kralik, line 207: We shall make this error more descriptive. It's not clear what is expected of the user.

managed/services/management/node.go
```diff
@@ -121,9 +131,16 @@ func (s *NodeService) Register(_ context.Context, req *managementpb.RegisterNode
...
-	}); e != nil {
+	})
+	if e != nil {
 		return nil, e
 	}
 
+	apiKeyName := fmt.Sprintf("pmm-agent-%s-%d", req.NodeName, rand.Int63())
+	_, res.Token, e = s.akp.CreateAdminAPIKey(ctx, apiKeyName)
+	if e != nil {
+		return res, e
```
michal-kralik, line 142: Is it intended we return `res` as not `nil`?

managed/services/management/node.go
```diff
@@ -121,9 +131,16 @@ func (s *NodeService) Register(_ context.Context, req *managementpb.RegisterNode
 			CreateNodeExporter(tx.Querier, pmmAgent.AgentID, nil, isPushMode(req.MetricsMode), req.DisableCollectors,
 				pointer.ToStringOrNil(req.AgentPassword), "")
 		return err
-	}); e != nil {
+	})
+	if e != nil {
 		return nil, e
 	}
 
+	apiKeyName := fmt.Sprintf("pmm-agent-%s-%d", req.NodeName, rand.Int63())
```
michal-kralik, line 139: Can you please confirm we initialize `rand` with `seed` or in other ways so we actually get random integers when `managed` runs?

michal-kralik (changes requested):

agent/commands/clients.go
BupycHuk: let's not return an error, but just use the username and password that we already have.

BupycHuk: Maybe write some warning message, but let it work

managed/services/management/node.go
BupycHuk: I think it's better to return nil

michal-kralik (changes requested):
Let's go with Nurlan's suggestion of not returning error when token is not available. Let's just log it.

managed/services/management/node.go
rishat-ishbulatov: done init rand seed in main

agent/commands/setup.go
```diff
@@ -155,9 +155,14 @@ func register(cfg *config.Config, l *logrus.Entry) {
...
 		os.Exit(1)
 	}
-
-	fmt.Printf("Registered.\n")
 	cfg.ID = agentID
+	if token != "" {
+		cfg.Server.Username = "api_key"
+		cfg.Server.Password = token
+	} else {
+		fmt.Printf("PMM server responded with empty api-key token (please consider upgrade pmm-server to the latest version).\n")
```
michal-kralik, line 163:
```suggestion
		l.Info("PMM Server responded with an empty api key token. Consider upgrading PMM Server to the latest version.")
```

michal-kralik (changes requested):
Looks good, one last change re logged message

BupycHuk (approved):

michal-kralik (approved):

PR #1477, opened 2022-12-05
PMM-11189: Build and enable vmproxy

michal-kralik:
PMM-11189

Build: SUBMODULES-2930


managed/services/supervisord/supervisord.go
```diff
@@ -680,6 +680,28 @@ stdout_logfile_backups = 3
 redirect_stderr = true
 {{end}}
 
+{{define "vmproxy"}}
+[program:vmproxy]
+priority = 7
```
atymchuk, line 685: Too many programs have the same `priority = 7` value. Let's give it priority 9, so we don't end up guessing who should start when.

atymchuk, line 685:
```suggestion
priority = 9
```

managed/testdata/supervisord.d/vmproxy.ini
```diff
@@ -0,0 +1,21 @@
+; Managed by pmm-managed. DO NOT EDIT.
+
+[program:vmproxy]
+priority = 7
```
atymchuk, line 4:
```suggestion
priority = 9
```
Ditto

update/ansible/playbook/tasks/files/datasources.yml
```diff
@@ -5,7 +5,7 @@ datasources:
   orgId: 1
   type: prometheus
   access: proxy
-  url: http://127.0.0.1:9090/prometheus/
+  url: http://127.0.0.1:1280/prometheus/
```
atymchuk, line 8: @puneet0191 @yurkovychv FYI, be aware of this change :)

atymchuk (changes requested):

docker-compose.yml
```diff
@@ -21,6 +21,7 @@ services:
       - AWS_ACCESS_KEY=${AWS_ACCESS_KEY}
       - AWS_SECRET_KEY=${AWS_SECRET_KEY}
       - ENABLE_BACKUP_MANAGEMENT=1
+      - ENABLE_RBAC=${ENABLE_RBAC}
```
BupycHuk, line 24:
```suggestion
      - ENABLE_AC=${ENABLE_AC}
```

managed/services/user/user.go
```diff
@@ -125,3 +125,23 @@ func (s *Service) UpdateUser(ctx context.Context, req *userpb.UserUpdateRequest)
 	}
 	return resp, nil
 }
+
+// ListUsers lists all users and their details.
+func (s *Service) ListUsers(_ context.Context, _ *userpb.ListUsersRequest) (*userpb.ListUsersResponse, error) {
```
BupycHuk, line 130: seems like we should call this endpoint somehow else since we don't return any other details except role ID

update/ansible/playbook/tasks/files/datasources.yml
BupycHuk, line 8: does VM gateway use the same port?

BupycHuk (approved):

docker-compose.yml
michal-kralik: This is how it's in the requirements. The flag will be removed with GA.
I did not want to change it so QA and docs is not confused.

update/ansible/playbook/tasks/files/datasources.yml
michal-kralik: No, they use different.

managed/services/user/user.go
michal-kralik: I imagine this endpoint will be used in the future and we'll add more details to the response.
I did not want to create a separate endpoint just for the IDs and then create a new endpoint (or migrate) when we need more details.

Happy to move it to its own endpoint though if it's preferred.

docker-compose.yml
BupycHuk: Yeah, but since we changed the naming shouldn't we change the environment variable as well? seems like it was forgotten during migration from RBAC to AC

managed/services/user/user.go
BupycHuk: I mean that we already have more details we can gather from grafana, but don't include them, even the Username.

update/ansible/playbook/tasks/files/datasources.yml
michal-kralik: Sure, that's configurable. Any specific reason for this change?

managed/services/user/user.go
michal-kralik: Do you have in mind we shall retrieve the relevant details by default and provide them in the response?

docker-compose.yml
michal-kralik: @maxkondr any thoughts about switching to `ENABLE_ACCESS_CONTROL`?

managed/services/user/user.go
BupycHuk: let's add it to TODO and discuss with @matejkubinec whether will it be easier for him as a consumer of this API endpoint.

matejkubinec: Yeah, currently the only use for it for the roles. We could evaulate this decision in the future, but I think we want to keep as much stuff separate from grafana as possible.

BupycHuk: ok, let's keep it as is then.

update/ansible/playbook/tasks/files/datasources.yml
michal-kralik: Changed to `8430`.
`8431` is used by vmgateway.

PR #1481, opened 2022-12-06
PMM-11151 Check pg_stat_statements ext version instead of postgres

tshcherban:
PMM-11151

Build: SUBMODULES-2936

We need to check pg_stat_statements extension version, not postgres server one, because we might do postgres update, which leaves extension installed version intact, so query would refer to invalid columns.


agent/agents/postgres/pgstatstatements/pgstatstatements.go
```diff
@@ -113,15 +113,24 @@ func getPGVersion(q *reform.Querier) (pgVersion float64, err error) {
 	return strconv.ParseFloat(v, 64)
 }
 
+func getPgStatVersion(q *reform.Querier) (pgVersion float64, err error) {
+	var v string
+	err = q.QueryRow(fmt.Sprintf("SELECT /* %s */ extVersion from pg_extension where pg_extension.extname = 'pg_stat_statements'", queryTag)).Scan(&v)
```
YaroslavPodorvanov, line 118: minor **from** and **where** to upper case **FROM** and **WHERE**

YaroslavPodorvanov (approved):
LGTM

agent/agents/postgres/pgstatstatements/pgstatstatements.go
```diff
@@ -113,15 +113,24 @@ func getPGVersion(q *reform.Querier) (pgVersion float64, err error) {
...
+	if err != nil {
+		return
+	}
+	return strconv.ParseFloat(v, 64)
```
michal-kralik, line 122: Are you sure we can parse the version to float?
Wouldn't a semver comparison make more sense?

michal-kralik (changes requested):

agent/agents/postgres/pgstatstatements/pgstatstatements.go
```diff
@@ -113,15 +113,24 @@ func getPGVersion(q *reform.Querier) (pgVersion float64, err error) {
...
+}
+
 func rowsByVersion(q *reform.Querier, tail string) (*sql.Rows, error) {
-	pgVersion, err := getPGVersion(q)
+	pgStatVersion, err := getPgStatVersion(q)
```
michal-kralik, line 126: Can we remove the original `getPGVersion` function? It seems to be unused.

agent/agents/postgres/pgstatstatements/pgstatstatements.go
tshcherban: for sure, that was borrowed from the old code. ill fix to semver

tshcherban: done
however had to append '.0' to the query result because postgres returns two-component version

michal-kralik (approved):

ShashankSinha252 (approved):

PR #1484, opened 2022-12-07
PMM-7: less log noise in dev

michal-kralik:
PMM-7

- Less noise in the logs in dev environment
- Moved round-trip time to `debug` level


docker-compose.yml
```diff
@@ -21,6 +21,7 @@ services:
       - AWS_ACCESS_KEY=${AWS_ACCESS_KEY}
       - AWS_SECRET_KEY=${AWS_SECRET_KEY}
       - ENABLE_BACKUP_MANAGEMENT=1
+      - LESS_LOG_NOISE=1
```
artemgavrilov, line 24: Maybe better to invert this flag? By default suppress noisy logs but allow to enable them with `LOG_VERBOSE` flag?

michal-kralik: I'd love to do that, if possible.
This would remove some of the logs from production environments though. Not sure if the logs help in production in any way though.

These logs would be gone from production unless enabled.
```
INFO[2022-12-15T14:33:37.297+00:00] Starting RPC /server.Server/Readiness ...     request=7624e3a4-7c85-11ed-b003-0242ac150002
INFO[2022-12-15T14:33:37.300+00:00] RPC /server.Server/Readiness done in 2.384712ms.  request=7624e3a4-7c85-11ed-b003-0242ac150002
INFO[2022-12-15T14:33:37.684+00:00] Round-trip time: 752.923µs. Estimated clock drift: 32.329µs.  agent_id=pmm-server request=28e30aa5-7c85-11ed-b003-0242ac150002
```

artemgavrilov: I wonder why we have them as `Info`? Debug should fit better IMHO.

@BupycHuk any thoughts?

BupycHuk: Regarding the round-trip message I agree, regarding RPC calls, I'm not sure.

michal-kralik:
@BupycHuk @artemgavrilov since you both agree about the roundtrip calls, can you please consider approving the PR?

BupycHuk (approved):

artemgavrilov (approved):

PR #1496, opened 2022-12-12
PMM-11219: telemetry-ia-fix

ritbl:
[PMM-11219](https://jira.percona.com/browse/PMM-11219)

Build:

- [ ] https://github.com/Percona-Lab/pmm-submodules/pull/2946


managed/services/telemetry/config.default.yml
```diff
@@ -2,7 +2,7 @@ telemetry:
   # PMM - generic information
   - id: PMMServerIAEnabled
     source: PMMDB_SELECT
-    query: (CASE WHEN alerting->'disabled' = 'false' THEN '1' ELSE '0' END) AS ia_enabled FROM settings s, jsonb_extract_path(s.settings, 'alerting') AS ia
+    query: (CASE WHEN ia -> 'enabled' = 'true' THEN '1' ELSE '0' END) AS ia_enabled FROM settings s, jsonb_extract_path(s.settings, 'ia') AS ia
```
artemgavrilov, line 5: `ia` in settings was replaced with `alerting`. This change was added in 2.32 as I remember.

artemgavrilov (changes requested):

managed/services/telemetry/config.default.yml
ritbl: for some reason we have old grafana in dev-container (grafana 8.3.10), while it should be 9.2.5.

managed/services/telemetry/config.default.yml
```diff
@@ -2,7 +2,7 @@ telemetry:
   # PMM - generic information
   - id: PMMServerIAEnabled
     source: PMMDB_SELECT
-    query: (CASE WHEN alerting->'disabled' = 'false' THEN '1' ELSE '0' END) AS ia_enabled FROM settings s, jsonb_extract_path(s.settings, 'alerting') AS ia
+    query: (CASE WHEN ia->'disabled' = 'false' THEN '1' ELSE '0' END) AS ia_enabled FROM settings s, jsonb_extract_path(s.settings, 'alerting') AS ia
```
artemgavrilov, line 5: I suggest to eliminate `ia` abbreviation, there is no more such system.
```suggestion
    query: (CASE WHEN alerting->'disabled' = 'false' THEN '1' ELSE '0' END) AS ia_enabled FROM settings s, jsonb_extract_path(s.settings, 'alerting') AS alerting
```

ritbl: sure, good idea

dliakhov (approved):

pkadej (approved):

PR #1497, opened 2022-12-12
PMM-11216-server-upgrade-side-container

michal-kralik:
PMM-11216

Build: SUBMODULES-2947


michal-kralik:
This is ready for review.
Any feedback is greatly appreciated.

Once we agree this approach works, I'm going to create a separate PR for documenting the solution.

admin/commands/base/setup.go
```diff
@@ -84,14 +84,30 @@ func SetupClients(ctx context.Context, globalFlags *flags.GlobalFlags) {
...
+		globalFlags.SkipTLSCertificateCheck,
+		logrus.Fields{"component": "server-transport"})
+
+	inventorypb.Default.SetTransport(transport)
+	managementpb.Default.SetTransport(transport)
+	serverpb.Default.SetTransport(transport)
+}
+
+// GetGRPCTransport returns a configured transport for gRPC.
+func GetGRPCTransport(ctx context.Context, url *url.URL, enableDebug, skipTLSCertificateCheck bool, logFields logrus.Fields) *httptransport.Runtime {
```
BupycHuk, line 100: Seems like this function isn't used outside of this package and can be private.

admin/commands/pmm/server/docker/deps.go
```diff
@@ -27,12 +27,12 @@ import (
...
 
-//go:generate ../../../../../bin/mockery -name=Functions -case=snake -inpkg -testonly
+//go:generate ../../../../../bin/mockery -name=functions -case=snake -inpkg -testonly
 
-// Functions contain methods required to interact with Docker.
-type Functions interface {
-	Imager
-	Installer
+// functions contain methods required to interact with Docker.
+type functions interface {
```
BupycHuk, line 33: maybe it's better to rename it to dockerFunctions or something else?

BupycHuk: the same with other usages of `functions` naming

api/swagger/swagger.json
```diff
@@ -1081,6 +1081,11 @@
                   "type": "string",
                   "format": "date-time",
                   "x-order": 4
+                },
+                "updater_ready": {
+                  "description": "True if pmm-server-upgrade is ready for an update.",
```
BupycHuk, line 1086: what does that mean?

managed/services/server/server.go
```diff
@@ -322,106 +345,81 @@ func (s *Server) StartUpdate(ctx context.Context, req *serverpb.StartUpdateReque
 		return nil, status.Error(codes.FailedPrecondition, "Updates are disabled via DISABLE_UPDATES environment variable.")
 	}
 
-	offset, err := s.supervisord.StartUpdate()
```
BupycHuk, line 325: let's have some period for having both types of updates(supervisor and pmm-server-update) because we can't just stop supporting upgrades for big part of our customers suddenly.

admin/services/status/status.go
```diff
@@ -0,0 +1,36 @@
...
+	updatepb.UnimplementedStatusServer
+}
+
+// New returns new instance of Server.
+func New() *Server {
+	return &Server{}
+}
+
+// Ready returns no error if the daemon is ready.
+func (s *Server) Ready(ctx context.Context, req *updatepb.ReadyRequest) (*updatepb.ReadyResponse, error) {
```
BupycHuk, line 34: should we check if we have access to docker and if we can find the docker container of PMM since it's a `Ready` function, not just `Available`?

BupycHuk (changes requested):
let's give some period to migrate to new process for our customers, we can't suddenly stop supporting default UI upgrades.
https://github.com/percona/pmm/pull/1497/files#diff-341c8a36cd13c4ece0a68dff863a27a2bcb6ca0b1823c644ae4c07d35faa7ad6R40 it doesn't look like `Cli`, maybe `dockerClient`?

admin/commands/pmm/server/docker/upgrade.go
```diff
@@ -71,9 +73,20 @@ func (u *upgradeResult) String() string {
...
+	}
+}
+
 // RunCmdWithContext runs upgrade command.
-func (c *UpgradeCommand) RunCmdWithContext(ctx context.Context, globals *flags.GlobalFlags) (commands.Result, error) { //nolint:unparam
-	logrus.Info("Starting PMM Server upgrade via Docker")
+func (c *UpgradeCommand) RunCmdWithContext(ctx context.Context, _ *flags.GlobalFlags) (commands.Result, error) { //nolint:unparam
+	if c.l == nil {
+		c.l = logrus.NewEntry(logrus.StandardLogger())
+	}
```
artemgavrilov, line 87: Is it really needed?

admin/commands/pmm/server/docker/upgrade.go
```diff
@@ -90,12 +103,17 @@ func (c *UpgradeCommand) RunCmdWithContext(ctx context.Context, globals *flags.G
...
 	if err := c.pullImage(ctx, c.DockerImage); err != nil {
 		return nil, err
 	}
 
-	logrus.Infof("Stopping PMM Server in container %q", currentContainer.Name)
+	c.l.Infof("Stopping PMM Server in container %q", currentContainer.Name)
+
+	// We wait a bit so the logs can be requested by pmm-managed before
+	// the container is stopped.
+	<-time.After(c.waitBeforeContainerStop)
```
artemgavrilov, line 115: This is a good place also to check for context cancellation.

admin/commands/pmm/server/docker/upgrade.go
```diff
@@ -114,28 +132,33 @@ func (c *UpgradeCommand) RunCmdWithContext(ctx context.Context, globals *flags.G
...
 	if err != nil {
-		logrus.Info("We could not disable restart policy in the old container.")
-		logrus.Infof(`We strongly recommend removing the old container manually with "docker rm %s"`, currentContainer.Name)
-		logrus.Errorf("Error for reference: %#v", err)
+		c.l.Info("We could not disable restart policy in the old container.")
+		c.l.Infof(`We strongly recommend removing the old container manually with "docker rm %s"`, currentContainer.Name)
+		c.l.Errorf("Error for reference: %#v", err)
 	}
 
+	c.l.Info("PMM Server update finished successfully")
```
artemgavrilov, line 140:
```suggestion
	c.l.Info("PMM Server update finished successfully.")
```

admin/commands/pmm/server/docker/upgrade.go
```diff
@@ -114,28 +132,33 @@ func (c *UpgradeCommand) RunCmdWithContext(ctx context.Context, globals *flags.G
 		RestartPolicy: container.RestartPolicy{Name: "no"},
 	})
 	if err != nil {
-		logrus.Info("We could not disable restart policy in the old container.")
-		logrus.Infof(`We strongly recommend removing the old container manually with "docker rm %s"`, currentContainer.Name)
-		logrus.Errorf("Error for reference: %#v", err)
+		c.l.Info("We could not disable restart policy in the old container.")
+		c.l.Infof(`We strongly recommend removing the old container manually with "docker rm %s"`, currentContainer.Name)
```
artemgavrilov, line 136:
```suggestion
		c.l.Infof(`We strongly recommend removing the old container manually with "docker rm %s".`, currentContainer.Name)
```

admin/commands/pmm/server/docker/upgrade.go
```diff
@@ -71,9 +73,20 @@ func (u *upgradeResult) String() string {
...
+
 // RunCmdWithContext runs upgrade command.
-func (c *UpgradeCommand) RunCmdWithContext(ctx context.Context, globals *flags.GlobalFlags) (commands.Result, error) { //nolint:unparam
-	logrus.Info("Starting PMM Server upgrade via Docker")
+func (c *UpgradeCommand) RunCmdWithContext(ctx context.Context, _ *flags.GlobalFlags) (commands.Result, error) { //nolint:unparam
+	if c.l == nil {
+		c.l = logrus.NewEntry(logrus.StandardLogger())
+	}
+
+	c.l.Info("Starting PMM Server upgrade via Docker")
```
artemgavrilov, line 89:
```suggestion
	c.l.Info("Starting PMM Server upgrade via Docker.")
```

admin/commands/pmm/server/docker/upgrade.go
```diff
@@ -114,28 +132,33 @@ func (c *UpgradeCommand) RunCmdWithContext(ctx context.Context, globals *flags.G
...
+	if ok && v == "cli" {
+		return true
 	}
 
 	return false
 }
 
 func (c *UpgradeCommand) backupVolumes(ctx context.Context, container *types.ContainerJSON) error {
-	logrus.Info("Starting backup of volumes")
+	c.l.Info("Starting backup of volumes")
```
artemgavrilov, line 159:
```suggestion
	c.l.Info("Starting backup of volumes.")
```

artemgavrilov: And everywhere else.

admin/commands/pmm_server_upgrade/start/start.go
```diff
@@ -0,0 +1,159 @@
...
+
+// BeforeApply is run before the command is applied.
+func (c *StartCommand) BeforeApply() error {
+	commands.SetupClientsEnabled = false
+	return nil
+}
+
+// RunCmdWithContext runs command
+func (c *StartCommand) RunCmdWithContext(ctx context.Context, globals *flags.GlobalFlags) (commands.Result, error) {
+	logrus.Info("Starting PMM Server Upgrade")
```
artemgavrilov, line 73: In `docker upgrade` command logger was set in constructor.

admin/commands/pmm_server_upgrade/start/start.go
```diff
@@ -0,0 +1,159 @@
...
+		reflection.Register(gRPCServer)
+		channelz.RegisterChannelzServiceToServer(gRPCServer)
+	}
+
+	// run server until it is stopped gracefully
+	go func() {
+		var err error
+		for {
+			l.Infof("Starting gRPC server on unix://%s", socketPath)
+			err := os.Remove(socketPath)
```
artemgavrilov, line 124:
```suggestion
			err = os.Remove(socketPath)
```

admin/commands/pmm_server_upgrade/start/start.go
```diff
@@ -0,0 +1,159 @@
...
+			if err != nil && !errors.Is(err, os.ErrNotExist) {
+				logrus.Panic(err)
+			}
+
+			listener, err := net.Listen("unix", socketPath)
+			if err != nil {
+				logrus.Panic(err)
```
artemgavrilov, line 131:
```suggestion
				l.Panic(err)
```

admin/commands/pmm_server_upgrade/start/start.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	// run server until it is stopped gracefully
+	go func() {
+		var err error
+		for {
+			l.Infof("Starting gRPC server on unix://%s", socketPath)
+			err := os.Remove(socketPath)
+			if err != nil && !errors.Is(err, os.ErrNotExist) {
+				logrus.Panic(err)
```
artemgavrilov, line 126:
```suggestion
				l.Panic(err)
```

admin/commands/pmm_server_upgrade/start/start.go
```diff
@@ -0,0 +1,159 @@
...
+	gRPCServer := grpc.NewServer(
+		grpc.MaxRecvMsgSize(gRPCMessageMaxSize),
+		grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(
+			grpc_validator.UnaryServerInterceptor(),
+		)),
+	)
+
+	u, err := update.New(ctx, c.DockerImage, gRPCMessageMaxSize)
+	if err != nil {
+		logrus.Fatal(err)
```
artemgavrilov, line 107:
```suggestion
		l.Fatal(err)
```

admin/pkg/docker/errors.go
```diff
@@ -0,0 +1,21 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package docker
+
+import "github.com/docker/docker/client"
+
+func (b *Base) IsErrNotFound(err error) bool {
```
artemgavrilov, line 19: I can't find where it's used

admin/services/update/update.go
```diff
@@ -0,0 +1,224 @@
...
+func (s *Server) UpdateStatus(ctx context.Context, req *updatepb.UpdateStatusRequest) (*updatepb.UpdateStatusResponse, error) { //nolint:unparam
+	var err error
+	var lines []string
+	var newOffset uint32
+	var done bool
+
+	// wait up to 30 seconds for new log lines
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+	for ctx.Err() == nil {
```
artemgavrilov, line 140: We are loosing context error message, is it ok?

api/updatepb/status.proto
```diff
@@ -0,0 +1,27 @@
...
+package update;
+
+option go_package = "api/updatepb;updatepb";
+
+import "protoc-gen-openapiv2/options/annotations.proto";
+import "google/api/annotations.proto";
+
+message ReadyRequest {}
+
+message ReadyResponse {}
```
artemgavrilov, line 12: I missed this thing, when updater is not ready we returning an error?

michal-kralik:
@artemgavrilov I see you prefer ending log messages with a dot. Could you elaborate on why you prefer this approach?

In general, I believe it's a common practice not to end log messages with a dot.
Our codebase uses a mix of both approaches while it seems we prefer log messages without a trailing dot.

Looking for feedback here.

artemgavrilov:
> @artemgavrilov I see you prefer ending log messages with a dot. Could you elaborate on why you prefer this approach?
> 
> In general, I believe it's a common practice not to end log messages with a dot. Our codebase uses a mix of both approaches while it seems we prefer log messages without a trailing dot.
> 
> Looking for feedback here.

It seems it's not mentioned in our code style docs, but I believe it's how I was told to do that when I joined Percona. Idea behind that is simple: everything that starts with capital letter opens a sentence, and every sentence should be closed. Also some logs consists from multiple sentences, so it looks weird when first sentence has dot, and second - not. Example: `Something bad happened. Check settings`

In general I don't have any strong opinion on that, so if you know better practice you always welcome to present it to the team. The only rule here that we should accept something one and follow it, otherwise it will be mess.

admin/commands/pmm/server/docker/deps.go
michal-kralik: Originally I had it with the name `dockerFunctions` but golint complained `docker.dockerFunctions` stutters :)
I'd prefer `dockerFunctions` as well.

api/swagger/swagger.json
michal-kralik: When this field is `true`, it means the following:
- `pmm-server-upgrade` runs
- It is reachable
- Responds to the `ready` request without an error

admin/services/status/status.go
michal-kralik: Originally I expected to perform the actions you suggested in `Ready`.
Later I realized it may be better to have this in the `update` flow instead. The advantage is we can provide more information what's wrong through the logs.

If we were to return we're not "ready" because the Docker socket is not available, the only way to check for the user what's wrong is to go through the logs of the container.

Looking for feedback here though.

admin/commands/pmm/server/docker/upgrade.go
michal-kralik: The `UpgradeCommand` is used both by `pmm` and `pmm-server-upgrade`.
The `pmm-server-upgrade` redirects logs to a file and provides its own instance of logger.
`pmm` on the other hand does not initialize the logger when the command runs, thus the initialization.

admin/commands/pmm_server_upgrade/start/start.go
michal-kralik: That's correct. See the comment for the `UpgradeCommand` above. It should clarify why we don't need to initialize it here.

api/updatepb/status.proto
michal-kralik: At the moment we never return an error. We will though if there's some additional logic in the future.
`managed` expects no error in case `pmm-server-upgrade` is ready.

michal-kralik:
@atymchuk @artemgavrilov @BupycHuk do you guys have bandwidth to do a review here?

api/updatepb/update.proto
```diff
@@ -0,0 +1,67 @@
...
+message UpdateStatusRequest {
+  // Token as returned by StartUpdateResponse.
+  string logs_token = 1;
+  // Offset where to start reading logs.
+  uint32 offset = 2;
+}
+
+message UpdateStatusResponse {
+  // Individual log lines.
+  repeated string lines = 1;
```
atymchuk, line 36: Should it be called `lines` or just `logs` to be more explicit? It'd echo the `logs_token` mentioned above.

managed/services/server/server.go
```diff
@@ -35,18 +35,24 @@ import (
...
 	"gopkg.in/reform.v1"
 
 	"github.com/percona/pmm/api/serverpb"
+	"github.com/percona/pmm/api/updatepb"
 	"github.com/percona/pmm/managed/models"
 	"github.com/percona/pmm/managed/utils/envvars"
 	"github.com/percona/pmm/version"
 )
 
+const updaterSocketPath = "/srv/pmm-server-upgrade.sock"
```
atymchuk, line 54: Do we want to put the socket in `/srv`? How about storing it in `/tmp` directory?

@BupycHuk what is your opinion in view of our HA spec?

managed/utils/envvars/parser.go
```diff
@@ -98,6 +99,11 @@ func ParseEnvVars(envs []string) (envSettings *models.ChangeSettingsParams, errs
 			if err != nil {
 				err = fmt.Errorf("invalid value %q for environment variable %q", v, k)
 			}
+		case "DISABLE_LEGACY_UPDATES":
```
atymchuk, line 102: IMO this needs to be covered with a unit test, i.e. `parser_test.go`.

atymchuk (changes requested):

managed/services/server/server.go
BupycHuk: `/srv` directory is already mounted, so we can reuse it to set up the second container without re-running PMM Server container

BupycHuk (commented):
Haven't reviewed whole PR, it's what I found so far.

admin/services/update/update.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (s *Server) isValidLogPath(filePath string) error {
+	filename := path.Base(filePath)
+	match, err := path.Match(logsFileNamePattern, filename)
+	if err != nil {
+		return err
+	}
+
+	if !match {
+		return fmt.Errorf("invalid log file path provided")
```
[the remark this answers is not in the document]
michal-kralik: This has been added for security reasons.
The `logs_token` submitted for `pmm-server-upgrade` is a full path to a file. This code is meant to limit the filename which can be read.

api/updatepb/update.proto
michal-kralik: I opted for keeping this in sync with the legacy update API so there are minimal changes to the UI.

admin/commands/pmm/server/docker/deps.go
```diff
@@ -27,12 +27,12 @@ import (
 	"github.com/percona/pmm/admin/pkg/docker"
 )
 
-//go:generate ../../../../../bin/mockery -name=Functions -case=snake -inpkg -testonly
+//go:generate ../../../../../bin/mockery -name=functions -case=snake -inpkg -testonly
```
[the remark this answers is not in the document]
michal-kralik: Good catch

admin/commands/pmm/server/docker/upgrade.go
michal-kralik: `New` is not used when Kong calls this command. We need this for Kong to properly init logger.

admin/commands/pmm_server_upgrade/start/start.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+type startResult struct{}
+
+// Result is a command run result.
+func (r *startResult) Result() {}
+
+// String stringifies command result.
+func (r *startResult) String() string {
+	return "Exiting"
```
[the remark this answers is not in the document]
michal-kralik: Yes and no. I think it makes sense here because this is printed when the `start` command finishes which stops the daemon and exits
But there's still a lot of work to be done with the UI/UX in general. I'd like to address it all at the same time.

admin/services/update/update.go
```diff
@@ -0,0 +1,228 @@
...
+		dockerFn:           d,
+		dockerImage:        dockerImage,
+		gRPCMessageMaxSize: gRPCMessageMaxSize,
+		updateInProgress:   make(map[string]struct{}, 1),
+	}, nil
+}
+
+// StartUpdate starts PMM Server update.
+func (s *Server) StartUpdate(ctx context.Context, req *updatepb.StartUpdateRequest) (*updatepb.StartUpdateResponse, error) {
+	containerID := req.Hostname
```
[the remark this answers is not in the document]
michal-kralik: The value is provided by PMM Server. The server does not have a concept of a container ID, it only knows its hostname.
The hostname is the container ID in Docker, unless you override it.
The assignment thus makes sense. But I guess we could add detection based on hostname in case it's not the container ID. Not sure how many customers set a custom hostname though. I expect very little.

admin/services/update/update.go
```diff
@@ -0,0 +1,228 @@
...
+	logrus.Debugf("Inspecting container %s", containerID)
+	container, err := s.dockerFn.ContainerInspect(ctx, containerID)
+	if err != nil {
+		logrus.Errorf("Could not inspect container %s. Error: %v", containerID, err)
+		return nil, err
+	}
+
+	if !container.State.Running {
+		return nil, fmt.Errorf("container %s is not running", containerID)
+	}
```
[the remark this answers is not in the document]
michal-kralik: I think this is a good check.
You can be in a situation where you have multiple PMM Server containers in the system since you kept them after an upgrade, but they're stopped.
In these cases you may have accidentally specified the wrong container name. This error notifies you that you may have made a mistake.

In cases where you have just a single container and it's stopped for some reason, the command fails. A workaround is very simple though - you start the container and run the command again.

admin/services/status/status.go
michal-kralik: I suggest we stick with `Ready`.
I expect there's going to be more logic over time. If there was only `Available` and more logic is added, we need to introduce `Ready` and will need to change API calls. This approach seems easier to maintain in the future even though I agree, at this point it's technically more about availability than readiness.

admin/services/update/update.go
```diff
@@ -0,0 +1,228 @@
...
+
+	logFile, err := os.CreateTemp("", logsFileNamePattern)
```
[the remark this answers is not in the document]
BupycHuk: the log file name will be lost if the user restarts the browser.

admin/commands/pmm_server_upgrade/start/start.go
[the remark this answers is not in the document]
BupycHuk: shouldn't then the `start` command be renamed to `run` if it stops the daemon once it's finished?

admin/services/update/update.go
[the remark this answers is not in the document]
BupycHuk: But since it's API of `pmm-server-updater` it should be `containerID` here.

admin/services/status/status.go
BupycHuk: With any new logic coming here we can use your previous statement that it's better to check logs during the start of the upgrade.
I'm for making it `Ready`, but include check for docker, otherwise it's not ready.

admin/services/update/update.go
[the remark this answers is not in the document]
michal-kralik: If we rename it to `containerID`, PMM Server has no way to provide the ID of the container since it does not know the ID.
Maybe we keep it a `hostname` and I implement the extra logic as described above?

admin/services/status/status.go
michal-kralik: Ok. Renamed to `Available`

admin/services/update/update.go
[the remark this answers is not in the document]
michal-kralik: Is there a way to open the logs after Grafana is reloaded? I assumed not, but maybe I'm wrong.
The approach is identical to the previous update process in terms of generating the token.
The old process used a generated uuid.

https://github.com/percona/pmm/blob/309d5fd3c580c75b6f40fa3cf86f39e519940828/managed/services/server/server.go#L395

admin/commands/pmm_server_upgrade/start/start.go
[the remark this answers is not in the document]
michal-kralik: Happy to rename if `run` is more appropriate.
I'd prefer `start` or `daemon` but it's just a personal preference.

admin/services/update/update.go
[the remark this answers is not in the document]
BupycHuk: ok

api/updatepb/status.proto
```diff
@@ -0,0 +1,27 @@
...
+message AvailableRequest {}
+
+message AvailableResponse {}
+
+// Status service provides public status APIs for pmm-server-upgrade.
+service Status {
+  // Available checks if updater is available.
+  rpc Available(AvailableRequest) returns (AvailableResponse) {
+    option (google.api.http) = {
+      post: "/v1/Available"
```
artemgavrilov, line 19:
```suggestion
      post: "/v1/update/Status/Available"
```

managed/models/settings.go
```diff
@@ -52,7 +52,8 @@ type Settings struct {
 	PMMPublicAddress string `json:"pmm_public_address"`
 
 	Updates struct {
-		Disabled bool `json:"disabled"`
+		Disabled       bool `json:"disabled"`
+		LegacyDisabled bool `json:"legacy_disabled"`
```
artemgavrilov, line 56: Shouldn't we migrate existing data to match this?

api/updatepb/status.proto
michal-kralik: Generally I prefer for `ready`, `health`, `available`, etc. endpoints to be in root.
Any reason you'd prefer to nest it under `update/Status`?

managed/models/settings.go
michal-kralik: The zero value would be `false` which is correct.
Is there any settings specific migration we shall account for?

api/updatepb/status.proto
artemgavrilov: That's just rule that majority of existing endpoints follow. I guess better to follow same pattern everywhere.

P.S. checked sources, we have `/version/package/service/endpoint`, `/version/service/endpoint`, `/version/endpoint` patterns. Maybe it makes sense to discuss in slack with other devs. It's not clear for me what our rules regarding this.

artemgavrilov: BTW if to use `/v1/Available` then it's not clear what is available?

api/serverpb/server.proto
```diff
@@ -20,6 +20,13 @@ enum DistributionMethod {
   DO = 5;
 }
 
+// UpdateMethod defines the method for updating PMM Server.
+enum UpdateMethod {
+  INVALID = 0;
```
artemgavrilov, line 25: I don't know why, but for invalid options we have this format
```suggestion
  UPDATE_METHOD_INVALID = 0;
```

managed/models/settings.go
michal-kralik: The `disabled` flag is meant for disabling updates altogether while `legacy_disabled` is for disabling `pmm-update` only.
If `disabled: true`, then it's irrelevant how `legacy_disabled` is configured. It will be disabled.

I added comments for more clarity.

api/updatepb/status.proto
michal-kralik: The API is specifically for `pmm-server-upgrade`. The endpoint shall return the state of the app as a whole.
I'll check in Slack to see what the general guideline preference is.

atymchuk: To me the variant of `/version/endpoint` suffers from one major downside - all endpoints would end up in the root, providing no clue to which part of the app they refer.

For some of us the variant of `/v1/update/Status/Available`  may seem to be overly verbose. However, if we try to reduce it to just `/v1/update/Available` or `/v1/Status/Available`, it will suffer from the same problem as the shortest version above - lack of semantics and lack of origin (either package or service or both).

That said, I'm in favor of the long variant as the most semantic. Personally, the benefit is that I would have to refer less to the docs to understand what it does. 

I understand that, with a few exceptions, this is the convention we currently use.

api/updatepb/update.proto
```diff
@@ -0,0 +1,67 @@
...
+  // True when update has finished.
+  bool done = 3;
+}
+
+// Update service provides public APIs for pmm-server-upgrade.
+service Update {
+  // StartUpdate starts PMM Server update.
+  rpc StartUpdate(StartUpdateRequest) returns (StartUpdateResponse) {
+    option (google.api.http) = {
+      post: "/v1/Updates/Start"
```
atymchuk, line 48: I believe we need to follow an already established naming convention, which requires the second URL section to be lowercase. However good or bad it may be :)

Screenshot 2023-02-06 at 14 24 19

atymchuk, line 48:
```suggestion
      post: "/v1/update/Server/Start"
```

api/updatepb/update.proto
```diff
@@ -0,0 +1,67 @@
...
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "Start update"
+      description: "Starts PMM Server update."
+    };
+  }
+  // UpdateStatus returns PMM Server update status.
+  rpc UpdateStatus(UpdateStatusRequest) returns (UpdateStatusResponse) {
+    option (google.api.http) = {
+      post: "/v1/Updates/Status"
```
atymchuk, line 59:
```suggestion
      post: "/v1/update/Server/Status"
```

atymchuk (changes requested):

api/updatepb/status.proto
artemgavrilov: Gosh, I read method description twice and though that it checks that PMM server update are available. It it's just health check for `pmm-server-upgrade` service itself, then I think we should follow K8S naming rules here.
https://kubernetes.io/docs/reference/using-api/health-checks/#api-endpoints-for-health

Like we do it in PMM: https://github.com/percona/pmm/blob/0b9ef90cae43e04490a61b858073967ff0116067/api/serverpb/server.proto#L310

michal-kralik: I removed the JSON rpc altogether. It wasn't used either way. It's not needed for our services to communicate with each other.

atymchuk (approved):

admin/services/update/update.go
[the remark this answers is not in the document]
BupycHuk: API should be implemented based on server needs, not a client

admin/services/update/update.go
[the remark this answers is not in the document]
BupycHuk: but if you want to safely stop container and then update you won't be able to do that. Can we add a prompt instead of showing an error?

admin/commands/pmm_server_upgrade/start/start.go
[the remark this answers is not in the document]
BupycHuk: I think renaming to `run` makes sense, in the case of `start` or `daemon` we should just start the daemon in the background and return the result to the process which invoked this command.

admin/services/update/update.go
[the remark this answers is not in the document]
michal-kralik: Adding a prompt would not work in this case. The `pmm-server-upgrade` does not accept any stdin. It's basically a daemon.
We could add it to the cli though, if it's not already there.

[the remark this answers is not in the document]
BupycHuk: Got it, now it makes sense.
I think we should split PMM CLI and `pmm-server-upgrade` to make the code more obvious.
And we should extract exact upgrade logic to separate struct to be able to reuse it in PMM CLI and `pmm-server-upgrade`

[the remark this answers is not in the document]
michal-kralik: I believe that's what we're doing.
`pmm-server-upgrade` uses one command from PMM CLI which is a separate structure mapped as a command in cli.
https://github.com/percona/pmm/blob/2b18b80ee7288fabf2b6a8a9264a9906d91bcc32/admin/commands/pmm/server/docker/upgrade.go#L38-L48

It's initialized here
https://github.com/percona/pmm/blob/2b18b80ee7288fabf2b6a8a9264a9906d91bcc32/admin/services/update/update.go#L102-L105

Do you have in mind you recommend separating it even further so that it does not live in the command but is rather a completely separate structure?

Happy to split it but I believe it's going to add additional complexity without reducing maintenance complexity.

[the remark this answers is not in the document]
BupycHuk: it looks not obvious, to make it clear `dockerCmd` should be renamed to `dockerUpgradeCmd`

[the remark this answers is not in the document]
michal-kralik: The `dockerCmd` is related to all server docker commands such as install, upgrade, etc.
I see your point though. I renamed `New` to `NewUpgradeCommand` - makes sense?

BupycHuk (commented):
forgot to click "Submit review" button :)

api/serverpb/server.proto
```diff
@@ -73,9 +80,15 @@ message CheckUpdatesResponse {
   string latest_news_url = 4;
   // Last check time.
   google.protobuf.Timestamp last_check = 5;
+  // True if pmm-update can be used for an update.
+  bool pmm_update_available = 6;
+  // True if pmm-server-upgrade can be used for an update.
+  bool server_upgrade_available = 7;
```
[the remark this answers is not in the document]
michal-kralik: Yes. We can end up in a situation where `pmm-server-upgrade` is not running and `pmm-update` is either disabled or not available anymore.
In this situation it would be great if UI shows a message explaining how to deploy `pmm-server-upgrade` for example through a link to docs.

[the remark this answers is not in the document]
BupycHuk: ok, got it

api/serverpb/server.proto
```diff
@@ -83,7 +83,7 @@ message CheckUpdatesResponse {
   // True if pmm-update can be used for an update.
   bool pmm_update_available = 6;
```
BupycHuk, line 84: I would rename this field also to make it more clear.
maybe make them like `internal_pmm_update_available` and `external_pmm_update_available`?

BupycHuk (approved):

api/serverpb/server.proto
michal-kralik: `pmm_update_service_available`?
The word `service` implies to me it's not the act of an update, rather a service. At least for me this is helpful. Looking for feedback here though.

admin/commands/pmm/server/docker/upgrade.go
```diff
@@ -94,14 +108,27 @@ func (c *UpgradeCommand) RunCmdWithContext(ctx context.Context, globals *flags.G
...
+	// We wait a bit so the logs can be requested by pmm-managed before the container is stopped.
+	select {
+	case <-time.After(c.waitBeforeContainerStop):
+	case <-ctx.Done():
+		msg := "upgrade has been cancelled"
+		if err = ctx.Err(); err != nil {
+			msg = fmt.Sprintf(msg+". Error: %s", err)
+		}
+
+		return nil, fmt.Errorf(msg)
```
artemgavrilov, line 127:
```suggestion
		return nil, errors.New(msg)
```

artemgavrilov (approved):

api/serverpb/server.proto
BupycHuk: `pmm_update_available` sounds like a new update(version) for PMM is available which might be not correct.

PR #1515, opened 2022-12-19
PMM-11175 mongo restore logs

idoqo:
PMM-11175

This adds the ability to stream pbm logs during restore and retrieve them later using the restore ID.

Build: SUBMODULES-2962


PavelKhripkov:
Let's change to `JobID:  jobs[0].ID,` here to get the latest job, not the first one
https://github.com/percona/pmm/blob/8a7f0c9ea39e138abfeb4881f94895833aea1fdf/managed/services/management/backup/backups_service.go#L436

agent/runner/jobs/pbm_job_logger.go
```diff
@@ -27,30 +27,30 @@ import (
...
+	pbmRestoreJob pbmJob = "restore"
 )
 
-type pbmEventLogger struct {
-	dbURL       *url.URL
-	parentJobID string
-	logChunkID  uint32
+type pbmJobLogger struct {
+	dbURL      *url.URL
+	jobI       string
```
PavelKhripkov, line 39:
```suggestion
	jobID       string
```

agent/runner/jobs/pbm_job_logger.go
```diff
@@ -27,30 +27,30 @@ import (
...
 }
 
-func newPbmEventLogger(parentJobId string, mongoUrl *url.URL) *pbmEventLogger {
-	return &pbmEventLogger{
-		parentJobID: parentJobId,
-		logChunkID:  0,
-		dbURL:       mongoUrl,
+func newPbmJobLogger(parentJobId string, mongoUrl *url.URL) *pbmJobLogger {
+	return &pbmJobLogger{
+		jobI:       parentJobId,
```
PavelKhripkov, line 45:
```suggestion
		jobID:       jobID,
```

agent/runner/jobs/pbm_job_logger.go
```diff
@@ -27,30 +27,30 @@ import (
...
+	jobI       string
+	logChunkID uint32
 }
 
-func newPbmEventLogger(parentJobId string, mongoUrl *url.URL) *pbmEventLogger {
-	return &pbmEventLogger{
-		parentJobID: parentJobId,
-		logChunkID:  0,
-		dbURL:       mongoUrl,
+func newPbmJobLogger(parentJobId string, mongoUrl *url.URL) *pbmJobLogger {
```
PavelKhripkov, line 43:
```suggestion
func newPbmJobLogger(jobID string, mongoUrl *url.URL) *pbmJobLogger {
```

agent/runner/jobs/pbm_job_logger.go
```diff
@@ -27,30 +27,30 @@ import (
...
+		logChunkID: 0,
+		dbURL:      mongoUrl,
 	}
 }
 
-func (l *pbmEventLogger) sendLog(send Send, data string, done bool) {
+func (l *pbmJobLogger) sendLog(send Send, data string, done bool) {
 	send(&agentpb.JobProgress{
-		JobId:     l.parentJobID,
+		JobId:     l.jobI,
```
PavelKhripkov, line 53:
```suggestion
		JobID:     l.jobID,
```

managed/models/job_helpers.go
```diff
@@ -82,7 +82,6 @@ func FindJobs(q *reform.Querier, filters JobsFilter) ([]*Job, error) {
 		crossJoin = true
 		andConds = append(andConds, "value ->> 'restore_id' = "+q.Placeholder(idx))
 		args = append(args, filters.RestoreID)
-		idx++
```
PavelKhripkov: I think it's better to leave this increment, but comment it. So when we copy this part next time we don't forget about increment.

PavelKhripkov: I think it's better to leave this increment, but comment it. So when we copy this part next time we don't forget about increment.

managed/services/management/backup/backups_service.go
```diff
@@ -413,23 +413,35 @@ func (s *BackupsService) RemoveScheduledBackup(ctx context.Context, req *backupp
...
+	if req.GetRestoreId() != "" && req.GetBackupArtifactId() != "" {
+		return nil, status.Error(codes.InvalidArgument, "only one of artifact ID or restore ID should be set")
+	}
+	if req.GetBackupArtifactId() != "" {
+		jobsFilter.ArtifactID = req.GetBackupArtifactId()
+	} else if req.GetRestoreId() != "" {
+		jobsFilter.RestoreID = req.GetRestoreId()
+	} else {
+		return nil, status.Error(codes.InvalidArgument, "one of artifact ID or restore ID is required")
+	}
```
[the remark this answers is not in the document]
PavelKhripkov: I mean you don't need to use functions like `req.GetRestoreId()` and `req.GetBackupArtifactId()`. You can do like:
`switch jodID := req.JobId.(type)`

managed/services/management/backup/backups_service_test.go
```diff
@@ -457,9 +457,11 @@ func TestGetLogs(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		logs, err := backupSvc.GetLogs(ctx, &backuppb.GetLogsRequest{
-			ArtifactId: "artifact",
-			Offset:     tc.offset,
-			Limit:      tc.limit,
+			JobId: &backuppb.GetLogsRequest_BackupArtifactId{
+				BackupArtifactId: "artifact",
```
PavelKhripkov, line 461: Let's check this works for restore too.

agent/runner/jobs/pbm_job_logger.go
```diff
@@ -36,21 +36,21 @@ const (
 
 type pbmJobLogger struct {
 	dbURL      *url.URL
-	jobI       string
+	jobID      string
 	logChunkID uint32
 }
 
-func newPbmJobLogger(parentJobId string, mongoUrl *url.URL) *pbmJobLogger {
+func newPbmJobLogger(jobID string, mongoUrl *url.URL) *pbmJobLogger {
```
PavelKhripkov, line 43:
```suggestion
func newPbmJobLogger(jobID string, mongoURL *url.URL) *pbmJobLogger {
```

managed/services/management/backup/backups_service_test.go
```diff
@@ -408,66 +408,190 @@ func TestGetLogs(t *testing.T) {
...
+			{
+				offset: 1,
+				limit:  3,
+				expect: []uint32{1, 2, 3},
+			},
+			{
+				offset: 5,
+				expect: []uint32{},
+			},
+		}
```
PavelKhripkov, line 459: Looks like we have a lot of duplicated code here and can optimize tests.

PavelKhripkov (approved):

agent/runner/jobs/pbm_job_logger.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"github.com/percona/pmm/api/agentpb"
+)
+
+type pbmJob string
+
+const (
+	pbmBackupJob  pbmJob = "backup"
+	pbmRestoreJob pbmJob = "restore"
+)
```
artemgavrilov, line 35: You can use job types instead: https://github.com/percona/pmm/blob/d7fd4bc9fa0ff08923785903ca0db574e68c07aa/agent/runner/jobs/job.go#L29-L34

agent/runner/jobs/pbm_job_logger.go
```diff
@@ -0,0 +1,113 @@
...
+
+func newPbmJobLogger(jobID string, mongoURL *url.URL) *pbmJobLogger {
+	return &pbmJobLogger{
+		jobID:      jobID,
+		logChunkID: 0,
+		dbURL:      mongoURL,
+	}
+}
+
+func (l *pbmJobLogger) sendLog(send Send, data string, done bool) {
```
artemgavrilov, line 51: I think we can move `send` to constructor

agent/runner/jobs/pbm_job_logger.go
```diff
@@ -0,0 +1,113 @@
...
+			Logs: &agentpb.JobProgress_Logs{
+				ChunkId: atomic.AddUint32(&l.logChunkID, 1) - 1,
+				Data:    data,
+				Done:    done,
+			},
+		},
+	})
+}
+
+func (l *pbmJobLogger) streamLogs(ctx context.Context, send Send, jobType pbmJob, name string) error {
```
artemgavrilov, line 65: Same with `send` and `jobType`

api/managementpb/backup/backups.proto
```diff
@@ -206,12 +206,14 @@ message RemoveScheduledBackupResponse {}
...
-      string_not_empty: true
-    }
+    deprecated = true
   ];
   uint32 offset = 2;
   uint32 limit = 3;
+  oneof job_id {
+    string backup_artifact_id = 4;
+    string restore_id = 5;
+  };
```
artemgavrilov, line 216: That's really confusing. JobID is job id, how we can put there completely different IDs?
Also this breaks backward compatibility.

artemgavrilov (changes requested):

agent/runner/jobs/pbm_job_logger.go
idoqo: the `pbmJob` are specific events for pbm though. They are used as prefix when filtering logs (https://docs.percona.com/percona-backup-mongodb/reference/pbm-commands.html#pbm-logs)

agent/runner/jobs/pbm_job_logger.go
idoqo: Got it, I've moved `jobType` to constructor. `send` is currently passed to the `Run` method only, and not initialized when we call constructor though. We can refactor it but it leads to too many changes and I don't think it's worth it at the moment.

api/managementpb/backup/backups.proto
idoqo: I've changed it to `log_job_id` - does that help?

Re backward compatibility, we only deprecated the original `artifact_id`, so old clients should still work fine (but they won't be able to get restore logs, only backup logs).

artemgavrilov: > I've changed it to log_job_id - does that help?

Not really, this is container for two ids(artifact_id and restore_id), it has nothing to do with jobs. We don't expose job concept on API level at all yet. Once it will happen then we will add `job_id` to that endpoint and all other IDs will be deprecated.

For now I think we can just add plain `string restore_id = 4`, that should be good enough for now.

agent/runner/jobs/pbm_job_logger.go
artemgavrilov: Ok-doki. Let's think about it later.

managed/services/management/backup/backups_service.go
```diff
@@ -429,19 +429,12 @@ func (s *BackupsService) GetLogs(ctx context.Context, req *backuppb.GetLogsReque
 			models.MongoDBRestoreBackupJob,
 		},
 	}
-
 	if req.ArtifactId != "" {
-		// this allows us to maintain backward compatibility.
 		jobsFilter.ArtifactID = req.ArtifactId
+	} else if req.RestoreId != "" {
+		jobsFilter.RestoreID = req.RestoreId
```
artemgavrilov, line 435: What if both artifact ID and restore ID set? Or none of them?

artemgavrilov (approved):

managed/services/management/backup/backups_service.go
idoqo: If none is set, we throw an error (in the else case). If both are set, we default to only using the artifact ID - though I guess we can send an error in that case as well.

PR #1521, opened 2022-12-21
PMM-10913 server with external postgres server

rishat-ishbulatov:
[PMM-10913](https://jira.percona.com/browse/PMM-10913)

Build: [PMM-10913](https://github.com/Percona-Lab/pmm-submodules/pull/3185)

[Doc1](https://github.com/percona/pmm-doc/pull/959) [Doc2](https://github.com/percona/pmm-doc/pull/1049)


docker-compose.yml
```diff
@@ -144,6 +144,103 @@ services:
...
+      -c pg_stat_statements.track=all 
+      -c pg_stat_statements.save=off 
+      -c ssl=on 
+      -c ssl_ca_file=/certs/root.crt 
+      -c ssl_cert_file=/certs/server.crt 
+      -c ssl_key_file=/certs/server.key
+      -c hba_file=/conf/pg_hba.conf
+    networks:
+      - ${NETWORK:-default}
+  pmm-managed-server-pg:
```
BupycHuk, line 174: I think we don't need it, let's just migrate all required environment variables to the main one and comment

BupycHuk (approved):
Good job, Looks good to me

docker-compose.yml
rishat-ishbulatov: moved to main pmm profile with commented env fields

managed/models/database.go
```diff
@@ -893,7 +900,7 @@ func SetupDB(sqlDB *sql.DB, params *SetupDBParams) (*reform.DB, error) {
 			params.Logf("Creating database %s and role %s", databaseName, roleName)
 		}
 		// we use empty password/db and postgres user for creating database
-		db, err := OpenDB(params.Address, "", "postgres", "")
+		db, err := OpenDB(params.Address, "", "postgres", "", "", "")
```
michal-kralik, line 903: We shall use SSL here if it was provided by the user.

managed/services/supervisord/supervisord.go
```diff
@@ -491,6 +493,14 @@ func addAlertManagerParams(alertManagerURL string, templateParams map[string]int
 	return nil
 }
 
+// addPostgresParams adds pmm-server postgres database params to template config for grafana.
+func addPostgresParams(templateParams map[string]interface{}) {
+	templateParams["PostgresAddr"] = getValueFromENV("POSTGRES_ADDR", "127.0.0.1:5432")
+	templateParams["PostgresDBName"] = getValueFromENV("POSTGRES_DBNAME", "pmm-managed")
+	templateParams["PostgresDBUsername"] = getValueFromENV("POSTGRES_USERNAME", "pmm-managed")
+	templateParams["PostgresDBPassword"] = getValueFromENV("POSTGRES_DBPASSWORD", "pmm-managed")
+}
```
michal-kralik, line 502: Shall we store SSL here as well?

managed/testdata/pg/certs/pmm-managed.cnf
```diff
@@ -0,0 +1,10 @@
+[req]
```
michal-kralik, line 1: Can you please add instructions how to generate the certificates from the config files?
This will be helpful later when no one remembers :)

update/ansible/playbook/tasks/files/datasources.yml
```diff
@@ -16,12 +16,14 @@ datasources:
   orgId: 1
   type: postgres
   access: proxy
-  url: localhost:5432
-  user: postgres
-  database: pmm-managed
+  url: ${POSTGRES_ADDR}
+  user: ${POSTGRES_USERNAME}
+  database: ${POSTGRES_DBNAME}
```
michal-kralik, line 21: Do we need SSL certificates here?

managed/models/database.go
```diff
@@ -820,9 +820,16 @@ var databaseSchema = [][]string{
 // aleksi: Go's zero values and non-zero default values in database do play nicely together in INSERTs and UPDATEs.
 
 // OpenDB returns configured connection pool for PostgreSQL.
-func OpenDB(address, name, username, password string) (*sql.DB, error) {
+func OpenDB(address, name, username, password, sslKeyPath, sslCertPath string) (*sql.DB, error) {
```
michal-kralik, line 823: Since we are going to support connecting to external Postgres, is there a way for us to check if the minimum Postgres version requirement is met and show a friendly error message if it's not?

managed/cmd/pmm-managed/main.go
```diff
@@ -663,10 +663,28 @@ func main() {
 	versionServiceAPIURLF := kingpin.Flag("version-service-api-url", "Version Service API URL").
 		Default("https://check.percona.com/versions/v1").Envar("PERCONA_TEST_VERSION_SERVICE_URL").String()
 
-	postgresAddrF := kingpin.Flag("postgres-addr", "PostgreSQL address").Default("127.0.0.1:5432").String()
-	postgresDBNameF := kingpin.Flag("postgres-name", "PostgreSQL database name").Required().String()
-	postgresDBUsernameF := kingpin.Flag("postgres-username", "PostgreSQL database username").Default("pmm-managed").String()
-	postgresDBPasswordF := kingpin.Flag("postgres-password", "PostgreSQL database password").Default("pmm-managed").String()
+	postgresAddrF := kingpin.Flag("postgres-addr", "PostgreSQL address").
```
michal-kralik, line 666: What do you think about using Postgres connection string instead of separating individual pieces of information?

michal-kralik (changes requested):

managed/cmd/pmm-managed/main.go
rishat-ishbulatov: This can be discussed with leads. In my opinion, it's better not to.
1. With environment variables approach, there is no need to rewrite the logic of connection functions.
2. For the client this will be obvious, since we initially designate the minimum set of values necessary for us. The rest we configure ourselves for our needs.
3. Otherwise, we will need to have something like a black and white list of parameters that will work for us. And most likely for different versions of databases.
4. And if these parameters are not suitable, we will need to report it, otherwise the process for the end user will be non-obvious and most likely inconvenient.

michal-kralik: The connection string provides more flexibility in terms of specifying connection parameters, such as supporting [multiple hosts](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-MULTIPLE-HOSTS).

I think it's worth considering using this approach so we don't limit our users from using common connection options.

managed/models/database.go
rishat-ishbulatov: We use this place only for our internal needs. This place essentially does the following:
1. If a user with such parameters and a database has not been created, we create a user and a database under root rights.
2. We give the user rights to this database and subsequently use this user for interaction. I'm not sure if it would be a good idea to ask the client for root permissions in order to create the database and user we need. I suppose if the client does it himself and gives the rights necessary in his opinion, it will be better.

docker-compose.yml
```diff
@@ -148,6 +155,34 @@ services:
...
+      postgres 
+      -c shared_preload_libraries=pg_stat_statements 
+      -c pg_stat_statements.max=10000 
+      -c pg_stat_statements.track=all 
+      -c pg_stat_statements.save=off 
+      -c ssl=on 
+      -c ssl_ca_file=/certs/root.crt 
+      -c ssl_cert_file=/certs/server.crt 
+      -c ssl_key_file=/certs/server.key
+      -c hba_file=/conf/pg_hba.conf
```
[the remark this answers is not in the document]
rishat-ishbulatov: fsync off added

Makefile.devcontainer
```diff
@@ -33,6 +36,9 @@ run-vmproxy: release-vmproxy
 	cp $(PMM_RELEASE_PATH)/vmproxy /usr/sbin/vmproxy
 	supervisorctl start vmproxy
 
+run-update: release-dev-update      ## Replace pmm-update from build
+	cp $(PMM_RELEASE_PATH)/pmm-update /usr/sbin/pmm-update
+	pmm-update -run-playbook -playbook /root/go/src/github.com/percona/pmm/update/ansible/playbook/tasks/update.yml
```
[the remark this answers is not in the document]
rishat-ishbulatov: update lines removed

managed/models/database.go
michal-kralik: Yes, let's do this please. With external databases we cannot expect to have `postgres` user with no password available.

rishat-ishbulatov: error message added

managed/testdata/pg/certs/pmm-managed.cnf
rishat-ishbulatov: makefile with instructions added

managed/cmd/pmm-managed/main.go
BupycHuk: we can't use just the connection string, because grafana doesn't support connection strings, so we will have to add additional logic to parse the connection string in ansible.

managed/models/database.go
BupycHuk: let's make current version(PG 14) as minimum

managed/cmd/pmm-managed/main.go
rishat-ishbulatov: then what about services that do not support multiple hosts in this case?

rishat-ishbulatov: my comment is outdated. Yes, so as you say that grafana doesn't support connection string...  I also meant it.

update/ansible/playbook/tasks/files/datasources.yml
rishat-ishbulatov: certificates fields added

managed/models/database.go
rishat-ishbulatov: PG version 14 check added

managed/services/supervisord/supervisord.go
rishat-ishbulatov: suggested fields added

managed/models/database.go
```diff
@@ -893,6 +894,10 @@ func SetupDB(sqlDB *sql.DB, params *SetupDBParams) (*reform.DB, error) {
 
 	if pErr, ok := errDB.(*pq.Error); ok && pErr.Code == "28000" {
 		// invalid_authorization_specification	(see https://www.postgresql.org/docs/current/errcodes-appendix.html)
+		if host, _, _ := net.SplitHostPort(params.Address); host != "127.0.0.1" {
```
michal-kralik, line 897: Is there a better way to do this?
This is non-deterministic. The fact the host is `127.0.0.1` does not mean we do or don't have access to `root`.

managed/cmd/pmm-managed/main.go
```diff
@@ -743,6 +744,10 @@ func main() {
 	}
 	defer sqlDB.Close() //nolint:errcheck
 
+	if postgresVersion, err := ver.GetPostgreSQLVersion(ctx, sqlDB); err != nil || postgresVersion.Float() < 14 {
```
michal-kralik, line 747: Let's make the minimum version `14` a constant and use the constant instead of the actual number.
This will help us change the version in the future more easily.

managed/cmd/pmm-managed/main.go
```diff
@@ -743,6 +744,10 @@ func main() {
...
+		l.Panicf("Unsupported PMM-Server PostgreSQL server version: %s. Consider using version 14 or newer", postgresVersion)
```
michal-kralik, line 748: Instead of `Consider ...`, let's use

```
l.Panicf("Unsupported PMM Server PostgreSQL server version: %s. Please upgrade to version 14 or newer", postgresVersion)
```

michal-kralik (changes requested):
Looks good. Just a couple minor changes.

managed/cmd/pmm-managed/main.go
rishat-ishbulatov: constant added

managed/cmd/pmm-managed/main.go
rishat-ishbulatov: after discussion in slack, it was decided to move to migrateDB and output as an error

managed/models/database.go
rishat-ishbulatov: generalized error added

rishat-ishbulatov: after discussion in slack, it was decided to add user friendly message on auth fail and attempt to create user and db.

managed/models/database.go
```diff
@@ -869,81 +884,115 @@ type SetupDBParams struct {
...
 	if params.Logf != nil {
 		logger = reform.NewPrintfLogger(params.Logf)
 	}
+
 	db := reform.NewDB(sqlDB, postgresql.Dialect, logger)
+	errCV := checkVersion(ctx, db)
+	if pErr, ok := errCV.(*pq.Error); ok && pErr.Code == "28000" {
+		// invalid_authorization_specification	(see https://www.postgresql.org/docs/current/errcodes-appendix.html)
+		if err := initWithRoot(params); err != nil {
+			return nil, errors.Wrapf(err, "couldn't connect do database with provided credentials: %s; tried to create user and database", errCV)
```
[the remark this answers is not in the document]
michal-kralik: Let's do one more thing.
In the case the user-provided credentials are incorrect, we are hiding the actual error message from line. We shall expose it.

Can you double-check what we print out in this case? I think it's going to be an error from `initWithRoot` instead of the `errCV`.

michal-kralik (changes requested):

managed/models/database.go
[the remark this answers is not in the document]
rishat-ishbulatov: `WARN[2023-01-17T11:16:15.559+00:00] Failed to migrate database: couldn't connect to database with provided credentials. Tried to create user and database. Error: pq: pg_hba.conf rejects connection for host "172.30.0.3", user "pmm-managed", database "pmm-managed", no encryption: pq: no pg_hba.conf entry for host "172.30.0.3", user "postgres", database "postgres", no encryption.  component=migration`

michal-kralik (approved):

agent/utils/version/postgresql.go
```diff
@@ -30,15 +34,45 @@ func ParsePostgreSQLVersion(v string) string {
...
-	switch len(parts) {
-	case 1: // major only
-		return parts[0]
-	case 2: // major and patch
-		return parts[0]
-	case 3: // major, minor, and patch
-		return parts[0] + "." + parts[1]
-	default:
-		return ""
+	return m[1]
```
BupycHuk, line 37: this change breaks test on line https://github.com/percona/pmm/pull/1521/files#diff-7cc90b7313706bd058a0be08fa6588b152d8ddde9bf9f4229c16dddf59ebbdd9R30

BupycHuk: it might break logic on pmm-agent side

managed/models/database.go
```diff
@@ -29,10 +30,16 @@ import (
 	"google.golang.org/grpc/status"
 	"gopkg.in/reform.v1"
 	"gopkg.in/reform.v1/dialects/postgresql"
+
+	"github.com/percona/pmm/agent/utils/version"
```
BupycHuk, line 34: don't import `agent` packages into `managed`

BupycHuk: if you would like to re-use something, extract it from there

docker-compose.yml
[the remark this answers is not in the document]
BupycHuk: it should be added to documentation

managed/models/database.go
```diff
@@ -869,81 +884,115 @@ type SetupDBParams struct {
...
+
 	db := reform.NewDB(sqlDB, postgresql.Dialect, logger)
+	errCV := checkVersion(ctx, db)
+	if pErr, ok := errCV.(*pq.Error); ok && pErr.Code == "28000" {
+		// invalid_authorization_specification	(see https://www.postgresql.org/docs/current/errcodes-appendix.html)
+		if err := initWithRoot(params); err != nil {
+			return nil, errors.Wrapf(err, "couldn't connect to database with provided credentials. Tried to create user and database. Error: %s", errCV)
+		}
+		errCV = checkVersion(ctx, db)
+	}
```
BupycHuk, line 902: 👍🏼

managed/services/supervisord/supervisord.go
```diff
@@ -491,6 +493,16 @@ func addAlertManagerParams(alertManagerURL string, templateParams map[string]int
...
 }
 
+// addPostgresParams adds pmm-server postgres database params to template config for grafana.
+func addPostgresParams(templateParams map[string]interface{}) {
+	templateParams["PostgresAddr"] = getValueFromENV("POSTGRES_ADDR", "127.0.0.1:5432")
+	templateParams["PostgresDBName"] = getValueFromENV("POSTGRES_DBNAME", "pmm-managed")
+	templateParams["PostgresDBUsername"] = getValueFromENV("POSTGRES_USERNAME", "pmm-managed")
+	templateParams["PostgresDBPassword"] = getValueFromENV("POSTGRES_DBPASSWORD", "pmm-managed")
+	templateParams["PostgresSSLKeyPath"] = getValueFromENV("POSTGRES_SSL_KEY_PATH", "")
+	templateParams["PostgresSSLCertPath"] = getValueFromENV("POSTGRES_SSL_CERT_PATH", "")
```
BupycHuk, line 503: I thought that we discussed getting these values from the kingpin.

BupycHuk (changes requested):

agent/utils/version/postgresql.go
rishat-ishbulatov: move back to previous version

managed/models/database.go
rishat-ishbulatov: managed model for postgres versioner added

managed/services/supervisord/supervisord.go
rishat-ishbulatov: supervisord with grafana params from kingpin done

managed/models/database.go
```diff
@@ -820,15 +825,33 @@ var databaseSchema = [][]string{
...
+	default:
+		q.Set("sslmode", "disable")
+		addSSL = false
+	}
+
+	if addSSL {
+		q.Set("sslrootcert", params.SSLCAPath)
+		q.Set("sslcert", params.SSLCertPath)
+		q.Set("sslkey", params.SSLKeyPath)
+	}
```
BupycHuk, line 848:
```suggestion
	```suggestion
        sslMode = params.SSLMode
        if sslMode == "" {
            sslMode = "disable"
        }
        q.Set("sslmode", sslMode)

	if sslMode != "disable" {
		q.Set("sslrootcert", params.SSLCAPath)
		q.Set("sslcert", params.SSLCertPath)
		q.Set("sslkey", params.SSLKeyPath)
	}
```

BupycHuk (approved):

managed/services/supervisord/supervisord.go
```diff
@@ -778,6 +794,14 @@ command =
 environment=GF_AUTH_SIGNOUT_REDIRECT_URL="https://{{ .IssuerDomain }}/login/signout?fromURI=https://{{ .PMMServerAddress }}/graph/login"
         {{- end}}
 environment =
+    POSTGRES_ADDR="{{ .PostgresAddr }}",
```
atymchuk, line 797: Can `supervisord` handle multiple `environment` statements? I doubt it can. At the very least, the [documentation](http://supervisord.org/subprocess.html#subprocess-environment) doesn't say so. 

I'm afraid it could cause trouble even if it appears to work. 

I see it was there prior to this PR, but we'd better address it now.

atymchuk (changes requested):

managed/services/supervisord/supervisord.go
rishat-ishbulatov: In fact, the documentation doesn't say otherwise. There are just a few variables involved. Which is reflected in the example of how to install them.
```set by a shell invocation for a particular user, you must do it explicitly within the environment= program config option. An example of setting these environment variables is as below.

[program:apache2]
command=/home/chrism/bin/httpd -c "ErrorLog /dev/stdout" -DFOREGROUND
user=chrism
environment=HOME="/home/chrism",USER="chrism"
```

rishat-ishbulatov: [environment](http://supervisord.org/configuration.html?highlight=configuration)
    A list of key/value pairs in the form KEY="val",KEY2="val2" that will be placed in the environment of all child processes. This does not change the environment of supervisord itself. This option can include the value %(here)s, which expands to the directory in which the supervisord configuration file was found. Values containing non-alphanumeric characters should be quoted (e.g. KEY="val:123",KEY2="val,456"). Otherwise, quoting the values is optional but recommended. To escape percent characters, simply use two. (e.g. URI="/first%%20name") Note that subprocesses will inherit the environment variables of the shell used to start supervisord except for the ones overridden here and within the program’s environment option. See [Subprocess Environment](http://supervisord.org/subprocess.html#subprocess-environment).
    Default: no values
    Required: No.
    Introduced: 3.0

atymchuk: @rishat-ishbulatov Maybe I wan't very clear. It's not about the ability to use multiple KEY/VALUE pairs in one `environment`, it's about multiple environment statements. I see two: one on line 794, the other on line 796. 

My fear is that the last statement would win. WDYT?

rishat-ishbulatov: redundant config line removed

atymchuk (approved):

managed/services/supervisord/supervisord.go
```diff
@@ -436,6 +438,8 @@ func (s *Service) marshalConfig(tmpl *template.Template, settings *models.Settin
 		"ClickhouseBlockSize":      clickhouseBlockSize,
 	}
 
+	s.addPostgresParams(templateParams)
```
artemgavrilov, line 441: To be honest I'd prefer to not have this func and fill templateParams right here, same as we do for CH right above.

managed/models/postgresql_version_test.go
```diff
@@ -0,0 +1,102 @@
...
+			wantVersion: PostgreSQLVersion{},
+			wantError:   true,
+		},
+	}
+
+	column := []string{"version"}
+
+	//nolint:paralleltest
+	for i := range testCases {
+		tc := testCases[i]
```
artemgavrilov, line 85: This is unusual way to use range loops. We are doing it like this in most cases:
```suggestion
	for _, tc := range testCases {
		tc := tc
```

managed/models/postgresql_version_test.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Run(tc.name, func(t *testing.T) {
```
artemgavrilov, line 86: Can we run these tests in parallel?

managed/models/postgresql_version_test.go
```diff
@@ -0,0 +1,102 @@
...
+	ctx := context.Background()
+
+	type testingCase struct {
+		name        string
+		mockedData  []string
+		wantVersion PostgreSQLVersion
+		wantError   bool
+	}
+
+	testCases := []testingCase{
```
artemgavrilov, line 46:
```suggestion
	var testCases := []struct {
		name        string
		mockedData  []string
		wantVersion PostgreSQLVersion
		wantError   bool
	}{
```

managed/models/postgresql_version_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"gopkg.in/reform.v1"
+	"gopkg.in/reform.v1/dialects/postgresql"
+)
+
+func TestGetPostgreSQLVersion(t *testing.T) {
+	sqlDB, mock, err := sqlmock.New()
+	if err != nil {
+		t.Log("error creating mock database")
+		return
+	}
```
artemgavrilov, line 33:
```suggestion
	require.NoError(t, err)
```

managed/models/grafana.go
```diff
@@ -0,0 +1,36 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package models
+
+// GrafanaParams - defines flags and settings for grafana.
+type GrafanaParams struct {
```
artemgavrilov, line 19: It looks more like PG params. Is it correct naming?

managed/services/supervisord/supervisord.go
rishat-ishbulatov: same with me. Only supervisor refactoring goes beyond this task. PR is too big already.

managed/models/postgresql_version_test.go
rishat-ishbulatov: Actually most cases it's a better way to range loops according to [that](https://www.youtube.com/watch?v=kClQ7rM5uBY&t=3113s) golang tutorial.

managed/models/grafana.go
rishat-ishbulatov: these options are used to configure graphana datasource to use postgres. I suppose it is worth naming the parameters according to the place of their application and not about what they are. A matter of taste. It's not for me to make these decisions. If worth renaming ok for me.

managed/models/postgresql_version_test.go
rishat-ishbulatov: this place redone

managed/models/postgresql_version_test.go
rishat-ishbulatov: redundant log removed

managed/models/postgresql_version_test.go
rishat-ishbulatov: redone to parallel

managed/models/postgresql_version_test.go
artemgavrilov: This may be a nice optimization for performance critical functions. But this is test)

artemgavrilov: Also `testCases` is a slice, not array

managed/models/grafana.go
rishat-ishbulatov: grafana params renamed to pg params

docker-compose.yml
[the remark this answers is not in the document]
rishat-ishbulatov: done

managed/services/supervisord/supervisord.go
```diff
@@ -777,14 +794,22 @@ command =
         cfg:default.auth.generic_oauth.auth_url="{{ .PerconaSSODetails.IssuerURL }}/authorize"
         cfg:default.auth.generic_oauth.token_url="{{ .PerconaSSODetails.IssuerURL }}/token"
         cfg:default.auth.generic_oauth.api_url="{{ .PerconaSSODetails.IssuerURL }}/userinfo"
-		cfg:default.auth.generic_oauth.role_attribute_path="(contains(portal_admin_orgs[*], '{{ .PerconaSSODetails.OrganizationID }}') || contains(pmm_demo_ids[*], '{{ .PMMServerID }}')) && 'Admin' || 'Viewer'"
-		cfg:default.auth.generic_oauth.use_pkce="true"
-
-environment=GF_AUTH_SIGNOUT_REDIRECT_URL="https://{{ .IssuerDomain }}/login/signout?fromURI=https://{{ .PMMServerAddress }}/graph/login"
+        cfg:default.auth.generic_oauth.role_attribute_path="contains(portal_admin_orgs[*], '{{ .PerconaSSODetails.OrganizationID }}') && 'Admin' || 'Viewer'"
```
BupycHuk, line 797: @rishat-ishbulatov looks like you removed changes made by another developer

rishat-ishbulatov: actually I did that change 2 month ago and Oksana 1 month ago)

BupycHuk: something went wrong during conflict resolution

PR #1529, opened 2022-12-23
PMM-11023 Move OLM from dbaas-controller to pmm

percona-csalguero:
PMM-11023 Move OLM from dbaas-controller to pmm.

- Moved code from dbaas-controller to pmm-managed.
- Deprecated the use of kubectl in favor of Kubernetes go-client.
- Fixed timeout when generating Grafana API keys.
- Fixed error because of an empy env-var to set the pmm client image.


managed/services/management/dbaas/kubernetes_server.go
```diff
@@ -270,37 +272,8 @@ func replaceAWSAuthIfPresent(kubeconfig string, keyID, key string) (string, erro
 	return string(c), err
 }
 
-// TODO: @Carlos do we still need it?
```
gen1us2k, line 273: No. We do not need it I think

managed/services/management/dbaas/kubernetes_server.go
```diff
@@ -341,159 +316,129 @@ func (k kubernetesServer) RegisterKubernetesCluster(ctx context.Context, req *db
...
 			}
+			k.l.Errorf("couldn't start monitoring of the kubernetes cluster: %s", err)
+			return nil, errors.Wrap(err, "couldn't start monitoring of the kubernetes cluster")
 		}
+	}
 
-		if clusterInfo.Operators == nil || clusterInfo.Operators.OlmOperatorVersion == "" {
-			operator := "victoriametrics-operator"
+	// Install the operators in the background.
+	go k.installDefaultOperators(operatorsToInstall)
```
gen1us2k, line 370: We may have a situation when we run StartMonitoring while VM operator is not installed, don't we?

gen1us2k (commented):
That's weird that PR has changes from different PRs I suppose.

managed/services/management/dbaas/kubernetes_server.go
percona-csalguero: You are right. Thanks.

gen1us2k (approved):

PR #1531, opened 2022-12-24
PMM-7 Fix linters CI configuration; enable gosec, depguard, stylecheck and fix their warnings

artemgavrilov:
PMM-7

Build: SUBMODULES-0


.github/workflows/common.yml
```diff
@@ -85,16 +85,16 @@ jobs:
...
         with:
           github_token: ${{ secrets.ROBOT_TOKEN || secrets.GITHUB_TOKEN }}
           go_version: ${{ env.GO_VERSION }}
           reporter: github-pr-review
           fail_on_error: true
           cache: false
-          golangci_lint_flags: "-c=.golangci-required.yml"
-          golangci_lint_version: v1.49.0
+          golangci_lint_flags: "-c=.golangci.yml"
+          golangci_lint_version: v1.50.1 # Version should match specified in Makefile
```
artemgavrilov, line 97: @oksana-grishchenko Missed thing)

managed/services/management/dbaas/kubernetes_server.go
```diff
@@ -271,22 +270,23 @@ func replaceAWSAuthIfPresent(kubeconfig string, keyID, key string) (string, erro
 	return string(c), err
 }
 
-func installOLMOperator(ctx context.Context, client dbaasClient, kubeconfig, version string) error {
```
artemgavrilov, line 274: @cfsalguero  it's unused code, do we need it?

.golangci.yml
```diff
@@ -128,7 +126,6 @@ linters:
     - forbidigo #temporary, needs to be processed
     - errcheck #temporary, needs to be processed
     - dupl #temporary, needs to be processed
-    - depguard #temporary, needs to be processed
     - containedctx #temporary, needs to be processed
```
atymchuk, line 129: I suggest we remove the repeated comment `#temporary, needs to be processed`, it became irrelevant.

managed/services/management/dbaas/kubernetes_server.go
atymchuk: If unused, let's remove - we can add it back when we actually need it.

atymchuk (approved):

.golangci.yml
artemgavrilov: These linters were disalbled when @oksana-grishchenko switched this repo to single golangci-lint configuration. This should not leave like this, we should review each linter one by one and decide do we want to enable it or disable it completely. This PR is a first step.

This comments highlight that these linters disabled temporary, better to keep them for now.

atymchuk: Yes, I guessed it :) How about one comment above them all?

gen1us2k (approved):

.golangci.yml
BupycHuk: maybe we shouldn't enable all by default and include into configuration only ones we want to enable, what do you think?

BupycHuk (approved):
Great job

.golangci.yml
atymchuk: IMO it allows us to discover what we can use, actually a great idea to turn them on all and then disable the ones we don't want 👍

oksana-grishchenko: As we agreed on the BE meetings, the initial idea was to have more strict linter config in CI - enable all linters by default and disable only the particular linters. The linter step is required in CI, which means if we enable all linters now, the CI will fail with tons of linter errors. 
So as I see this PR enables several new linters, which moves the repo forward to a more reliable and maintainable state, which is great 👍

oksana-grishchenko (approved):

PR #1536, opened 2022-12-27
PMM-11116 Push agent status on reconnection

ShashankSinha252:
[PMM-11116](https://jira.percona.com/browse/PMM-11116) : Missing status update from PMM agent to PMM server in case of server restart.

Build: SUBMODULES-0


BupycHuk (changes requested):
if channel `changes` is full it leads to deadlock.

BupycHuk (commented):
we are on the right way. we also should cleanup agent statuses channel to not resend the same statuses again.

agent/agents/supervisor/supervisor.go
```diff
@@ -186,6 +186,11 @@ func (s *Supervisor) AgentLogByID(id string) ([]string, uint) {
 	return nil, 0
 }
 
+// ClearChanges returns channel with Agent's state changes.
+func (s *Supervisor) ClearChanges() {
+	s.changes = make(chan *agentpb.StateChangedRequest, 100)
```
BupycHuk, line 191: it should be locked to be safe from simultaneous access

percona-csalguero: The read below in the `Changes` func should also have a lock.

percona-csalguero (commented):
LGTM. Just need to add locks.

agent/agents/supervisor/supervisor.go
ShashankSinha252: Done

BupycHuk: don't forget to close the old channel if it's being used by someone.

BupycHuk: the channel is used here https://github.com/percona/pmm/pull/1536/files#diff-e84ef06ecfdbc9bd2dfd4ca3aad8a21039e5c5c1766386735710c724ece67a05R446 as well

BupycHuk (changes requested):

agent/client/client.go
```diff
@@ -220,6 +222,40 @@ func (c *Client) Run(ctx context.Context) error {
 	return nil
 }
 
+// GetAgentsStatus sends status of running agents to server
+func (c *Client) GetAgentsStatus() {
+	// Drain all existing messages
+L:
```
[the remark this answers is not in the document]
ShashankSinha252: Done

agent/client/client.go
```diff
@@ -220,6 +222,40 @@ func (c *Client) Run(ctx context.Context) error {
...
+	for {
+		select {
+		case _, ok := <-c.supervisor.Changes():
+			if !ok {
+				break loop
+			}
+		default:
+			break loop
+		}
+	}
```
BupycHuk, line 238: I would extract this logic to a separate method, cleaning the `changes` channel seems out of responsibility of `GetAgentsStatus` method and instead of `break` we will be able to use just `return`. I would even move it inside `supervisor` and call it before calling `GetAgentsStatus` method.
BTW `GetAgentsStatus` seems like incorrect naming, maybe `SendActualStatuses`?

BupycHuk: don't see changes

BupycHuk (changes requested):
waiting for the new commit

agent/agents/supervisor/supervisor.go
```diff
@@ -186,6 +186,20 @@ func (s *Supervisor) AgentLogByID(id string) ([]string, uint) {
 	return nil, 0
 }
 
+// ClearChangesChannel drains state change channel
+func (s *Supervisor) ClearChangesChannel() {
+	for {
+		select {
+		case _, ok := <-s.changes:
```
JiriCtvrtka, line 193: Maybe we can even use just one switch, since there is no other case.

BupycHuk (approved):

PR #1542, opened 2023-01-04
PMM-10684 Clickhouse v2.

JiriCtvrtka:
PMM-10684

Build: https://github.com/Percona-Lab/pmm-submodules/pull/3596


managed/services/supervisord/supervisord.go
```diff
@@ -425,8 +425,6 @@ func (s *Service) marshalConfig(tmpl *template.Template, settings *models.Settin
 	clickhouseDatabase := getValueFromENV("PERCONA_TEST_PMM_CLICKHOUSE_DATABASE", defaultClickhouseDatabase)
 	clickhouseAddr := getValueFromENV("PERCONA_TEST_PMM_CLICKHOUSE_ADDR", defaultClickhouseAddr)
 	clickhouseDataSourceAddr := getValueFromENV("PERCONA_TEST_PMM_CLICKHOUSE_DATASOURCE_ADDR", defaultClickhouseDataSourceAddr)
-	clickhousePoolSize := getValueFromENV("PERCONA_TEST_PMM_CLICKHOUSE_POOL_SIZE", "")
```
ademidoff, line 428: Did they remove these two params from the client?

ademidoff (approved):
👍

managed/services/supervisord/supervisord.go
ademidoff: OK, I see the commit message - it's no longer supported )

JiriCtvrtka: In v2 pool size and block size changed. It is also recommended to set them agains exact DB. 

See: https://clickhouse.com/docs/en/integrations/go
MaxIdleConns should be alternative for pool size
BlockBufferSize for block size

But our block size default was 10 000 and BlockBufferSize did not accept such number, only lower ones, like 100. 

`time="2024-03-22T07:35:19.278+00:00" level=info msg="DSN: clickhouse://127.0.0.1:9000/pmm?block_buffer_size=10000&max_idle_conns=2" component=main
stdlog: Connection: strconv.ParseUint: parsing "10000": value out of range`

So seems it is slightly different. MaxIdleConns should behave same like pool size, but since both of them are with prefix "PERCONA_TEST" I decided to remove them.

BupycHuk (commented):
looks good to me, but I think it's better to do it in v3 branch since we are removing some env variables.

qan-api2/db_test.go
```diff
@@ -113,13 +113,13 @@ func TestCreateDbIfNotExists(t *testing.T) {
...
+		dsn = strings.Replace(dsn, "/pmm_test", "/pmm_created_db", 1)
 		if !ok {
-			dsn = "clickhouse://127.0.0.1:19000?database=pmm_created_db"
+			dsn = "clickhouse://127.0.0.1:19000/pmm_created_db"
 		}
 
 		db := createDB(dsn)
 
-		require.Equal(t, db, nil, "Check connection after we create database")
+		require.Equal(t, nil, db, "Check connection after we create database")
```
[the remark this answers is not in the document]
JiriCtvrtka: Changed.

managed/services/supervisord/supervisord.go
JiriCtvrtka: My fault. Thanks for notice that.

ademidoff: Don't we treat variables prefixed with `PERCONA_TEST_` as non-public, internal ones?

I'm referring to https://github.com/percona/pmm/blob/70d5afdf8744bda5ab0563e868dfc2c063cd0aba/managed/CONTRIBUTING.md?plain=1#L54

ademidoff: I mean they should be used in non-prod environments, i.e. test environments.

BupycHuk: You are right, seems like we can merge it to PMM 2 as well

BupycHuk (approved):

managed/services/supervisord/supervisord.go
BupycHuk: let's do it for v3 now, later we can backport to v2

docs/process/v2_to_v3_environment_variables.md
```diff
@@ -26,11 +26,9 @@ Below is a list of affected variables and their new names.
 | `PERCONA_TEST_CHECKS_PUBLIC_KEY`              |                                            | Removed in PMM v3, use `PMM_DEV_PERCONA_PLATFORM_PUBLIC_KEY` |
 | `PERCONA_TEST_NICER_API`                      | `PMM_NICER_API`                            |                                                              |
 | `PERCONA_TEST_PMM_CLICKHOUSE_ADDR`            | `PMM_CLICKHOUSE_ADDR`                      |                                                              |
-| `PERCONA_TEST_PMM_CLICKHOUSE_BLOCK_SIZE`      | `PMM_CLICKHOUSE_BLOCK_SIZE`                |                                                              |
```
idoqo, line 29: I think that since the variables still exist in v2, it's better we keep them here and mention that they're now removed in v3 in the comments column.

idoqo (approved):

docs/process/v2_to_v3_environment_variables.md
JiriCtvrtka: Added back with description.

PR #1553, opened 2023-01-10
PMM-11362: support role replacement during removal

michal-kralik:
PMM-11362

Build: SUBMODULES-2994


api/managementpb/role/role.proto
```diff
@@ -38,12 +38,21 @@ message UpdateRoleRequest {
 }
 message UpdateRoleResponse {}
 
+enum ReplacementRoleMode {
+  // LAST_ROLE replaces a role only in case it's the last role to be removed and no other roles would be left for the user.
+  LAST_ROLE = 0;
+  // ALWAYS replaces a role for all users.
+  ALWAYS = 1;
+}
+
```
BupycHuk, line 47: it seems like it goes against what's written in the ticket. I see that we have only one option like LAST_ROLE in the ticket description.

michal-kralik: yes, this is on top of the ticket.
I expect at some point this behavior will be requested. It made sense to implement it while I was on it rather than later.

BupycHuk: let's remove it for now, it might be not requested at all.

BupycHuk (approved):

gen1us2k (approved):

PR #1556, opened 2023-01-11
PMM-11371: support for sockets during agent setup

michal-kralik:
PMM-11371
PMM-3996

Build: SUBMODULES-3009


admin/cmd/bootstrap.go
```diff
@@ -211,3 +222,81 @@ func processFinalError(err error, isJSON bool) {
...
+
+		err := checkConnection(ctx, "tcp", net.JoinHostPort(agentlocal.Localhost, strconv.Itoa(int(localPort))))
+		if err == nil {
+			port = localPort
+		}
+
+		portChan <- port
+	}()
+
+	if sock := <-socketChan; sock != "" {
```
idoqo, line 276: I assumed we want to use either socket or a port. It looks like this will block until it gets a message on `socketChan`? This means the port check doesn't happen until we have detected a socket, correct?

agent/config/config.go
```diff
@@ -311,6 +216,157 @@ func get(args []string, cfg *Config, l *logrus.Entry) (configFileF string, err e
...
+}
+
+func configureListenDefaults(cfg *Config) {
+	if cfg.ListenSocket == "" {
+		switch {
+		case cfg.ListenAddress == "" && cfg.ListenPort == 0:
+			cfg.ListenSocket = filepath.Join(cfg.Paths.PathsBase, defaultListenSocketJSONFilename)
+			if cfg.ListenSocketGRPC == "" {
+				cfg.ListenSocketGRPC = filepath.Join(cfg.Paths.PathsBase, defaultListenSocketGRPCFilename)
+			}
```
idoqo, line 272: this looks like the same code in `configureListenGRPCDefaults()`, is it a duplicate? or it has to be here too for some reason?

admin/cli/flags/global.go
```diff
@@ -25,13 +25,17 @@ import (
...
+const SocketPath = "/usr/local/percona/pmm2/pmm-agent.sock"
+
 // GlobalFlags stores flags global to all commands.
 type GlobalFlags struct {
 	ServerURL               *url.URL    `placeholder:"SERVER-URL" help:"PMM Server URL in https://username:password@pmm-server-host/ format"`
 	SkipTLSCertificateCheck bool        `name:"server-insecure-tls" help:"Skip PMM Server TLS certificate validation"`
 	EnableDebug             bool        `name:"debug" help:"Enable debug logging"`
 	EnableTrace             bool        `name:"trace" help:"Enable trace logging (implies debug)"`
-	PMMAgentListenPort      uint32      `default:"${defaultListenPort}" help:"Set listen port of pmm-agent"`
+	PMMAgentListenPort      uint32      `placeholder:"PORT" help:"Port of local pmm-agent to connect to"`
```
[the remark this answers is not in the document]
michal-kralik: After this change pmm-agent listens on a socket by default.
For this reason there's no default for the port.

agent/config/config.go
michal-kralik: we need this part in both functions.
The agent listens both for grpc and json. grpc is always listening on a socket while json could be either socket or ip:port.
If json is ip:port, we need to configure the grpc defaults in `configureListenGRPCDefaults()` because `configureListenDefaults()` does not take care of the default in this case.

idoqo (approved):

JiriCtvrtka (approved):

PR #1590, opened 2023-01-19
PMM-11423: support for updating a service

michal-kralik:
PMM-11423

Requires the following to be merged at the same time:
- https://github.com/percona/pmm/pull/1662
- https://github.com/percona-platform/grafana/pull/599

Build: SUBMODULES-3016


managed/services/inventory/services.go
```diff
@@ -434,17 +426,72 @@ func (ss *ServicesService) RemoveCustomLabels(ctx context.Context, req *inventor
...
+
+		if req.ExternalGroup != nil {
+			columns = append(columns, "external_group")
+			service.ExternalGroup = *req.ExternalGroup
+		}
+
+		err = tx.UpdateColumns(service, columns...)
+		if err != nil {
+			return err
+		}
```
BupycHuk, line 469: I would move all this logic into model package

denisok:
@PavelKhripkov @BupycHuk please review.

api/inventorypb/services.proto
```diff
@@ -468,6 +468,20 @@ message RemoveCustomLabelsRequest {
 
 message RemoveCustomLabelsResponse {}
 
+message UpdateServiceRequest {
```
BupycHuk, line 471: Since we have `ChangeXXX` naming for agents, I would suggest having the same naming for services and nodes to keep it consistent

BupycHuk: And TBH we might need to separate these endpoints for each service type since other features Inventory Management is coming.

michal-kralik: Splitting as in MySQL, MongoDB, Postgres, etc?

BupycHuk: I suppose so but doesn't have a strong opinion about that. Not sure if we will require it at all.

michal-kralik: I suggest we stick with `ChangeService` then.
These are common settings for any service. Similar to custom labels.

PavelKhripkov:
@michal-kralik we have some functionality related to `cluster` label. Probably we need some changes in other parts of the code before merging your changes. Let me check it.

managed/services/inventory/services.go
```diff
@@ -434,17 +426,56 @@ func (ss *ServicesService) RemoveCustomLabels(ctx context.Context, req *inventor
...
+		return nil, errTx
+	}
+
+	if err := ss.updateScrapeConfig(ctx, req.ServiceId); err != nil {
+		return nil, err
+	}
+
+	return &inventorypb.ChangeServiceResponse{}, nil
+}
+
```
PavelKhripkov, line 467: I'm not really understand the approach used in this function.

Why do we use `UpdateColumns` instead of just `Update`? Do we really need to complicate the code?

Why do we use direct orm calls (again `UpdateColumns`) from business-logic level? We can do it from the models level.

Usually we unpacking req structure but in this function we're passing it directly. Can we be consistent here?

managed/models/service_model.go
```diff
@@ -99,6 +106,33 @@ func (s *Service) SetCustomLabels(m map[string]string) error {
...
+		s.ReplicationSet = *labels.ReplicationSet
+	}
+
+	if labels.ExternalGroup != nil {
+		columns = append(columns, "external_group")
+		s.ExternalGroup = *labels.ExternalGroup
+	}
+
+	return columns
+}
```
PavelKhripkov, line 134: If we really need this function it should be placed in the service_helpers.go file since we have it.

managed/models/service_model.go
```diff
@@ -99,6 +106,33 @@ func (s *Service) SetCustomLabels(m map[string]string) error {
 	return setLabels(m, &s.CustomLabels)
 }
 
+// ChangeStandardLabels changes standard labels and returns a list of columns which were updated.
+func (s *Service) ChangeStandardLabels(labels ServiceStandardLabelsParams) []string {
+	columns := []string{}
+
+	if labels.Cluster != nil {
+		columns = append(columns, "cluster")
```
PavelKhripkov, line 114: We can use either typed or untyped constants instead of strings like "cluster" cuz it's easy to make a typo and compiler cannot help in such case.

api/inventorypb/services.proto
BupycHuk: ok

managed/services/inventory/services.go
BupycHuk: Agree, I think it should be in the models package to separate model and business logic levels.

managed/models/service_model.go
BupycHuk: totally agree

managed/services/inventory/services.go
michal-kralik: How would you suggest we address concurrency issues when using `Update` instead of `UpdateColumns`?
In case the service changes between we queried for data and updated it, the database will be in an inconsistent state.

`Update` updates all columns for a record, while `UpdateColumns` just the specific columns we want to touch.

michal-kralik:
> @michal-kralik we have some functionality related to `cluster` label. Probably we need some changes in other parts of the code before merging your changes. Let me check it.
> 
> upd: We really need changes otherwise our mongo backup logic will be broken.

Sounds good. What kind of changes do you have in mind?

managed/services/inventory/services.go
PavelKhripkov: Updating only specific columns doesn't solve the problem with inconsistency. We'll end up with the same if several transactions updating the same columns. I suggest to use `Serializable` level for such transactions.

managed/models/service_model.go
PavelKhripkov: If you still really need column names for some reason, them should be kept as constants in only one package which is responsible for the specific functionality.

PavelKhripkov:
> > @michal-kralik we have some functionality related to `cluster` label. Probably we need some changes in other parts of the code before merging your changes. Let me check it.
> > upd: We really need changes otherwise our mongo backup logic will be broken.
> 
> Sounds good. What kind of changes do you have in mind?

I'm waiting response from @rnovikovP regarding this thing. Need to set priority and discuss possible solutions

managed/services/inventory/services.go
```diff
@@ -434,17 +426,50 @@ func (ss *ServicesService) RemoveCustomLabels(ctx context.Context, req *inventor
...
+			Cluster:        params.Cluster,
+			Environment:    params.Environment,
+			ReplicationSet: params.ReplicationSet,
+			ExternalGroup:  params.ExternalGroup,
+		})
+		if err != nil {
+			return err
+		}
+
+		return nil
```
PavelKhripkov, line 449:
```suggestion
		return err
```
Or even don't need err variable at all, just `return foo()`

managed/services/inventory/services.go
michal-kralik: The update operation is atomic. Multiple transactions changing the same columns shall always produce a consistent result.
Setting transaction isolation to serializable could help as well. I'm just not sure if we're ready, across the codebase, to retry all failed transactions automatically.
If you think this is feasible, I suggest we create a separate PR for the change.

PavelKhripkov: Of course it's up to you but imo we're not solving the problem, just hiding it and all this stuff with specifiyng changed fields is error prone.

PavelKhripkov:
> > > @michal-kralik we have some functionality related to `cluster` label. Probably we need some changes in other parts of the code before merging your changes. Let me check it.
> > > upd: We really need changes otherwise our mongo backup logic will be broken.
> > 
> > 
> > Sounds good. What kind of changes do you have in mind?
> 
> I'm waiting response from @rnovikovP regarding this thing. Need to set priority and discuss possible solutions

upd: We decided to not block these changes and we'll prepare fixes to cover cases when cluster is changed.

managed/models/service_model.go
```diff
@@ -99,6 +106,33 @@ func (s *Service) SetCustomLabels(m map[string]string) error {
 	return setLabels(m, &s.CustomLabels)
 }
 
+// ChangeStandardLabels changes standard labels and returns a list of columns that have been updated.
+func (s *Service) ChangeStandardLabels(labels ServiceStandardLabelsParams) []string {
```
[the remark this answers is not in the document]
michal-kralik: Good catch

managed/models/service_helpers.go
```diff
@@ -405,3 +405,48 @@ func ValidateServiceType(serviceType ServiceType) error {
...
+func ChangeStandardLabels(q *reform.Querier, id string, labels ServiceStandardLabelsParams) error {
+	s, err := FindServiceByID(q, id)
+	if err != nil {
+		return err
+	}
+
+	columns := []string{}
+
+	if labels.Cluster != nil {
+		columns = append(columns, "cluster")
```
PavelKhripkov, line 428: I'm still suggesting to use constants at the top of file like `const serviceClusterField = "cluster"` and cover all such parts with tests. Or at least just add tests. Otherwise if somebody changes string values and we miss it on review, we will never know about that until users report bug.

If you have any other ideas how to protect from bugs, please implement them.

PavelKhripkov (changes requested):

managed/models/service_helpers.go
michal-kralik: I understand the intention. This PR does not make the situation worse though.
There's numerous places where we already do the same, such as:

https://github.com/percona/pmm/blob/a9716c8232669deea5c6eb548982e4bc9fcaa36c/managed/models/node_helpers.go#L139
https://github.com/percona/pmm/blob/a9716c8232669deea5c6eb548982e4bc9fcaa36c/managed/models/node_helpers.go#L163

We are not going to improve the current state by applying a new approach to the column names here.
I don't expect we'll be able to change the column names in the future without a major refactoring of the whole codebase.

Either way, I added tests which shall cover changing any label and making sure the column names match.

PavelKhripkov (approved):

managed/services/inventory/services.go
```diff
@@ -434,7 +435,11 @@ func (ss *ServicesService) RemoveCustomLabels(ctx context.Context, req *inventor
 }
 
 // ChangeService changes service configuration.
-func (ss *ServicesService) ChangeService(ctx context.Context, params *models.ChangeStandardLabelsParams) (*inventorypb.ChangeServiceResponse, error) {
+func (ss *ServicesService) ChangeService(ctx context.Context, mgmtServices common.MgmtServices, params *models.ChangeStandardLabelsParams) error {
+	if err := mgmtServices.RemoveScheduledTasks(ctx, ss.db, params); err != nil {
```
michal-kralik, line 439: @PavelKhripkov would it make more sense to execute this in a transaction along with the labels change?
In case the labels change fail, we already removed scheduled tasks and there's no way back.

PavelKhripkov: For sure, but the calls made on api-level that doesn't allow to rollback. To implement the same on a deeper level, we need to duplicate alot of code lines. And without proper testing It won't be reliable.

gen1us2k (approved):

BupycHuk (approved):

PR #1591, opened 2023-01-19
PMM-7: add role API to swagger

michal-kralik:
PMM-7

Build: SUBMODULES-3022


BupycHuk (approved):

PavelKhripkov:
Don't we need to add new path in the **clean** section?
https://github.com/percona/pmm/blob/b99d37f3c1feaf4e5a52622f24e2736deeee228a/Makefile.include#L144

michal-kralik:
> Don't we need to add new path in the **clean** section?
> 
> https://github.com/percona/pmm/blob/b99d37f3c1feaf4e5a52622f24e2736deeee228a/Makefile.include#L144

Good catch. It does.

PavelKhripkov (approved):

PR #1603, opened 2023-01-20
PMM-7: fix race condition in config reload

michal-kralik:
PMM-7

Build: SUBMODULES-3019

Fixes a race condition with config reload in agent.
Since we pass around a pointer to the config, including go routines, `go -race` complains about concurrent access to the config.

This PR aims to allow reload of the config while preventing concurrent access through a mutex.


agent/client/client.go
```diff
@@ -82,9 +82,9 @@ type Client struct {
 // New creates new client.
 //
 // Caller should call Run.
-func New(cfg *config.Config, supervisor supervisor, r *runner.Runner, connectionChecker connectionChecker, sv softwareVersioner, cus *connectionuptime.Service, logStore *tailog.Store) *Client { //nolint:lll
+func New(getCfg func() *config.Config, supervisor supervisor, r *runner.Runner, connectionChecker connectionChecker, sv softwareVersioner, cus *connectionuptime.Service, logStore *tailog.Store) *Client { //nolint:lll
```
BupycHuk, line 85: let's make it an interface instead of a just function

agent/config/storage.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/sirupsen/logrus"
+)
+
+type storage struct {
+	cfg *Config
+	mu  sync.RWMutex
+}
+
+var cfgStorage = storage{}
```
BupycHuk, line 28: let's not make global variables.

BupycHuk (changes requested):

agent/config/storage.go
BupycHuk: can we make this storage a structure and then use it everywhere.

recharte (approved):
Looks good

agent/config/storage.go
```diff
@@ -0,0 +1,73 @@
...
+// Reload reloads config into the global object.
+func (s *Storage) Reload(l *logrus.Entry) (string, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	newCfg := &Config{}
+	cfgPath, err := getFromCmdLine(newCfg, l)
+	if err != nil {
+		s.cfg = newCfg
+	}
```
BupycHuk, line 70: uhm...
why don't we set `s.cfg` always?
it leads us to outdated config being stored in this storage

BupycHuk (changes requested):

agent/config/storage.go
```diff
@@ -0,0 +1,73 @@
...
+// Getter allows for getting a config.
+type Getter interface {
+	Get() *Config
+}
+
+// GetReloader allows for getting and reloading a config.
+type GetReloader interface {
+	Get() *Config
+	Reload(l *logrus.Entry) (string, error)
+}
```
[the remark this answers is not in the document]
michal-kralik: Makes sense. Great concept. Thank you for sharing.

agent/config/storage.go
```diff
@@ -0,0 +1,73 @@
...
+// Get returns a global config object.
+func (s *Storage) Get() *Config {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	return s.cfg
+}
+
+// Reload reloads config into the global object.
+func (s *Storage) Reload(l *logrus.Entry) (string, error) {
```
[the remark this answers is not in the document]
michal-kralik: Not sure I understand. Do you mean why it can't reload automatically or are you referring to the comment? I updated the comment, it was outdated.

agent/config/storage.go
michal-kralik: Good catch, did not realize there's an edge case.
Do you think checking for `ConfigFileDoesNotExistError` is sufficient here or are there any other cases where we shall reload the config regardless of errors?

agent/config/storage.go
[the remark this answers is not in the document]
BupycHuk: ignore it

agent/agentlocal/agent_local.go
```diff
@@ -61,9 +61,15 @@ const (
 	serverZipFile   = "pmm-agent.log"
 )
 
+// getReloader allows for getting and reloading a config.
+type getReloader interface {
```
BupycHuk, line 65: here it can be called `configGetter` or something like that

agent/config/storage.go
BupycHuk: I'm not sure, I would suggest logging and returning an error and not replacing the configuration at all.

michal-kralik: Sorry, I see what you mean now. That's a bug, it should have returned, not replace the config.
A implemented a check for `ConfigFileDoesNotExistError` - I think this is the correct behavior since config shall be reloaded in case of this error.

agent/agentlocal/agent_local.go
```diff
@@ -61,9 +61,15 @@ const (
 	serverZipFile   = "pmm-agent.log"
 )
 
+// configGetReloader allows for getting and reloading a config.
```
atymchuk, line 64:
```suggestion
// configGetReloader allows to get and reload a config.
```
nit: I was looking for simplicity :)

agent/client/client.go
```diff
@@ -55,9 +55,14 @@ const (
 	clockDriftWarning = 5 * time.Second
 )
 
+// configGetter allows for getting a config.
```
atymchuk, line 58:
```suggestion
// configGetter allows to get a config.
```

atymchuk (approved):

BupycHuk (approved):

PR #1618, opened 2023-01-24
PMM-11421 Fix proxy misconfiguration in platform requests

palash25:
Platform requests were using two different HTTP clients to make requests one for refresh token and the other for the rest of the platform endpoints The client for refresh token requests was the default HTTP client which respected proxy and the other ignored proxy settings. This commit adds the default env proxy settings to the second client to make the request paths consistent. In the future we should consider unifying these two HTTP clients and just use one.

PMM-11421

Build: SUBMODULES-0


oksana-grishchenko (approved):

BupycHuk (approved):

managed/models/percona_sso_model_helpers.go
```diff
@@ -23,25 +23,18 @@ import (
...
-	perconaSSOMtx           sync.Mutex
-	ErrNotConnectedToPortal = errors.New("PMM Server is not connected to Portal")
-)
+var ErrNotConnectedToPortal = errors.New("PMM Server is not connected to Portal")
 
 // GetPerconaSSODetails returns PerconaSSODetails if there are any, error otherwise.
 // Access token is automatically refreshed if it is expired.
 // Get, check eventually refresh done in one tx.
 func GetPerconaSSODetails(ctx context.Context, q *reform.Querier) (*PerconaSSODetails, error) {
-	perconaSSOMtx.Lock()
```
palash25, line 42: removed this mutex since it was not guarding anything in memory

managed/services/platform/platform.go
```diff
@@ -37,7 +38,10 @@ import (
 
 const rollbackFailed = "Failed to rollback:"
 
-var errGetSSODetailsFailed = status.Error(codes.Aborted, "Failed to fetch SSO details.")
+var (
+	errGetSSODetailsFailed = status.Error(codes.Aborted, "Failed to fetch SSO details.")
+	errProxyNotSupported   = status.Error(codes.Aborted, "PMM Platform connection does not support proxy.")
```
palash25, line 43: added this error since the ticket asks for a more descritptive error instead of Internal Server https://jira.percona.com/browse/PMM-11421

managed/models/percona_sso_model_helpers.go
palash25: @oksana-grishchenko could you confirm this? i don't see any concurrent access here? why was this mutex added do you remember?

oksana-grishchenko: During the refresh token process the the AccessToken that stored in DB [got updated](https://github.com/percona/pmm/blob/3df82544be0ab59b259c3df2fac5ba71f3c20d0d/managed/models/percona_sso_model_helpers.go#L106). I suppose this was why this mutex were there

palash25: i still don't understand if the write is happening on disk then why we need a mutex, but i will keep it for now just so that this doesn't break anything. thanks.

PR #1638, opened 2023-01-29
GH-1436 use embed to package binary data

### LINKED ISSUE #1436: Replace github.com/kevinburke/go-bindata lib with go.dev/pkg/embed
### Description

Go has embedding functionality in standard lib now. So we can get rid of https://github.com/kevinburke/go-bindata in our code base.

### Suggested solution

Find and replace all usages of `go-bindata` library with `embed`. Remove this lib from project dependencies.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

atymchuk:
Closes #1436

Note: this migration required to upgrade v3.x of github.com/golang-migrate/migrate to v4, since v3 does not support the file system source.

@JiriCtvrtka Does this PR need a complete manual test of QAN?


atymchuk:
Damn, mend's license scan still freak out on our own library even though we allow-listed it twice! What an annoyance :( 

I ended up by filing a second (sic!) support case with them, let's see...
https://whitesourcesoftware.force.com/SupportPortal/s/case/5005p00002wiKxNAAU/inability-to-whitelist-a-library?tabset-1f16f=2

.github/workflows/common.yml
```diff
@@ -132,12 +132,13 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Run Merge Gatekeeper
-        uses: upsidr/merge-gatekeeper@v1
+        uses: upsidr/merge-gatekeeper@v1.2.0
```
artemgavrilov, line 135: Why you used full version here?

atymchuk: Because they don't use the tag `latest` which is why we end up using an older version.

artemgavrilov: Ah, that's because they have exactly `v1` version tag alongside with proper `v1.0.0`.

go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	github.com/go-openapi/validate v0.22.1
 	github.com/go-sql-driver/mysql v1.7.0
 	github.com/golang-migrate/migrate v3.5.4+incompatible
```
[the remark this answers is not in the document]
atymchuk: Good catch, rectified now.

artemgavrilov (approved):

PavelKhripkov (approved):

PR #1642, opened 2023-01-30
Fix CI

artemgavrilov:
PMM-7

Remove cli tests exception from admin CI


michal-kralik (approved):

artemgavrilov:
@atymchuk I can't merge because of license check. Maybe we can add it to exceptions too?

atymchuk:
> @atymchuk I can't merge because of license check. Maybe we can add it to exceptions too?

I did, but it still doesn't pass. I opened a ticket, meanwhile let's merge it.

PR #1651, opened 2023-02-02
PMM-11525 postgres file store and endpoints

rishat-ishbulatov:
[PMM-11525](https://jira.percona.com/browse/PMM-11525)

[Build](https://github.com/Percona-Lab/pmm-submodules/pull/3037)


api/serverpb/server.proto
```diff
@@ -291,6 +291,61 @@ message AWSInstanceCheckRequest {
...
+  string name = 1 [
+    (validator.field) = {
+      string_not_empty: true
+    }
+  ];
+}
+
+message ListFilesResponse {
+  repeated string names = 1;
+}
```
BupycHuk, line 342: we don't need CRUD, because we have a hardcoded list of files, so we don't need so much endpoints. List and Update should be enough

BupycHuk (commented):
left only one comment, it should decrease amount of changes.

api/serverpb/server.proto
rishat-ishbulatov: What should list return then? Names would be enough? Update should contain old name, new name if we hardcode file names?

rishat-ishbulatov: Which fields do you think would be useful in the future to update?

BupycHuk: for both endpoints, there might be just 2 fields `prometheus` and `alertmanager`

rishat-ishbulatov: I would like to discuss model will looks like then.

rishat-ishbulatov: after approvals in the slack, it was decided to leave only 2 endpoints and models for them

recharte (approved):

PR #1662, opened 2023-02-05
PMM-11546 remove scheduled backups when cluster changed

PavelKhripkov:
PMM-11546

Build: SUBMODULES-0


michal-kralik (changes requested):
Ideally changing a service's cluster name shall not remove any scheduled tasks. They shall continue working.

Is there a way for us to update the scheduled tasks instead?

recharte (approved):

PavelKhripkov:
> Ideally changing a service's cluster name shall not remove any scheduled tasks. They shall continue working.
> 
> Is there a way for us to update the scheduled tasks instead?

Yes, there is such a way. But it would take more afforts and we still won't have full control over clusters. So we agreed with @rnovikovP that for now we're just removing them.

managed/services/management/management.go
```diff
@@ -25,8 +25,15 @@ import (
 	"github.com/percona/pmm/api/inventorypb"
 	"github.com/percona/pmm/api/managementpb"
 	"github.com/percona/pmm/managed/models"
+	managementbackup "github.com/percona/pmm/managed/services/management/backup"
 )
 
+type MgmtServices struct {
+	BackupsService        *managementbackup.BackupsService
+	ArtifactsService      *managementbackup.ArtifactsService
+	RestoreHistoryService *managementbackup.RestoreHistoryService
```
artemgavrilov, line 34: Maybe it makes sense to merge these 3 services into single one and do the same with API. Not sure, but it seems that we have too much in common between them.

managed/services/inventory/misc.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Check no restore tasks running.
+	restores, err := mgmtServices.RestoreHistoryService.ListRestoreHistory(ctx, &backuppb.ListRestoreHistoryRequest{})
+	if err != nil {
+		return err
+	}
+
+	for _, restoreItem := range restores.Items {
+		if _, ok := sMap[restoreItem.ServiceId]; ok && restoreItem.Status == backuppb.RestoreStatus_RESTORE_STATUS_IN_PROGRESS {
+			return errors.Wrapf(ErrClusterBlocked, "there is a not finished restore job for service %s or other service in the same cluster", s.ServiceID)
```
artemgavrilov, line 114:
```suggestion
			return errors.Wrapf(ErrClusterBlocked, "there is a non finished restore job for service %s or other service in the same cluster", s.ServiceID)
```

managed/services/inventory/misc.go
```diff
@@ -0,0 +1,119 @@
...
+	backuppb "github.com/percona/pmm/api/managementpb/backup"
+	"github.com/percona/pmm/managed/models"
+	"github.com/percona/pmm/managed/services/management"
+)
+
+// ErrClusterBlocked is returned when there is a not finished job that doesn't allow to change service cluster name.
+var ErrClusterBlocked = errors.New("cluster/service is blocked")
+
+// removeScheduledTasks removes scheduled backup tasks and check there are no running backup/restore tasks in case user changes service cluster label.
+func removeScheduledTasks(ctx context.Context, db *reform.DB, mgmtServices management.MgmtServices, params *models.ChangeStandardLabelsParams) error {
```
artemgavrilov, line 33: Can it be a method of `mgmtServices`?

artemgavrilov (approved):

michal-kralik:
> Yes, there is such a way. But it would take more afforts and we still won't have full control over clusters. So we agreed with @rnovikovP that for now we're just removing them.

This behavior is certainly not expected by the users. What other changes to the product are necessary to make sure users are aware this breaks any scheduled jobs?

PavelKhripkov:
> This behavior is certainly not expected by the users. What other changes to the product are necessary to make sure users are aware this breaks any scheduled jobs?

Let's see.
1. Documentation
2. UI warning
3. Probably description in API
4. If feature accessible via CLI we need to describe the behaviour or drop warning/question before calling method. Or add a `--force` flag and handle it.

managed/services/management/management.go
PavelKhripkov: Agree, on API level we can merge the functionality.

michal-kralik:
> > This behavior is certainly not expected by the users. What other changes to the product are necessary to make sure users are aware this breaks any scheduled jobs?
> 
> Let's see.
> 
> 1. Documentation
> 2. UI warning
> 3. Probably description in API
> 4. If feature accessible via CLI we need to describe the behaviour or drop warning/question before calling method. Or add a `--force` flag and handle it.

Sounds good.

managed/services/inventory/misc.go
PavelKhripkov: Done

api/inventorypb/services.proto
```diff
@@ -624,7 +624,7 @@ service Services {
     };
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
       summary: "Change service"
-      description: "Changes service configuration."
+      description: "Changes service configuration. If new cluster label specified, removes all existing scheduled backup/restore tasks for the related services. Fails if there are any backup/restore running tasks."
```
atymchuk, line 627:
```suggestion
      description: "Changes service configuration. If a new cluster label is specified, it removes all backup/restore tasks scheduled for the related services. Fails if there are running backup/restore tasks."
```

api/swagger/swagger-dev.json
```diff
@@ -15542,7 +15542,7 @@
     },
     "/v1/inventory/Services/Change": {
       "post": {
-        "description": "Changes service configuration.",
+        "description": "Changes service configuration. If new cluster label specified, removes all existing scheduled backup/restore tasks for the related services. Fails if there are any backup/restore running tasks.",
```
atymchuk, line 15545:
```suggestion
        "description": "Changes service configuration. If a new cluster label is specified, it removes all backup/restore tasks scheduled for the related services. Fails if there are running backup/restore tasks.",
```

managed/services/inventory/grpc/services_server.go
```diff
@@ -20,21 +20,29 @@ import (
...
+	s            *inventory.ServicesService
+	mgmtServices common.MgmtServices
 
 	inventorypb.UnimplementedServicesServer
 }
 
 // NewServicesServer returns Inventory API handler for managing Services.
-func NewServicesServer(s *inventory.ServicesService) inventorypb.ServicesServer {
-	return &servicesServer{s: s}
+func NewServicesServer(s *inventory.ServicesService, mgmtServices common.MgmtServices) inventorypb.ServicesServer {
```
atymchuk, line 41: Plural of plurals is not a good language construct. It should be rather `NewServiceServer`.

managed/services/management/common/common.go
```diff
@@ -0,0 +1,127 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+// Package common contains common and cross-service logics.
+
+package common
```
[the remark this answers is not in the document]
atymchuk: Can we address this one as well?

managed/services/management/common/common.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/pkg/errors"
+	"gopkg.in/reform.v1"
+
+	backuppb "github.com/percona/pmm/api/managementpb/backup"
+	"github.com/percona/pmm/managed/models"
+	managementbackup "github.com/percona/pmm/managed/services/management/backup"
+)
+
+// ErrClusterBlocked is returned when there is an unfinished job that doesn't allow to change service cluster name.
+var ErrClusterBlocked = errors.New("cluster/service is blocked")
```
atymchuk, line 31: Shouldn't it be `Locked` vs `Blocked`? IMO it better reflects our intent to temporarily disallow certain changes.

atymchuk (changes requested):

managed/services/management/common/common.go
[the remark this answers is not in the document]
PavelKhripkov: I did couple of times but it still has empty string betwee comment and package name. Not sure what is going on, but l'll try again)

managed/services/inventory/grpc/services_server.go
PavelKhripkov: Not sure where exactly the plural of plurals, I see only one plural, but I agree it sounds a little strange. Yet we have plenty of them like `NewNodesServer`, `NewAgentsServer` etc. So I'm not sure we need to do refactoring in the scope of this pr.

managed/services/management/common/common.go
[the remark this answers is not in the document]
PavelKhripkov: Looks like it just wasn't marked as outdated.

managed/services/inventory/grpc/services_server.go
atymchuk: I'm not in favor of refactoring now. Rather, we have a good opportunity to do it right this time.

PavelKhripkov: Ok, I've just checked. This notation comes from grpc api, and if we just rename one func we'd get inconsistency. We can discuss it on a weekly call, but for now I don't think we really need change something here.

atymchuk: It's not about renaming - we are creating a brand new function :)

atymchuk: Apart from that, do we need to be consistent in poor naming?

atymchuk (approved):

managed/services/inventory/grpc/services_server.go
PavelKhripkov: Actually this is an old function, we just need to add new dependency. About poor naming: let's discuss it with the team)

PR #1665, opened 2023-02-06
PMM-11517 resync local storage before restore

idoqo:
PMM-11517

With local storage, backup artifacts goes missing before after we restart pbm/pbm-agents (i.e., after a physical restore). This causes subsequent restores to fail as `pbm` cannot correctly detect such artifact.

This adds the `--force-resync` option to pbm config before restoring to update the artifact list.

Build: SUBMODULES-0

### DIFF
--- a/agent/runner/jobs/mongodb_backup_job.go
+++ b/agent/runner/jobs/mongodb_backup_job.go
@@ -115,7 +115,7 @@ func (j *MongoDBBackupJob) Run(ctx context.Context, send Send) error {
 	}
 	defer os.Remove(confFile) //nolint:errcheck
 
-	if err := pbmConfigure(ctx, j.l, j.dbURL, confFile); err != nil {
+	if err := pbmConfigure(ctx, j.l, j.dbURL, false, confFile); err != nil {
 		return errors.Wrap(err, "failed to configure pbm")
 	}
 
--- a/agent/runner/jobs/mongodb_restore_job.go
+++ b/agent/runner/jobs/mongodb_restore_job.go
@@ -107,7 +107,8 @@ func (j *MongoDBRestoreJob) Run(ctx context.Context, send Send) error {
 	}
 	defer os.Remove(confFile) //nolint:errcheck
 
-	if err := pbmConfigure(ctx, j.l, j.dbURL, confFile); err != nil {
+	forceResync := conf.Storage.FileSystem.Path != ""
+	if err := pbmConfigure(ctx, j.l, j.dbURL, forceResync, confFile); err != nil {
 		return errors.Wrap(err, "failed to configure pbm")
 	}
 
@@ -136,7 +137,6 @@ func (j *MongoDBRestoreJob) Run(ctx context.Context, send Send) error {
 	go func() {
 		err := j.jobLogger.streamLogs(streamCtx, send, restoreOut.Name)
 		if err != nil && !errors.Is(err, io.EOF) && !errors.Is(err, context.Canceled) {
-			j.jobLogger.sendLog(send, err.Error(), false)
 			j.l.Errorf("stream logs: %v", err)
 		}
 	}()
@@ -154,6 +154,10 @@ func (j *MongoDBRestoreJob) Run(ctx context.Context, send Send) error {
 		},
 	})
 
+	select {
+	case <-ctx.Done():
+	case <-time.After(waitForLogs):
+	}
 	return nil
 }
 
--- a/agent/runner/jobs/pbm_helpers.go
+++ b/agent/runner/jobs/pbm_helpers.go
@@ -311,7 +311,6 @@ func findPITRRestoreName(ctx context.Context, dbURL *url.URL, restoreInfo *pbmRe
 }
 
 func waitForPBMRestore(ctx context.Context, l logrus.FieldLogger, dbURL *url.URL, restoreInfo *pbmRestore, backupType, confFile string) error {
-	l.Infof("waiting for pbm restore")
 	var name string
 	var err error
 
@@ -325,6 +324,8 @@ func waitForPBMRestore(ctx context.Context, l logrus.FieldLogger, dbURL *url.URL
 		name = restoreInfo.Name
 	}
 
+	l.Infof("waiting for pbm restore: %s", restoreInfo.Name)
+
 	ticker := time.NewTicker(statusCheckInterval)
 	defer ticker.Stop()
 
@@ -334,7 +335,7 @@ func waitForPBMRestore(ctx context.Context, l logrus.FieldLogger, dbURL *url.URL
 		case <-ticker.C:
 			var info describeInfo
 			if backupType == "physical" {
-				err = execPBMCommand(ctx, dbURL, &info, "describe-restore", "--config="+confFile, name)
+				err = execPBMCommand(ctx, dbURL, &info, "describe-restore", name, "--config="+confFile)
 			} else {
 				err = execPBMCommand(ctx, dbURL, &info, "describe-restore", name)
 			}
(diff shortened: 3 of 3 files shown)


agent/runner/jobs/pbm_helpers.go
```diff
@@ -365,17 +366,21 @@ func waitForPBMRestore(ctx context.Context, l logrus.FieldLogger, dbURL *url.URL
 	}
 }
 
-func pbmConfigure(ctx context.Context, l logrus.FieldLogger, dbURL *url.URL, confFile string) error {
-	l.Info("Configuring S3 location.")
+func pbmConfigure(ctx context.Context, l logrus.FieldLogger, dbURL *url.URL, forceResync bool, confFile string) error {
```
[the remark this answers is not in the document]
PavelKhripkov: I would say we have two options:
1. pass struct (not interface) to the function
2. create a service (class) that would be responsible for all that pbm staff and get rid of at least three parameters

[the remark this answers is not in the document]
idoqo: We already have a struct for pbm configuration, in that case, we can follow the same practice and create one for the config flags here.

agent/runner/jobs/mongodb_restore_job.go
```diff
@@ -154,6 +154,10 @@ func (j *MongoDBRestoreJob) Run(ctx context.Context, send Send) error {
 		},
 	})
 
+	select {
+	case <-ctx.Done():
+	case <-time.After(waitForLogs):
+	}
```
PavelKhripkov, line 160: These changes are not related to the scope of the ticket. Could you describe why we need them?

agent/runner/jobs/pbm_helpers.go
```diff
@@ -325,6 +324,8 @@ func waitForPBMRestore(ctx context.Context, l logrus.FieldLogger, dbURL *url.URL
 		name = restoreInfo.Name
 	}
 
+	l.Infof("waiting for pbm restore: %s", restoreInfo.Name)
```
PavelKhripkov, line 327:
```suggestion
	l.Infof("waiting for pbm restore: %s", name)
```

`restoreInfo.Name` is not always valid.

agent/runner/jobs/pbm_helpers.go
```diff
@@ -311,7 +311,6 @@ func findPITRRestoreName(ctx context.Context, dbURL *url.URL, restoreInfo *pbmRe
 }
 
 func waitForPBMRestore(ctx context.Context, l logrus.FieldLogger, dbURL *url.URL, restoreInfo *pbmRestore, backupType, confFile string) error {
-	l.Infof("waiting for pbm restore")
```
PavelKhripkov, line 314: We can replace it with something like "Defining restore name"

agent/runner/jobs/mongodb_restore_job.go
```diff
@@ -107,7 +107,8 @@ func (j *MongoDBRestoreJob) Run(ctx context.Context, send Send) error {
 	}
 	defer os.Remove(confFile) //nolint:errcheck
 
-	if err := pbmConfigure(ctx, j.l, j.dbURL, confFile); err != nil {
+	forceResync := conf.Storage.FileSystem.Path != ""
```
PavelKhripkov, line 110: Sometimes pbm is unpredictable. What if we resyncing for both storages?

PavelKhripkov (changes requested):

agent/runner/jobs/mongodb_restore_job.go
idoqo: Right - they were leftovers from trying to figure out the problem, removed them now.

atymchuk (approved):

agent/runner/jobs/mongodb_restore_job.go
```diff
@@ -136,7 +137,6 @@ func (j *MongoDBRestoreJob) Run(ctx context.Context, send Send) error {
 	go func() {
 		err := j.jobLogger.streamLogs(streamCtx, send, restoreOut.Name)
 		if err != nil && !errors.Is(err, io.EOF) && !errors.Is(err, context.Canceled) {
-			j.jobLogger.sendLog(send, err.Error(), false)
```
PavelKhripkov, line 139: Do we really need remove this line?

agent/runner/jobs/pbm_helpers.go
```diff
@@ -163,6 +163,21 @@ type pbmError struct {
 	Error string `json:"Error"`
 }
 
+// pbmConfigFlags groups the flags/options for configuring PBM.
+type pbmConfigFlag struct {
```
PavelKhripkov, line 167:
```suggestion
type pbmConfigParams struct {
```

agent/runner/jobs/mongodb_restore_job.go
```diff
@@ -107,7 +107,8 @@ func (j *MongoDBRestoreJob) Run(ctx context.Context, send Send) error {
 	}
 	defer os.Remove(confFile) //nolint:errcheck
 
-	if err := pbmConfigure(ctx, j.l, j.dbURL, confFile); err != nil {
+	configFlag := createPBMConfigFlag(confFile, j.dbURL, true)
```
PavelKhripkov, line 110:
```suggestion
	configParams := createPBMConfigFlag(confFile, j.dbURL, true)
```

agent/runner/jobs/mongodb_backup_job.go
```diff
@@ -115,7 +115,8 @@ func (j *MongoDBBackupJob) Run(ctx context.Context, send Send) error {
 	}
 	defer os.Remove(confFile) //nolint:errcheck
 
-	if err := pbmConfigure(ctx, j.l, j.dbURL, confFile); err != nil {
+	configFlag := createPBMConfigFlag(confFile, j.dbURL, false)
```
PavelKhripkov, line 118:
```suggestion
	configParams := createPBMConfigFlag(confFile, j.dbURL, false)
```

agent/runner/jobs/pbm_helpers.go
```diff
@@ -163,6 +163,21 @@ type pbmError struct {
...
+	dbURL          *url.URL
+}
+
+func createPBMConfigFlag(confFile string, dbURL *url.URL, forceResync bool) pbmConfigFlag {
+	return pbmConfigFlag{
+		forceResync:    forceResync,
+		configFilePath: confFile,
+		dbURL:          dbURL,
+	}
+}
```
PavelKhripkov, line 179: Not sure we need this func. We can fill structure in-line without this helper.

agent/runner/jobs/pbm_helpers.go
```diff
@@ -365,17 +382,21 @@ func waitForPBMRestore(ctx context.Context, l logrus.FieldLogger, dbURL *url.URL
 	}
 }
 
-func pbmConfigure(ctx context.Context, l logrus.FieldLogger, dbURL *url.URL, confFile string) error {
-	l.Info("Configuring S3 location.")
+func pbmConfigure(ctx context.Context, l logrus.FieldLogger, flag pbmConfigFlag) error {
```
PavelKhripkov, line 385:
```suggestion
func pbmConfigure(ctx context.Context, l logrus.FieldLogger, params pbmConfigFlag) error {
```

agent/runner/jobs/mongodb_restore_job.go
idoqo: yeah, we currently try to send only pbm logs to job logs - this breaks that. Also, if stream logs failed, it makes sense to log directly (instead of attempting to send another log to the PMM server which might also fail).

agent/runner/jobs/pbm_helpers.go
idoqo: I think using this function gives us guarantee that all fields are set as intended.

PavelKhripkov: Usually we're using Validator for that. I just think this func guarantees that values provided but doesn't check they're correct. So not a big difference.

agent/runner/jobs/mongodb_restore_job.go
PavelKhripkov: My concern is: will it be tested in the scope of these changes?

idoqo: Can you clarify what you mean by "tested" here? If you mean tested by QA, then we could inform them to check logs again

PavelKhripkov (approved):

artemgavrilov (approved):

PR #1666, opened 2023-02-06
PMM-10571 backup management telemetry

PavelKhripkov:
PMM-10571

Build: SUBMODULES-3054


rnovikovP (approved):
I checked sanity and presence of descriptions - lgtm

managed/services/telemetry/config.default.yml
```diff
@@ -720,3 +720,86 @@ telemetry:
...
+      - metric_name: "3"
+        column: "bm_data_model"
+      - metric_name: "4"
+        column: "bm_job_run_type"
+      - metric_name: "5"
+        column: "bm_job_status"
+      - metric_name: "6"
+        column: "bm_location_type"
+      - metric_name: "7"
+        column: "count"
```
artemgavrilov, line 805: Some metrics in this file has sting name, some numeric. I don't remember what was the reason to use numeric names?

PavelKhripkov: As I understood, numbers used when we choose json format.

ritbl: it has to be numeric because JSON will be sorted by keys (it is default behavior with serialization), plus you will be to access in materialized view by number, you will need to declare ([example](https://github.com/percona-platform/telemetryd/blob/main/migrations/011_api_usage_view.up.sql#L31)), like this one `Array(Tuple(String,String,String,String,String,String,String))` and you will refer to to them by number.
Another concern is storage size.
image

ritbl (approved):

PR #1673, opened 2023-02-07
PMM-11336 UUID generation ami

talhabinrizwan:
PMM-11336


build/ansible/pmm2/post-build-actions.yml
```diff
@@ -87,6 +87,45 @@
...
+        name: pmm-managed
+        state: present
+      register: db_exists
+
+    - name: Remove pmm-managed database
+      postgresql_db:
+        login_user: postgres
+        name: pmm-managed
+        state: absent
+      when: db_exists.db | default(false)
```
atymchuk, line 111: `when: db_exists.db` - would that suffice?

atymchuk: Frankly, I would assume that the role exists after a successful build and:
1/ remove the same tasks from above, i.e. [lines](https://github.com/percona/pmm/pull/1673/files#diff-ca10aefe274646dba1a8c6cf00ce819d94ba0f6e40b3050eeffd9d54643f9046L70-L80)
2/ remove all when statements in db and role deletion tasks

atymchuk: That would be much simpler to read and understand.

build/packer/pmm2.json
```diff
@@ -136,7 +136,7 @@
         {
             "type": "ansible-local",
             "playbook_file": "ansible/pmm2/post-build-actions.yml",
-            "extra_arguments": ["-v"]
+            "extra_arguments": ["-vvv"]
```
atymchuk, line 139: That type of verbosity is almost impossible to read. Is it a leftover?

atymchuk (changes requested):

build/packer/pmm2.json
talhabinrizwan: Since 'update.yml' already had it, I believed it would be appropriate to include it for the rest as well.

build/ansible/pmm2/post-build-actions.yml
talhabinrizwan: I've committed a simplified version.

build/packer/pmm2.json
atymchuk: After a month of non-stop debugging our RHEL9 build I will still disagree :)

atymchuk (approved):

gen1us2k (approved):

EvgeniyPatlan (approved):

PR #1674, opened 2023-02-07
PMM-11566 Demo mode

oksana-grishchenko:
PMM-11566 

**_Context_**
Portal provides an option to create a Civo k8s cluster for 3 hours.
In the future releases we are going to create an already connected to the Portal PMM instance inside that cluster automatically to provide to the end users an easy way to test-drive DBaaS and the connected PMM.

**_Problem_**
Roles mapping. Technical users from the Portal become Viewers in PMM, which restricts the available options inside PMM significantly, including the DBaaS availability. So we are looking for a way to provide Admin rights to all users of that temporary PMM instances.

**_Solution_**
Grafana config: check if the user profile `pmm_demo_ids` field contains the `PMMServerID` value and provide Admin access in case it does. The `pmm_demo_ids` value themselves are managed by Portal. 

[Build](https://github.com/Percona-Lab/pmm-submodules/pull/3059)

-  https://github.com/percona-platform/okta/pull/18


gen1us2k (approved):

managed/services/supervisord/supervisord.go
```diff
@@ -772,7 +774,7 @@ command =
         cfg:default.auth.generic_oauth.auth_url="{{ .PerconaSSODetails.IssuerURL }}/authorize"
         cfg:default.auth.generic_oauth.token_url="{{ .PerconaSSODetails.IssuerURL }}/token"
         cfg:default.auth.generic_oauth.api_url="{{ .PerconaSSODetails.IssuerURL }}/userinfo"
-		cfg:default.auth.generic_oauth.role_attribute_path="contains(portal_admin_orgs[*], '{{ .PerconaSSODetails.OrganizationID }}') && 'Admin' || 'Viewer'"
+		cfg:default.auth.generic_oauth.role_attribute_path="(contains(portal_admin_orgs[*], '{{ .PerconaSSODetails }}') || contains(pmm_demo_ids[*], '{{ .PMMServerID }}')) && 'Admin' || 'Viewer'"
```
BupycHuk, line 777: should the `.OrganizationID` be removed here?

oksana-grishchenko: No, bc the `PMMServerID` will appear only in profiles of users that connected the _temporary Demo instances_. The normal instances will use the old way to map roles.

BupycHuk: Agree, but you removed `.OrganizationID`

oksana-grishchenko: Oh, thanks, I didn't notice that. That's why my tests are failing 😄

BupycHuk (approved):

PR #1675, opened 2023-02-07
PMM-7 Fix linters

artemgavrilov:
PMM-7

Fixes all warnings except one, it requires some effort to fix.


managed/services/telemetry/config.go
```diff
@@ -79,7 +79,7 @@ type DSGrafanaSqliteDB struct {
 }
 
 // DSConfigPMMDB telemetry config.
-type DSConfigPMMDB struct {
+type DSConfigPMMDB struct { //nolint:musttag
```
BupycHuk, line 82: wouldn't it be better to really tag fields?

BupycHuk (approved):

managed/services/telemetry/config.go
artemgavrilov: I don't know if it was made like this on purpose or not?

BupycHuk: @ritbl ^

artemgavrilov: Ah, right. All untagged fields are part of structs that marked as for `no serialization`

oksana-grishchenko (changes requested):
Thanks so much for the quick fix! 

One thing to ask. With go 1.19.5 there is one more error left, could you please also fix it? 

```
LOG_LEVEL=error bin/golangci-lint run
managed/services/management/rds.go:123:40: SA1019: endpoints.RdsServiceID is deprecated: Use client package's EndpointsID value instead of these ServiceIDs. These IDs are not maintained, and are out of date. (staticcheck)
                        for r := range partition.Services()[endpoints.RdsServiceID].Regions() {
```

artemgavrilov:
> Thanks so much for the quick fix!
> 
> One thing to ask. With go 1.19.5 there is one more error left, could you please also fix it?
> 
> ```
> LOG_LEVEL=error bin/golangci-lint run
> managed/services/management/rds.go:123:40: SA1019: endpoints.RdsServiceID is deprecated: Use client package's EndpointsID value instead of these ServiceIDs. These IDs are not maintained, and are out of date. (staticcheck)
>                         for r := range partition.Services()[endpoints.RdsServiceID].Regions() {
> ```

I added mention about it to PR header. This warning is more complex and requires some effort, unfortunately I don't have time for that now.

PR #1690, opened 2023-02-09
PMM-11472: telemetry for access control

michal-kralik:
PMM-11472

Build: SUBMODULES-0


BupycHuk (approved):

idoqo (approved):

managed/services/telemetry/config.default.yml
```diff
@@ -348,6 +356,14 @@ telemetry:
       - metric_name: "pmm_server_usage_environments_count"
         column: "environments_count"
 
+  - id: PMMUsagePerconaRolesCount
```
atymchuk, line 359: nit: this is a compound noun, and, if we want to use the rules, it has to be in singular vs plural, i.e. `PMMUsagePerconaRoleCount`.

michal-kralik: good point. For consistency purposes I'd opt for keeping it this way, since we use plural in other places.

atymchuk: Do we need consistency when it increases the number of things that look weird and don't comply with language rules?

michal-kralik: I see there's a strong preference in the team for consistency over correctness.
Happy to rename this if there's consensus though.

atymchuk: Actually, there is no consistency in applying that particular rule. A few examples fro the code: `scheduleService`, `backupService`, `removalService` are respecting the rules even though they also refer to multiples of schedules, backups, etc.

That said, making it right will not affect any consistency because there is none :)

michal-kralik: I was referring to other metrics, where we already use plural
- `PMMUsageEnvironmentsCount`
- `PMMUsageServicesCount`
- `PMMUsageNodesCount`
- etc.

atymchuk: Sure, I saw that as well... All I'm saying is that getting it right doesn't break anything.

atymchuk: Just an extract, no offense :)

"A tree that has apples has many apples, but we say an apple tree, not apples tree; matchbox not matchesbox; toothbrush not teethbrush."

percona-csalguero (approved):

PR #1691, opened 2023-02-09
PMM-11438 stop internal postgresql server

rishat-ishbulatov:
[PMM-11438](https://jira.percona.com/browse/PMM-11438)

[Build](https://github.com/Percona-Lab/pmm-submodules/pull/3053)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,6 +36,7 @@ services:
     #      - PERCONA_TEST_PMM_CLICKHOUSE_BLOCK_SIZE=10000
     #      - PERCONA_TEST_PMM_CLICKHOUSE_POOL_SIZE=2
     #      - PMM_DEBUG=1
+    #      - POSTGRES_STOP=true
 
     extra_hosts:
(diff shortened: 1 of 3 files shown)


recharte (commented):
I have a couple of questions:
1. Are we aware of all the possible side-effects scattered across PMM of not having the internal postgres server running? Is it safe to having it stopped?
2. Is stopping the service after running the supervisord the best solution? Would it be possible/better to remove it from the list of services to be started before launching supervisord?

rishat-ishbulatov:
> I have a couple of questions:
> 
>     1. Are we aware of all the possible side-effects scattered across PMM of not having the internal postgres server running? Is it safe to have it stopped?
> 
>     2. Is stopping the service after running the supervisord the best solution? Would it be possible/better to remove it from the list of services to be started before launching supervisord?

This decision was made in order to be able to turn off the internal postgress in case the user wants to use the [external one](https://jira.percona.com/browse/PMM-10913). I agree that it is better not to enable the service at all in this case. I'll see how feasible this is.

PR #1701, opened 2023-02-11
PMM-11438 Stop Internal PostgreSQL Server

rishat-ishbulatov:
[PMM-11438](https://jira.percona.com/browse/PMM-11438)

[Build](https://github.com/Percona-Lab/pmm-submodules/pull/3053)


managed/services/supervisord/pmm_config.go
```diff
@@ -0,0 +1,219 @@
...
+	return nil
+}
+
+func marshalConfig(disableDB bool) ([]byte, error) {
+	var buf bytes.Buffer
+	if err := pmmTemplate.Execute(&buf, map[string]bool{"DisableInternalDB": disableDB}); err != nil {
+		return nil, errors.Wrapf(err, "failed to render pmm template")
+	}
+	return buf.Bytes(), nil
+}
```
recharte, line 45: Having a boolean input param is a code smell that usually severely hinders code readability.
What's the difference between `supervisord.SavePMMConfig(false)` and `supervisord.SavePMMConfig(true)`?
You don't know unless you know the implementation of that function.

Often, the need for a bool param means that the function does more than it should and should therefore be split into two (or more) functions.

In this case, I don't think that the function should be split but I would change the input param from a `bool` to a `map[string]any` that will ultimately be the input of `Template.Execute`.
I see 2 advantages of this approach:
1. Extract the template configuration to the caller, increasing readability `supervisord.SavePMMConfig(map[string]bool{"DisableInternalDB": disableDB})` 
2. The template becomes extensible allowing for future customization of other fields other that `DisableInternalDB`

What do you think?

recharte (approved):
Well done @rishat-ishbulatov :clap:

JiriCtvrtka (approved):
Good job. Just please take a look on problem with tests.

managed/cmd/pmm-managed-init/main.go
```diff
@@ -34,6 +35,12 @@ func main() {
 	if on, _ := strconv.ParseBool(os.Getenv("PMM_TRACE")); on {
 		logrus.SetLevel(logrus.TraceLevel)
 	}
+	if ok, _ := strconv.ParseBool(os.Getenv("DISABLE_INTERNAL_DB")); ok {
```
atymchuk, line 38:
```suggestion
	if ok, _ := strconv.ParseBool(os.Getenv("PMM_DISABLE_INTERNAL_DB")); ok {
```
Can we start applying a `PMM_` prefix to all variables? I provided some background in [this doc](https://confluence.percona.com/display/PMM/PMM+v3.0) (section 3).

managed/services/supervisord/pmm_config.go
```diff
@@ -0,0 +1,219 @@
...
+startretries = 3
+startsecs = 1
+stopsignal = TERM
+stopwaitsecs = 300
+stdout_logfile = /srv/logs/pmm-update-perform-init.log
+stdout_logfile_maxbytes = 50MB
+stdout_logfile_backups = 3
+redirect_stderr = true
+{{ if not .DisableInternalDB}}
+[program:postgresql]
```
atymchuk, line 118: Can we have empty lines between neighbouring sections (or `programs` in supervisor's terminology)?

atymchuk:
Rishat, let's address the linter warnings as well

managed/utils/envvars/parser.go
```diff
@@ -179,6 +179,9 @@ func ParseEnvVars(envs []string) (envSettings *models.ChangeSettingsParams, errs
 			// This variable is not part of the settings and is parsed separately.
 			continue
 
+		case "DISABLE_INTERNAL_DB":
```
atymchuk, line 182: That seems to be ambiguous given that we have three "internal" databases. How about "PMM_DISABLE_BUILTIN_POSTGRES"?

managed/services/supervisord/pmm_config.go
rishat-ishbulatov: spaces between sections added

managed/utils/envvars/parser.go
rishat-ishbulatov: env renamed

PavelKhripkov: Looks like not all of them. Can we create dedicated type for envs and use constants instead of strings?

managed/services/supervisord/pmm_config.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	return
+}
+
+var pmmTemplate = template.Must(template.New("").Option("missingkey=error").Parse(`
+[unix_http_server]
+chmod = 0700
+username = dummy
+password = dummy
+
```
PavelKhripkov, line 85: Isn't it better to have config in a separate ini file?

rishat-ishbulatov: What you mean? )

managed/testdata/supervisord.d/pmm-db_enabled.ini
```diff
@@ -0,0 +1,138 @@
...
+autostart = false
+startretries = 3
+startsecs = 1
+stopsignal = TERM
+stopwaitsecs = 300
+stdout_logfile = /srv/logs/pmm-update-perform-init.log
+stdout_logfile_maxbytes = 50MB
+stdout_logfile_backups = 3
+redirect_stderr = true
+[program:postgresql]
```
PavelKhripkov, line 38:
```suggestion

[program:postgresql]
```

PavelKhripkov:
If I understood correctly, now we have 2 config files: one original and one from template. It'd hard to maintain them. Is it possible to not duplicate config?

managed/services/supervisord/pmm_config.go
```diff
@@ -0,0 +1,221 @@
...
+		return nil
+	}
+
+	// restore old content
+	defer func() {
+		if err == nil {
+			return
+		}
+		if resErr := os.WriteFile(path, oldCfg, 0o644); resErr != nil { //nolint:gosec
+			err = errors.Wrapf(resErr, "failed to restore config: %s", err.Error())
```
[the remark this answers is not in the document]
rishat-ishbulatov: This place redone

managed/services/supervisord/pmm_config.go
```diff
@@ -0,0 +1,221 @@
...
+		// nothing to change
+		return nil
+	}
+
+	// restore old content
+	defer func() {
+		if err == nil {
+			return
+		}
+		if resErr := os.WriteFile(path, oldCfg, 0o644); resErr != nil { //nolint:gosec
```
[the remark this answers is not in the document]
rishat-ishbulatov: after negotiations in the slack we decided to leave this place as it is since in any case there will be an error message.

rishat-ishbulatov:
> If I understood correctly, now we have 2 config files: one original and one from template. It'd hard to maintain them. Is it possible to not duplicate config?

It's good point. I personally think that reading from a template would be better. That is, there will be a single approach in this for the supervisor. I wonder what @atymchuk and @BupycHuk  think about this

managed/testdata/supervisord.d/pmm-db_enabled.ini
rishat-ishbulatov: Ok I'll look for a suitable solution

atymchuk:
> > If I understood correctly, now we have 2 config files: one original and one from template. It'd hard to maintain them. Is it possible to not duplicate config?
> 
> It's good point. I personally think that reading from a template would be better. That is, there will be a single approach in this for the supervisor. I wonder what @atymchuk and @BupycHuk think about this

Unfortunately, we need both. The problem is that `supervisord` always starts first. It needs to see this config (pmm.ini) on the filesystem, otherwise it won't be able to properly launch the processes. This may trigger bigger problems and may need refactoring, which may easily lead to scope creep.

I understand the desire to keep things DRY, but I'm not sure this is feasible in this case. Btw, this is not the only instance where we are forced to sacrifice DRY.

atymchuk:
Rishat, we usually try to fix the linter warnings before asking for review, but I realized that you may not necessarily see them :)

Check them out anyway, they look like this one below
Screenshot 2023-02-16 at 01 45 33

managed/cmd/pmm-managed-init/main.go
```diff
@@ -34,6 +35,12 @@ func main() {
 	if on, _ := strconv.ParseBool(os.Getenv("PMM_TRACE")); on {
 		logrus.SetLevel(logrus.TraceLevel)
 	}
+	if ok, _ := strconv.ParseBool(os.Getenv("PMM_DISABLE_BUILTIN_POSTGRES")); ok {
+		if err := supervisord.SavePMMConfig(map[string]any{"DisableInternalDB": ok}); err != nil {
+			logrus.Errorf("PMM Server configuration error: %s.", err)
+			os.Exit(1)
+		}
+	}
```
BupycHuk, line 43: we need to reload supervisord after that

managed/services/supervisord/pmm_config.go
```diff
@@ -0,0 +1,221 @@
...
+	}()
+
+	if err = os.WriteFile(path, cfg, 0o644); err != nil { //nolint:gosec
+		err = errors.Wrap(err, "failed to write new config")
+	}
+	return
+}
+
+var pmmTemplate = template.Must(template.New("").Option("missingkey=error").Parse(`
+[unix_http_server]
```
BupycHuk, line 81: since we have it here, should we remove it from ansible?

managed/cmd/pmm-managed-init/main.go
rishat-ishbulatov: In general, this is not required there, since pmm-managed-init, judging by [entrypoint.sh](https://github.com/percona/pmm/blob/main/build/docker/server/entrypoint.sh#L25-L30), is started before supervisord.
`# pmm-managed-init validates environment variables.
pmm-managed-init

# Start supervisor in foreground
exec supervisord -n -c /etc/supervisord.conf`

managed/cmd/pmm-managed-init/main.go
```diff
@@ -34,6 +35,12 @@ func main() {
 	if on, _ := strconv.ParseBool(os.Getenv("PMM_TRACE")); on {
 		logrus.SetLevel(logrus.TraceLevel)
 	}
+	if ok, _ := strconv.ParseBool(os.Getenv("PMM_DISABLE_BUILTIN_POSTGRES")); ok {
```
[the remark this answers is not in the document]
rishat-ishbulatov: this place refactored

rishat-ishbulatov:
> Rishat, we usually try to fix the linter warnings before asking for review, but I realized that you may not necessarily see them :)
> 
> Check them out anyway, they look like this one below Screenshot 2023-02-16 at 01 45 33

linter errors fixed

managed/testdata/supervisord.d/pmm-db_enabled.ini
rishat-ishbulatov: this place fixed

managed/services/supervisord/pmm_config.go
```diff
@@ -93,19 +94,19 @@ nodaemon = true
...
 
-; we rewrite autostart to true during update or build.
 [program:pmm-update-perform-init]
 command = /usr/sbin/pmm-update -run-playbook -playbook=/usr/share/pmm-update/ansible/playbook/tasks/init.yml
 directory = /
 autorestart = unexpected
 priority=-1
 exitcodes = 0
-autostart = false
+autostart = true
```
BupycHuk, line 109: uhm, please revert it

managed/cmd/pmm-managed-init/main.go
```diff
@@ -58,4 +51,20 @@ func main() {
 		logrus.Errorf("Configuration error: %s.", err)
 		os.Exit(1)
 	}
+
+	pmmConfigParams := make(map[string]any, 2)
+	pmmConfigParams["DisableInternalDB"], _ = strconv.ParseBool(os.Getenv("PMM_DISABLE_BUILTIN_POSTGRES"))
+	pmmConfigParams["DisableSupervisor"] = !isDocker()
```
BupycHuk, line 57: why do we do it?

managed/services/supervisord/pmm_config.go
```diff
@@ -93,19 +94,19 @@ nodaemon = true
 nocleanup = false
 user = root
 strip_ansi = false
+{{- end }}
 
 [supervisorctl]
 username = dummy
 password = dummy
 
-; we rewrite autostart to true during update or build.
```
BupycHuk: did you find out why do we do it?

rishat-ishbulatov: reverted to previous state

managed/cmd/pmm-managed-init/main.go
rishat-ishbulatov: line removed

managed/services/supervisord/pmm_config.go
rishat-ishbulatov: reverted to previous state

PR #1703, opened 2023-02-13
PMM-11439 internal pmm agent collects from external db

rishat-ishbulatov:
[PMM-11439](https://jira.percona.com/browse/PMM-11439)

[Build](https://github.com/Percona-Lab/pmm-submodules/pull/3061)


managed/models/database.go
```diff
@@ -1057,30 +1060,66 @@ func setupFixture1(q *reform.Querier, username, password string) error {
 		return err
 	}
 
+	address := strings.Split(params.Address, ":") // TODO: find better solution
```
JiriCtvrtka, line 1063: Maybe 

`host, port, err := net.SplitHostPort("hello.com:81")`

managed/models/agent_model.go
```diff
@@ -474,14 +474,14 @@ func (s *Agent) DSN(service *Service, dialTimeout time.Duration, database string
...
 					switch key {
 					case caFilePlaceholder:
-						q.Add("sslrootcert", tdp.Left+".TextFiles."+caFilePlaceholder+tdp.Right)
+						q.Add("sslrootcert", value)
 					case certificateFilePlaceholder:
-						q.Add("sslcert", tdp.Left+".TextFiles."+certificateFilePlaceholder+tdp.Right)
+						q.Add("sslcert", value)
 					case certificateKeyFilePlaceholder:
-						q.Add("sslkey", tdp.Left+".TextFiles."+certificateKeyFilePlaceholder+tdp.Right)
+						q.Add("sslkey", value)
```
BupycHuk, line 484: it will work only for this case and breaks logic for other usages

managed/models/database.go
```diff
@@ -37,6 +38,8 @@ const (
 	PMMServerPostgreSQLServiceName = "pmm-server-postgresql"
 	// minPGVersion stands for minimal required PostgreSQL server version for PMM Server.
 	minPGVersion float64 = 14
+	// DefataultPostgreSQLAddr represent default local PostgreSQL database server address.
+	DefataultPostgreSQLAddr = "127.0.0.1:5432"
```
BupycHuk, line 42:
```suggestion
	DefaultPostgreSQLAddr = "127.0.0.1:5432"
```

managed/models/database.go
```diff
@@ -1057,30 +1060,66 @@ func setupFixture1(q *reform.Querier, username, password string) error {
...
 	}
+
+	tls := params.SSLMode != "disable"
+	tlsSkipVerify := params.SSLMode == "disable" || params.SSLMode == "verify-ca"
+	var psqlOps *PostgreSQLOptions
+	if tls {
+		psqlOps = &PostgreSQLOptions{
+			SSLCa:   params.SSLCAPath,
+			SSLCert: params.SSLCertPath,
+			SSLKey:  params.SSLKeyPath,
```
BupycHuk, line 1100: we should read these files and store content in DB.

BupycHuk (changes requested):

managed/models/database.go
rishat-ishbulatov: in that case if we pass docker-compose like service name "some-server" without port we get an error that port is not provided

managed/models/database.go
rishat-ishbulatov: redone to store content of this files

managed/models/database.go
rishat-ishbulatov: typo fixed

managed/models/agent_model.go
rishat-ishbulatov: moved to initial state

managed/models/database.go
rishat-ishbulatov: split host port to parse address done

JiriCtvrtka: So then maybe it can be:
If it contains : then return service name
otherwise net.SplitHostPort. 

If its still actual after changes and its up to you.

rishat-ishbulatov: proposal is implemented)

gen1us2k (approved):

managed/models/database.go
```diff
@@ -1058,36 +1061,64 @@ func setupFixture1(q *reform.Querier, username, password string) error {
 	if _, err = CreateNodeExporter(q, PMMServerAgentID, nil, false, []string{}, nil, ""); err != nil {
 		return err
 	}
+	address, port, err := preparePGAddress(params.Address)
+	if err != nil {
+		return err
+	}
+	if params.Address != DefaultPostgreSQLAddr {
+		if node, err = CreateNode(q, RemoteNodeType, &CreateNodeParams{
+			NodeName: "pmm-server-db",
```
PavelKhripkov, line 1070: It's better to keep strings in constants as it's done for `PMMServerPostgreSQLServiceName`.

managed/models/database.go
```diff
@@ -1058,36 +1061,64 @@ func setupFixture1(q *reform.Querier, username, password string) error {
...
 
 	// create PostgreSQL Service and associated Agents
 	service, err := AddNewService(q, PostgreSQLServiceType, &AddDBMSServiceParams{
 		ServiceName: PMMServerPostgreSQLServiceName,
 		NodeID:      node.NodeID,
-		Address:     pointer.ToString("127.0.0.1"),
-		Port:        pointer.ToUint16(5432),
+		Database:    params.Name,
+		Address:     &node.Address,
+		Port:        pointer.ToUint16(port),
```
PavelKhripkov, line 1083:
```suggestion
		Port:        &port,
```

managed/models/database.go
```diff
@@ -1096,3 +1127,19 @@ func setupFixture2(q *reform.Querier, username, password string) error {
 
 	return nil
 }
+
+// preparePGAddress prepares PostgreSQL address.
+func preparePGAddress(address string) (string, uint16, error) {
+	if idx := strings.LastIndexByte(address, ':'); idx < 0 {
```
PavelKhripkov, line 1133:
```suggestion
	if !strings.Contains(addressStr, ":") {
```

managed/models/database.go
```diff
@@ -1096,3 +1127,19 @@ func setupFixture2(q *reform.Querier, username, password string) error {
 
 	return nil
 }
+
+// preparePGAddress prepares PostgreSQL address.
+func preparePGAddress(address string) (string, uint16, error) {
```
PavelKhripkov, line 1132:
```suggestion
// parsePGAddress parses PostgreSQL address into address:port; if no port specified returns default port number.
func parsePGAddress(address string) (string, uint16, error) {
```

managed/models/database.go
```diff
@@ -1058,36 +1061,64 @@ func setupFixture1(q *reform.Querier, username, password string) error {
...
+			if len(path) == 0 {
+				continue
+			}
+			content, err := os.ReadFile(path) //nolint:gosec
+			if err != nil {
+				return err
+			}
+			*field = string(content)
+		}
+	}
```
PavelKhripkov, line 1113: Shouldn't we give user the option to not cache certs?

managed/models/database.go
```diff
@@ -1058,36 +1061,64 @@ func setupFixture1(q *reform.Querier, username, password string) error {
...
+	_, err = CreateAgent(q, PostgresExporterType, ap)
```
PavelKhripkov, line 1114: We're creating agent model here. Also we need running agents on the node side. In case user specifies external db, would this logic work as well?

managed/models/database.go
rishat-ishbulatov: constant added

managed/models/database.go
rishat-ishbulatov: This case is not described in the task. I wonder what @atymchuk and @BupycHuk think about this.

managed/models/database.go
rishat-ishbulatov: What can interfere?

PavelKhripkov: I don't know how that part works, so I just wonder. If there are no problems then ok)

managed/models/database.go
```diff
@@ -39,6 +41,8 @@ const (
 	minPGVersion float64 = 14
 	// DefataultPostgreSQLAddr represent default local PostgreSQL database server address.
```
PavelKhripkov, line 42:
```suggestion
	// DefaultPostgreSQLAddr represent default local PostgreSQL database server address.
```

PavelKhripkov (approved):

managed/models/database.go
BupycHuk: what do you mean by not cache certs? we store custom certs to be able to connect to PostgreSQL with custom certs.
BTW @rishat-ishbulatov it's better to use `path == ""` instead of `len(path) == 0`

managed/models/database.go
BupycHuk: Hi, in this case, the agent runs on pmm-server and collects metrics from external PostgreSQL as we do for remote instances.

managed/models/database.go
```diff
@@ -1067,36 +1071,64 @@ func setupFixture1(q *reform.Querier, username, password string) error {
...
+	if ap.TLS {
+		ap.PostgreSQLOptions = &PostgreSQLOptions{}
+		for path, field := range map[string]*string{
+			params.SSLCAPath:   &ap.PostgreSQLOptions.SSLCa,
+			params.SSLCertPath: &ap.PostgreSQLOptions.SSLCert,
+			params.SSLKeyPath:  &ap.PostgreSQLOptions.SSLKey,
+		} {
+			if len(path) == 0 {
+				continue
+			}
```
BupycHuk, line 1116:
```suggestion
			if path == "" {
				continue
			}
```

BupycHuk (approved):

JiriCtvrtka (approved):

Issue #1541, opened 2023-01-04
Enable temporarily disabled linters


BupycHuk:
### Description

In the golangci configuration file, we have a bunch of temporarily [disabled](https://github.com/percona/pmm/blob/main/.golangci.yml#L93-L129) [linters](https://golangci-lint.run/usage/linters/).

### Suggested solution

We want to review those disabled, one by one. Our options are: 
1) either enable the rule and fix potential warning messages or 
2) disable the rule and leave a comment explaining why.

It's not impossible to disable a couple of rules if we realize we don't need them.

### Additional context

list of currently disabled linters that
- [x] bodyclose
- [x] containedctx
- [x] contextcheck
- [x] deadcode
- [x] dupl
- [x] errcheck
- [x] errorlint
- [x] execinquery
- [x] forbidigo
- [x] forcetypeassert
- [x] gocognit
- [x] gocritic
- [x] godot
- [x] godox
- [x] golint
- [x] gosimple
- [x] interfacebloat
- [x] ireturn
- [x] maintidx
- [x] nakedret
- [x] nestif
- [x] nilnil
- [x] nilerr
- [x] nonamedreturns
- [x] noctx
- [x] nosnakecase
- [x] nosprintfhostport
- [x] paralleltest
- [x] prealloc
- [x] predeclared
- [x] revive
- [x] tagliatelle
- [x] thelper
- [x] tparallel
- [x] usestdlibvars
- [x] whitespace

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

sajdakabir:
Hey, @BupycHuk I would love to this issue can you please assign it to me?

artemgavrilov:
Hi @sajdakabir, thank you for your interest! This is cumulative issue, we don't expect fixing all linters at once. Moreover it will be very hard to review. So please open individual PR for every linter fix. If you afraid that someone else will start working on the same linter, you can just add comment here to inform others that you already started (BTW we don't have much activity here).

atymchuk:
@JiriCtvrtka Could you please take care of the `execinquery` rule? It seems to be mostly related to slowlog functionality, which is part of QAN I'd assume.

https://pasteboard.co/QfvSXujpeNQ4.png

JiriCtvrtka:
@atymchuk sure. I will take a look right after priorities for this week.

PR #1711, opened 2023-02-13
PMM-7 re-enable the `nilnil` linter rule

atymchuk:
PMM-7

Refers to: https://github.com/percona/pmm/issues/1541

Note: this PR re-enables the `ninil` linter rule and fixes the warnings thereof. Even though I mostly had to resort to disabling all warnings with `nolint`, IMO it is still a good rule to follow for the sake of more semantic code. 

Fixing this rule in the code would have required us to re-test the code in a number of components, which wasn't the gole of this PR.

### DIFF
--- a/managed/services/checks/funcs.go
+++ b/managed/services/checks/funcs.go
@@ -111,7 +111,7 @@ func GetAdditionalContext() map[string]starlark.GoFunc {
 }
 
 // ipIsPrivate accepts a single string argument (IP address or a network) and
-// returns true for a private address, otherwise false. In case of an invalid argument nil.
+// returns true for a private address, otherwise false. It returns nil in case of an invalid argument.
 func ipIsPrivate(args ...interface{}) (interface{}, error) {
 	log := logrus.WithField("component", "checks")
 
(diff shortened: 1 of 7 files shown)


managed/services/checks/funcs.go
```diff
@@ -130,7 +130,7 @@ func ipIsPrivate(args ...interface{}) (interface{}, error) {
 		_, net, err := net.ParseCIDR(ip)
 		if err != nil {
 			log.Errorf("invalid ip/network address: %q", ip)
-			return nil, nil
+			return nil, errors.Errorf("invalid ip/network address: %q", ip)
```
atymchuk, line 133: @artemgavrilov It seems like it's legit to return an error in this one case. WDYT?

artemgavrilov: This behavior was intentional as it's described if func description. I don't know why it was done like this, but this function was made for advisors (it used by advisors itself), so most likely we can't change it just like this. It may break existing advisors that rely on such output.

atymchuk: Yes, it actually breaks the tests, so I reverted.

BupycHuk (approved):

artemgavrilov (approved):

idoqo (approved):

PR #1720, opened 2023-02-14
PMM-11651: fix full access role bug

michal-kralik:
When full access role is combined with limited role, full access was not granted

PMM-11651

Build: SUBMODULES-3330


recharte (approved):

managed/services/grafana/auth_server.go
```diff
@@ -356,9 +356,14 @@ func (s *AuthServer) getFiltersForVMProxy(userID int) ([]string, error) {
 
 	filters := make([]string, 0, len(roles))
 	for _, r := range roles {
-		if r.Filter != "" {
-			filters = append(filters, r.Filter)
+		if r.Filter == "" {
+			// Special case when a user has assigned a role with no filters.
+			// In this case it's irrelevant what other roles the user has assigned.
+			// The user shall have full access.
+			return nil, nil
```
atymchuk, line 363:
```suggestion
			return nil, nil //nolint:nilnil
```
We have a rule now :)

managed/services/grafana/auth_server.go
```diff
@@ -356,9 +356,14 @@ func (s *AuthServer) getFiltersForVMProxy(userID int) ([]string, error) {
 
 	filters := make([]string, 0, len(roles))
 	for _, r := range roles {
-		if r.Filter != "" {
-			filters = append(filters, r.Filter)
+		if r.Filter == "" {
+			// Special case when a user has assigned a role with no filters.
+			// In this case it's irrelevant what other roles the user has assigned.
```
atymchuk, line 361:
```suggestion
			// In this case it's irrelevant what other roles are assigned to the user.
```

managed/services/grafana/auth_server.go
michal-kralik: I changed it to return empty slice.
I'm wondering why nothing complained about the `nilnil` rule.

atymchuk: Because you haven't rebased after I merged that rule, no?

atymchuk (approved):

PR #1726, opened 2023-02-15
PMM-7 fix the use of deprecated `RdsServiceID` constant

atymchuk:
PMM-7

Why:
It was showing up in every PR as a linter warning, too annoying...

### DIFF
--- a/managed/services/management/rds.go
+++ b/managed/services/management/rds.go
@@ -120,7 +120,7 @@ func listRegions(partitions []string) []string {
 				continue
 			}
 
-			for r := range partition.Services()[endpoints.RdsServiceID].Regions() {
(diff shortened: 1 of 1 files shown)


managed/services/management/rds.go
```diff
@@ -120,7 +120,7 @@ func listRegions(partitions []string) []string {
 				continue
 			}
 
-			for r := range partition.Services()[endpoints.RdsServiceID].Regions() {
+			for r := range partition.Services()[rds.EndpointsID].Regions() {
```
atymchuk, line 123: Note: those two constants are equivalent, so this replacement should't change the behavior.

BupycHuk (approved):

JiriCtvrtka (approved):

PR #1796, opened 2023-03-01
PMM-11742 cli tests refactor

vyarosh:
PMM-11742


cli-tests/tests/pmm-cli/server/install.spec.ts
```diff
@@ -5,57 +5,60 @@ import { teardown } from '@tests/helpers/containers'
...
+        --admin-password=${adminPw}
+        --docker-image="${imageName}"
+        --https-listen-port=1443
+        --http-listen-port=1080
+        --container-name=${containerName}
+        --volume-name=${volumeName}`
+    )
+    await output.assertSuccess()
+    expect(output.stderr).toContain('Starting PMM Server')
+  });
```
michal-kralik, line 27: Does this approach work with the parallel mode?
When writing the tests I experienced issues with `beforeAll` and the parallel mode since it is executed by every worker.

vyarosh: It does, the fluent wait has been missing. But it's not effective. See: https://github.com/percona/pmm/actions/runs/4801870077/jobs/8544628562?pr=1796
The image has been downloaded 5 times. To have an optimization I would ask for a "download" command, so that the image will be downloaded once and the reused for setup without downloading. Or this spec could be switched to chain execution.

michal-kralik: Can we use project dependencies to run this only once before tests? 
https://playwright.dev/docs/test-projects#dependencies

vyarosh: We will to perform "docker pull" to download all images before tests. Though it's limited since setup result is not visible in tests. The official recommended way to pass results to tests via env.VAR which makes limitations and requires some redesign

michal-kralik: I don't think we need to pull the images. cli pulls the images automatically and we shall test this part as well

cli-tests/helpers/cliHelper.ts
```diff
@@ -1,34 +1,43 @@
...
-  } else {
-    assert.ok(code !== 0, `The command ${command} was expected to return with failure but found to be executing without any error, the return code found ${code}`);
-  }
-
-  if (!getError) return stdout as string;
-
-  return stderr as string;
-}
+/**
+ * Shell(sh) exec() wrapper to use outside outside {@link test}
```
michal-kralik, line 6:
```suggestion
 * Shell(sh) exec() wrapper to use outside {@link test}
```

cli-tests/tests/pmm-cli/server/install.spec.ts
```diff
@@ -1,61 +1,68 @@
...
-      await teardown([`^${containerName}$`], [volumeName])
-    }
-  })
-})
+import { test, expect } from '@playwright/test';
+import * as cli from '@helpers/cliHelper';
+import PMMRestClient from '@tests/support/types/request';
+import { teardown } from '@tests/helpers/containers';
+
+test.describe('Install PMM Server respects relevant flags', async () => {
```
michal-kralik, line 6: Let's run the tests in parallel. Otherwise it takes too long to finish.

michal-kralik (changes requested):

cli-tests/playwright.config.ts
```diff
@@ -1,5 +1,5 @@
-import type { PlaywrightTestConfig } from '@playwright/test';
-
+import { defineConfig, devices } from '@playwright/test';
+// import { devices } from '@playwright/test';
```
michal-kralik, line 2: A leftover

cli-tests/package.json
```diff
@@ -13,6 +13,7 @@
   },
   "dependencies": {
     "@playwright/test": "^1.32.3",
+    "@types/shelljs": "^0.8.12",
```
michal-kralik, line 16: This can go to `devDependencies`

cli-tests/support/pmm-bin.setup.ts
```diff
@@ -0,0 +1,18 @@
...
+  process.env.server_image = newImage;
+  const output = await cli.exec(`docker pull ${newImage}`);
+  await output.assertSuccess();
+});
+
+setup(`pull ${oldImage}`, async () => {
+  process.env.server_image_old = oldImage;
+  const output = await cli.exec(`docker pull ${oldImage}`);
+  await output.assertSuccess();
+});
```
michal-kralik, line 18: Let's not pull the images here. We want to make sure `pmm` cli can pull them properly.

cli-tests/tests/pmm-cli/server/install.spec.ts
```diff
@@ -1,61 +1,63 @@
...
+        --https-listen-port=1443
+        --http-listen-port=1080
+        --container-name=${containerName}
+        --volume-name=${volumeName}`);
+    await output.assertSuccess();
+
+    await expect(async () => {
+      expect(output.stderr).toContain('Starting PMM Server');
+    }).toPass({
+      // Probe, wait 1s, probe, wait 2s, probe, wait 2s, probe, wait 2s, probe, ....
```
michal-kralik, line 33: Why do we need to retry here? I believe the command has already finished, correct?

cli-tests/tests/pmm-cli/server/install.spec.ts
```diff
@@ -1,61 +1,63 @@
...
+    output = await cli.exec('docker ps --format="{{.Names}}"');
+    await output.outContains(containerName);
+
+    // verify --volume-name
+    output = await cli.exec('docker volume ls --format="{{.Name}}"');
+    await output.outContains(volumeName);
+
+    // verify --docker-image
+    output = await cli.exec('docker ps --format="{{.Names}} {{.Image}}"');
+    await output.outContains(`${containerName} ${imageName}`);
```
michal-kralik, line 61: Using `outContains` does not properly check for the provided value.
Since more tests are running in parallel, a substring will provide a false successful check.
Such as
```js
const containerName1 = "container"
const containerName2 = "container-with-data"

// This will match containerName2 as well
await output.outContains(containerName1)
```

cli-tests/tests/pmm-cli/server/upgrade.spec.ts
```diff
@@ -1,17 +1,17 @@
...
-    const oldContainerName = 'pmm-server-upgrade-1'
-    const newContainerName = 'pmm-server-upgrade-1-new'
-    const oldImage = 'percona/pmm-server:2.32.0'
-    const newImage = 'percona/pmm-server:2.33.0'
-    const volumeName = 'pmm-data-upgrade-1'
+    const oldContainerName = 'pmm-server-upgrade-1';
+    const newContainerName = 'pmm-server-upgrade-1-new';
+    const oldImage = 'percona/pmm-server:2.32.0';
+    const newImage = 'percona/pmm-server:2.33.0';
+    const volumeName = 'pmm-data-upgrade-1';
```
michal-kralik, line 14: There's a lot of these "styling" updates.
I recommend using a library to check and apply styles, such as `Standard.js`

michal-kralik (changes requested):

cli-tests/tests/pmm-cli/server/install.spec.ts
vyarosh: It has failed on "parallel" command run on this check. So it has been left as is - it will exit on the 1st try and will apply reties only on fail. Though I left is as a safe block to move for any test structure.

cli-tests/package.json
vyarosh: I'm not sure really sure. Since this is a subproject implementing "tests" only. So it's original flow is "dev" only. And we would need to add and use "dev build" for tests. And we don't have `prod` and `dev` state for tests subprojects in our build process. Or I'm missing this?

cli-tests/support/pmm-bin.setup.ts
vyarosh: I cannot agree on given parallel run. Why would we want tests to download the same image in 6 thread same time?
I will ad a single test to check the download. It actually present - test which install defaults - it downloads the latest image.
Also please consider this as improvement point to un-hardcode pmm-server versions.

cli-tests/playwright.config.ts
vyarosh: fixed

cli-tests/tests/pmm-cli/server/install.spec.ts
vyarosh: Added quick fix without test report wrapper

cli-tests/package.json
michal-kralik: All dependencies which are not required to run the app are considered `devDependencies`.
Since `@types/*` is not required to run the app, it shall be moved to `devDependencies`

cli-tests/support/pmm-bin.setup.ts
michal-kralik: afaik `docker` cli downloads the image only once, even with multiple concurrent requests to download the image.
It may not be the same case when using API but I did not notice the image being downloaded multiple times when running locally.

cli-tests/tests/pmm-cli/server/install.spec.ts
michal-kralik: The code appears to be trying to find `Starting PMM Server` string in the `output.stderr` which does not change when it's retried.
Am I missing anything?

cli-tests/tests/pmm-cli/server/install.spec.ts
michal-kralik: The same issue applies to `volumeNam` and `imageName`

cli-tests/tests/pmm-cli/server/install.spec.ts
vyarosh: No, you are not missing anything. Please see the original code added by you: https://github.com/percona/pmm/blob/main/cli-tests/tests/pmm-cli/server/install.spec.ts#L29
Also please see the logs: https://github.com/percona/pmm/actions/runs/4830926936/jobs/8607722662
Here is a linked screenshot: https://www.screencast.com/t/bjYMvObHdrd
Green section is "stdout" pipe and red section is "stderr" pipe. pmm binary makes non-error output in the error pipe.
Would you like me to rise a defect for this?

michal-kralik: It should not be stderr, not sure why it is. Let's open a ticket for it.
The original comment is still valid though - I don't think retrying this part adds any value since the output is always the same during retries.

cli-tests/package.json
vyarosh: Got it. I've missed to think from this point. Moved

cli-tests/support/pmm-bin.setup.ts
vyarosh: I've did a research. Indeed docker downloads same image in the single thread for any amount of started to pull terminals.
Also downloading progress does not trash the logs.
These were 2 main reason to have pull before the tests. Removed.

cli-tests/tests/pmm-cli/server/install.spec.ts
vyarosh: fixed.

cli-tests/tests/pmm-cli/server/install.spec.ts
vyarosh: Added same check

cli-tests/tests/pmm-cli/server/install.spec.ts
vyarosh: PMM-12084 - added ticket. And removed retries.

michal-kralik (approved):

PR #1810, opened 2023-03-02
PMM-10974 PGSM as default query source for PG.

JiriCtvrtka:
PMM-10974

- [ ] API Docs updated


percona-csalguero (approved):

michal-kralik (changes requested):
I suggest we document this falls back to `pgstatements`.

JiriCtvrtka:
@michal-kralik What do you mean please?

rishat-ishbulatov (approved):

JiriCtvrtka:
@michal-kralik We cannot easily verify if PGSM is available or not for now. So without this it cannot simply fallback to pgstatstatements. Right now, customer needs to be sure that PGSM is available in his PG. For Percona distributions its built-in. Its same behaviour for like MySQL.

michal-kralik:
Ok, that's not clear from the ticket.
This means it's a breaking change for some customers. Is it expected?

JiriCtvrtka:
Breaking changes in case of readding service by script for example.

michal-kralik (approved):

PR #1825, opened 2023-03-06
PMM-11707: format json result without marshaler

michal-kralik:
PMM-11707

Link to the Feature Build: SUBMODULES-3113

### DIFF
--- a/admin/cli/cli.go
+++ b/admin/cli/cli.go
@@ -161,6 +161,8 @@ func printSuccessResult(opts *flags.GlobalFlags, res commands.Result) {
 			fmt.Printf("%s\n", b) //nolint:forbidigo
 		case fmt.Stringer:
 			logrus.Info(res.String())
+		default:
+			logrus.Infof("Result: %#v.", res)
 		}
 	} else {
 		fmt.Println(res.String()) //nolint:forbidigo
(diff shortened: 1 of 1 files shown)


BupycHuk:
can we write unit tests for this output?

admin/cli/cli.go
```diff
@@ -151,12 +151,17 @@ func printResponse(opts *flags.GlobalFlags, res commands.Result, err error) erro
 
 func printSuccessResult(opts *flags.GlobalFlags, res commands.Result) {
 	if opts.JSON {
-		b, jErr := json.Marshal(res)
-		if jErr != nil {
-			logrus.Infof("Result: %#v.", res)
-			logrus.Panicf("Failed to marshal result to JSON.\n%s.\nPlease report this bug.", jErr)
+		switch res.(type) {
```
artemgavrilov, line 154: I guess we need some `default` here.

admin/cli/cli.go
```diff
@@ -151,12 +151,19 @@ func printResponse(opts *flags.GlobalFlags, res commands.Result, err error) erro
 
 func printSuccessResult(opts *flags.GlobalFlags, res commands.Result) {
 	if opts.JSON {
-		b, jErr := json.Marshal(res)
-		if jErr != nil {
+		switch res.(type) {
+		case json.Marshaler:
+			b, jErr := json.Marshal(res)
```
percona-csalguero, line 156: if `res` is a `json.Marshaler`, insn't there a `res.Marshal` instead of calling `json.Marshal(res)`?

michal-kralik:
@percona-csalguero is right. This is not a correct approach. Closing in favor of #1827

PR #1849, opened 2023-03-09
PMM-2852 Comments in QAN filter.

JiriCtvrtka:
PMM-2852

Link to the Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/3235

- [ ] API Docs updated


qan-api2/models/reporter.go
```diff
@@ -510,8 +511,17 @@ func (r *Reporter) SelectFilters(ctx context.Context, periodStartFromSec, period
 			if mainMetricPerSec == 0 {
 				total = totals[label.key]
 			}
+
+			var value string
+			switch v := label.value.(type) {
+			case string:
+				value = v
+			case []string:
+				value = strings.Join(v, ",")
```
michal-kralik, line 520: Is it worth adding a `default` here and at least log we received unexpected value?

agent/queryparser/helpers.go
```diff
@@ -0,0 +1,174 @@
...
+	// to compile regexp only once
+	multilineOnce.Do(func() {
+		multilineRegexp, errMultiline = regexp.Compile(`(?s)\/\*(.*?)\*\/`)
+	})
+	if errMultiline != nil {
+		return errMultiline
+	}
+
+	return nil
+}
```
michal-kralik, line 145: What about calling this in `init()` for the package?

agent/queryparser/helpers.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+
+	return comments, nil
+}
+
+func prepareMultilineRegexp() error {
+	// to compile regexp only once
+	multilineOnce.Do(func() {
+		multilineRegexp, errMultiline = regexp.Compile(`(?s)\/\*(.*?)\*\/`)
```
michal-kralik, line 138: Can the comment blocks be escaped, such as `\/*`?

agent/queryparser/parser.go
```diff
@@ -39,3 +39,36 @@ func MySQL(q string) (string, uint32, error) {
...
+// MySQLComments parse query and return its comments. Can parse multi comments.
+// Multi comments support should be dropped in future MySQL versions.
+// Doc: https://dev.mysql.com/doc/refman/8.0/en/comments.html
+func MySQLComments(q string) ([]string, error) {
+	comments, err := parseMySQLComments(q)
+	if err != nil {
+		return nil, err
+	}
+
+	var res []string //nolint:prealloc
```
michal-kralik, line 52: Any reason we don't allocate the slice on initialization?

agent/queryparser/parser.go
```diff
@@ -39,3 +39,36 @@ func MySQL(q string) (string, uint32, error) {
...
+
+// PostgreSQLComments parse query and return its comments. Can parse multi comments.
+// Doc: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-COMMENTS
+func PostgreSQLComments(q string) ([]string, error) {
+	comments, err := parsePostgreSQLComments(q)
+	if err != nil {
+		return nil, err
+	}
+
+	var res []string //nolint:prealloc
```
michal-kralik, line 68: Any reason we don't allocate the slice on initialization?

agent/queryparser/helpers.go
JiriCtvrtka: I dont like init because it no so transparent and maybe user will not use action at all. So in case of init it will be triggered for no reason. I am suggesting to have it like that.

agent/queryparser/helpers.go
JiriCtvrtka: Since we scrapping query and comments from DB I think its not valid case.

agent/agents/postgres/pgstatstatements/models.go
```diff
@@ -72,9 +72,10 @@ type pgStatStatementsExtended struct {
...
 	Tables           []string
 	IsQueryTruncated bool
+	Comments         []string
 }
 
 func (e *pgStatStatementsExtended) String() string {
-	return fmt.Sprintf("%q %q %v: %d: %s (truncated = %t)",
-		e.Database, e.Username, e.Tables, e.QueryID, e.Query, e.IsQueryTruncated)
+	return fmt.Sprintf("%q %q %v: %d: %s (truncated = %t) %v",
+		e.Database, e.Username, e.Tables, e.QueryID, e.Query, e.IsQueryTruncated, e.Comments)
```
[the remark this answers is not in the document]
JiriCtvrtka: Comments are already trimmed here. It coming from: https://github.com/percona/pmm/pull/1849/files#diff-cec84468e3ff94b502a61a7030e454e709d31b2038eac3d0f0089a55f9b5b1bfR152-R164

agent/agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -548,6 +551,13 @@ func (m *PGStatMonitorQAN) makeBuckets(current, cache map[time.Time]map[string]*
 				mb.Common.ExampleType = agentpb.ExampleType_RANDOM
 			}
 
+			if !m.disableCommentsParsing && currentPSM.Comments != nil {
+				value := strings.ReplaceAll(*currentPSM.Comments, "/*", "")
+				value = strings.ReplaceAll(value, "*/", "")
+				value = strings.TrimSpace(value)
```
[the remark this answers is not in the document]
JiriCtvrtka: In this specific case comments are provided by PGSM and there is no case to have it like that.

qan-api2/models/reporter.go
JiriCtvrtka: For unexpected values value will be empty anyway, but sure why not.

agent/queryparser/parser.go
JiriCtvrtka: I rather doing this when its really needed, but it you want I can change it for sure.

agent/queryparser/parser.go
JiriCtvrtka: I rather doing this when its really needed, but it you want I can change it for sure.

qan-api2/models/reporter.go
JiriCtvrtka: @michal-kralik I took a look and its inside of loop. We dont have logger here passed and it can return only error. I am suggesting to keep it as it is to dont break whole loop and everything else in case of problem. Now we just skipping it.

agent/queryparser/helpers.go
```diff
@@ -0,0 +1,174 @@
...
+	for c := range dashComments {
+		comments[c] = true
+	}
+
+	return comments, nil
+}
+
+func parseSinglelineComments(q, startChar string) (map[string]bool, error) {
+	r, err := regexp.Compile(fmt.Sprintf("%s.*", startChar))
+	if err != nil {
```
rishat-ishbulatov, line 116: IMHO don't see much reason checking regexp compile errors on runtime. Maybe 
```
var compiled = regexpr.MustCompite(payload) 
```
at package level would be better? I wonder what is the probability of an error appearing there? )
Here performance [example](https://github.com/golang/go/issues/26623)
```
cat f_test.go
package p

import (
        "regexp"
        "testing"
)

var Sink bool

func BenchmarkCompileRun(b *testing.B) {
        for i := 0; i < b.N; i++ {
                rx := regexp.MustCompile(`[\w\.+-]+@[\w\.-]+\.[\w\.-]+`)
                Sink = rx.MatchString("123456789 foo@bar.etc")
        }
}

func BenchmarkRun(b *testing.B) {
        rx := regexp.MustCompile(`[\w\.+-]+@[\w\.-]+\.[\w\.-]+`)
        for i := 0; i < b.N; i++ {
                Sink = rx.MatchString("123456789 foo@bar.etc")
        }
}
$ go test -bench=.
goos: linux
goarch: amd64
pkg: mvdan.cc/p
BenchmarkCompileRun-4             100000             14160 ns/op
BenchmarkRun-4                   1000000              1121 ns/op
PASS
ok      mvdan.cc/p      2.693s
```

rishat-ishbulatov (approved):

api/agentpb/collector.proto
```diff
@@ -35,6 +35,8 @@ message MetricsBucket {
     string explain_fingerprint = 25;
     // ammount of variables in query.
     uint32 placeholders_count = 26;
+    // Contains extracted comments from queries.
+    repeated string comments = 27;
```
BupycHuk, line 39: why don't we make it a map? in that case whole logic of comments parsing would be in agent instead of splitting it across pmm-managed and pmm-agent

JiriCtvrtka: Yes, right. Let me fix it.

BupycHuk (approved):

PR #1854, opened 2023-03-09
PMM-5081: Prometheus metrics for inventory data

xgrommx:
[PMM-5081](https://jira.percona.com/browse/PMM-5081)

Build:
- https://github.com/Percona-Lab/pmm-submodules/pull/3126


managed/services/inventory/inventory.go
```diff
@@ -0,0 +1,152 @@
...
+
+func NewInventory(db *reform.DB, r agentsRegistry) *Inventory {
+	i := &Inventory{
+		db: db,
+		r:  r,
+	}
+	return i
+}
+
+func (i *Inventory) Describe(chan<- *prom.Desc) {}
```
percona-csalguero, line 69: Just a question because I really don't remember how it works internally.
If describe doesn't return anything, Collect is still going to be called once or twice?
I remember we had a problem with MongoDB exporter, Prometheus was calling DescribeByCollect, resulting in calling Collect twice per cycle.

xgrommx: What should be returned in `Describe`?

percona-csalguero: The list of metrics.

gen1us2k (commented):
can we have some tests please?

managed/services/inventory/inventory.go
ritbl: Why we don't send descriptors in Describe?

ritbl (changes requested):
please see my comment

ritbl:
@xgrommx have you created a dashboard that uses these metrics?

managed/services/inventory/inventory.go
xgrommx: But metrics builded here
https://github.com/percona/pmm/pull/1854/files#diff-e9b7475ea5592bfc2b7287010a7920f4ed8b3d3c632d19d3b34fac7b4a53ce48R131
For any item of items - new metrics 
Because each of metrics have unique labels (unique name of agent, service, etc)

ritbl: @percona-csalguero
yes, it can be called 2 times in case use this one for example `prom.DescribeByCollect` in Descibe.
internally it works this way:
```
func DescribeByCollect(c Collector, descs chan<- *Desc) {
	metrics := make(chan Metric)
	go func() {
		c.Collect(metrics)
		close(metrics)
	}()
	for m := range metrics {
		descs <- m.Desc()
	}
}
```
i.e. it calls collect and resends these to desc.
it is not actually 2 collections. Describe called 1 time, on start.
---
@xgrommx I am not following your explanation.
what don't we use prom.DescribeByCollect?
of course we can do
```
	ch <- mAgentsDesc
	ch <- mNodesDesc
	ch <- mServicesDesc
```
but in this case we are not checking if we can collect metrics, with actual collection we see potential issues on startup, not first scrape.

xgrommx:
> can we have some tests please?

ok

xgrommx:
> can we have some tests please?
can u check my tests?

.github/workflows/managed.yml
```diff
@@ -91,6 +91,9 @@ jobs:
         continue-on-error: true
         run: docker cp ~/go/pkg/mod pmm-managed-server:/root/go/pkg/mod
 
+      - name: Update binaries
```
ritbl, line 94: we need to update binaries so that we can run integration tests

managed/services/inventory/inventory_metrics.go
```diff
@@ -0,0 +1,231 @@
...
+	ctx, cancelCtx := context.WithTimeout(context.Background(), requestTimeout)
+	defer cancelCtx()
+
+	i.mutex.Lock()
+	defer i.mutex.Unlock()
+
+	agentMetrics, agentError := i.metrics.GetAgentMetrics(ctx)
+
+	if agentError != nil {
+		logrus.Error(agentError)
```
ritbl, line 200: please take logger from context

managed/services/inventory/inventory_metrics.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	for _, agentMetric := range agentMetrics {
+		ch <- prom.MustNewConstMetric(i.mAgentsDesc, prom.GaugeValue, agentMetric.value, agentMetric.labels...)
+	}
+
+	nodeMetrics, nodeError := i.metrics.GetNodeMetrics(ctx)
+
+	if nodeError != nil {
+		logrus.Error(nodeError)
```
ritbl, line 211: please take logger from context

managed/services/inventory/inventory_metrics.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (i *InventoryMetricsCollector) Describe(ch chan<- *prom.Desc) {
+	prom.DescribeByCollect(i, ch)
+}
+
+func (i *InventoryMetricsCollector) Collect(ch chan<- prom.Metric) {
+	ctx, cancelCtx := context.WithTimeout(context.Background(), requestTimeout)
+	defer cancelCtx()
+
+	i.mutex.Lock()
```
ritbl, line 194: why do we lock?

managed/services/inventory/inventory_metrics.go
```diff
@@ -0,0 +1,231 @@
...
+	err = i.db.InTransactionContext(ctx, nil, func(tx *reform.TX) error {
+		dbAgents, dbAgentsError := models.FindAgents(tx.Querier, models.AgentFilters{})
+
+		if dbAgentsError != nil {
+			return dbAgentsError
+		}
+
+		for _, agent := range dbAgents {
+			disabled := "0"
+			connected := float64(0)
```
ritbl, line 101: why in one case it is string, while int in the other

managed/services/inventory/inventory_metrics.go
xgrommx: what do u mean?

managed/services/inventory/inventory_metrics.go
xgrommx: because `connected` is a value, but `disabled` is label (which should be as string)

managed/services/inventory/inventory_metrics.go
ritbl: why it is a problem?

managed/services/inventory/inventory_metrics.go
ritbl: see managed/utils/logger

managed/services/inventory/inventory_metrics.go
xgrommx: do u think that it don't needed?

managed/services/inventory/inventory_metrics.go
xgrommx: but we create own context here
https://github.com/percona/pmm/pull/1854/files#diff-0c01ef95d8b74b9e9f04d3173af69a708935b99a7802facb2209f2cd74c0e04bR191

xgrommx:
@ritbl any updates?

pkadej (approved):

dliakhov (approved):

ritbl (approved):

managed/services/inventory/inventory_metrics.go
```diff
@@ -0,0 +1,245 @@
...
+		db:       db,
+		registry: registry,
+	}
+}
+
+func NewInventoryMetricsCollector(metrics inventoryMetrics) *InventoryMetricsCollector {
+	return &InventoryMetricsCollector{
+		mAgentsDesc: prom.NewDesc(
+			prom.BuildFQName(prometheusNamespace, prometheusSubsystem, "agents"),
+			"Inventory Agent",
```
idoqo, line 69: is there any reason why we can't expose the `node_name` label here (and in services metrics too)?

managed/services/inventory/inventory_metrics.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func (i *InventoryMetricsCollector) Collect(ch chan<- prom.Metric) {
+	ctx, cancelCtx := context.WithTimeout(context.Background(), requestTimeout)
+	defer cancelCtx()
+
+	ctx = logger.Set(ctx, "inventoryMetrics")
+	l := logger.Get(ctx)
+
+	agentMetrics, agentError := i.metrics.GetAgentMetrics(ctx)
```
artemgavrilov, line 211: Why not just `err`?

managed/services/inventory/inventory_metrics.go
```diff
@@ -0,0 +1,245 @@
...
+	return metrics, err
+}
+
+func (i *InventoryMetricsCollector) Describe(ch chan<- *prom.Desc) {
+	prom.DescribeByCollect(i, ch)
+}
+
+func (i *InventoryMetricsCollector) Collect(ch chan<- prom.Metric) {
+	ctx, cancelCtx := context.WithTimeout(context.Background(), requestTimeout)
+	defer cancelCtx()
```
artemgavrilov, line 206: One context with one timeout for all requests, and these requests are sequential. It may cause problems if for example first request was too slow. Maybe it makes sense to start collection for every type in separate goroutine?

managed/services/inventory/inventory_metrics.go
```diff
@@ -0,0 +1,245 @@
+// Copyright (C) 2017 Percona LLC
```
artemgavrilov, line 1:
```suggestion
// Copyright (C) 2023 Percona LLC
```

managed/services/inventory/inventory_metrics_test.go
```diff
@@ -0,0 +1,128 @@
...
+		defer cancelCtx()
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, "http://localhost:7773/debug/metrics", nil)
+		require.NoError(t, err)
+		resp, err := client.Do(req)
+		require.NoError(t, err)
+		defer resp.Body.Close() //nolint:gosec
+
+		body, err := io.ReadAll(resp.Body)
+		assert.NoError(t, err)
```
artemgavrilov, line 73:
```suggestion
		require.NoError(t, err)
```

idoqo:
Hi @xgrommx. Thanks so much for your work on this. Will be closing this in favor of #2214

PR #1855, opened 2023-03-09
PMM-11640 Enable backup management feature by default

artemgavrilov:
PMM-11640

Link to the Feature Build: SUBMODULES-3127


atymchuk (approved):

artemgavrilov:
> It looks like frontend needs to consider the change in Settings as well.

Nope, there we can use same toggle, but by default it will be in `On` state.

PavelKhripkov (approved):

PavelKhripkov:
Some thoughts.
1. Do the changes affect other parts? I remember we had problems with telemetry when made similar changes for alerts.
2. Looks like something wrong with design if we need to change so many files just to change default settings.

artemgavrilov:
> Some thoughts.
> 
>     1. Do the changes affect other parts? I remember we had problems with telemetry when made similar changes for alerts.
> 
>     2. Looks like something wrong with design if we need to change so many files just to change default settings.

Damn! I forgot about telemetry!

managed/services/telemetry/config.default.yml
```diff
@@ -34,7 +34,7 @@ telemetry:
 
   - id: PMMServerBackupManagementEnabled
     source: PMMDB_SELECT
-    query: (CASE WHEN backup_management->'enabled' = 'true' THEN '1' ELSE '0' END) AS backup_management_enabled FROM settings s, jsonb_extract_path(s.settings, 'backup_management') AS backup_management
+    query: (CASE WHEN backup_management->'disabled' = 'false' THEN '1' ELSE '0' END) AS backup_management_enabled FROM settings s, jsonb_extract_path(s.settings, 'backup_management') AS backup_management
```
artemgavrilov, line 37: @ritbl Please check this

artemgavrilov:
> Looks like something wrong with design if we need to change so many files just to change default settings.

@PavelKhripkov That's true. This happens that we can't distinguish default state, we need to address this.

PR #1862, opened 2023-03-13
PMM-11727: make promscrape.maxScrapeSize 64MiB by default, allow override via ENV

ritbl:
[PMM-11727](https://jira.percona.com/browse/PMM-11727)

FB:
- https://github.com/Percona-Lab/pmm-submodules/pull/3133


managed/services/agents/vmagent.go
```diff
@@ -16,19 +16,28 @@
...
 
 import (
+	"os"
 	"sort"
 
 	"github.com/percona/pmm/api/agentpb"
 	"github.com/percona/pmm/api/inventorypb"
 )
 
+var maxScrapeSizeEnv = "PMM_PROMSCRAPE_MAX_SCRAPE_SIZE"
```
ritbl, line 26: not sure if we have agreement on how we name env vars for config like this

managed/services/agents/vmagent.go
```diff
@@ -16,19 +16,28 @@
...
+
 // vmAgentConfig returns desired configuration of vmagent process.
 func vmAgentConfig(scrapeCfg string) *agentpb.SetStateRequest_AgentProcess {
+	maxScrapeSize := "64MiB"
```
pkadej, line 30: Default value in function is not intuitive for me. Maybe we should consider moving it to the top of file as global variable and naming it with `defaultMaxScrapSize`?

managed/services/agents/vmagent_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMaxScrapeSize(t *testing.T) {
+	t.Run("by default 64MiB", func(t *testing.T) {
```
pkadej, line 26: Maybe default value should be exported in managed/services/agents/vmagent.go and used here?
Every change of default value will cause this test fail. 
What do you think?

pkadej (approved):

managed/services/agents/vmagent.go
```diff
@@ -16,19 +16,31 @@
...
 import (
+	"os"
 	"sort"
 
 	"github.com/percona/pmm/api/agentpb"
 	"github.com/percona/pmm/api/inventorypb"
 )
 
+var (
+	maxScrapeSizeEnv     = "PMM_PROMSCRAPE_MAX_SCRAPE_SIZE"
```
ritbl, line 27: not sure if we have agreement on how we name env vars for config like this
can you suggest what it is better to put in pmm doc?
@BupycHuk @artemgavrilov

artemgavrilov: I don't know any conventions about env vars naming. You variant looks good for me, probably it can be even simplified to `PMM_MAX_SCRAPE_SIZE`

managed/services/agents/vmagent.go
```diff
@@ -16,19 +16,31 @@
...
+	maxScrapeSizeDefault = "64MiB"
+)
+
 // vmAgentConfig returns desired configuration of vmagent process.
 func vmAgentConfig(scrapeCfg string) *agentpb.SetStateRequest_AgentProcess {
+	maxScrapeSize := maxScrapeSizeDefault
+	if space := os.Getenv(maxScrapeSizeEnv); space != "" {
+		maxScrapeSize = space
+	}
```
artemgavrilov, line 36: How server behave on invalid string here? Is it clear from logs what is the problem?

artemgavrilov (approved):

gen1us2k (approved):

managed/services/agents/vmagent.go
ritbl: INFO[2023-03-13T11:46:08.732+01:00] invalid value "64ABC" for flag -promscrape.maxScrapeSize: strconv.ParseFloat: parsing "64ABC": invalid syntax  agentID=/agent_id/c717a23c-bfdd-4c36-9723-5d4a6388fde5 component=agent-process type=vm_agent

PR #1814, opened 2023-03-02
PMM-9474 Support PostgreSQL queries for multiple databases in Advisor Checks

artemgavrilov:
PMM-9474

Link to the Feature Build: SUBMODULES-3106

If this PR adds or removes or alters one or more API endpoints, please review and add or update the relevant [API documents](https://github.com/percona/pmm/tree/main/docs/api) as well:

- [ ] API Docs updated

If this PR is related to some other PRs in this or other repositories, please provide links to those PRs:

- https://github.com/percona-platform/platform/pull/649
- https://github.com/percona-platform/saas/pull/69

PR #1877, opened 2023-03-16
PMM-11808 Fix query results deserialization bug

artemgavrilov:
PMM-11808

Link to the Feature Build: SUBMODULES-3142

This PR reverts some changes related to data transferring introduced in https://github.com/percona/pmm/pull/1814
Also it adds base64 encoding for binary date before json serialization.

### DIFF
--- a/managed/cmd/pmm-managed-starlark/main.go
+++ b/managed/cmd/pmm-managed-starlark/main.go
@@ -25,13 +25,13 @@ import (
 
 	"github.com/percona-platform/saas/pkg/check"
 	"github.com/percona-platform/saas/pkg/starlark"
-	"github.com/percona/pmm/api/agentpb"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"go.starlark.net/resolve"
 	"golang.org/x/sys/unix"
 	"gopkg.in/alecthomas/kingpin.v2"
 
+	"github.com/percona/pmm/api/agentpb"
 	"github.com/percona/pmm/managed/services/checks"
 	"github.com/percona/pmm/managed/utils/logger"
 	"github.com/percona/pmm/version"
@@ -137,7 +137,6 @@ func runChecks(l *logrus.Entry, data *checks.StarlarkScriptData) ([]check.Result
 		default:
 			l.Errorf("unknown query result type %T", qr)
 		}
-
 	}
 
 	var results []check.Result
--- a/managed/cmd/pmm-managed-starlark/main.go
+++ b/managed/cmd/pmm-managed-starlark/main.go
@@ -136,6 +136,7 @@ func runChecks(l *logrus.Entry, data *checks.StarlarkScriptData) ([]check.Result
 			}
 		default:
 			l.Errorf("unknown query result type %T", qr)
+			os.Exit(1)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


managed/services/checks/checks.go
```diff
@@ -1266,8 +1279,18 @@ WHERE datallowconn = true AND datistemplate = false AND has_database_privilege(c
...
 
+	dec, err := b64.DecodeString(res)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to decode database discovery results")
+	}
+
+	data, err := agentpb.UnmarshalActionQueryResult(dec)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshall database discovery results")
+	}
```
artemgavrilov, line 1290: Inconvenient part, but I will think how we can eliminate it later

managed/cmd/pmm-managed-starlark/main.go
```diff
@@ -117,17 +119,52 @@ func runChecks(l *logrus.Entry, data *checks.StarlarkScriptData) ([]check.Result
...
+		results, err = env.Run(data.Name, res, contextFuncs, l.Debugln)
 	}
 	if err != nil {
 		return nil, errors.Wrap(err, "error running starlark env")
 	}
 
 	return results, nil
 }
+
+func unmarshallQueryResult(qr string) ([]map[string]any, error) {
```
idoqo, line 158:
```suggestion
func unmarshalQueryResult(qr string) ([]map[string]any, error) {
```

managed/services/checks/checks.go
idoqo, line 1290: This block is similar to the `unmarshalQueryResult` function in `main.go`. I wonder if we can export bring the function to this package and export it for `main.go` to use.

idoqo (approved):

managed/services/checks/checks.go
artemgavrilov: I will address these comments in another PR. For now we need just to fix RC ASAP.

PavelKhripkov (approved):

managed/cmd/pmm-managed-starlark/main.go
```diff
@@ -117,17 +119,55 @@ func runChecks(l *logrus.Entry, data *checks.StarlarkScriptData) ([]check.Result
...
+	b, err := base64.StdEncoding.DecodeString(qr)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to decode base64 encoded query result")
+	}
+
+	res, err := agentpb.UnmarshalActionQueryResult(b)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshall query result")
```
atymchuk, line 169:
```suggestion
		return nil, errors.Wrap(err, "failed to unmarshal query result")
```
There should be a single `l` everywhere )

managed/services/checks/checks.go
```diff
@@ -1266,8 +1279,18 @@ WHERE datallowconn = true AND datistemplate = false AND has_database_privilege(c
...
-	r := make([]string, len(res))
-	for i, row := range res {
+	dec, err := b64.DecodeString(res)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to decode database discovery results")
+	}
+
+	data, err := agentpb.UnmarshalActionQueryResult(dec)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshall database discovery results")
```
atymchuk, line 1289:
```suggestion
		return nil, errors.Wrap(err, "failed to unmarshal database discovery results")
```

atymchuk (approved):

PR #1880, opened 2023-03-17
PMM-11078: mark IA-related code as deprecated

idoqo:
PMM-11078

- [ ] API Docs updated


artemgavrilov (changes requested):
@idoqo and what about `channels.proto`, `rules.proto`, etc? We need to deprecate whole set of `IA` related APIs.

api/swagger/swagger-dev.json
```diff
@@ -32652,7 +32652,7 @@
                             "description": "Channels.",
                             "type": "array",
                             "items": {
-                              "description": "Channel represents a single Notification Channel.\n\nreserved channels\n pushover_config = 5;\n opsgenie_config = 7;\n victorops_config = 9;\n wechat_config = 10;",
+                              "description": "Channel represents a single Notification Channel.",
```
artemgavrilov, line 32655: There is no notifications in the swagger.json that this API is deprecated. It seems that we can add it here by using ` option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {deprecated: true}` option.
Please take a look: https://github.com/grpc-ecosystem/grpc-gateway/blob/f31ce88d9dbe898943943139ce726d7842582fa1/protoc-gen-openapiv2/options/openapiv2.proto#L106

idoqo: Right - though seems this option is only supported for RPCs and not for messages (so we can't mark specific messages/schemas as deprecated in swagger).

managed/services/management/ia/rules_service.go
PavelKhripkov, line 1: Why don't we add `Deprecated. Do not use.` comment to the funcs without comments?

PavelKhripkov (approved):

managed/services/management/ia/rules_service.go
idoqo: They were unexported so not really significant but added them in any case.

PR #1881, opened 2023-03-17
PMM-9574 VM Operator registered multiple times for the same cluster

percona-csalguero:
[PMM-9574](https://jira.percona.com/browse/PMM-9574) VM Operator registered multiple times for the same cluster

Link to the Feature Build: N/A yet

1. Moved code from dbaas-controller to PMM
2. Added checks to avoid double installation of all operators
3. If the VM operator is installed, it will be uninstalled and installed again to update possible configuration changes like a new PMM Server IP address. 
4. Added a new label to the subscriptions to make it easier to identify.


managed/services/management/dbaas/kubernetes_server.go
```diff
@@ -424,6 +435,16 @@ func (k kubernetesServer) installDefaultOperators(operatorsToInstall map[string]
 			channel = "stable-v0"
 		}
 		operatorName := "victoriametrics-operator"
+		subs, err := kubeClient.GetSubscription(ctx, namespace, operatorName)
+
+		// If the VM operator already exists, we must delete it in order to reconfigure it.
+		// It migh be the case were the cluster is being re-registered and the server's IP has changed.
```
gen1us2k, line 441: Just out of curiosity but what about the option when we list existing agents running and if they are okay we do nothing? If we have less than 1 agent running then we can provision a new VM agent?

managed/services/management/dbaas/kubernetes_server.go
```diff
@@ -446,19 +467,22 @@ func (k kubernetesServer) installDefaultOperators(operatorsToInstall map[string]
...
-		params := kubernetes.InstallOperatorRequest{
-			Namespace:              namespace,
-			Name:                   operatorName,
-			OperatorGroup:          operatorGroup,
-			CatalogSource:          catalogSource,
-			CatalogSourceNamespace: catalogSourceNamespace,
-			Channel:                channel,
-			InstallPlanApproval:    v1alpha1.ApprovalManual,
-		}
+		_, err := kubeClient.GetSubscription(ctx, namespace, operatorName)
```
gen1us2k, line 470: It would be great to wait until VM's CRDs will be present in a cluster. There can be a corner case when we start provisioning monitoring and applying rbac rules or starting kube-state-metrics will fail because CSV is in Ready to install or installing state.

managed/services/management/dbaas/kubernetes_server.go
```diff
@@ -490,25 +517,68 @@ func (k kubernetesServer) installDefaultOperators(operatorsToInstall map[string]
...
 
 	return retval
 }
 
+// uninstallVM unistalls Victoria Metrics Operator.
+// To delete the vm operator we must:
+//  1. Delete the subscription.
+//  2. Delete the CSV.
+func uninstallVM(ctx context.Context, kubeClient kubernetesClient, subscription *v1alpha1.Subscription) error {
+	key := types.NamespacedName{
```
gen1us2k, line 547: Actually, I'm not sure that's a great option because there can be already existing VM agents running in the cluster (e.g. we're registering k8s cluster with the 4 db clusters and one VM agent running. In that case we may have broken monitoring for these database clusters). 

Yet, I don't have any recommendations here. Maybe getting running vmagents and removing agents only and reconfiguring them would be a better solution. What do you think?

managed/services/management/dbaas/kubernetes_server.go
```diff
@@ -490,25 +517,68 @@ func (k kubernetesServer) installDefaultOperators(operatorsToInstall map[string]
...
-		params := kubernetes.InstallOperatorRequest{
-			Namespace:              namespace,
-			Name:                   operatorName,
-			OperatorGroup:          operatorGroup,
-			CatalogSource:          "percona-dbaas-catalog",
-			CatalogSourceNamespace: catalogSourceNamespace,
-			Channel:                channel,
-			InstallPlanApproval:    v1alpha1.ApprovalManual,
-		}
+		_, err := kubeClient.GetSubscription(ctx, namespace, operatorName)
```
gen1us2k, line 520: I like the idea to get a subscription first but I think that we need to get CSV and check that it has the 'Installed' or 'Succeeded' phase in the status field. Getting a CSV object is a better option to understand that operator is installed. 

As an example of the corner case that might be in the cluster. Let's imagine that we have two channels fast and stable (or, let's say that we have one channel per PMM's release). And dbaas-catalog wasn't installed, so there will be a subscription present in the cluster and it uses as a source percona-dbaas-catalog, yet CSV won't be created because dbaas-catalog is not present in the cluster. 

Getting the state of csv (and this is what we need to have in the end) will tell us more about the state of the operator's installation. What do you think about moving from GetSubscriotions to GetCSV?

managed/services/management/dbaas/kubernetes_server.go
```diff
@@ -367,6 +370,14 @@ func (k kubernetesServer) setupMonitoring(ctx context.Context, operatorsToInstal
 		k.l.Errorf("cannot start monitoring the clusdter: %s", err)
 	}
 
+	if err := kubeClient.ProvisionMonitoring("api_key", apiKey); err != nil {
+		e := k.grafanaClient.DeleteAPIKeyByID(ctx, apiKeyID)
+		if e != nil {
+			k.l.Warnf("couldn't delete created API Key %v: %s", apiKeyID, e)
+		}
+		k.l.Errorf("couldn't start monitoring of the kubernetes cluster: %s", err)
```
gen1us2k, line 378: Do we need to cleanupmonitoring when we can't provision it?  What do you think?

managed/services/management/dbaas/kubernetes_server.go
```diff
@@ -490,25 +517,68 @@ func (k kubernetesServer) installDefaultOperators(operatorsToInstall map[string]
...
+	err = kubeClient.DeleteObject(csv)
+	if err != nil {
+		return errors.Wrapf(err, "cannot delete VM clusterserviceversion %s for reinstallation", subscription.Status.CurrentCSV)
+	}
+
+	return nil
+}
+
 // UnregisterKubernetesCluster removes a registered Kubernetes cluster from PMM.
 func (k kubernetesServer) UnregisterKubernetesCluster(ctx context.Context, req *dbaasv1beta1.UnregisterKubernetesClusterRequest) (*dbaasv1beta1.UnregisterKubernetesClusterResponse, error) { //nolint:lll
```
gen1us2k, line 583: I think that we miss cleaningup monitoring on the Unregistering

gen1us2k (commented):
Thanks for the changes. I left some minor comments and started some discussions. Looking forward to discussing them.

managed/services/management/dbaas/kubernetes_server.go
percona-csalguero: How can we know if the agent that's already installed belongs to the same PMM server? 
Does it make sense? Maybe we shouldn't do anything and it should be only removed when you unregister the K8s server. 
I also added a new label `app.kubernetes.io/part-of:pmm` to be able to check that the operator/s were installed by pmm. Maybe we could also use `app.kubernetes.io/managed-by` with the PMM server id and only uninstall things that were installed by this PMM installation. 

DOC REF: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

managed/services/management/dbaas/kubernetes_server.go
percona-csalguero: by ` yet CSV won't be created because dbaas-catalog is not present in the cluster.` you wanted to say subscription, right?

percona-csalguero:
Closing in favor of https://github.com/percona/pmm/pull/1979

PR #1884, opened 2023-03-17
PMM-11456: Add endpoints to work with tips

dliakhov:
PMM-11456

This PR introduces new endpoints for working with Help Center tips, which are related to the Onboarding functionality. There are two types of tips: system tips and user tips.

System tips can only be completed once per PMM instance, and their completion is based on hardcoded logic (at least for now). Currently, there are only three system tips:
- Installing PMM server. This tip is always completed and only exists to show that the user has installed PMM.
- Installing PMM client. This tip is completed when the user connects an external PMM agent. Once an admin adds the external agent, this tip will be marked as completed.
- Connecting service. This tip is completed when the user connects an external database. Once an admin connects the database, this tip will be marked as completed.

User tips, on the other hand, can only be completed manually by the user. 
System tip IDs start from 1 and User tip IDs start from 1000.

This PR includes implementation of two endpoints:
1. Retrieve status of tip.
2. Complete user tip (can be completed only user tip, for system tip endpoint will return error).


api/onboardingpb/json/onboardingpb.json
```diff
@@ -0,0 +1,203 @@
...
+                  "x-order": 2
+                },
+                "errorMessage": {
+                  "type": "string",
+                  "title": "container error message. it's empty if there is no error",
+                  "x-order": 3
+                },
+                "isCompleted": {
+                  "type": "boolean",
+                  "title": "show whether t ip is completed or not",
```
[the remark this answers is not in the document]
dliakhov: Fixed this.

It's autogenerated code which was taken from comment of the proto file.

api/onboardingpb/tips.proto
```diff
@@ -0,0 +1,72 @@
...
+
+option go_package = "api/onboardingpb;onboardingpb";
+
+enum TipType {
+  SYSTEM = 0;
+  USER = 1;
+}
+
+// Request message for GetTipStatus method.
+message GetTipfRequest {
```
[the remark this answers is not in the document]
dliakhov: Fixed.

managed/models/database.go
```diff
@@ -838,6 +838,25 @@ var databaseSchema = [][]string{
 		FROM services
         WHERE service_type = 'mongodb';`,
 	},
+	79: {
+		`CREATE TABLE system_tip (
```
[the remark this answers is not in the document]
dliakhov: good catch

fixed

managed/services/onboarding/tips.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+	}
+	return &onboardingpb.GetTipResponse{
+		TipId:       tip.UserTipID,
+		IsCompleted: tip.IsCompleted,
+	}, nil
+}
+
+func (t *TipsService) retrieveSystemTip(tipID int32) (models.SystemTip, error) {
+	res, err := t.db.Querier.FindOneFrom(models.SystemTipTable, "id", tipID)
```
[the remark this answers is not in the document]
dliakhov: yeah, fixed

managed/services/onboarding/tips.go
```diff
@@ -0,0 +1,232 @@
...
+	default:
+		return nil, errors.New("Tip type is not correct")
+	}
+}
+
+func (t *TipsService) retrieveOrCreateUserTip(tipRequest *onboardingpb.GetTipfRequest) (*onboardingpb.GetTipResponse, error) {
+	tip, err := t.retrieveUserTip(tipRequest.TipId, tipRequest.UserId)
+	if err != nil {
+		if err == reform.ErrNoRows {
+			tip, err = t.createUserTip(tipRequest.TipId, tipRequest.UserId)
```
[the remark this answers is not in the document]
dliakhov: if I add constraint that user_id and user_tip_id should be unique, then this case won't be relevant, right? https://github.com/percona/pmm/blob/1e7805b988ac8a4db31b1769ee64ecb040ebc834/managed/models/database.go#L853

like this

```
CREATE TABLE user_tips (
		    id SERIAL PRIMARY KEY,
			user_id PRIMARY KEY INTEGER NOT NULL,
			user_tip_id INTEGER NOT NULL,
			is_completed BOOLEAN NOT NULL,
	
			created_at TIMESTAMP NOT NULL,
			updated_at TIMESTAMP NOT NULL
		   UNIQUE (user_id, user_tip_id)
		);
```

because in case if during request we would try to create user tip with existing ids then database should handle attempt of creation tip with the same id.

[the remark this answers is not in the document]
dliakhov: added this constraint and wrapped this code into db transaction

api/onboardingpb/json/onboardingpb.json
```diff
@@ -0,0 +1,192 @@
...
+                  "type": "string",
+                  "x-order": 1
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/v1/onboarding/tips/{tipId}/type/{tipType}/user/{userId}": {
```
ritbl, line 94: Is it a sub-resource or filtering?

https://stackoverflow.com/a/31261026/1065759

api/onboardingpb/json/onboardingpb.json
```diff
@@ -0,0 +1,192 @@
...
+                  "description": "user id which completes user tip.",
+                  "type": "integer",
+                  "format": "int32",
+                  "x-order": 1
+                }
+              }
+            }
+          }
+        ],
+        "responses": {
```
ritbl, line 52: 404?

api/onboardingpb/json/onboardingpb.json
```diff
@@ -0,0 +1,192 @@
...
+        }
+      }
+    }
+  },
+  "tags": [
+    {
+      "name": "TipService"
+    }
+  ]
+}
```
ritbl, line 192: please add last empty line

ritbl (changes requested):
we need to discuss API

api/onboardingpb/json/onboardingpb.json
dliakhov: how can we do that? it's generated automatically and if you take a look on other swagger files it's always only 200 or unexpected error.

api/onboardingpb/json/onboardingpb.json
dliakhov: it's generated automatically, if I modify it someone else will update it again.

managed/services/onboarding/tips.go
```diff
@@ -0,0 +1,245 @@
...
+		tip, err := t.retrieveUserTip(tx, userTipRequest.TipId, userTipRequest.UserId)
+		if err != nil {
+			if err == reform.ErrNoRows {
+				tip, err = t.createUserTip(tx, userTipRequest.TipId, userTipRequest.UserId)
+				if err != nil {
+					return errors.Wrap(err, fmt.Sprintf("cannot create user tip by id: %d", userTipRequest.TipId))
+				}
+			} else {
+				return errors.Wrap(err, fmt.Sprintf("cannot retrieve user tip by id: %d", userTipRequest.TipId))
+			}
```
JiriCtvrtka, line 225:
```suggestion
			return errors.Wrap(err, fmt.Sprintf("cannot retrieve user tip by id: %d", userTipRequest.TipId))
```

JiriCtvrtka (approved):
Good job.

api/onboardingpb/json/onboardingpb.json
```diff
@@ -0,0 +1,204 @@
...
+                  "type": "string",
+                  "x-order": 1
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/v1/onboarding/{userId}": {
```
ritbl, line 94: not good from API design

`/v1/onboarding/{userId}` - is not subresource, so userId should not be a part for path.

but `userId` should not be present because we take it from contex

api/onboardingpb/tips.proto
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// Response message for CompleteUserTip method.
+message CompleteUserTipResponse {}
+
+// Service service provides public methods for managing Services.
+service TipService {
+  // GetTipStatus returns info about tip status
+  rpc GetOnboardingStatus(GetOnboardingStatusRequest) returns (GetOnboardingStatusResponse) {
+    option (google.api.http) = {get: "/v1/onboarding/{userId}"};
```
ritbl, line 49: this endpoint to be called by authorized user, we will have userId in the context.
we don't need to provide API with userId

managed/models/database.go
```diff
@@ -838,6 +838,50 @@ var databaseSchema = [][]string{
 		FROM services
         WHERE service_type = 'mongodb';`,
 	},
```
ritbl, line 840: I think we need to prefix all onboarding related tables with `onboarding_`

managed/models/database.go
```diff
@@ -838,6 +838,50 @@ var databaseSchema = [][]string{
...
+			created_at TIMESTAMP NOT NULL,
+			updated_at TIMESTAMP NOT NULL
+		);
+		
+		CREATE TABLE system_tips (
+			 id SERIAL PRIMARY KEY,
+			 is_completed BOOLEAN NOT NULL,
+		
+			 created_at TIMESTAMP NOT NULL,
+			 updated_at TIMESTAMP NOT NULL
```
ritbl, line 855: available_tips - can be an enum.

why `system_tips` are stored? it is something we check every time, we don't need to store that

dliakhov: because when user disconnects agent/database if we don't store the status of the tip it will be again uncompleted.

To avoid this we store the status when user connects/disconnects the agent or external db.

api/onboardingpb/json/onboardingpb.json
dliakhov: yes, it's not good design, but we don't have ability to read http headers in our code when we use GRPC over HTTP (like it works now), because this code is autogenerated.

What can we do it's to put the userId to request body and change method to POST, but not sure that it will look better.

api/onboardingpb/tips.proto
dliakhov: I don't know how to retrieve it. If you take a look how other endpoints implemented - they do the same, path user id via request param or body

https://github.com/percona/pmm/blob/8ba03139901af250e2180f006a8c55651daaf6a3/api/userpb/user.proto#L20

managed/models/available_tip.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"gopkg.in/reform.v1"
+)
+
+//go:generate ../../bin/reform
+
+// AvailableTip list of tips which ids are valid for the PMM
+//
+//reform:available_tips
+type AvailableTip struct {
```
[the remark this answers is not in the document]
dliakhov: This table contains ids of available tips which are valid for PMM instance. It's used generating data in the user_tips table when they doesn't exist yet.

managed/models/database.go
dliakhov: okay

PR #1885, opened 2023-03-19
PMM-11820: make binding to interface configurable

ritbl:
[PMM-11820](https://jira.percona.com/browse/PMM-11820)

### DIFF
--- a/managed/cmd/pmm-managed/main.go
+++ b/managed/cmd/pmm-managed/main.go
@@ -25,7 +25,7 @@ import (
 	"log"
 	"net"
 	"net/http"
-	_ "net/http/pprof" //nolint:gosec // register /debug/pprof
+	_ "net/http/pprof"
 	"net/url"
 	"os"
 	"os/signal"
@@ -99,6 +99,7 @@ import (
 	"github.com/percona/pmm/managed/services/victoriametrics"
 	"github.com/percona/pmm/managed/services/vmalert"
 	"github.com/percona/pmm/managed/utils/clean"
+	"github.com/percona/pmm/managed/utils/env"
 	"github.com/percona/pmm/managed/utils/envvars"
 	"github.com/percona/pmm/managed/utils/interceptors"
 	"github.com/percona/pmm/managed/utils/logger"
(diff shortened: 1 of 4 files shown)


dliakhov (approved):

managed/cmd/pmm-managed/main.go
```diff
@@ -108,14 +109,17 @@ import (
 	"github.com/percona/pmm/version"
 )
 
+var (
+	interfaceToBind = env.GetEnv("PMM_DEV_BIND_INTERFACE", "127.0.0.1")
```
artemgavrilov, line 113: Let's use `PERCONA_TEST_` prefix, it's standard naming in PMM for env variables that should not be used by user.

managed/utils/env/env.go
```diff
@@ -0,0 +1,25 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package env
```
artemgavrilov, line 16: I think we can just put in `envvars` package.

managed/cmd/pmm-managed/main.go
artemgavrilov: Also probably it makes sense to validate value

managed/cmd/pmm-managed/main.go
```diff
@@ -108,14 +109,17 @@ import (
...
+	gRPCAddr        = interfaceToBind + ":7771"
```
artemgavrilov, line 114: Better use `net.JoinHostPort` func for that.

managed/services/agents/vmagent.go
```diff
@@ -35,6 +36,8 @@ func vmAgentConfig(scrapeCfg string) *agentpb.SetStateRequest_AgentProcess {
 		maxScrapeSize = space
 	}
 
+	interfaceToBind := env.GetEnv("PMM_DEV_BIND_INTERFACE", "127.0.0.1")
```
artemgavrilov, line 39: If we need this in multiple places I suggest to create function `GetDevInterface` or something like this in `envvars` package

managed/cmd/pmm-managed/main.go
ritbl: sure

managed/services/agents/vmagent.go
ritbl: okay

artemgavrilov (approved):

PR #1920, opened 2023-03-29
PMM-11878: refactor user roles in access control

michal-kralik:
- Prepares access roles for other entities, such as teams and orgs

PMM-11878

Link to the Feature Build: SUBMODULES-3168


api/managementpb/role/role.proto
```diff
@@ -51,9 +51,17 @@ message SetDefaultRoleRequest {
...
+  USER = 1;
+}
+
 message AssignRolesRequest {
   repeated uint32 role_ids = 1;
-  uint32 user_id = 2 [(validator.field) = {int_gt: 0}];
+  // Deprecated: do not use
+  uint32 user_id = 2 [deprecated = true];
+  uint32 entity_id = 3 [(validator.field) = {int_gt: 0}];
+  EntityType entity_type = 4 [(validator.field) = {int_gt: 0}];
```
BupycHuk, line 64: it will be good to add descriptions to these fields

BupycHuk: Can we use `oneof` for it?

managed/cmd/pmm-managed/main.go
```diff
@@ -951,7 +958,7 @@ func main() {
 		l.Fatalf("Failed to get settings: %+v.", err)
 	}
 
-	authServer := grafana.NewAuthServer(grafanaClient, awsInstanceChecker, db)
+	authServer := grafana.NewAuthServer(grafanaClient, awsInstanceChecker, db, roleRegistry, roleRegistry)
```
BupycHuk, line 961: why do we pass it twice?

managed/models/role_helpers_test.go
```diff
@@ -27,10 +26,11 @@ import (
...
 
-const (
-	roleATitle = "Role A"
-	roleBTitle = "Role B"
-)
+type noOpBeforeDelete struct{}
+
+func (n *noOpBeforeDelete) BeforeDeleteRole(tx *reform.TX, roleID, newRoleID int) error {
+	return nil
+}
```
BupycHuk, line 33: wouldn't it be better to use mocks instead of these structures?

managed/models/role_helpers.go
```diff
@@ -112,72 +76,6 @@ func DeleteRole(tx *reform.TX, roleID, replacementRoleID int) error {
...
-		if err != nil {
-			return err
-		}
-	}
-
-	return nil
-}
-
 // FindAndLockRole retrieves a role by ID and locks it for update.
 func FindAndLockRole(tx *reform.TX, roleID int, role *Role) error {
```
BupycHuk, line 80: in other places, we return the value, but here we pass it as one of the parameters. Any specific reason for inconsistency?

managed/services/management/roles/generic.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"gopkg.in/reform.v1"
+
+	"github.com/percona/pmm/managed/models"
+)
+
+// EntityModel defines methods required on a model to be used with a generic struct.
```
BupycHuk, line 27: having this model-related structure here doesn't seem correct to me.

managed/models/user_roles_model.go
```diff
@@ -34,6 +34,16 @@ type UserRoles struct {
 	UpdatedAt time.Time `reform:"updated_at"`
 }
 
+// SetEntityID sets user ID
```
BupycHuk, line 37: Could we refactor this model to include entity ID, entity type, and role here instead of creating another model which will be pretty similar to this one?

BupycHuk: in that case we wouldn't need all these hacks.

managed/services/grafana/auth_server.go
```diff
@@ -144,6 +144,14 @@ type clientInterface interface {
 	getAuthUser(context.Context, http.Header) (authUser, error)
 }
 
+type defaultRoleAssigner interface {
+	AssignDefaultRole(tx *reform.TX, userID int) error
+}
+
+type userRolesGetter interface {
+	GetUserRoles(q *reform.Querier, userID int) ([]models.Role, error)
+}
```
BupycHuk, line 153: can we merge these two interfaces to not pass the same object as two arguments

managed/services/grafana/auth_server_test.go
```diff
@@ -167,7 +183,13 @@ func TestAuthServerAuthenticate(t *testing.T) {
 
 	ctx := context.Background()
 	c := NewClient("127.0.0.1:3000")
-	s := NewAuthServer(c, checker, nil)
+
+	roleRegistry := roles.NewRegistry(map[roles.EntityType]roles.EntityService{
+		roles.EntityUser: roles.NewUser("user_id", func() roles.EntityModel {
+			return &models.UserRoles{}
+		}, models.UserRolesView),
+	})
```
BupycHuk, line 191: we already have interfaces in this package, so we can mock this registry in this test.

BupycHuk (changes requested):
I think if we will rework model layer then code will be changed significantly and a lot of structs can be removed.

BupycHuk:
can we write API tests?

managed/models/user_roles_model.go
michal-kralik: Usually I try to avoid polymorphic relationships in database schemas unless there's great support in the ORM for them.
Without a great support, it's easy to overlook issues with queries for developers without experience in this part of codebase.

Personally I'd prefer individual tables for the relationships. Let me know if we shall move forward with `entity_type`, `entity_id` and `role_id` table instead. Happy to refactor.

BupycHuk: even if polymorphic relationships don't exist in a database schema, they still exist in the code. it makes the code much harder to read. We use these kinds of columns everywhere, agent_type, service_type and etc.
maybe we can get another opinion from other developers?
@artemgavrilov or @PavelKhripkov your thoughts?

artemgavrilov: I can't understand what you're discussing without the context? What is the entity, and why method `SetEntityID` uses `userID` and an argument?

michal-kralik: We need to store the following relationships:
- `user_id` to `role_id` (one to many)
- `team_id` to `role_id` (one to many)

We can either have polymorphic relationships in a single table, such as with `entity_id`, `entity_type` and `role_id` colums where `entity_id` refers to the appropriate entity based on `entity_type`.

Or we can have two separate tables `user_roles` and `team_roles` and handle the rest in the code similar to how the PR handles it.

BupycHuk: I'm for polymorphic relationships because here are too many hacks around the usage of multiple tables.

PavelKhripkov: @michal-kralik as I understood, you're only preparing to introduce Team entity. And it's not clear what is the relationship between User and Team models?

michal-kralik:
Closing as it has been decided we'll introduce polymorphic relationships between entities via `entity_type`.
There will be a new PR for this.

managed/models/user_roles_model.go
michal-kralik: A user can belong to one or many teams.
Each user and team can have a set of roles assigned and the resulting set of roles for a user is a union of all the roles from a user and a team.

PavelKhripkov: Polymorphic relationships usually used to store entities of the same level. E.g. you have a table with history of actions and then you can store records like:
`id=1, type=service, action=delete, time=123`
`id=8, type=node, action=add, time=456`

There are no relations between records in the `history` table. But as I see you're trying to store in a single table entities of different levels, of different meaning and with relations between them. To me it looks extremly confusing. By the meaning entity `Team` is close to `Role` rather than to `User` because it provides access level to the User. But storing `Role` and `Team` in a single table not much better.

BupycHuk: a role can be assigned to a user as well as it can be assigned to a team.

PavelKhripkov: Role can be assigned to a user as well as Team can be assigned to a User)

BupycHuk: yes, but a team and a role aren't the same :)
a team is a group of users whereas a role is a group of permissions :)

PavelKhripkov: Right. And if we're thinking about Team as about marker or label then there is no need to add new entity, we can just add labels to User or Role. But I imagine Team as not something to only group Roles. Team could have other properties hepling to manage groups of Users. And if we put User and Team in the same table we need much efforts later to extend Teams functionality.

BupycHuk: currently, Teams and users are stored in the Grafana database and this table is used only to assign a role to a team or a user. It's a table for many-to-many connections. There shouldn't be any new columns in the future.

PavelKhripkov: Ah, I see, you already have a data model and trying to implement new functionality on top of that. Then it's not so bad as I thought)

BupycHuk: Yes)

PR #1950, opened 2023-04-05
PMM-11522 add AdvisorsDisabled telemetry metric

rnovikovP:
PMM-11522

Add datapoint to track disabled Advisors since they are Enabled by default now.


idoqo (approved):

artemgavrilov:
@rnovikovP shouldn't we remove `PMMServerSTTEnabled` metric then? It has the same meaning, but boolean value there is reversed

artemgavrilov (approved):

idoqo:
This was merged into #1935.

PR #1957, opened 2023-04-06
PMM-7438: fix reading long log lines

michal-kralik:
PMM-7438

Link to the Feature Build: SUBMODULES-3200

### DIFF
--- a/managed/services/supervisord/logs.go
+++ b/managed/services/supervisord/logs.go
@@ -302,13 +302,17 @@ func readLog(name string, maxLines int, maxBytes int64) ([]byte, time.Time, erro
 	}
 
 	r := ring.New(maxLines)
-	s := bufio.NewScanner(f)
-	for s.Scan() {
-		r.Value = []byte(s.Text() + "\n")
+	reader := bufio.NewReader(f)
+	for {
+		r.Value, err = reader.ReadBytes('\n')
 		r = r.Next()
-	}
-	if err = s.Err(); err != nil {
-		return nil, m, errors.WithStack(err)
+		if err == io.EOF {
+			break
+		}
+
+		if err != nil {
+			return nil, m, errors.WithStack(err)
+		}
 	}
 
 	res := make([]byte, 0, maxBytes)
--- a/managed/services/supervisord/logs.go
+++ b/managed/services/supervisord/logs.go
@@ -306,8 +306,14 @@ func readLog(name string, maxLines int, maxBytes int64) ([]byte, time.Time, erro
 	for {
 		b, err := reader.ReadBytes('\n')
 		if err == io.EOF {
+			// A special case when the last line does not end with a new line
+			if len(b) != 0 {
+				r.Value = b
+				r = r.Next()
+			}
 			break
 		}
+
 		r.Value = b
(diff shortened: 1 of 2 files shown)


BupycHuk:
@michal-kralik PTAL tests, one of them failed.

rishat-ishbulatov (approved):

managed/services/supervisord/logs.go
```diff
@@ -302,13 +302,18 @@ func readLog(name string, maxLines int, maxBytes int64) ([]byte, time.Time, erro
...
 	r := ring.New(maxLines)
-	s := bufio.NewScanner(f)
-	for s.Scan() {
-		r.Value = []byte(s.Text() + "\n")
+	reader := bufio.NewReader(f)
+	for {
+		b, err := reader.ReadBytes('\n')
+		if err == io.EOF {
+			break
+		}
```
PavelKhripkov, line 310: Looks like we're losing the last part if source doen't end with new line.

PavelKhripkov (changes requested):

managed/services/supervisord/logs.go
michal-kralik: Good catch

michal-kralik:
> 1. We're using Scanner in other parts of the code. Are we going to change them as well?
Not as part of this PR.

> 2. Don't we need to add tests to check that bug is really gone?
I personally don't think such test adds value. It only checks if someone uses an incorrect way of reading a file.

BupycHuk (approved):

managed/services/supervisord/logs.go
```diff
@@ -302,13 +302,24 @@ func readLog(name string, maxLines int, maxBytes int64) ([]byte, time.Time, erro
...
-	for s.Scan() {
-		r.Value = []byte(s.Text() + "\n")
+	reader := bufio.NewReader(f)
+	for {
+		b, err := reader.ReadBytes('\n')
+		if err == io.EOF {
+			// A special case when the last line does not end with a new line
+			if len(b) != 0 {
+				r.Value = b
+				r = r.Next()
```
PavelKhripkov, line 312: Not sure we need `r.Next()` here since we're exiting the loop.

PavelKhripkov:
> > 1. We're using Scanner in other parts of the code. Are we going to change them as well?
> >    Not as part of this PR.
We can create a tech debt ticket then to not forget.


> > 2. Don't we need to add tests to check that bug is really gone?
> >    I personally don't think such test adds value. It only checks if someone uses an incorrect way of reading a file.
Not sure I understood the phrase "an incorrect way of reading a file". Did you check that the described problem has gone after the changes you made?

managed/services/supervisord/logs.go
michal-kralik: We do, otherwise we don't get correct results.

michal-kralik:
> Not sure I understood the phrase "an incorrect way of reading a file". Did you check that the described problem has gone after the changes you made?

Yes, this fixes the issue.
An "incorrect way" would be to use `bufio.NewScanner()` which states it stops on tokens/lines which are too long to fit into a buffer.

PR #2008, opened 2023-04-14
Update config.default.yml

rnovikovP:
PMM-0

- [ ] API Docs updated

### DIFF
--- a/managed/services/telemetry/config.default.yml
+++ b/managed/services/telemetry/config.default.yml
@@ -600,6 +600,15 @@ telemetry:
       - metric_name: "pmm_server_node_os_version"
         label: "release"
 
+  - id: PMMNodeOSVersionInside
+    source: VM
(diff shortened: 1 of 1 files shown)


managed/services/telemetry/config.default.yml
```diff
@@ -600,6 +600,15 @@ telemetry:
       - metric_name: "pmm_server_node_os_version"
         label: "release"
 
+  - id: PMMNodeOSVersionInside
+    source: VM
+    query: node_os
+    summary: "PMM node OS version Inside"
```
atymchuk, line 606:
```suggestion
    summary: "PMM node OS version"
```
I guess it's obvious, seems redundant :)

atymchuk: How are they different? The former is inside and the latter - outside?

PR #2009, opened 2023-04-16
PMM-12016: allow override deps with ENV vars

ritbl:
[PMM-12016](https://jira.percona.com/browse/PMM-12016)

FB:
 - https://github.com/Percona-Lab/pmm-submodules/pull/3219


BupycHuk:
We are working on it. please, don't create unnecessary conflicts

BupycHuk (changes requested):
Core team is working on it and Victoria Metrics doesn't work that way.

PR #2066, opened 2023-05-01
PMM-7 disable `golint` rule

atymchuk:
PMM-7

Ref: issue #1541 

Changes:
- disable the `golint` rule as [unmaintained](https://github.com/golang/lint)
- fix a couple of lint warnings that are present in `main` branch
- add a possibility to pass arbitrary files to managed's `make test ...` command

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -50,11 +50,6 @@ linters-settings:
     line-length: 170
     tab-width: 4
 
-  staticcheck:
-    checks:
-      - all
-      - "-SA1019"
-
   tagliatelle:
     # Check the struck tag name case.
     case:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,6 +2,7 @@
 linters-settings:
   exhaustive:
     default-signifies-exhaustive: true
+
   cyclop:
     max-complexity: 30
 
@@ -49,6 +50,11 @@ linters-settings:
     line-length: 170
     tab-width: 4
 
+  staticcheck:
+    checks:
+      - all
+      - "-SA1019"
+
   tagliatelle:
     # Check the struck tag name case.
     case:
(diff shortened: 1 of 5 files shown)


.golangci.yml
```diff
@@ -49,6 +50,11 @@ linters-settings:
     line-length: 170
     tab-width: 4
 
+  staticcheck:
+    checks:
+      - all
+      - "-SA1019"
```
artemgavrilov, line 56: That's actually good rule and we should address highlighted problems.

atymchuk: > That's actually good rule and we should address highlighted problems.

I also agree it's a good rule, if it wasn't for two facts:
1/ we deprecate things and we never fix the linter warnings that appear as a result of deprecation; I even [asked the author](https://github.com/percona/pmm/issues/1541#issuecomment-1436500268) to address them once, but they are still there :)
2/ deprecation warnings are so numerous that they make it super challenging to find what you need in the logs (pls see attachement)
[14_Run linters.txt](https://github.com/percona/pmm/files/11372274/14_Run.linters.txt)

I suggest we temporarily disable this rule until we fix all deprecation warnings.

I believe we need to hear more opinions.  @BupycHuk @michal-kralik @PavelKhripkov 
What do you think?

artemgavrilov: If we will keep this linter disabled we need ticket to enable it back. Otherwise we will forget about it.

atymchuk: > If we will keep this linter disabled we need ticket to enable it back. Otherwise we will forget about it.

Sure, I see no problem :)

BupycHuk: I agree with Artem that we should keep this linter even if we ignore it for now.

atymchuk: > I agree with Artem that we should keep this linter even if we ignore it for now.

So, what do we do now? Disable, create a ticket for re-enabling it and then address the linter warnings?

BupycHuk: No, we shouldn't disable it, instead, we should address warnings.

BupycHuk: @JiriCtvrtka already fixed some of them, and @cfsalguero is supposed to fix the ones related to DBaaS. So we are fixing them.

artemgavrilov (approved):

BupycHuk (approved):

PR #2128, opened 2023-05-15
PMM-7 fix caching of modules

atymchuk:
PMM-7

Why:
We see a lot of errors when github `actions/cache@v3` tries to restore files from cache. Meanwhile, [v4](https://github.com/actions/setup-go#v4) of `actions/setup-go` enables cache by default, which make the use of a separate action redundant.

In this PR we will try to fix that. We'll still use `actions/cache@v3` for tip versions of `go`.

Note:
It seems like modules in directories `/tools` and `api-tests/tools` won't be cached, since `actions/setup-go` is able to use only one `go.sum` file, which in our case resides in the root folder. TODO: see if/how we can leverage the `cache-dependency-path` [parameter](https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs).


artemgavrilov (changes requested):
Please don't merge, I have concerns.

atymchuk:
Closing this in favor of #2137

PR #2131, opened 2023-05-15
PMM-12119 fix rds_exporter status update

atymchuk:
PMM-12119

Link to the Feature Build: SUBMODULES-3280

- [ ] API Docs updated


managed/services/agents/roster.go
```diff
@@ -35,34 +39,62 @@ const (
...
+
+	sort.Strings(exporterIDs)
+
+	r.m[groupID] = exporterIDs
+	r.l.Debugf("add: %s = %v", groupID, exporterIDs)
+	return groupID
 }
 
-func (r *roster) get(groupID string) (agentIDs []string) { //nolint:nonamedreturns
+func (r *roster) get(groupID string) (string, []string, error) {
```
idoqo, line 72: i know the linter frowns at named returns - maybe we can add method comments to describe what each return parameter is/does?

idoqo (approved):

managed/services/agents/roster.go
atymchuk: Sure. Done!

managed/services/agents/roster.go
```diff
@@ -35,36 +39,68 @@ const (
...
 
-func (r *roster) get(groupID string) (agentIDs []string) { //nolint:nonamedreturns
+// get returns a PMMAgentID and a group of exporter IDs for a given Group ID.
+func (r *roster) get(groupID string) (string, []string, error) {
 	r.rw.RLock()
 	defer r.rw.RUnlock()
 
-	agentIDs = r.m[groupID]
+	suffix := "/" + string(rdsGroup)
+	PMMAgentID, ok := strings.CutSuffix(groupID, suffix)
```
rishat-ishbulatov, line 80: I think this is not critical, a matter of taste, but it looks like we don’t need to allocate memory for a `suffix` every time and we could create it once using const at the compile time with `const suffix = "/" + string(rdsGroup)` [This](https://stackoverflow.com/questions/35026810/runtime-const-in-golang) place describes this minor tweak.

rishat-ishbulatov (approved):

managed/services/agents/roster.go
atymchuk: No problem, done!

PR #2135, opened 2023-05-16
PMM-12118 Add support for the official clickhouse plugin

0leksii:
Do not strip binaries while building the percona-dashboards RPMs
to avoid checksum inconsistencies for grafana plugins. Add new
options to pass to grafana for setting up the new datasource.
Setup new datasource for Grafana.

PMM-12118

Link to the Feature Build: SUBMODULES-3269

Related PRs:
- https://github.com/percona/grafana-dashboards/pull/1503

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -100,7 +100,6 @@ linters:
     # or leave disabled and provide a reason why
     - paralleltest
     - tagliatelle
-    - thelper
     - tparallel
     - gocritic
     - godot
--- a/agent/config/config_test.go
+++ b/agent/config/config_test.go
@@ -29,6 +29,7 @@ import (
 const defaultWindowPeriod = time.Hour
 
 func writeConfig(t *testing.T, cfg *Config) string {
+	t.Helper()
 	f, err := os.CreateTemp("", "pmm-agent-test-")
 	require.NoError(t, err)
 	require.NoError(t, f.Close())
@@ -37,6 +38,7 @@ func writeConfig(t *testing.T, cfg *Config) string {
 }
 
 func removeConfig(t *testing.T, name string) {
+	t.Helper()
 	require.NoError(t, os.Remove(name))
 }
 
--- a/api-tests/helpers.go
+++ b/api-tests/helpers.go
@@ -81,6 +81,7 @@ func AssertAPIErrorf(t TestingT, actual error, httpStatus int, grpcCode codes.Co
 }
 
 func ExpectFailure(t *testing.T, link string) *expectedFailureTestingT {
+	t.Helper()
 	return &expectedFailureTestingT{
 		t:    t,
 		link: link,
--- a/api-tests/management/alerting/alerting_test.go
+++ b/api-tests/management/alerting/alerting_test.go
@@ -38,6 +38,7 @@ import (
 // we don't enable or disable IA explicit in our tests since it is enabled by default through
 // ENABLE_ALERTING env var.
 func assertTemplate(t *testing.T, expectedTemplate alert.Template, listTemplates []*alerting.ListTemplatesOKBodyTemplatesItems0) {
+	t.Helper()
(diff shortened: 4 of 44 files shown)


BupycHuk:
Shouldn't we remove the old one?

update/ansible/playbook/tasks/files/datasources.yml
```diff
@@ -48,3 +48,9 @@ datasources:
   url: http://localhost:9093/alertmanager/
   jsonData:
     keepCookies: []
+- name: ClickHouse (New)
+  type: grafana-clickhouse-datasource
+  jsonData:
+    port: ${PERCONA_TEST_PMM_CLICKHOUSE_PORT}
+    server: ${PERCONA_TEST_PMM_CLICKHOUSE_HOST}
+    tlsSkipVerify: true
```
atymchuk, line 56: What if the database is provisioned externally? It seems like we may need TLS.
@0leksii @BupycHuk What do you think?

atymchuk:
> Shouldn't we remove the old one?

I also think we should.

0leksii:
> Shouldn't we remove the old one?

My understanding was that if it is in use by any of the dashboards - we cannot simply remove it, but deprecate and remove in 5 releases. 
From what I see, we use an old plugin in `Environments Overview (Designed for PMM)` dashboard. This one is in experimental, so should we consider the plugin in use, or for experimental dashboards we can ignore that?
@rnovikovP ^^ ??

0leksii:
> The main purpose of this task was to get rid of CVE. And not removing the old plugin doesn't fix the main problem. In that case we need to double an effort and update the old plugin as well.

I agree, but updating the plugin to the latest version won't help as it still have reported CVEs. You can check in comments here: https://jira.percona.com/browse/PMM-12118

atymchuk:
> > The main purpose of this task was to get rid of CVE. And not removing the old plugin doesn't fix the main problem. In that case we need to double an effort and update the old plugin as well.
> 
> I agree, but updating the plugin to the latest version won't help as it still have reported CVEs. You can check in comments here: https://jira.percona.com/browse/PMM-12118

This is true. However, native Grafana plugins usually run a higher chance of getting CVEs fixes asap, compared to plugins contributed by the community. We need to remember that our codebase is a few minor versions behind upstream, where those CVEs may have already been fixed, especially the critical and high severity ones.

update/ansible/playbook/tasks/files/datasources.yml
0leksii: Done

atymchuk: > Done

I'm curious: how did we address it?

update/ansible/playbook/tasks/files/datasources.yml
```diff
@@ -48,3 +40,9 @@ datasources:
   url: http://localhost:9093/alertmanager/
   jsonData:
     keepCookies: []
+- name: ClickHouse
```
atymchuk, line 43: I believe we need to delete the previous one by inserting these lines at the very top:
```
deleteDatasources:
  - name: ClickHouse
    orgId: 1
```

update/ansible/playbook/tasks/files/datasources.yml
```diff
@@ -48,3 +40,9 @@ datasources:
...
+  type: grafana-clickhouse-datasource
```
atymchuk, line 44: Can we add `orgId: 1` here as well?

update/ansible/playbook/tasks/files/datasources.yml
0leksii: Ouch.... It seems that I disabled it. And then thought I forgot to switch the flag and switched it back....
Sorry for that. Will disable it.

ritbl (approved):

atymchuk (approved):

PR #2159, opened 2023-05-19
PMM-7 clean up some gh actions

atymchuk:
PMM-7

This PR follows up on the recent refactor of go module and build cache. It fixes a few misses and combines a couple of sbom actions.

### DIFF
--- a/.github/workflows/managed.yml
+++ b/.github/workflows/managed.yml
@@ -1,9 +1,5 @@
 name: Managed
 on:
-  schedule:
-    # run every Sunday to re-populate caches after they are cleaned on Saturday
-    - cron: "0 12 * * 0"
-
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


.github/workflows/managed.yml
```diff
@@ -52,15 +47,13 @@ jobs:
             ${{ runner.os }}-go-build-
 
       - name: Enable Go modules cache
-        if: matrix.use-cache
```
atymchuk, line 55: These prevented the cache from being used in this action.

.github/workflows/managed.yml
```diff
@@ -1,9 +1,5 @@
 name: Managed
 on:
-  schedule:
-    # run every Sunday to re-populate caches after they are cleaned on Saturday
-    - cron: "0 12 * * 0"
-
```
atymchuk, line 6: The schedule is no longer relevant, since the cleanup action was removed earlier.

BupycHuk (approved):

.github/workflows/managed.yml
atymchuk: Ok, it's reverted.

artemgavrilov (approved):

PR #2186, opened 2023-05-25
PMM-12155 Use new option name for node_exporter

0leksii:
Since the new version of exporter-toolkit used in the node exporter the new option `--web.config.file` should be used instead to pass the TLS configuration.

PMM-12155

Link to the Feature Build: SUBMODULES-3284


JiriCtvrtka (approved):

managed/services/agents/node.go
```diff
@@ -30,7 +30,9 @@ import (
 // The node exporter prior 2.28 use exporter_shared and gets basic auth config from env.
 // Starting with pmm 2.28, the exporter uses Prometheus Web Toolkit and needs a config file
 // with the basic auth users.
-var v2_27_99 = version.MustParse("2.27.99")
+var (
+	v2_28_00 = version.MustParse("2.28.0-0")
```
idoqo, line 34: Btw do you think we should eventually move this to features for consistency?

idoqo (approved):

ademidoff (approved):

managed/services/agents/node.go
BupycHuk: I think we need to drop it eventually :)

version/features.go
```diff
@@ -0,0 +1,34 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package version
+
+// Versions list.
+var (
+	V3_0_0 = MustParse("3.0.0") //nolint:revive,stylecheck
```
BupycHuk, line 20:
```suggestion
	V3_0_0 = MustParse("3.0.0-0") //nolint:revive,stylecheck
```

PR #2217, opened 2023-06-01
PMM-7 disable `dupl` linter rule

atymchuk:
PMM-7

Ref: #1541 
Ref: https://github.com/mibk/dupl

This PR disables the rule since we can't really avoid code duplication between different services of a component as well as between different components of this monorepo.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -79,6 +79,7 @@ linters:
     - exhaustivestruct # too annoying
     - exhaustruct      # too many files to fix/nolint
     - deadcode         # unmaintained, we leverage `unused`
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -98,7 +99,6 @@ linters:
 
     # TODO: carefully review all the rules below and either fix the code
     # or leave disabled and provide a reason why
-    - paralleltest
```
atymchuk, line 101: This one was a duplicate entry.

BupycHuk (approved):

artemgavrilov (approved):

PR #2254, opened 2023-06-09
PMM-12175 fix supervisorctl path for EL7.

BupycHuk:
PMM-12175

Link to the Feature Build: SUBMODULES-0

If this PR adds or removes or alters one or more API endpoints, please review and add or update the relevant [API documents](https://github.com/percona/pmm/tree/main/docs/api) as well:

- [ ] API Docs updated

If this PR is related to some other PRs in this or other repositories, please provide links to those PRs:

- Links to related pull requests (optional).

PR #2257, opened 2023-06-09
Revert "PMM-12175 fix supervisorctl path for EL7."

talhabinrizwan:
Reverts percona/pmm#2254


BupycHuk:
let's fix this instead of reverting it

talhabinrizwan:
> let's fix this instead of reverting it

I'm just checking if it works without this commit in following RP:
https://github.com/Percona-Lab/pmm-submodules/pull/3299

BupycHuk:
Fixed in #2258

PR #2264, opened 2023-06-13
PMM-7 fix several CI issues

atymchuk:
PMM-7

This PR should make out pipelines green again. In particular, it:
- installs `vi` to PMM Server docker container
- fixes the PMM Update pipeline, which was failing (ansible-lint)
- bumps up Go version to 1.20.5 in Dockerfile(s) 
- fixes `yum clean plugins` call on OL9
- fixes git directory (pmm) ownership issue
- fixes a few `ansible-lint` errors and issues


build/docker/server/Dockerfile.el9
```diff
@@ -24,7 +24,7 @@ WORKDIR /opt
 RUN microdnf -y install yum && yum -y install python3-pip && \
     yum -y install oracle-epel-release-el9 ansible-core && \
     yum -y install epel-release && \
-    yum -y install ansible
+    yum -y install ansible vi
```
artemgavrilov, line 27: Let's install `vim` instead

update/.ansible-lint
```diff
@@ -1,4 +1,13 @@
...
-  - "403"  # "Package installs should not use latest" - but we need it
+  - "303"  # "supervisorctl used in place of supervisorctl module" - we need more control
+  - "403"  # "Package installs should not use latest" - we need latest
+
+warn_list:  # don't move to 'skip_list' - it will silence them completely
+  - literal-compare  # Don't compare to literal True/False
+  - risky-file-permissions # File permissions unset or incorrect
+  - no-changed-when  # Commands should not change things if nothing needs doing
+  - parser-error  # AnsibleParserError
+  - ignore-errors  # Use failed_when and specify error conditions instead of using ignore_errors
```
artemgavrilov, line 13: Missing newline

update/Makefile
```diff
@@ -23,14 +23,16 @@ LD_FLAGS = -ldflags " \
 			-X 'github.com/percona/pmm/version.FullCommit=$(PMM_RELEASE_FULLCOMMIT)' \
 			-X 'github.com/percona/pmm/version.Branch=$(PMM_RELEASE_BRANCH)' \
 			"
+prep: 													## OL9-specific
```
artemgavrilov, line 26: Can you please add better comment and align it. We use them for `make help` output generation.

artemgavrilov: BTW why it's needed for `9` but not needed for `7`? What is the difference?

update/Makefile
```diff
@@ -68,7 +70,7 @@ env-down:                       ## Stop development environment
 	docker-compose down --volumes --remove-orphans
 
 install-dev-tools:
-	docker exec pmm-update-server /root/go/src/github.com/percona/pmm/update/.devcontainer/install-dev-tools.sh
+	.devcontainer/install-dev-tools.sh
```
artemgavrilov, line 73: This target should be moved to `Make.devcontainer` then

update/ansible/playbook/tasks/update.yml
```diff
@@ -259,6 +262,7 @@
         - tzdata
         - libssh2
         - sshpass
+        - vi
```
artemgavrilov, line 265:
```suggestion
        - vim
```

build/docker/server/Dockerfile.el9
atymchuk: Vim is much heavier in terms of binary size. Given that we install it to the base image, I was trying not to blow up the image size.

update/Makefile
atymchuk: Ok, will do. RHEL 7 and 9 use different versions of git, so it's mainly because git was erroring out.

build/docker/server/Dockerfile.el9
atymchuk: Agreed to go ahead with `vi`.

update/.ansible-lint
atymchuk: Fixed.

update/Makefile
atymchuk: Commented.

update/Makefile
atymchuk: Not sure I understand. Can you suggest a fix?

atymchuk: Would that be a proper solution? https://github.com/percona/pmm/pull/2264/commits/fff4312b409b82ce7d317ac94ff1a8b7159e95e0

artemgavrilov: `Makefile` used on the host machine, if user will execute it then `install-dev-tools.sh` will try to install tools on the host machine, but it should happen only inside devcontainer.

atymchuk: Well, I doubt that things like `check`, `test`, `test-race` will run outside of the container. There is a good mix of targets that will and won't. 

Anyway, this seems to be blocking the review so I'm revering it for now.

PR #2298, opened 2023-06-22
PMM-7 Add backups dev env bootstrap

artemgavrilov:
PMM-7

1) Docker compose and scripts for easy backups dev environment setup
2) Rename `pmm-managed-server` -> `pmm-server`


.github/workflows/managed.yml
```diff
@@ -68,21 +68,21 @@ jobs:
       - name: Restore Go build cache
         if: ${{ fromJSON(env.DEVCONTAINER_CACHE_ENABLED) }}
         continue-on-error: true
-        run: docker cp ~/.cache/go-build pmm-managed-server:/root/.cache/go-build
+        run: docker cp ~/.cache/go-build pmm-server:/root/.cache/go-build
```
atymchuk, line 71: Can't we just mount this directory and the one below to the docker container instead of copying?

dev/mongo-rs-backups/rs-bootstrap.js
```diff
@@ -0,0 +1,34 @@
...
+        conn = new Mongo("localhost:27017");
+    } catch (Error) {
+        attempt++;
+    }
+
+    if (attempt >= 10) {
+        print("Max connection attempts exceeded.");
+        break;
+    }
+    sleep(100);
```
atymchuk, line 13: 10 times 100 would make only one second. Is that OK to try to connect 10 times within one second?

dev/mongo-rs-backups/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
+version: "4"
```
atymchuk, line 1: It's deprecated, see https://docs.docker.com/compose/compose-file/04-version-and-name/

.github/workflows/managed.yml
artemgavrilov: Probably, but not in the scope of this PR. Also now these caches are disabled.

atymchuk (approved):

dev/mongo-rs-backups/README.md
```diff
@@ -0,0 +1,25 @@
+## MongoDB replica set with PBM and PMM Agent
+
+This directory contains docker compose and scripts for Backups Dev environment bootstrap. This environment based
+on mongoDB replica set running in docker, as result it supports only logical backups/restores.
```
idoqo, line 4:
```suggestion
This directory contains docker compose and scripts for Backups Dev environment bootstrap. This environment is based
on MongoDB replica set running in docker, as a result, it only supports logical backups/restores.
```

dev/mongo-rs-backups/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM percona/percona-server-mongodb:4.2
```
idoqo, line 1: Maybe we can make the image tag configurable? Not super important, but e.g 4.4 is also out (and they are other versions)

idoqo (approved):

dev/mongo-rs-backups/Dockerfile
artemgavrilov: Probably, but for now let's leave it as is. It's just starting point to simplify dev environment creation, we will improve it over time.

PR #2302, opened 2023-06-23
PMM-7 enable `forbidigo` linter rule

atymchuk:
PMM-7

Ref: #1541 
Ref: https://github.com/ashanbrown/forbidigo

This PR:
- enables the rule
- suppresses the warnings as they seem to be all legitimate

Note: it seems like the linter doesn't allow to ignore files or directories with a pattern.

### DIFF
--- a/agent/commands/setup.go
+++ b/agent/commands/setup.go
@@ -50,7 +50,7 @@ func Setup() {
 
 	var e config.ConfigFileDoesNotExistError
 	if err != nil && !errors.As(err, &e) {
-		fmtp.Printf("Failed to load configuration: %s.\n", err) //nolint:forbidigo
+		fmtp.Printf("Failed to load configuration: %s.\n", err)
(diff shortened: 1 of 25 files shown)


agent/commands/setup.go
```diff
@@ -15,7 +15,7 @@
 package commands
 
 import (
-	"fmt"
+	fmtp "fmt"
```
artemgavrilov, line 18: Instead we can apply `nolint` to the whole file. Renaming std package just to bypass linter is not best idea.

atymchuk: [Done](https://github.com/percona/pmm/pull/2302/files#diff-10b7abc985dfc8f94129edff8605bf58f6264ebf6f2bf30a0f5e180827c9bb2dR15)

gen1us2k (approved):

artemgavrilov (approved):

PR #2324, opened 2023-06-28
PMM-6279 Allow show multiple metadata values.

JiriCtvrtka:
PMM-6279

- [ ] API Docs updated


qan-api2/models/metrics.go
```diff
@@ -1120,26 +1116,63 @@ func (m *Metrics) GetSelectedQueryMetadata(ctx context.Context, periodStartFromS
...
-			&res.ServiceType,
-			&res.ServiceId,
-			&res.Environment,
-			&res.NodeId,
-			&res.NodeName,
-			&res.NodeType)
+		row := make([]any, len(columnNames))
+		for i := range row {
+			row[i] = &row[i]
+		}
```
artemgavrilov, line 1129: It seems we have some problem here.

qan-api2/models/metrics.go
```diff
@@ -1120,26 +1116,63 @@ func (m *Metrics) GetSelectedQueryMetadata(ctx context.Context, periodStartFromS
...
+func valueIsInArray(value string, array []string) bool {
+	for _, v := range array {
+		if v != value {
+			continue
+		}
+
+		return true
+	}
+
+	return false
```
artemgavrilov, line 1177:
```suggestion
	for _, v := range array {
		if v == value {
			return true
		}
	}

	return false
```

artemgavrilov (changes requested):

qan-api2/models/metrics.go
JiriCtvrtka: Without that I cannot read it into map. Any ideas how to handle this better?

artemgavrilov: Aha, it seems that I got what you're trying to do. Try this, it should be easier to understand:
```suggestion
		row := make([]any, len(columnNames))
		for i := range columnNames {
			row[i] = new(sql.RawBytes)
		}
```

artemgavrilov: Or `new(any)`, probably will work as well

JiriCtvrtka: Nice thanks. I will test it next week right after national holidays.

qan-api2/models/metrics.go
```diff
@@ -1120,26 +1116,61 @@ func (m *Metrics) GetSelectedQueryMetadata(ctx context.Context, periodStartFromS
 	}
 	defer rows.Close()
 
+	metadata := make(map[string][]string)
```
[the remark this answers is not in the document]
BupycHuk: in that case we don't need to check if the value is exist

BupycHuk (changes requested):
Suggested performance improvements

qan-api2/models/metrics.go
JiriCtvrtka: In the end I used new(string).

JiriCtvrtka:
All required changes should be addressed now.

artemgavrilov (approved):

BupycHuk (approved):
👍🏼

PR #2341, opened 2023-07-03
PMM-11658 move the exporter configs away from /tmp

atymchuk:
PMM-11658

- [ ] API Docs updated


.github/workflows/admin.yml
```diff
@@ -126,7 +126,6 @@ jobs:
 
       - name: Setup tools
         run: |
-          sudo apt-get install -y wget jq
```
atymchuk, line 129: This was redundant as both binaries are already installed by default.

managed/cmd/pmm-managed/main.go
```diff
@@ -596,9 +597,14 @@ func setup(ctx context.Context, deps *setupDeps) bool {
 }
 
 func getQANClient(ctx context.Context, sqlDB *sql.DB, dbName, qanAPIAddr string) *qan.Client {
+	bc := backoff.DefaultConfig
```
atymchuk, line 600: This migrates from the deprecated `grpc.WithBackoffMaxDelay`.

managed/cmd/pmm-managed/main.go
```diff
@@ -558,15 +559,15 @@ func setup(ctx context.Context, deps *setupDeps) bool {
 	deps.l.Infof("Updating supervisord configuration...")
 	settings, err := models.GetSettings(db.Querier)
 	if err != nil {
-		deps.l.Warnf("Failed to get settings: %+v.", err)
+		deps.l.Warnf("Failed to get settings: %s.", err)
```
atymchuk, line 562: It was a bit weird to see a part of the stack trace printed at the start of pmm-managed.

`
[2023-06-26T14:35:15.947+00:00] Failed to get Percona SSO Details: PMM Server is not connected to Portal
github.com/percona/pmm/managed/models.init
        /home/builder/rpm/BUILD/pmm-7a7431afdc35098f665dbd8d33ccf914fad8b0b5/src/github.com/percona/pmm/managed/models/percona_sso_model_helpers.go:35
runtime.doInit
        /usr/local/go/src/runtime/proc.go:6507
runtime.doInit
        /usr/local/go/src/runtime/proc.go:6484
runtime.main
        /usr/local/go/src/runtime/proc.go:233
runtime.goexit
        /usr/local/go/src/runtime/asm_amd64.s:1598.  ESC[33mcomponentESC[0m=setup
`

.golangci.yml
```diff
@@ -7,7 +7,7 @@ linters-settings:
     max-complexity: 30
 
   depguard:
-    list-type: blacklist
+    list-type: denylist
```
atymchuk, line 10: Used to be part of offensive terminology :)

agent/config/config.go
```diff
@@ -474,7 +489,7 @@ func Application(cfg *Config) (*kingpin.Application, *string) {
 // Other errors are returned if file exists, but configuration can't be loaded due to permission problems,
 // YAML parsing problems, etc.
 func loadFromFile(path string) (*Config, error) {
-	if _, err := os.Stat(path); os.IsNotExist(err) {
+	if _, err := os.Stat(path); errors.Is(err, fs.ErrNotExist) {
```
atymchuk, line 492: `os.IsNotExist` is deprecated

agent/config/config.go
```diff
@@ -237,6 +241,25 @@ func get(args []string, cfg *Config, l *logrus.Entry) (configFileF string, err e
 			cfg.Paths.ExportersBase = abs
 		}
 
+		if cfg.Paths.TempDir == "" {
+			wd, err := os.Getwd()
+			if err == nil {
+				cfg.Paths.TempDir = filepath.Join(wd, agentTmpPath)
+			} else {
+				// As a last resort, if unable to get the workdir, use system temp directory.
+				cfg.Paths.TempDir = filepath.Join(os.TempDir(), agentTmpPath)
```
BupycHuk, line 250: I think we need to warn about that.

agent/config/config.go
```diff
@@ -237,6 +241,25 @@ func get(args []string, cfg *Config, l *logrus.Entry) (configFileF string, err e
 			cfg.Paths.ExportersBase = abs
 		}
 
+		if cfg.Paths.TempDir == "" {
+			wd, err := os.Getwd()
```
BupycHuk, line 245: it should be relative to `cfg.Paths.PathsBase`

agent/config/config_test.go
```diff
@@ -92,6 +104,9 @@ func TestGet(t *testing.T) {
 		}, &actual, logrus.WithField("test", t.Name()))
 		require.NoError(t, err)
 
+		tmpDir := generateTempDir(t)
+		defer removeTempDir(t, tmpDir)
```
BupycHuk, line 108: let's use t.Cleanup

agent/config/config_test.go
```diff
@@ -42,6 +42,18 @@ func removeConfig(t *testing.T, name string) {
 	require.NoError(t, os.Remove(name))
 }
 
+func generateTempDir(t *testing.T) string {
```
BupycHuk, line 45: it seems more like generateTempDirPath

agent/config/config_test.go
```diff
@@ -143,6 +158,9 @@ func TestGet(t *testing.T) {
 		}, &actual, logrus.WithField("test", t.Name()))
 		require.NoError(t, err)
 
+		tmpDir := generateTempDir(t)
+		defer removeTempDir(t, tmpDir)
```
BupycHuk, line 162: seems like we can move all these calls to the top level of these tests in TestGet

BupycHuk (changes requested):

agent/config/config_test.go
atymchuk: Done.

agent/config/config_test.go
atymchuk: Done.

agent/config/config_test.go
atymchuk: Done.

agent/config/config.go
atymchuk: What if we do not try to put the config file to `os.TempDir()` since this is essentially where the issue originated from?

BupycHuk: that sounds good to me

agent/config/config.go
atymchuk: I will discuss with some heavy users of PMM to see if this is a good choice.

agent/config/config.go
```diff
@@ -237,6 +241,25 @@ func get(args []string, cfg *Config, l *logrus.Entry) (configFileF string, err e
...
+			if err == nil {
+				cfg.Paths.TempDir = filepath.Join(wd, agentTmpPath)
+			} else {
+				// As a last resort, if unable to get the workdir, use system temp directory.
+				cfg.Paths.TempDir = filepath.Join(os.TempDir(), agentTmpPath)
+			}
+			l.Infof("Temporary directory is undefined, will create one at %q", cfg.Paths.TempDir)
+			createTempDir(cfg.Paths.TempDir, l)
+		} else {
+			l.Infof("Temporary directory is defined as %q", cfg.Paths.TempDir)
```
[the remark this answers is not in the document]
atymchuk: Done.

agent/config/config.go
BupycHuk: other fields works that way, so it should work the same

atymchuk: Done, pending validation from MS.

agent/config/config.go
atymchuk: Done.

build/docker/server/Dockerfile.el9
```diff
@@ -6,6 +6,7 @@ ARG BUILD_DATE
 ENV LANG=en_US.utf8
 ENV LC_ALL=en_US.utf8
 ENV GF_PLUGIN_DIR=/srv/grafana/plugins
+ENV PS1="[\u@\h \W]\$ "
```
atymchuk, line 9: Make the prompt more informative.

agent/agents/supervisor/supervisor.go
```diff
@@ -359,6 +359,12 @@ func (s *Supervisor) setBuiltinAgents(builtinAgents map[string]*agentpb.SetState
 		<-agent.done
 
 		delete(s.builtinAgents, agentID)
+
+		agentTmp := filepath.Join(s.cfg.Get().Paths.TempDir, strings.ToLower(agent.requestedState.Type.String()), agentID)
```
JiriCtvrtka, line 363: Are we sure that Paths cannot be nil?

atymchuk: Pretty sure. If the agent is about to be stopped, which means it is running, its `cfg.Paths` must exist.

BupycHuk (approved):

JiriCtvrtka (approved):

agent/config/config.go
```diff
@@ -237,6 +241,21 @@ func get(args []string, cfg *Config, l *logrus.Entry) (configFileF string, err e
 			cfg.Paths.ExportersBase = abs
 		}
 
+		if cfg.Paths.TempDir == "" {
+			cfg.Paths.TempDir = filepath.Join(cfg.Paths.PathsBase, agentTmpPath)
+			l.Infof("Temporary directory is not configured, will attempt to create it at %s", cfg.Paths.TempDir)
+			createTempDir(cfg.Paths.TempDir, l)
```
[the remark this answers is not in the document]
atymchuk: > What if we remove this line? Do features keep working?

Everything keeps working, however the exporter configs still go to `os.TempDir()` :(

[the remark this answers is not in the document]
BupycHuk: that's weird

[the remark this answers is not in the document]
atymchuk: > I'm not sure that we should create a directory in getConfig function

I do agree, that reads bad given the function name and purpose. I was actually looking to put the creation of the TempDir inside of agent's Setup, but then decided to put it close to where we defined the paths. Let me see if I can find a better place.

agent/config/config.go
```diff
@@ -510,11 +529,18 @@ func SaveToFile(path string, cfg *Config, comment string) error {
...
 		}
 		return err
 	}
 	return unix.Access(path, unix.W_OK)
 }
+
+func createTempDir(path string, l *logrus.Entry) {
+	err := os.MkdirAll(path, 0o700)
+	if err != nil {
+		l.WithError(err).Panicf("Unable to create the temporary directory at %s", path)
```
artemgavrilov, line 544: Wow, I think it's better to let top level code to decide what to do in case of error rather than panic in this function

atymchuk: > It looks like depguard configuration format has changed: https://golangci-lint.run/usage/linters/#depguard

That must be the latest one, which we don't use yet.

atymchuk: > Wow, I think it's better to let top level code to decide what to do in case of error rather than panic in this function

Then we don't need a function. The whole point was to invariably panic. If the TempDir can't be created, there is no point to continue. That would avoid repeating lines 543 through 545.

agent/config/config.go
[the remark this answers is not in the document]
atymchuk: It turned out to be unnecessary, so now the code looks much cleaner :)

agent/config/config.go
```diff
@@ -237,6 +241,21 @@ func get(args []string, cfg *Config, l *logrus.Entry) (configFileF string, err e
 			cfg.Paths.ExportersBase = abs
 		}
 
+		if cfg.Paths.TempDir == "" {
+			cfg.Paths.TempDir = filepath.Join(cfg.Paths.PathsBase, agentTmpPath)
+			l.Infof("Temporary directory is not configured, will attempt to create it at %s", cfg.Paths.TempDir)
+			createTempDir(cfg.Paths.TempDir, l)
+		} else {
+			l.Debugf("Temporary directory is configured as %s", cfg.Paths.TempDir)
+			_, err := os.Stat(cfg.Paths.TempDir)
```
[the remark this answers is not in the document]
atymchuk: Ok, thanks! Now resolved.

artemgavrilov (approved):

.golangci.yml
atymchuk: OK, I'm removing this change so we can address it separately.

PR #2365, opened 2023-07-12
PMM-12231 set grafana as owner of the plugin dir

atymchuk:
PMM-12231

Link to the Feature Build: SUBMODULES-3342


update/ansible/playbook/tasks/roles/dashboards_upgrade/tasks/main.yml
```diff
@@ -50,6 +50,14 @@
     src: /usr/share/percona-dashboards/panels/
     dest: /srv/grafana/plugins/
 
+- name: Set permissions for the plugin directory
```
atymchuk, line 53: Note: we are not going recursive here, just need the top-level to be owned by `grafana` user. The files underneath will have the proper permissions set by the spec.

BupycHuk (approved):

PR #2370, opened 2023-07-17
PMM-11770: Use PMM public address as Grafana domain

idoqo:
PMM-11770

Link to the Feature Build: SUBMODULES-3348

- [ ] API Docs updated


managed/services/supervisord/supervisord.go
```diff
@@ -48,6 +48,7 @@ import (
 
 const (
 	defaultClickhouseDatabase       = "pmm"
+	defaultHostAddress              = "localhost"
```
atymchuk, line 51: It seems like this should be 127.0.0.1, since `localhost` could, in theory, map to anything through `/etc/hosts`.

atymchuk: Also, our nginx.conf operates mostly with `127.0.0.1`.

BupycHuk (changes requested):
please fix the test related to supervisord

JiriCtvrtka (approved):

managed/services/server/server.go
```diff
@@ -510,9 +511,20 @@ func (s *Server) validateChangeSettingsRequest(ctx context.Context, req *serverp
...
+			return status.Error(codes.InvalidArgument, "Both pmm_public_address and remove_pmm_public_address are present.")
+		}
+
+		publicURL, err := url.Parse(req.PmmPublicAddress)
+		if err != nil {
+			return status.Error(codes.InvalidArgument, "Provided PMM public address is not a valid URL.")
+		}
+		if publicURL.Scheme != "" {
+			return status.Error(codes.InvalidArgument, "PMM public address should not include the URL scheme")
+		}
```
artemgavrilov, line 525: This is the new condition as I see, right? If so, what users which already saved public address with scheme before should do?

idoqo: PMM will continue to work fine for them, but Grafana domain will still be localhost.

They will need to update the public address to update Grafana domain, then we will apply this check.

artemgavrilov: I mean user with previous PMM version was able to save public address with schema, right?

artemgavrilov: I checked, it works in recent version: 
image

atymchuk (approved):

atymchuk:
@idoqo Don't we want to test it with a feature build?

managed/services/server/server.go
idoqo: I've reworked it to only use the domain part for grafana.ini

managed/services/supervisord/supervisord.go
```diff
@@ -445,13 +446,20 @@ func (s *Service) marshalConfig(tmpl *template.Template, settings *models.Settin
 
 	s.addPostgresParams(templateParams)
 
+	templateParams["PMMServerAddress"] = defaultHostAddress
```
artemgavrilov, line 449: Minor: I think it's also makes sense to rename this template param to `PMMServerHost`  or something like this.

artemgavrilov (approved):

artemgavrilov:
@idoqo tests are failing

managed/services/supervisord/supervisord.go
```diff
@@ -782,8 +790,8 @@ command =
     /usr/sbin/grafana server
         --homepath=/usr/share/grafana
         --config=/etc/grafana/grafana.ini
+        cfg:default.server.domain="{{ .PMMServerHost }}"
```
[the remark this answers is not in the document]
idoqo: Yeah, that is what we try to do here https://github.com/percona/pmm/blob/3bbd0ff9ac088931c43463fb87c7fab15abb03f9/managed/services/supervisord/supervisord.go#L449-L450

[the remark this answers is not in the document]
BupycHuk: actually no, there still will be a line
```
cfg:default.server.domain=""
```

BupycHuk (approved):

PR #2376, opened 2023-07-19
PMM-4817 move git tags script to pmm

atymchuk:
PMM-4817

- https://github.com/Percona-Lab/jenkins-pipelines/pull/2217

Note: please review the linked PR as well.


artemgavrilov (approved):

build/scripts/create-tags
```diff
@@ -0,0 +1,67 @@
...
+    DIR_NAME=${REPO##*/} # grab the repo name from <company>/<repo>
+    rm -fr "$DIR_NAME" || true
+    mkdir -p "$DIR_NAME"
+
+    pushd "$DIR_NAME" >/dev/null
+        git clone https://github.com/${REPO} ./
+        # The default is https, and we want to set it to ssh
+        git remote set-url origin git@github.com:${REPO}.git
+
+        BRANCH="pmm-${VERSION}"
```
BupycHuk, line 44: we are planning to change our branching logic from `2.39.0` to `2.39.x` where `x` is just symbol `x`, not a number, so this code will be outdated pretty soon. In Python script we were relying on pmm-submodules which uses git submodule and knows exact commit which was used

build/scripts/create-tags
```diff
@@ -0,0 +1,67 @@
...
+        if ! git checkout "$BRANCH"; then
+            echo "Warning: failed to checkout the repository $REPO to $BRANCH"
+            popd >/dev/null
+            rm -rf "$DIR_NAME"
+            continue
+        fi
+        echo "SHA: $(git rev-parse HEAD)"
+
+        # If the tag already exists, we want to delete it and re-tag this SHA
```
BupycHuk, line 53: do we? it's a bad practice

build/scripts/create-tags
```diff
@@ -0,0 +1,67 @@
...
+        if git tag -l "$TAG"; then
+            git tag --delete "$TAG"
+            git push --delete origin "$TAG"
+        fi
+
+        git tag --message="Version $TAG." "$TAG"
+        if ! git push origin "$TAG"; then
+            echo "Warning: failed to tag the repository $REPO with $TAG"
+        fi
```
BupycHuk, line 62: it shouldn't be just a warning, the script should fail

build/scripts/create-tags
BupycHuk: it's better to warn and skip in case we already have a tag

BupycHuk (commented):
what do you think if we write the code in Go instead of Bash, we don't have enough Bash experts in our team to maintain it

build/scripts/create-tags
atymchuk: > do we? it's a bad practice

Well, that has already happened a few times in the past. But OK, no problem to bail out. I will update the code.

build/scripts/create-tags
atymchuk: > it shouldn't be just a warning, the script should fail

I believe it should not fail - it is part of the requirement as per the ticket.

build/scripts/create-tags
BupycHuk: When? That was referred by @EvgeniyPatlan as a bad practice, but I don't remember we've done it for pmm. Only for mongodb_exporter, but their initial tags were done by mistake.

build/scripts/create-tags
atymchuk: I see no value added if all that Go has to do is to run the very same commands using "os/exec". We can certainly use submodules for this. Should we?

build/scripts/create-tags
BupycHuk: > Do tagging as the last step, so if it fails, we can still do it by hand. It should not break RPM pushing, Docker image tagging, etc.
it doesn't say that it shouldn't fail. It says that if it fails it shouldn't break RPM pushing, and it can be done simply by making it another Jenkins job which is called if the main one finished successfully. We already have this kind of job for devcontainer

build/scripts/create-tags
BupycHuk: Yes, I think. if we will migrate to new branching logic it will make more sense

build/scripts/create-tags
atymchuk: Pushing a tag to one repo could fail, the other three could succeed.

BupycHuk: Yes, and the whole job should fail if any of repos weren't tagged

build/scripts/create-tags
atymchuk: Are you saying we will tag releases with `2.40.x`? How is that?

BupycHuk: no, branch name will be 2.40.x, but tags will be 2.40.0, 2.40.1, 2.40.2, etc...

atymchuk: Done.

artemgavrilov:
> @artemgavrilov The codecov check is disrupting the flow. Can we do something about it?

I have an idea, but it's based on non-GA feature in go. In general root of the problem that we run agent tests multiple times on different DB versions and generate one report. So last run wins, because of that we have this flip-flop coverage for agent. In recent go versions were added binary format for coverage reports, and there is a command that can merge multiple reports into one. But for unit tests binary format is not GA option yet.

atymchuk:
> > @artemgavrilov The codecov check is disrupting the flow. Can we do something about it?
> 
> I have an idea, but it's based on non-GA feature in go. In general root of the problem that we run agent tests multiple times on different DB versions and generate one report. So last run wins, because of that we have this flip-flop coverage for agent. In recent go versions were added binary format for coverage reports, and there is a command that can merge multiple reports into one. But for unit tests binary format is not GA option yet.

Do you think we can use this feature for the tests?

artemgavrilov:
> > > @artemgavrilov The codecov check is disrupting the flow. Can we do something about it?
> > 
> > 
> > I have an idea, but it's based on non-GA feature in go. In general root of the problem that we run agent tests multiple times on different DB versions and generate one report. So last run wins, because of that we have this flip-flop coverage for agent. In recent go versions were added binary format for coverage reports, and there is a command that can merge multiple reports into one. But for unit tests binary format is not GA option yet.
> 
> Do you think we can use this feature for the tests?

I created ticket for that: https://jira.percona.com/browse/PMM-12365
You can try to solve this problem if you want.

BupycHuk (approved):

PR #2393, opened 2023-08-07
PMM-12361 enable pmm-admin inventory help command

atymchuk:
PMM-12361

Link to the Feature Build: SUBMODULES-3358

The top-level `pmm-admin inventory` is a "hidden" command, which was done on purpose so that user favor as shorter and therefore the more memorable  `pmm-admin add/remove service` commands. However, `pmm-admin inventory list/add/remove` were also hidden, which was contributing to the reported problem, since Kong was unable to execute the `pmm-admin inventory` command. 

The solution was to implement the inventory's `RunCmd()` method and "unhide" `pmm-admin inventory list/add/remove` comands, while keeping `pmm-admin inventory` hidden.


admin/commands/inventory/inventory.go
```diff
@@ -23,9 +23,9 @@ import (
 
 // InventoryCommand is used by Kong for CLI flags and commands.
 type InventoryCommand struct {
-	List   ListCommand   `cmd:"" hidden:"" help:"List inventory commands"`
```
idoqo, line 26: Was the `hidden` tag meant to prevent the command/subcommand from being shown in the help text?

idoqo (approved):

admin/commands/inventory/inventory.go
atymchuk: Yes, it was meant not to show up, but such an approach turned out to be counter productive and was reported as a bug. It's quite enough to hide the top-level `pmm-admin inventory` command, which is in effect hidden right now.

JiriCtvrtka (approved):

PR #2398, opened 2023-08-08
PMM-12393 can't login to PMM with SSO

atymchuk:
PMM-12393


update/ansible/playbook/tasks/roles/grafana/files/grafana.ini
```diff
@@ -24,7 +24,7 @@ format = console
 # enable gzip
 enable_gzip = true
 # The public facing domain name used to access grafana from a browser
-domain = 127.0.0.1
+domain = localhost
```
BupycHuk, line 27: as far as I see, we didn't have these lines at all

atymchuk: Not really, they have always come from the config overrides, which has the same effect but is much les maintainable. Let me try the feature build - https://github.com/Percona-Lab/pmm-submodules/pull/3359

BupycHuk: no, in config overrides, we had that only if a public address is provided.

atymchuk: yes, which is exactly when we need to override it, otherwise it should always be 127.0.0.1

BupycHuk: oh, wait I was wrong it was set only when PMM connected to the portal.

atymchuk: yeah, and that's the only case when we need to change it

BupycHuk (approved):

PR #2405, opened 2023-08-09
PMM-12405 Handle empty schema captured and not included in query.

JiriCtvrtka:
PMM-12405

- [ ] API Docs updated


agent/runner/actions/mysql_explain_action.go
```diff
@@ -83,6 +85,9 @@ func (a *mysqlExplainAction) Run(ctx context.Context) ([]byte, error) {
 		return nil, errors.New("Query to EXPLAIN is empty")
 	}
 
+	logrus.Infof("Schema used in explain: %s", a.params.Schema)
```
atymchuk, line 88: I wonder if those two info log messages should not be invoked with `.Debugf`? Won't these make the logs too verbose?

agent/runner/actions/mysql_explain_action.go
```diff
@@ -258,3 +254,17 @@ func (a *mysqlExplainAction) explainTraditionalJSON(ctx context.Context, tx *sql
 	}
 	return jsonRows(columns, dataRows)
 }
+
+func checkExplainError(err error, schema string) error {
+	if err != nil {
+		if strings.Contains(err.Error(), errNoDatabaseSelectedCode) {
```
atymchuk, line 260: Is there a way to avoid string comparison? Can't we leverage `errors.Is` or `errors.As` somehow? I believe it will make the code more maintainable.

agent/runner/actions/mysql_explain_action.go
```diff
@@ -245,11 +244,8 @@ func (a *mysqlExplainAction) explainJSON(ctx context.Context, tx *sql.Tx) ([]byt
 
 func (a *mysqlExplainAction) explainTraditionalJSON(ctx context.Context, tx *sql.Tx) ([]byte, error) {
 	rows, err := tx.QueryContext(ctx, fmt.Sprintf("EXPLAIN /* pmm-agent */ %s", a.params.Query), prepareValues(a.params.Values)...)
-	if err != nil {
-		if strings.Contains(err.Error(), errNoDatabaseSelectedCode) {
-			return nil, errors.Wrap(err, errNoDatabaseSelectedMessage)
-		}
-		return nil, err
+	if e := checkExplainError(err, a.params.Schema); e != nil {
```
atymchuk, line 247: nit: I'd call `checkExplainError` only if `err != nil`

artemgavrilov: I'm agree with Alex

agent/runner/actions/mysql_explain_action.go
```diff
@@ -35,6 +36,7 @@ import (
 const (
 	errNoDatabaseSelectedCode    = "Error 1046 (3D000)"
 	errNoDatabaseSelectedMessage = "Database name is not included in this query. Explain could not be triggered without this info"
+	errNoDatabaseProvidedMessage = "Captured database name is empty and it is not even included in query. Run explain is not possible"
```
artemgavrilov, line 39: TBH I don't really understand what is the difference between these two errors

JiriCtvrtka: We need two different error messages for:
1. Case schema is captured by MySQL and not included in query (explain should be possible), but explain failed anyway for some reason. Could be bug.
2. Case schema is not captured by MySQL and not included in query then explain is not possible at all. Not bug.

Without two different messages its very hard to figure out what happened from customers reports and then we need ask them to debug problem to decide which situation happened.

BupycHuk, line 39: let's make a message more friendly
```suggestion
	errNoDatabaseProvidedMessage = "Captured database name is empty in query source and not included in the query. it is not possible to execute explain."
```

BupycHuk (approved):

agent/runner/actions/mysql_explain_action.go
```diff
@@ -258,3 +254,14 @@ func (a *mysqlExplainAction) explainTraditionalJSON(ctx context.Context, tx *sql
...
+
+func checkExplainError(err error, schema string) error {
+	if !strings.Contains(err.Error(), errNoDatabaseSelectedCode) {
+		return err
+	}
+	if schema == "" {
+		return errors.Wrap(err, errNoDatabaseProvidedMessage)
+	}
+
+	return errors.Wrap(err, errNoDatabaseSelectedMessage)
```
BupycHuk, line 266: how does it's possible to come to this line?

PR #2459, opened 2023-09-04
PMM-11670 Set the cluster label if empty

atymchuk:
PMM-11670

Link to the Feature Build: SUBMODULES-3378

Note: the test coverage was already more than sufficient, so there are no new tests, only updates to make them pass.


managed/models/service_helpers_test.go
```diff
@@ -158,10 +158,11 @@ func TestServiceHelpers(t *testing.T) {
 		services, err = models.FindServices(q, models.ServiceFilters{NodeID: "N1"})
 		assert.NoError(t, err)
 		assert.Equal(t, 3, len(services))
-		assert.Equal(t, services, []*models.Service{{
+		assert.Equal(t, []*models.Service{{
```
atymchuk, line 161: Note: the signature of `assert.Equal` is `(t, expected, actual, ...)` and not `(t, actual, expected)`. This puts it in the right order.

idoqo: I wonder if `assert.ElementsMatch` is a better option here (but fine if we can always guarantee that both services will be in this same order).

idoqo (approved):

managed/models/service_helpers_test.go
atymchuk: We chose `assert.Equal` because we want to check whether services are appearing in the exact order they were added, and not in a random order.

BupycHuk (approved):

.github/workflows/managed.yml
```diff
@@ -112,4 +112,4 @@ jobs:
           go env
           pwd
           git status
-          kubectl version --short --output json
+          kubectl version --output json
```
artemgavrilov, line 115: We don't use `kubernetes` at all in this CI as I see, I guess we can remove this line completely.

artemgavrilov (approved):

.github/workflows/managed.yml
atymchuk: Done.

PR #2474, opened 2023-09-11
PMM-12457 add node_name to agent metrics

atymchuk:
PMM-12457

Link to the Feature Build: SUBMODULES-3383


managed/data/iatemplates/agent_down.yml
```diff
@@ -8,6 +8,5 @@ templates:
     severity: critical
     annotations:
       description: |-
-        PMM agent on node '{{ $labels.node_id }}' cannot be reached. Host may be down.
-      summary: PMM agent on node '{{ $labels.node_id }}' cannot be reached. Host may be down.
-
+        PMM agent on node '{{ $labels.node_name }}', node ID '{{ $labels.node_id }}', cannot be reached. Host may be down.
```
atymchuk, line 11: @cezmunsta Could you please check the description and the summary and let us know if they need more love?

managed/services/inventory/inventory_metrics.go
```diff
@@ -103,6 +103,16 @@ func (i *InventoryMetrics) GetAgentMetrics(ctx context.Context) ([]Metric, error
 			return err
 		}
 
+		dbNodes, err := models.FindNodes(tx.Querier, models.NodeFilters{})
```
JiriCtvrtka, line 106: Maybe we can use FindNodesByID
https://github.com/percona/pmm/blob/6425880d68e8a2c9e0db38b2837175aca791f1a5/managed/models/node_helpers.go#L117

JiriCtvrtka (approved):
LGTM, just a minor comment.

managed/services/inventory/inventory_metrics.go
atymchuk: Not in this case - we are fetching all nodes to create a map with nodeID and nodeName to be able then to pull the nodeName for each of the agents. This is much faster compared to querying nodes one-by-one. :)

managed/data/iatemplates/agent_down.yml
cezmunsta: It wasn't so much about the description and summary, since the user can edit those, but that they could not add the information as it is not present. So long as the user isn't forced to try to manually determine which instance to look at, it seems fine

cezmunsta (approved):

BupycHuk (approved):

artemgavrilov (approved):

PR #2505, opened 2023-09-27
PMM-7 Fix the upgrade test (dry-run)

atymchuk:
The test fails with "ERROR: no such process/group: clickhouse"

Screenshot 2023-09-27 at 12 37 08

### DIFF
--- a/update/ansible/playbook/tasks/roles/clickhouse/tasks/main.yml
+++ b/update/ansible/playbook/tasks/roles/clickhouse/tasks/main.yml
@@ -1,6 +1,8 @@
 ---
 - name: Stop and remove clickhouse before update | EL7
-  when: ansible_distribution == 'CentOS' and ansible_distribution_major_version == '7'
+  when: 
+    - ansible_distribution == 'CentOS'
+    - ansible_distribution_major_version == '7'
   command: supervisorctl {{ item }} clickhouse
   changed_when: True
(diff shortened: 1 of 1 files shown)


artemgavrilov (approved):

update/ansible/playbook/tasks/roles/clickhouse/tasks/main.yml
```diff
@@ -13,7 +15,11 @@
   register: supervisord_socket
 
 - name: Stop and remove clickhouse before update | EL9
-  when: supervisord_socket.stat.exists and (ansible_distribution == 'OracleLinux' or ansible_distribution == 'AlmaLinux') and ansible_distribution_major_version == '9'
+  when:
+    - supervisord_socket.stat.exists
+    - ansible_distribution == 'OracleLinux' or ansible_distribution == 'AlmaLinux'
+    - ansible_distribution_major_version == '9'
+    - not ansible_check_mode
```
BupycHuk, line 22: should we add the same for EL7? How does it fixes upgrade?

atymchuk: We can, but it's hardly needed, since we only run this upgrade test in dry-run mode on EL9.

BupycHuk (approved):

PR #2531, opened 2023-10-09
PMM-12576 Change admin ID on update.

BupycHuk:
PMM-12576

Link to the Feature Build: SUBMODULES-3411

- [ ] API Docs updated


ademidoff:
@BupycHuk I see the FB container failed... I suppose we need to be sure it builds OK to be able to test it.

update/ansible/playbook/tasks/roles/initialization/tasks/main.yml
```diff
@@ -160,6 +163,12 @@
             - skip_ansible_lint # '503 Tasks that run when changed should likely be handlers'.
       when: is_database_sqlite.stat.exists
 
+    - name: Change default admin id
```
ademidoff, line 166: I don't think we need it here since the same will run on each container start in `update/ansible/playbook/tasks/update.yml`.

ademidoff (changes requested):

update/ansible/playbook/tasks/roles/initialization/tasks/main.yml
BupycHuk: it shouldn't, because there are `when` condition for the whole block and `update.yml` doesn't work for docker-way upgrade.

artemgavrilov (approved):

PR #2534, opened 2023-10-10
PMM-12422 restrict agent interface

idoqo:
PMM-12422

Link to the Feature Build: SUBMODULES-3420

- [ ] API Docs updated


api/managementpb/agent/agent.proto
```diff
@@ -140,6 +140,8 @@ message UniversalAgent {
   string version = 37;
   // True if Agent is running and connected to pmm-managed.
   bool is_connected = 38;
+  // True if an exporter agent is exposed on all host addresses.
+  bool expose_exporter_address = 40;
```
ademidoff, line 144: I wonder if 39 is already reserved?

api/managementpb/mongodb.proto
```diff
@@ -92,6 +92,8 @@ message AddMongoDBRequest {
   bool enable_all_collectors = 30;
   // Exporter log level
   inventory.LogLevel log_level = 31;
+  // Optionally expose the node_exporter process on 0.0.0.0
+  bool expose_exporter_address = 33;
```
ademidoff, line 96: Same here for 32?

managed/models/database.go
```diff
@@ -915,6 +915,12 @@ var databaseSchema = [][]string{
 		`ALTER TABLE agents
 		ALTER COLUMN comments_parsing_disabled DROP DEFAULT`,
 	},
+	85: {
+		`ALTER TABLE agents 
+		ADD COLUMN expose_exporter_address BOOLEAN NOT NULL DEFAULT FALSE;`,
```
ademidoff, line 920: I believe the recommendation is not to use defaults, as per line 920. Let's see if we get away without it.

api/managementpb/agent/agent.proto
idoqo: Yeah... it was already been used by `bool comments_parsing_disabled = 39;` - same with the other case too.

admin/commands/management/add.go
```diff
@@ -97,5 +108,14 @@ func processGlobalAddFlagsWithSocket(cmd connectionGetter, opts AddCommonFlags)
 		portI = int(opts.AddPortFlag)
 	}
 
-	return serviceName, socket, host, uint16(portI), nil
+	if opts.AddExposeExporterAddressFlag {
+		exposeExporterAddress = opts.AddExposeExporterAddressFlag
+	}
```
BupycHuk, line 113: I think it can be simplified by removing `exposeExporterAddress` and using `opts.AddExposeExporterAddressFlag` directly

BupycHuk: Can we use `opts.AddExposeExporterAddressFlag` in the place we call it without putting it into another struct?

admin/commands/management/add.go
```diff
@@ -30,11 +30,21 @@ type AddCommand struct {
...
+
 // AddCommonFlags is used by Kong for CLI flags and commands.
 type AddCommonFlags struct {
-	AddServiceNameFlag string `name:"service-name" placeholder:"NAME" help:"Service name (overrides positional argument)"`
-	AddHostFlag        string `name:"host" placeholder:"HOST" help:"Service hostname or IP address (overrides positional argument)"`
-	AddPortFlag        uint16 `name:"port" placeholder:"PORT" help:"Service port number (overrides positional argument)"`
+	AddServiceNameFlag           string `name:"service-name" placeholder:"NAME" help:"Service name (overrides positional argument)"`
+	AddHostFlag                  string `name:"host" placeholder:"HOST" help:"Service hostname or IP address (overrides positional argument)"`
+	AddPortFlag                  uint16 `name:"port" placeholder:"PORT" help:"Service port number (overrides positional argument)"`
+	AddExposeExporterAddressFlag bool   `name:"expose-exporter-address" placeholder:"EXPOSE-EXPORTER" help:"Expose the address of the exporter publicly on 0.0.0.0"`
```
BupycHuk, line 47: we need to add this flag to inventory commands too

managed/models/database.go
BupycHuk: we have to use default to set default value for `not null` columns, but we need to remove that once values are set.
So it should look like the migration above

BupycHuk (commented):
we need to implement this flag for inventory commands and flags as well.
Regarding variable name maybe it's better to keep it shorter like `--expose-exporter`?

ademidoff:
> we need to implement this flag for inventory commands and flags as well. Regarding variable name maybe it's better to keep it shorter like `--expose-exporter`?

An even better term would use the verb `bind`, which seems to be heavily used in such cases. I guess something like `--bind-all-interfaces` or `--bind-all` would sound reasonable and more semantic.

idoqo:
> > we need to implement this flag for inventory commands and flags as well. Regarding variable name maybe it's better to keep it shorter like `--expose-exporter`?
> 
> An even better term would use the verb `bind`.

I went with this initially too, but then felt like we may want to have `"exporter"` somewhere in the flag to reflect the fact that only exporters will be affected.

admin/commands/management/add.go
idoqo: > Can we use `opts.AddExposeExporterAddressFlag` in the place we call it without putting it into another struct?

Do you mean without the `globalAddServiceParams` struct? In that case the function will return five values - I feel that's a lot and hard to keep track of their meanings (considering we don't use named returns)

BupycHuk:
it can be `--bind-exporter` with two options `all` and `local`

api/inventorypb/agents.proto
```diff
@@ -185,6 +190,8 @@ message MongoDBExporter {
   string process_exec_path = 16;
   // Log level for exporter.
   LogLevel log_level = 17;
+  // Optionally expose the exporter process on 0.0.0.0
```
BupycHuk, line 193: I believe there should be better wording for the description. Maybe `expose exporter APIs to the world`

BupycHuk (approved):

JiriCtvrtka (approved):

PR #2542, opened 2023-10-12
PMM-12185 migrate telemetry GrafanaDS to using PostgreSQL

ademidoff:
PMM-12185

Link to the Feature Build: SUBMODULES-3415


managed/services/telemetry/config.default.yml
```diff
@@ -761,47 +761,47 @@ telemetry:
...
+    query: count(u.id) as count from public.user u left join preferences p on p.user_id = u.id where p.theme = '' or p.theme is null
     summary: "Grafana Default Theme Users Count"
     data:
       - metric_name: "grafana_default_theme_users_count"
         column: "count"
 
   - id: GrafanaCustomDashboardsCount
     source: GRAFANADB_SELECT
-    query: count(*) as count from dashboard where created_by != -1 and is_folder = 0
+    query: count(*) as count from dashboard where created_by != -1 and is_folder = true
```
BupycHuk, line 796: 0=false
```suggestion
    query: count(*) as count from dashboard where created_by != -1 and is_folder = false
```

managed/services/telemetry/config.default.yml
```diff
@@ -761,47 +761,47 @@ telemetry:
...
     summary: "Grafana Custom Dashboards Count"
     data:
       - metric_name: "grafana_custom_dashboards_count"
         column: "count"
 
   - id: GrafanaCustomDashboardsPerPillar
     source: GRAFANADB_SELECT
-    query: inn.title as pillar, count(d.title) as count from dashboard d left join dashboard inn ON d.folder_id =  inn.id where d.created_by != -1 and d.is_folder = 0 group by inn.title
+    query: inn.title as pillar, count(d.title) as count from dashboard d left join dashboard inn ON d.folder_id = inn.id where d.created_by != -1 and d.is_folder = true group by inn.title
```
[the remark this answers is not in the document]
ademidoff: OMG :( Thank you!

BupycHuk (approved):

JiriCtvrtka (approved):

managed/services/config/pmm-managed.yaml
```diff
@@ -19,7 +19,10 @@ services:
       GRAFANADB_SELECT:
         enabled: true
         timeout: 5s
-        db_file: /srv/grafana/grafana.db
+        use_separate_credentials: true
+        separate_credentials:
+          username: grafana
+          password: grafana
```
artemgavrilov, line 25: This file has TODO comment on the top. Can we eliminate its extension and use just plain constants?

managed/services/telemetry/config.default.yml
```diff
@@ -761,47 +761,47 @@ telemetry:
   #Grafana
   - id: GrafanaUsersCount
     source: GRAFANADB_SELECT
-    query: count(*) AS count FROM user
+    query: count(*) AS count FROM public.user
```
artemgavrilov, line 764: Why did you specify schema? Grafana uses multiple schemas in their database?

managed/services/telemetry/telemetry.go
```diff
@@ -135,15 +135,15 @@ func (s *Service) Run(ctx context.Context) {
...
-		s.l.Debugf("\nTelemetry captured:\n%s\n", s.Format(report))
-
 		if s.config.Reporting.Send {
 			s.sendCh <- report
 		} else {
 			s.l.Info("Telemetry sent is disabled.")
 		}
 	}
 
+	s.l.Infof("\nTelemetry captured:\n%s\n", s.Format(s.prepareReport(ctx)))
```
artemgavrilov, line 145: Do we want to print this to use logs with info level?

managed/services/telemetry/config.default.yml
ademidoff: `user` is an internal function in PostgreSQL (and a reserved word) - it returns the current db user. 

While this query will run OK - `SELECT * FROM user;`, and even this will run too - `SELECT user;`, this one - `SELECT id FROM user` - will fail with `ERROR: column "id" does not exist` even though the column `id` does exist.

Therefore, to prevent those `user` table queries from failing, we have to prefix them with their schema.

managed/services/telemetry/telemetry.go
ademidoff: Man, I mistakenly remove the wrong log command. Thanks, it's rectified.

managed/services/config/pmm-managed.yaml
ademidoff: As per TODO, all of them need to be migrated. For the sake of scope and testing, isn't it better to do that in one PR?

managed/services/telemetry/config.default.yml
artemgavrilov: Aha, I got what is the issue here. `user` and `public.user` are two different things, first one in reserved word for current PG user and second one is ordinary grafana table with grafana users.

So all these queries were invalid until your changes. Good 👍

managed/services/config/pmm-managed.yaml
artemgavrilov: Yes, it's ok to do that later. Just wanted to get your attention to plans regarding this config.

artemgavrilov (approved):

PR #2582, opened 2023-10-30
PMM-12630 UI upgrade from 2.37.1 and lower does not work

adivinho:
PMM-12630

### DIFF
--- a/update/ansible/playbook/tasks/roles/clickhouse/tasks/main.yml
+++ b/update/ansible/playbook/tasks/roles/clickhouse/tasks/main.yml
@@ -33,6 +33,8 @@
       - percona-clickhouse-server
       - percona-clickhouse-common-static
       - percona-clickhouse-server-common
+      - clickhouse-client-21.3.20.1-2
+      - clickhouse-server-21.3.20.1-2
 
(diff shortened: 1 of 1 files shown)


adivinho:
It's a bug in el7. It can't update noarch to x86_64 package.
Screenshot 2023-10-30 at 10 23 58

BupycHuk (approved):

update/ansible/playbook/tasks/roles/clickhouse/tasks/main.yml
```diff
@@ -33,6 +33,8 @@
       - percona-clickhouse-server
       - percona-clickhouse-common-static
       - percona-clickhouse-server-common
+      - clickhouse-client-21.3.20.1-2
```
ademidoff, line 36: What about uninstalling versions prior to 21.3.20? Or was this the only one used in PMM?

BupycHuk: I think a few lines above does that trick.

ademidoff (approved):

BupycHuk:
merged the PR since it can be tested only on dev-latest

PR #2590, opened 2023-11-03
PMM-12529 Run pmm-agent as non root

ademidoff:
PMM-12529

Link to the Feature Build: SUBMODULES-3472


managed/models/postgresql_version_test.go
```diff
@@ -78,7 +78,7 @@ func TestGetPostgreSQLVersion(t *testing.T) {
 	for _, tc := range testCases {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
-			t.Parallel()
+			// t.Parallel()
```
artemgavrilov, line 81: Problematic test?

artemgavrilov (approved):

managed/models/postgresql_version_test.go
ademidoff: Yes, permanently racing, so this was the only way to make them pass :(

BupycHuk (commented):
Not sure if we need to make pmm2-client to be installed under pmm user for all installation. isn't it possible to do only for PMM Server?

ademidoff:
> Not sure if we need to make pmm2-client to be installed under pmm user for all installation. isn't it possible to do only for PMM Server?

Maybe we can. Let me try :)

BupycHuk (approved):
please fix merge conflicts

PR #2604, opened 2023-11-07
PMM-12660 do not output certs when listing agents

ademidoff:
PMM-12660

Link to the Feature Build: SUBMODULES-3438


BupycHuk (approved):
was it necessary to update descriptors?

ademidoff:
> was it necessary to update descriptors?

I believe yes, otherwise it bails out when you run `make gen` with an error message. 

BTW, it's easy to reproduce the failure by restoring `descriptor.bin` from the head and then running `bin/buf breaking --against descriptor.bin api` - it should output an error.

JiriCtvrtka (approved):

PR #2616, opened 2023-11-10
PMM-12643 consistent types in list output of pmm-admin

ademidoff:
PMM-12643

Link to the Feature Build: SUBMODULES-3445


admin/commands/list.go
```diff
@@ -222,7 +222,7 @@ func (cmd *ListCommand) RunCmd() (Result, error) { //nolint:cyclop,maintidx
 		return "pull"
 	}
 	pmmAgentIDs := make(map[string]struct{})
-	var agentsList []listResultAgent
+	agentsList := make([]listResultAgent, 0, len(agentsRes.Payload.PMMAgent))
```
BupycHuk, line 225: there are much more items then just PMM Agents

ademidoff: Fixed.

ademidoff:
> Can we add a test here?

Did you mean if we can add a test for RunCmd?

BupycHuk (approved):
wow, with these improvements code looks much more clear

JiriCtvrtka (approved):

PR #2621, opened 2023-11-14
PMM-12429 Use a configuration file for mockery

ademidoff:
PMM-12429

Note: this is intentionally a v3 ticket so we don't have to solve conflicts.


BupycHuk (approved):

.mockery.yaml
```diff
@@ -0,0 +1,110 @@
+# This is the top-level configuration file for mockery.
+disable-version-string: True
```
artemgavrilov, line 2: 🔥👍

artemgavrilov (approved):

PR #2625, opened 2023-11-17
PMM-12702 Fix SSLSni checker.

BupycHuk:
PMM-12702

Link to the Feature Build: SUBMODULES-3449

- [ ] API Docs updated

### DIFF
--- a/managed/models/agentversion.go
+++ b/managed/models/agentversion.go
@@ -33,14 +33,9 @@ type AgentNotSupportedError struct {
 	MinAgentVersion string
 }
 
-func (e *AgentNotSupportedError) Error() string {
+func (e AgentNotSupportedError) Error() string {
 	return fmt.Sprintf("'%s' functionality is not supported by pmm-agent %q version %q. Required minimum version is %q", e.Functionality,
 		e.AgentID, e.AgentVersion, e.MinAgentVersion)
-}
-
-func (e *AgentNotSupportedError) Is(err error) bool {
-	_, ok := err.(*AgentNotSupportedError)
-	return ok
 }
 
 // PMMAgentSupported checks if pmm agent version satisfies required min version.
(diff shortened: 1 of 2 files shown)


managed/models/agentversion.go
```diff
@@ -37,18 +36,22 @@ func (e *AgentNotSupportedError) Error() string {
 	return fmt.Sprintf("'%s' functionality is not supported by pmm-agent %q version %q. Required minimum version is %q", e.Functionality,
 		e.AgentID, e.AgentVersion, e.MinAgentVersion)
 }
+func (e *AgentNotSupportedError) Is(err error) bool {
+	_, ok := err.(*AgentNotSupportedError)
+	return ok
+}
```
artemgavrilov, line 42: Why it's needed? It's analog of `errors.Is()`, no?

managed/models/agentversion.go
```diff
@@ -17,7 +17,6 @@ package models
 
 import (
 	"fmt"
-
```
ademidoff, line 20: It looks like it needs to get `make format`'ed ❓

managed/models/agentversion.go
BupycHuk: errors.Is didn't work correctly and actually calls this method to compare, so I had to implement it this way.

BupycHuk: https://github.com/percona/pmm/pull/2625/files#diff-8792e067d067e501754c905b86a46d9101c2861f211cc265cdc5fd395a2e4bc4R81 here is where errors.Is is called

artemgavrilov: It seems that we need `As` there instead of `Is`. We need to check error type, but not exact copy.

BupycHuk: As didn't work as well

artemgavrilov: Why? What's so special about this case?

BupycHuk: I have no idea

BupycHuk: the reason was because `IsAgentSupported` was returning a pointer and that's why it didn't work.

ademidoff (approved):

artemgavrilov (approved):

PR #2636, opened 2023-11-22
PMM-12473 do not mangle the query string  for external exporters

ademidoff:
PMM-12473

Link to the Feature Build: SUBMODULES-3455


.devcontainer/setup.py
```diff
@@ -89,7 +89,7 @@ def setup():
         # Having fsync off in dev environment is fine.
         "sed -i -e \"s/#fsync = on/fsync = off/\" /srv/postgres14/postgresql.conf",
         "echo 'host    all         all     0.0.0.0/0     trust' >> /srv/postgres14/pg_hba.conf",
-        "supervisorctl restart postgresql",
+        # "supervisorctl restart postgresql",
```
ademidoff, line 92: This still fails, but we don't necessarily have to restart postgresql during build time - it will start on container launch.

BupycHuk (approved):

JiriCtvrtka (approved):

artemgavrilov (approved):

PR #2642, opened 2023-11-24
PMM-12422 use public IP for exporters in pull mode

idoqo:
PMM-12422

Due to issues with getting metrics when using the node IP, this exposes exporters running in pull mode on the public IP address.

### DIFF
--- a/managed/services/agents/agents.go
+++ b/managed/services/agents/agents.go
@@ -147,9 +147,11 @@ func getExporterListenAddress(node *models.Node, exporter *models.Agent) string 
 	switch {
 	case exporter.ExposeExporter:
 		return "0.0.0.0"
-	case !exporter.PushMetrics && node != nil:
+	case exporter.PushMetrics:
+		return "127.0.0.1"
+	case node != nil:
 		return node.Address
 	default:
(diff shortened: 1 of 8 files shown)


BupycHuk (approved):

managed/services/agents/agents.go
```diff
@@ -145,11 +145,11 @@ func ensureAuthParams(exporter *models.Agent, params *agentpb.SetStateRequest_Ag
...
 func getExporterListenAddress(node *models.Node, exporter *models.Agent) string {
 	switch {
-	case !exporter.PushMetrics && node != nil:
-		return node.Address
 	case exporter.ExposeExporter:
 		return "0.0.0.0"
-	default:
+	case exporter.PushMetrics:
 		return "127.0.0.1"
+	default:
```
ademidoff, line 152: It seems like this makes the first case redundant.

ademidoff (approved):

managed/services/agents/agents.go
idoqo: Yeah true - I wanted to keep the case where exposeExporter is enabled at the top so it overrides anything else.

PR #2667, opened 2023-12-04
PMM-7 cleanup docker-compose.yml for v3

ademidoff:
PMM-7

This PR removes a few lines related to DBAAS, which are now redundant and which may have sneaked in during the last conflict resolution. It also contains a couple of other fixes.


BupycHuk:
that was done for HA purpose and pmm-managed is supposed to start it. I would investigate that.

BupycHuk (approved):

PR #2701, opened 2023-12-15
PMM-12712 Shards collector.

JiriCtvrtka:
PMM-12712

https://github.com/Percona-Lab/pmm-submodules/pull/3485

https://github.com/percona/mongodb_exporter/pull/762


BupycHuk (approved):

idoqo (approved):

managed/services/agents/mongodb.go
```diff
@@ -252,6 +257,11 @@ func defaultCollectors(collectAll bool) map[string]collectorArgs {
 			enabled:     collectAll,
 			enableParam: "--collector.topmetrics",
 		},
+		// disabled until we have better information on the resources usage impact
```
artemgavrilov, line 260: Is this comment correct? Why we are saying that it's disables when it's configured by `collectAll` parameter in this func?

artemgavrilov (approved):

managed/services/agents/mongodb.go
JiriCtvrtka: collectAll is by default disabled, but yeah since it is parameter in func it is not fully true. This occurs on more places. Removed from all places where you can pass this as a param.

managed/services/agents/mongodb.go
```diff
@@ -187,6 +187,11 @@ func v225Args(exporter *models.Agent, tdp *models.DelimiterPair, listenAddress s
 			enabled:     false,
 			enableParam: "--collector.topmetrics",
 		},
+		// disabled until we have better information on the resources usage impact
+		"shards": {
+			enabled:     false,
+			enableParam: "--collector.shards",
+		},
```
BupycHuk, line 194: we don't have it in old versions, so let's drop it

managed/services/agents/mongodb.go
```diff
@@ -232,26 +237,26 @@ func defaultCollectors(collectAll bool) map[string]collectorArgs {
...
 		},
-		// disabled until we have better information on the resources usage impact
 		"topmetrics": {
 			enabled:     collectAll,
 			enableParam: "--collector.topmetrics",
 		},
+		"shards": {
+			enabled:     collectAll,
+			enableParam: "--collector.shards",
+		},
```
BupycHuk, line 259: we need to add check for version 2.41.1 to support this flag

managed/services/agents/mongodb.go
JiriCtvrtka: Removed.

managed/services/agents/mongodb.go
JiriCtvrtka: Added check for 2.41.1

managed/services/agents/mongodb_test.go
```diff
@@ -231,6 +231,161 @@ func TestMongodbExporterConfig226(t *testing.T) {
 	})
 }
 
+func TestMongodbExporterConfig2411(t *testing.T) {
+	pmmAgentVersion := version.MustParse("2.41.1")
```
BupycHuk, line 235: our dev-latest and Feature builds usually have version which looks like `2.41.1-HEAD-fasdjf`, please make sure that it works for that too, otherwise the feature will not pass QA.

JiriCtvrtka: Fixed and added short test which checking this.

BupycHuk (approved):

PR #2706, opened 2023-12-18
PMM-7 fix devcontainer build

ademidoff:
PMM-7

Now that `3-dev-latest` is rootless, we need root to build the devcontainer.

We are also temporarily disabling the PMM Server upgrade test.


artemgavrilov (approved):

BupycHuk (approved):
should we change user back to pmm once we installed all tools?

ademidoff:
> should we change user back to pmm once we installed all tools?

I'm not sure to be honest. `pmm` user is too restrictive and we sometimes need to do things that would lack permissions, like install a binary. I believe we need a bit more time to understand what the user should be in the devcontainer.

PR #2707, opened 2023-12-18
Fix migrations order

artemgavrilov:
PMM-0

- [ ] API Docs updated


ademidoff (approved):

managed/models/database.go
```diff
@@ -949,17 +949,17 @@ var databaseSchema = [][]string{
...
 		`DROP TABLE IF EXISTS ia_channels`,
 		`DROP TABLE IF EXISTS ia_rules`,
 		`ALTER TABLE ia_templates RENAME TO alert_rule_templates`,
 		`UPDATE settings SET settings = settings #- '{alerting, email_settings}';`,
 		`UPDATE settings SET settings = settings #- '{alerting, slack_settings}';`,
 	},
-	88: {
+	90: {
 		`UPDATE settings SET settings = settings - 'alert_manager_url'`,
 	},
```
BupycHuk, line 964: it was done on purpose to have some room for newer items that might appear in PMM 2.x to not have conflicts

artemgavrilov: Due merging PR's in `v3` branch we got broken order `87, 100, 101, 88`. But we will have conficts with `v2` anyway, because it's same lines, it doesn't matter what number specified there, no?

BupycHuk: I meant conflicts in runtime. when users in DB will have higher number of schema_migration then we have in the first migration of PMM 3 and it will be skipped. And also each time we create new migration in PMM 2 we will have to update all these values.

managed/models/database.go
```diff
@@ -949,17 +949,17 @@ var databaseSchema = [][]string{
 			PRIMARY KEY (dump_id, chunk_id)
 		)`,
 	},
-	100: {
+	88: {
```
artemgavrilov, line 952:
```suggestion
	100: {
```

managed/models/database.go
```diff
@@ -949,17 +949,17 @@ var databaseSchema = [][]string{
...
 		`DROP TABLE kubernetes_clusters`,
 	},
-	101: {
+	89: {
```
artemgavrilov, line 955:
```suggestion
	101: {
```

managed/models/database.go
```diff
@@ -949,17 +949,17 @@ var databaseSchema = [][]string{
...
 		`DROP TABLE IF EXISTS ia_channels`,
 		`DROP TABLE IF EXISTS ia_rules`,
 		`ALTER TABLE ia_templates RENAME TO alert_rule_templates`,
 		`UPDATE settings SET settings = settings #- '{alerting, email_settings}';`,
 		`UPDATE settings SET settings = settings #- '{alerting, slack_settings}';`,
 	},
-	88: {
+	90: {
```
artemgavrilov, line 962:
```suggestion
	102: {
```

managed/models/database.go
artemgavrilov: Ok, changed back to 1xx

BupycHuk (approved):

PR #2717, opened 2023-12-20
PMM-7 Lint fix.

JiriCtvrtka:


managed/utils/interceptors/interceptors.go
```diff
@@ -75,8 +76,7 @@ func logRequest(l *logrus.Entry, prefix string, f func() error) (err error) {
 		}
 	}()
 
-	err = f()
-	return //nolint:nakedret
+	return f()
```
BupycHuk, line 79: we still need `err = f()` here for logic in `defer`

BupycHuk: please check other places where you removed `err = .*` to make sure that `err` isn't used in `defer` function

JiriCtvrtka: Yeah right. Anyway PR is not ready yet.

JiriCtvrtka:
Fix for nonamedreturns and nakedret lints. Right now I am skipping it in interceptors files since it is creating bug with EOF instead on "internal server error".

BupycHuk (approved):

ademidoff (approved):

PR #2777, opened 2024-01-18
PMM-12811 single agents change endpoint

ademidoff:
PMM-12811

Link to the Feature Build: SUBMODULES-3508

This is what swagger looks like after refactoring:

Screenshot 2024-01-19 at 12 56 30

- [x] API Docs updated


.github/workflows/managed.yml
```diff
@@ -61,8 +61,14 @@ jobs:
           pushd tools && go mod download -x
           popd        && go mod download -x
 
-      - name: Initialize CI environment
-        run: make env-compose-up # the container workdir is /root/go/src/github.com/percona/pmm
+      - name: Launch PMM Server (see docker-compose.yml)
+        run: |
+          # Note: launching the container with --wait fails for an unknown reason.
```
idoqo, line 66: Yeah, I think is because docker-compose exits with a non-zero code when using wait, so the CI environment treats it as an error/failure.

idoqo (approved):

.github/workflows/managed.yml
ademidoff: Well, it used to work so far, I view it as a CI oddity, which should be temporary.

api-tests/inventory/agents_test.go
```diff
@@ -554,21 +554,23 @@ func TestQanAgentExporter(t *testing.T) {
...
-					AgentID: agentID,
-					Common: &agents.ChangeQANMySQLPerfSchemaAgentParamsBodyCommon{
-						Disable:            true,
-						RemoveCustomLabels: true,
+		changeQANMySQLPerfSchemaAgentOK, err := client.Default.AgentsService.ChangeAgent(
+			&agents.ChangeAgentParams{
+				Body: agents.ChangeAgentBody{
+					QANMysqlPerfschemaAgent: &agents.ChangeAgentParamsBodyQANMysqlPerfschemaAgent{
+						AgentID: agentID,
+						Common: &agents.ChangeAgentParamsBodyQANMysqlPerfschemaAgentCommon{
```
BupycHuk, line 562: can we move common fields one level up to reuse the same code?

api-tests/inventory/agents_test.go
```diff
@@ -554,21 +554,23 @@ func TestQanAgentExporter(t *testing.T) {
...
+							Disable:            true,
+							RemoveCustomLabels: true,
```
BupycHuk, line 564: can we use the same approach @artemgavrilov introduced to not have 2 fields for each enable/disable and set/remove operations

JiriCtvrtka:
@ademidoff LGTM, could you check managed failing tests?

ademidoff:
> @ademidoff LGTM, could you check managed failing tests?

I did. This is what fails - `TestClient/getRole/NoAnonymousAccess`, which seems unrelated to this PR.

JiriCtvrtka (approved):

BupycHuk (commented):
I'm curious if we can put all duplicated fields to highest level to not duplicate code, we at least have agent_id, enable/disable agent, enable/disable push metrics.

api-tests/inventory/agents_test.go
ademidoff: Good point - resolved with [this commit](https://github.com/percona/pmm/pull/2777/commits/1555a22ddc04e375463ead9663f40178b804ef7a)

api-tests/inventory/agents_test.go
ademidoff: I created a PR to cover that, but it seems like the benefit is not obvious. https://perconadev.atlassian.net/browse/PMM-12875
Let's discuss offline.

ademidoff: Decided to close that one as it may be difficult to refactor such a payload structure.

api/inventory/v1/agents.proto
```diff
@@ -576,18 +576,14 @@ message AzureDatabaseExporter {
...
   // Replace all custom user-assigned labels.
-  map<string, string> custom_labels = 3;
+  map<string, string> custom_labels = 2;
   // Remove all custom user-assigned labels.
-  bool remove_custom_labels = 4;
-  // Enables push metrics with vmagent, can't be used with disable_push_metrics.
-  bool enable_push_metrics = 5;
-  // Disables push metrics, pmm-server starts to pull it, can't be used with enable_push_metrics.
-  bool disable_push_metrics = 6;
+  bool remove_custom_labels = 3;
```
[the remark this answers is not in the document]
ademidoff: Done.

api-tests/management/action/explain_test.go
```diff
@@ -37,7 +37,7 @@ func TestRunExplain(t *testing.T) {
 			Body: actions.StartMySQLExplainActionBody{
 				// PMMAgentID: "/agent_id/f235005b-9cca-4b73-bbbd-1251067c3138",
 				ServiceID: "/service_id/5a9a7aa6-7af4-47be-817c-6d88e955bff2",
-				Query:     "SELECT `t` . * FROM `test` . `key_value` `t`",
+				QueryID:   "3923dace316a86643fcf87cb45e0558a",
```
BupycHuk, line 40: this test can be dropped, because we already have it below

api-tests/server/settings_test.go
```diff
@@ -39,7 +39,7 @@ func TestSettings(t *testing.T) {
 		res, err := serverClient.Default.ServerService.GetSettings(nil)
 		require.NoError(t, err)
 		assert.True(t, res.Payload.Settings.TelemetryEnabled)
-		assert.True(t, res.Payload.Settings.SttEnabled)
```
BupycHuk, line 42: we need sanity check in Github CI to check that API tests can be built.

api/inventory/v1/agents.proto
```diff
@@ -703,7 +686,7 @@ message AddAgentRequest {
 }
 
 message AddAgentResponse {
-  oneof exporter {
+  oneof agent {
```
BupycHuk, line 689: 👍🏼

api/inventory/v1/agents.proto
```diff
@@ -723,6 +706,57 @@ message AddAgentResponse {
   int32 table_count = 20;
```
BupycHuk, line 706: I'm wondering why do we need this field, can we include it to one of exporter messages? I believe that it's related to MySQLdExporter

api/inventory/v1/agents.proto
```diff
@@ -723,6 +706,57 @@ message AddAgentResponse {
...
+    RDSExporter rds_exporter = 8;
+    AzureDatabaseExporter azure_database_exporter = 9;
+    QANMySQLPerfSchemaAgent qan_mysql_perfschema_agent = 10;
+    QANMySQLSlowlogAgent qan_mysql_slowlog_agent = 11;
+    QANMongoDBProfilerAgent qan_mongodb_profiler_agent = 12;
+    QANPostgreSQLPgStatementsAgent qan_postgresql_pgstatements_agent = 13;
+    QANPostgreSQLPgStatMonitorAgent qan_postgresql_pgstatmonitor_agent = 14;
+  }
+  // Actual table count for SQL databases at the moment of adding.
+  int32 table_count = 20;
```
BupycHuk, line 746: probably we don't need it here

api/inventory/v1/agents.proto
```diff
@@ -666,25 +660,14 @@ message GetAgentLogsResponse {
...
-  string runs_on_node_id = 1 [(validate.rules).string.min_len = 1];
-  // Custom user-assigned labels.
-  map<string, string> custom_labels = 2;
-}
-
 // TODO Change PMMAgent?
 
 // TODO Add VMAgent?
 
 // TODO Change VMAgent?
```
BupycHuk, line 667: let's drop it

api/inventory/v1/agents.proto
```diff
@@ -723,6 +706,57 @@ message AddAgentResponse {
   int32 table_count = 20;
 }
 
+message ChangeAgentRequest {
+  oneof agent {
+    // ChangePMMAgentParams pmm_agent = 1;
```
BupycHuk, line 711: let's uncomment it and keep it empty for now

managed/services/inventory/agents.go
```diff
@@ -680,42 +701,48 @@ func (as *AgentsService) AddProxySQLExporter(ctx context.Context, p *inventoryv1
...
 		return nil, err
 	}
 
-	res := agent.(*inventoryv1.ProxySQLExporter) //nolint:forcetypeassert
-	as.state.RequestStateUpdate(ctx, res.PmmAgentId)
+	agent := ag.(*inventoryv1.ProxySQLExporter) //nolint:forcetypeassert
+	as.state.RequestStateUpdate(ctx, agent.PmmAgentId)
+
+	res := &inventoryv1.ChangeAgentResponse{
+		Agent: &inventoryv1.ChangeAgentResponse_ProxysqlExporter{
```
BupycHuk, line 714: I believe that we can return only Agent value here, so we don't have to wrap the message in each method

BupycHuk (commented):
We need to add information that only one of agent params should be passed into endpoint description in swagger.

api-tests/server/settings_test.go
ademidoff: This should be done within #2805 and it's out of scope of this ticket anyway. I had to make change because the code couldn't compile, but otherwise I didn't have to.

api/inventory/v1/agents.proto
ademidoff: Why drop? Do we not want to implement it?

managed/services/inventory/agents.go
ademidoff: These are `proto` types generated by grpc, I don't quite see how we can return arbitrary values here. Let's discuss.

api-tests/management/action/explain_test.go
ademidoff: Removed. FYI: all these tests are currently skipped.

managed/services/inventory/agents.go
BupycHuk: we can wrap it in the method where we call this method, as far as I see all these methods with duplicated wrap logic is called by ChangeAgent method

api/inventory/v1/agents.proto
BupycHuk: Probably no

api-tests/server/settings_test.go
BupycHuk: Agree, just wrote it here to not forget

api/inventory/v1/agents.proto
ademidoff: Done.

api/inventory/v1/agents.proto
ademidoff: Decided to leave for now.

managed/services/inventory/agents.go
ademidoff: I couldn't implement it quite as you requested, but I succeeded in further aligning `AddAgent` with `ChangeAgent` so they now look similar. A lot of redundant code was removed as well.

api/inventory/v1/agents.proto
ademidoff: Done.

api/inventory/v1/agents.proto
ademidoff: Moved to MySQLdExporter.

ademidoff:
> We need to add information that only one of agent params should be passed into endpoint description in swagger.

Done.

BupycHuk (approved):

PR #2803, opened 2024-01-31
PMM-12852: prefix PMM public URLs with scheme before parsing

idoqo:
PMM-12852

Link to the Feature Build: SUBMODULES-3526

- [ ] API Docs updated


managed/services/server/server_test.go
```diff
@@ -228,6 +228,12 @@ func TestServer(t *testing.T) {
 		assert.NoError(t, s.validateChangeSettingsRequest(ctx, &serverpb.ChangeSettingsRequest{
 			DisableStt: true,
 		}))
+
+		// fail if new public address is not a valid URL.
+		expected = status.New(codes.FailedPrecondition, "Provided PMM public address is invalid.")
+		tests.AssertGRPCError(t, expected, s.validateChangeSettingsRequest(ctx, &serverpb.ChangeSettingsRequest{
+			PmmPublicAddress: "192.168.55.1:5555",
```
BupycHuk, line 235: this format should be valid, because our UI uses this format

managed/services/supervisord/supervisord.go
```diff
@@ -443,15 +443,13 @@ func (s *Service) marshalConfig(tmpl *template.Template, settings *models.Settin
 
 	templateParams["PMMServerHost"] = ""
 	if settings.PMMPublicAddress != "" {
+		if !strings.HasPrefix(settings.PMMPublicAddress, "https://") && !strings.HasPrefix(settings.PMMPublicAddress, "http://") {
+			settings.PMMPublicAddress = fmt.Sprintf("https://%s", settings.PMMPublicAddress)
```
BupycHuk, line 447: `marshalConfig` shouldn't update value incoming arguments field, instead let's use separate variable.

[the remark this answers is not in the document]
idoqo: Yeah in this case, it doesn't really matter since we only add it for URL parsing to pass. One possible issue is if `settings.PMMPublicAddress` ends up being used somewhere else, but addressing this [comment](https://github.com/percona/pmm/pull/2803#discussion_r1473358397) should fix that.

BupycHuk (approved):

JiriCtvrtka (approved):

ademidoff (approved):

idoqo:
> @idoqo I think there is one place that also needs this fix. When [parsing PMM_PUBLIC_ADDRESS env variable](https://github.com/percona/pmm/blob/64ff314b25f2f0776d2f1fcacff7c3432849d880/managed/utils/envvars/parser.go#L159)
@yurkovychv the current fix applies to both ways of setting `PMM_PUBLIC_ADDRESS` actually (so it doesn't matter if it's from settings or via environment variables).

PR #2828, opened 2024-02-15
PMM-12913 Use regular http verbs

ademidoff:
PMM-12913

Link to the Feature Build: SUBMODULES-3556

This is a partial PR primarily aiming at joining a number of closely related services, and API endpoints thereof, to one.

It also provides a document where you can see the future migration of endpoints to make them more RESTful - [MIGRATION_TO_V3.md](https://github.com/percona/pmm/pull/2828/files#diff-6a4e2f775c4e73a53069a233e17199ab8bd1fd45bb7a1ceb8bd833d65426f4fd). This migration will go to the next PR.

You can visualize the current swagger APIs spec on https://petstore.swagger.io/?url=https://raw.githubusercontent.com/percona/pmm/facd6c6f86254d9d4fc6f7a14c70c67a90b55681/api/swagger/swagger-dev.json.


managed/services/management/node.go
```diff
@@ -34,27 +34,8 @@ import (
...
-	return &NodeService{
-		db:  db,
-		akp: akp,
-		l:   logrus.WithField("component", "node"),
-	}
-}
-
 // RegisterNode performs the registration of a new node.
-func (s *NodeService) RegisterNode(ctx context.Context, req *managementv1.RegisterNodeRequest) (*managementv1.RegisterNodeResponse, error) {
+func (s *ServiceService) RegisterNode(ctx context.Context, req *managementv1.RegisterNodeRequest) (*managementv1.RegisterNodeResponse, error) {
```
BupycHuk, line 38: it's not really `Service` related method :)

ademidoff: Right, it doesn't look semantic now :( 

I believe renaming it to ManagementService will be much better, WDYT?

ademidoff: Problem is that we have way too many services, which is why we end up running out of good names. Example: MgmtService, which is also `ManagementService v1beta1`, so the only difference between them is the version, i.e. `v1` versus `v1beta1`.

IMO this transition will necessitate making a few compromises or workarounds anyway :)

api/actions/v1/actions.proto
```diff
@@ -434,7 +433,7 @@ service ActionsService {
     };
   }
   // CancelAction stops an Action.
-  rpc CancelAction(CancelActionRequest) returns (CancelActionResponse) {
+  rpc CancelAction(CancelActionRequest) returns (google.protobuf.Empty) {
```
artemgavrilov, line 436: Bad thing about `Empty` type is that it cuts off extensibility. In the future there will be no option to return, for example, action progress that was done, or something else.

api/backup/v1/artifacts.proto
```diff
@@ -85,20 +82,19 @@ message ListPitrTimerangesResponse {
 // Artifacts service provides public methods for managing backup artifacts.
 service ArtifactsService {
   // ListArtifacts returns a list of all backup artifacts.
-  rpc ListArtifacts(ListArtifactsRequest) returns (ListArtifactsResponse) {
+  rpc ListArtifacts(google.protobuf.Empty) returns (ListArtifactsResponse) {
```
artemgavrilov, line 85: Same story, there will be no option to add filters parameters to this method. We will be forced to create new endpoint

api/actions/v1/actions.proto
BupycHuk: Agree with Artem, let's drop usage of `Empty` type

managed/services/management/node.go
BupycHuk: I think we need to differentiate them by using namings which makes it clear that it's beta like `ManagementService` and `ManagementServiceV1Beta`

api/backup/v1/artifacts.proto
ademidoff: @artemgavrilov @maxkondr I wonder how do folks who promote the use of `google.protobuf.Empty` get away in such scenarios?

Doesn't that qualify for the next version of API, say `v2`?

managed/services/management/node.go
ademidoff: OK, good suggestion! Will do.

api/backup/v1/artifacts.proto
artemgavrilov: Filters should be optional, so adding them doesn't break backward compatibility. So it has not much sense to create v2 for that.

artemgavrilov: For me it will be simpler to use something one, otherwise it became a mess over time.

maxkondr: @ademidoff 
could you please check what will happen in case:
1.
```
rpc ListArtifacts(google.protobuf.Empty) returns (ListArtifactsResponse) {
  option (google.api.http) = {
       post: "/v1/backup/Artifacts/List"
  };
}
```
and call endpoint as:
```
curl -XPOST http://<domain>/v1/backup/Artifacts/List?foo=bar
```

2. that change API definition to:
```
rpc ListArtifacts(ListArtifactsRequest) returns (ListArtifactsResponse) {
  option (google.api.http) = {
       post: "/v1/backup/Artifacts/List"
  };
}
```
and call endpoint as:
```
curl -XPOST http://<domain>/v1/backup/Artifacts/List?foo=bar
```

Will it break smh?

api/actions/v1/actions.proto
artemgavrilov: Yes: https://github.com/percona/pmm/pull/731/files

api/backup/v1/artifacts.proto
```diff
@@ -85,20 +82,19 @@ message ListPitrTimerangesResponse {
...
     option (google.api.http) = {
       post: "/v1/backup/Artifacts/List"
```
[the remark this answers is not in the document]
ademidoff: This PR is only half done, I wrote about it in the description. Otherwise, it's supposed to be a `POST`.

ademidoff:
> another POV - have we really had such need in the past of such extensibility?

@artemgavrilov Has mentioned that we'd rather be consistent in using empty messages for the sake of exensibility, otherwise we won't be able to establish a guide clearly stating when to use what.

api/actions/v1/actions.proto
ademidoff: > another POV - have we really had such need in the past of such extensibility?

Yes, we have changed the request and response types in the past. This was OK, since extending protobufs doesn't break anything.

@artemgavrilov has mentioned that we'd rather be consistent in using empty messages everywhere, otherwise we won't be able to establish a guide clearly stating when to use what.

api/backup/v1/artifacts.proto
ademidoff: If your question is `How to pass URL parameters to a gRPC POST request`, then we have follow a special syntax for both examples. This `limitation` comes from the `http.proto` [doc](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L176-L180).

```
message ListArtifactsRequest {
  string foo = 5;
}

rpc ListArtifacts(ListArtifactsRequest) returns (ListArtifactsResponse) {
  option (google.api.http) = {
       post: "/v1/backup/Artifacts/List"
       body: "listartifactrequest" # Note the absence of "*"
  };
}

```

maxkondr: my Q was: if we not utilize `google.protobuf.Empty` as request body, will it break smth in case later we need to change it to custom message in order to support list parameters?

ademidoff: My understanding is it won't break anything. We can replace `.Empty` with with a different proto message when we need to.

artemgavrilov: It will break gRPC API, which is public too

managed/services/management/node.go
ademidoff: Done!

Screenshot 2024-02-27 at 11 48 49

api/backup/v1/artifacts.proto
ademidoff: Changes reverted, `gogle.protob.Empty` is not used anymore.

api/management/v1/rds.proto
```diff
@@ -122,35 +121,9 @@ message AddRDSResponse {
...
-    option (google.api.http) = {
-      post: "/v1/management/RDS/Add"
-      body: "*"
-    };
-    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
-      summary: "Add RDS"
-      description: "Adds RDS instance."
-    };
-  }
+  int32 table_count = 9;
```
artemgavrilov, line 128: Maybe it makes sense to have this filed in the begining of the message with num `1` or have a gap and use number `20` for example. Just for case when there will be new RDS services?

artemgavrilov (approved):
LGTM

BupycHuk (commented):
not finished review yet

BupycHuk (approved):

api/MIGRATION_TO_V3.md
```diff
@@ -0,0 +1,167 @@
...
+POST /v1/backup/Artifacts/Delete                    DELETE /api/backups/v1/artifacts/{id}             ?remove_files=true
+POST /v1/backup/Artifacts/PITRTimeranges            GET /api/backups/v1/artifacts/{id}/pitr_timeranges
+
+**BackupsService**                                  **BackupService**                                 TODO: rename to singular
+POST /v1/backup/Backups/ChangeScheduled             PUT /api/backups/v1/backups:changeScheduled
+POST /v1/backup/Backups/GetLogs                     GET /api/backups/v1/backups/{id}/logs
+POST /v1/backup/Backups/ListArtifactCompatibleServices GET /api/backups/v1/backups/{id}/services      Could also be /compatible_services
+POST /v1/backup/Backups/ListScheduled               GET /api/backups/v1/backups/scheduled
+POST /v1/backup/Backups/RemoveScheduled             GET /api/backups/v1/backups/scheduled/{id}
+POST /v1/backup/Backups/Restore                     POST /api/backups/v1/backups:restore
```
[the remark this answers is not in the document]
ademidoff: Put both to the `RestoreService` 👍

api/management/v1/rds.proto
BupycHuk: I remember that we discussed moving table_count to MySQLdExporter

ademidoff: Yes, it's in there as well 

```
message AddMySQLResponse {
  inventory.v1.MySQLService service = 1;
  inventory.v1.MySQLdExporter mysqld_exporter = 2;
  inventory.v1.QANMySQLPerfSchemaAgent qan_mysql_perfschema = 3;
  inventory.v1.QANMySQLSlowlogAgent qan_mysql_slowlog = 4;
  // Actual table count at the moment of adding.
  int32 table_count = 5;
}
```

BupycHuk: not sure about the code you attached, because `AddMySQLResponse != MySQLdExporter`

ademidoff: Ah, right - it's not the same. Let me see...

ademidoff: I think what you remember from one of the previous rounds of API refactoring has been addressed - https://github.com/percona/pmm/blob/d3ad3cc1bfc5e1833a111b7e07e24d71caf12c8b/api/inventory/v1/agents.proto#L137

This, however, may be also valid, so I suggest we address it in the next round so keep making progress.

BupycHuk: since we already migrated that into MySQLdExporter, let's drop table_count from other messages

PR #2837, opened 2024-02-21
PMM-12894 Add max conneсtions config support for PG exporter

artemgavrilov:
PMM-12894

Link to the Feature Build: SUBMODULES-3559

- [ ] API Docs updated

- https://github.com/percona/postgres_exporter/pull/193


managed/models/agent_helpers.go
```diff
@@ -73,6 +74,7 @@ func PostgreSQLOptionsFromRequest(params PostgreSQLOptionsParams) *PostgreSQLOpt
 	// PostgreSQL exporter has these parameters but they are not needed for QAN agent.
 	if extendedOptions, ok := params.(PostgreSQLExtendedOptionsParams); ok && extendedOptions != nil {
 		res.AutoDiscoveryLimit = extendedOptions.GetAutoDiscoveryLimit()
+		res.MaxExporterConnections = extendedOptions.GetMaxConnections() // TODO: is it a right place for this?
```
artemgavrilov, line 77: Is it?

admin/commands/management/add_postgresql.go
```diff
@@ -61,24 +61,25 @@ type AddPostgreSQLCommand struct {
...
+	TLSSkipVerify          bool              `help:"Skip TLS certificates validation"`
+	MaxQueryLength         int32             `placeholder:"NUMBER" help:"Limit query length in QAN (default: server-defined; -1: no limit)"`
+	DisableQueryExamples   bool              `name:"disable-queryexamples" help:"Disable collection of query examples"`
+	MetricsMode            string            `enum:"${metricsModesEnum}" default:"auto" help:"Metrics flow mode, can be push - agent will push metrics, pull - server scrape metrics from agent or auto - chosen by server"`
+	DisableCollectors      []string          `help:"Comma-separated list of collector names to exclude from exporter"`
+	ExposeExporter         bool              `name:"expose-exporter" help:"Optionally expose the address of the exporter publicly on 0.0.0.0"`
+	AutoDiscoveryLimit     int32             `default:"0" placeholder:"NUMBER" help:"Auto-discovery will be disabled if there are more than that number of databases (default: server-defined, -1: always disabled)"`
+	MaxExporterConnections int32             `default:"0" placeholder:"NUMBER" help:"Maximum number of connections to the exporter (default: server-defined)"`
```
idoqo, line 82: Do we need to point out what the default value of 0 implies? (e.g it will be ignored vs meaning an unlimited max number of connections)?

managed/services/agents/postgresql_test.go
```diff
@@ -286,6 +287,84 @@ func TestAutoDiscovery(t *testing.T) {
...
+	t.Run("Max exporter connections set to 0 - ignore", func(t *testing.T) {
+		exporter.PostgreSQLOptions = &models.PostgreSQLOptions{
+			MaxExporterConnections: 0,
+		}
+		res, err := postgresExporterConfig(node, postgresql, exporter, redactSecrets, pmmAgentVersion)
+		assert.NoError(t, err)
+		assert.NotContains(t, res.Args, maxConnectionsFlag)
+	})
+
+	t.Run("Max exporter connections set to 0 - apply", func(t *testing.T) {
```
idoqo, line 358:
```suggestion
	t.Run("Max exporter connections set to 5 - apply", func(t *testing.T) {
```

idoqo (approved):

admin/commands/management/add_postgresql.go
artemgavrilov: It seems like no. `0` or `uset` means that server chooses appropriate value by itself, we have mention about that in the help string. Removed this `default` setting.

api-tests/management/postgresql_test.go
```diff
@@ -100,14 +101,15 @@ func TestAddPostgreSQL(t *testing.T) {
...
-					AutoDiscoveryLimit: 10,
+					AgentID:                listAgents.Payload.PostgresExporter[0].AgentID,
+					ServiceID:              serviceID,
+					PMMAgentID:             pmmAgentID,
+					Username:               "username",
+					DisabledCollectors:     []string{"custom_query.ml", "custom_query.mr.directory"},
+					PushMetricsEnabled:     true,
+					Status:                 &AgentStatusUnknown,
+					AutoDiscoveryLimit:     10,
+					MaxExporterConnections: 0,
```
BupycHuk, line 112: seems like behaviour of AutoDiscoveryLimit and MaxExporterConnections are different, can we have them consistent?

BupycHuk: I mean that above you added both items with value = `0`, but in response they have different values.

admin/commands/management/add_postgresql.go
```diff
@@ -61,24 +61,25 @@ type AddPostgreSQLCommand struct {
...
+	TLSSkipVerify          bool              `help:"Skip TLS certificates validation"`
+	MaxQueryLength         int32             `placeholder:"NUMBER" help:"Limit query length in QAN (default: server-defined; -1: no limit)"`
+	DisableQueryExamples   bool              `name:"disable-queryexamples" help:"Disable collection of query examples"`
+	MetricsMode            string            `enum:"${metricsModesEnum}" default:"auto" help:"Metrics flow mode, can be push - agent will push metrics, pull - server scrape metrics from agent or auto - chosen by server"`
+	DisableCollectors      []string          `help:"Comma-separated list of collector names to exclude from exporter"`
+	ExposeExporter         bool              `name:"expose-exporter" help:"Optionally expose the address of the exporter publicly on 0.0.0.0"`
+	AutoDiscoveryLimit     int32             `placeholder:"NUMBER" help:"Auto-discovery will be disabled if there are more than that number of databases (default: server-defined, -1: always disabled)"`
+	MaxExporterConnections int32             `placeholder:"NUMBER" help:"Maximum number of connections to the exporter (default: server-defined)"`
```
BupycHuk, line 82: as far as I see it's not really server-defined, but exporter defined.

api-tests/management/postgresql_test.go
artemgavrilov: Ah... with `max_exporter_connections` logic is that if it's not set, then it's up to exporter to decide (i.e. default value set on the exporter side). Not sure that it makes sense to have it on PMM side. With current logic we can easily change default value without DB migration.

admin/commands/management/add_postgresql.go
artemgavrilov:
```suggestion
	MaxExporterConnections int32             `placeholder:"NUMBER" help:"Maximum number of connections that exporter can open at the same time (default: exporter-defined)"`
```

Like this?

api-tests/management/postgresql_test.go
BupycHuk: Probably we still need it until we finish the last part.
We could achieve `server-defined` result without storing the value in DB, but having it hardcoded.

managed/services/agents/postgresql.go
```diff
@@ -34,6 +35,7 @@ var (
 	postgresExporterAutodiscoveryVersion = version.MustParse("2.15.99")
 	postgresExporterWebConfigVersion     = version.MustParse("2.30.99")
 	postgresSSLSniVersion                = version.MustParse("2.40.99")
+	postgresMaxExporterConnsVersion      = version.MustParse("2.41.99")
```
BupycHuk, line 38: it will be in 2.41.2 or 2.41.3, so I think better to use version `2.41.2-0`

managed/services/agents/postgresql.go
```diff
@@ -85,6 +87,12 @@ func postgresExporterConfig(node *models.Node, service *models.Service, exporter
 			fmt.Sprintf("--exclude-databases=%s", strings.Join(postgresExcludedDatabases(), ",")))
 	}
 
+	if !pmmAgentVersion.Less(postgresMaxExporterConnsVersion) &&
+		exporter.PostgreSQLOptions != nil &&
+		exporter.PostgreSQLOptions.MaxExporterConnections != 0 {
+		args = append(args, "--max-connections="+strconv.Itoa(int(exporter.PostgreSQLOptions.MaxExporterConnections)))
+	}
```
BupycHuk, line 94:
```suggestion
	if !pmmAgentVersion.Less(postgresMaxExporterConnsVersion) {
	    if exporter.PostgreSQLOptions != nil &&
		exporter.PostgreSQLOptions.MaxExporterConnections != 0 {
		    args = append(args, "--max-connections="+strconv.Itoa(int(exporter.PostgreSQLOptions.MaxExporterConnections)))
		} else {
		    args = append(args, "--max-connections=10")
		}
	}
```

managed/models/agent_helpers.go
artemgavrilov: @JiriCtvrtka Do you know why do we need extended options at all?

JiriCtvrtka: @artemgavrilov I far as I remember we using it for unique properties l. Like property is needed for postgres_exporter but not useful for pg_stat_monitor etc.

artemgavrilov: Ok, thanks. Then it's right place for new parameter

api-tests/management/postgresql_test.go
artemgavrilov: After discussion we decided to change `AutoDiscoveryLimit` behavior instead.

managed/services/agents/postgresql.go
artemgavrilov: Done

TODO: double check before merge

admin/commands/management/add_postgresql.go
artemgavrilov: Fixed description, but kept `server-defined`. Please let me know if you want to change it to something else.

managed/services/agents/postgresql.go
artemgavrilov: Not actual anymore

JiriCtvrtka (approved):

PR #2856, opened 2024-02-28
PMM-12333 Make pmm-client tarball installation more user-friendly

surbhat1595:
PMM-12333

Link to the Feature Build: SUBMODULES-3585

- [ ] API Docs updated


build/scripts/install_tarball
```diff
@@ -17,7 +43,15 @@ else
     INSTALL_COMMAND="install -o ${PMM_USER} -g ${PMM_GROUP}"
 fi
 
-echo "Installing into ${PMM_DIR}..."
+# Check if PMM_DIR has the right permission to install files into it
+mkdir -p "${PMM_DIR}" || true
+if [ -w ${PMM_DIR} ]; then
+    echo "Installing into ${PMM_DIR}...";
+else
+    echo -e "${PMM_DIR} is not writable. \nPlease use a directory with the right permissions\n\n"
```
ademidoff, line 51:
```suggestion
    echo -e "${PMM_DIR} is not writable. \nPlease make sure the user $(id -un) has permissions to write to this directory\n\n"
```

build/scripts/install_tarball
```diff
@@ -2,7 +2,33 @@
...
+do
+    case "$arg" in
+        "-u") UPDATE=1 ;;
+        "-help") usage ;;
+        *) echo "Invalid option:$arg"; usage;;
+    esac
+done
+
 CURRENT_DIR="$(pwd)"
+PMM_VER=$(echo "$CURRENT_DIR" | cut -d'-' -f3)
```
ademidoff, line 31: Why do we assume PMM version can be taken from current directory? What if the use runs the whole package to an arbitrary directory and then runs `./install_tarball`?

IMO `PMM_VER` will be an empty string at the very best.

ademidoff (changes requested):

build/scripts/install_tarball
surbhat1595: According to the installation guide https://docs.percona.com/percona-monitoring-and-management/setting-up/client/index.html#binary-package, we run the install_tarball script from the extracted PMM tarball directory (eg: pmm2-client-2.41.1), hence I fetched PMM_VER using this.

Also PMM_VER isn't really necessary. I have just used it to print a message when the installation is successful. We can remove the PMM_VER variable. Not sure if there is any other way to find the version.

ademidoff: > According to the installation guide https://docs.percona.com/percona-monitoring-and-management/setting-up/client/index.html#binary-package, we run the install_tarball script from the extracted PMM tarball directory (eg: pmm2-client-2.41.1) which is the current directory, hence I fetched PMM_VER using this.

I don't think this answers the question whether or not the user won't run the install script from an arbitrary directory :)

ademidoff: > Also PMM_VER isn't really necessary. I have just used it to print a message when the installation is successful. We can remove the PMM_VER variable. Not sure if there is any other way to find the version.

It should be possible to run `cat VERSION` from the directory where the user unpacks the tarball (see attachement). But that file is empty now for some reason, i.e. its length is just one byte :( FYI @talhabinrizwan 

Screenshot 2024-02-28 at 23 50 13

ademidoff: I suggest we skip printing the version for now.

surbhat1595: Yes, thought of fetching from the VERSION file but found it empty. Sure, I shall remove "PMM_VER".

build/scripts/install_tarball
```diff
@@ -64,19 +97,14 @@ for FILE in example.prom queries-mysqld.yml example-queries-postgres.yml; do
...
-
-if [ "${update}" = "1" ]; then
+if [ "${UPDATE}" = "1" ]; then
   echo "Config file was not removed!"
 else
   ${INSTALL_COMMAND} -m 0660 /dev/null "${PMM_DIR}"/config/pmm-agent.yaml
 fi
 
+if [ $? -eq 0 ]; then
+  echo "Successfully installed files to ${PMM_DIR}"
```
ademidoff, line 107:
```suggestion
  echo "Successfully installed PMM Client to ${PMM_DIR}"
```

build/scripts/install_tarball
```diff
@@ -2,6 +2,31 @@
...
+# Usage instruction for the install_tarball script
+usage () {
+    cat <<EOF
+Usage: $0 [OPTIONS]
+    The following options may be given :
+        -u                  Update PMM-Agent
+        -help) usage ;;
+Example:
+        For help command - $0 -help
+        For PMM tarball installation - $0
```
nastena1606, line 14:
```suggestion
        To install PMM tarball - $0
```

build/scripts/install_tarball
```diff
@@ -17,7 +42,15 @@ else
     INSTALL_COMMAND="install -o ${PMM_USER} -g ${PMM_GROUP}"
 fi
 
-echo "Installing into ${PMM_DIR}..."
+# Check if PMM_DIR has the right permission to install files into it
+mkdir -p "${PMM_DIR}" || true
+if [ -w ${PMM_DIR} ]; then
+    echo "Installing into ${PMM_DIR}...";
+else
+    echo -e "${PMM_DIR} is not writable. \nPlease make sure the user $(id -un) has permissions to write to this directory\n\n"
```
nastena1606, line 50:
```suggestion
    echo -e "Cannot write to ${PMM_DIR}. \nPlease make sure the user $(id -un) has permissions to write to this directory\n\n"
```

ademidoff: Quote from the source - https://www.gnu.org/software/bash/manual/bash.html

> -w file
True if file exists and is writable.

It seems like this is how the Linux community talks `bash` :)

surbhat1595: @ademidoff So, should we go with "writable" only?

EvgeniyPatlan: I agree with @nastena1606 's suggestion.

BupycHuk (approved):

BupycHuk:
@surbhat1595 do we have FB for it?

surbhat1595:
@BupycHuk Yes, I have created a FB for it.

BupycHuk:
@surbhat1595 what's the status of this task? As far as I see in jira it's not in progress, but `Open`

surbhat1595:
@BupycHuk  The status has to be changed to "In Progress", I think this should be changed by someone in the PMM team.  Once the changes are merged, the script can be tested by the QA.

PR #2882, opened 2024-03-07
PMM-12895 Add metrics resolutions configuration to agents

artemgavrilov:
PMM-12895

Link to the Feature Build: SUBMODULES-3578

- [ ] API Docs updated


api-tests/inventory/agents_test.go
```diff
@@ -1198,3 +1198,170 @@ func TestPGStatMonitorQanAgent(t *testing.T) {
...
+				MetricsResolutions: &agents.ChangePostgresExporterOKBodyPostgresExporterMetricsResolutions{
+					Hr: "600s",
+					Mr: "300s",
+					Lr: "0s",
+				},
+			},
+		},
+	}, changePostgresExporterOK)
+
+	// Reset part of metrics resolutions
```
artemgavrilov, line 1335:
```suggestion
	// Change part of metrics resolutions
```

api-tests/inventory/agents_azure_database_exporter_test.go
```diff
@@ -73,6 +73,11 @@ func TestAzureDatabaseExporter(t *testing.T) { //nolint:tparallel
 						"custom_label_azure_database_exporter": "azure_database_exporter",
 					},
 					Status: &AgentStatusUnknown,
+					MetricsResolutions: &agents.GetAgentOKBodyAzureDatabaseExporterMetricsResolutions{
+						Hr: "0s",
+						Lr: "0s",
+						Mr: "0s",
```
BupycHuk, line 79: what if they are just empty?

api-tests/management/external_test.go
```diff
@@ -95,6 +95,11 @@ func TestAddExternal(t *testing.T) {
 					RunsOnNodeID: nodeID,
 					Scheme:       "http",
 					MetricsPath:  "/metrics",
+					MetricsResolutions: &agents.ListAgentsOKBodyExternalExporterItems0MetricsResolutions{
+						Hr: "0s",
+						Lr: "0s",
+						Mr: "0s",
+					},
```
BupycHuk, line 102: I believe that any other places where we don't test MetricsResolution should support nil value for this field

descriptor.bin
BupycHuk, line 1: we shouldn't change descriptor in patch release, it means that we made breaking change

BupycHuk (changes requested):
what was the reason to update descriptor.bin? Does that mean that we introduced some backward-incompatible change?

descriptor.bin
ademidoff: I see a breaking change in `server.proto`. 

Maybe we can leave `server.proto` as is and just use the messages we added to `common.proto` where approriate (except server.proto) without breaking the backward compat?

artemgavrilov: Reverted

api-tests/management/external_test.go
artemgavrilov: Done. Not every resolution field is nullable as well as whole `MetricsResoultions` field

api-tests/inventory/agents_azure_database_exporter_test.go
artemgavrilov: Made them nullable

BupycHuk (approved):
LGTM, can we add a new test to scrape_configs_test with custom metrics resolution?

artemgavrilov:
> LGTM, can we add a new test to scrape_configs_test with custom metrics resolution?

Functions from `scrape_configs.go` accept metrics resolutions as a parameter, so it doesn't matter from where it comes.

JiriCtvrtka (approved):

PR #2898, opened 2024-03-13
PMM-12896 Add limit for actions/jobs executed on the same DB at the same time

artemgavrilov:
PMM-12896

Link to the Feature Build: SUBMODULES-3588


agent/runner/actions/mongodb_explain_action.go
```diff
@@ -31,23 +31,32 @@ import (
...
 
 var errCannotExplain = fmt.Errorf("cannot explain this type of query")
 
 // NewMongoDBExplainAction creates a MongoDB EXPLAIN query Action.
-func NewMongoDBExplainAction(id string, timeout time.Duration, params *agentpb.StartActionRequest_MongoDBExplainParams, tempDir string) Action {
+func NewMongoDBExplainAction(id string, timeout time.Duration, params *agentpb.StartActionRequest_MongoDBExplainParams, tempDir string) (Action, error) {
+	dsn, err := templates.RenderDSN(params.Dsn, params.TextFiles, filepath.Join(tempDir, strings.ToLower(mongoDBExplainActionType), id))
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
```
BupycHuk, line 51: I'm afraid that in case of SSL enabled it will be difference for each action. I would recommend to use params.Dsn instead of rendered one

artemgavrilov: Mmm, I didn't get it. This chunk was extracted from `Run` method just to be able to access `dsn` later (in ` DSN()` method).

artemgavrilov: We discussed this topic on call.

agent/runner/actions/mongodb_explain_action.go
```diff
@@ -31,23 +31,30 @@ import (
...
-	tempDir string
+	dsn     string
 }
 
 var errCannotExplain = fmt.Errorf("cannot explain this type of query")
 
 // NewMongoDBExplainAction creates a MongoDB EXPLAIN query Action.
-func NewMongoDBExplainAction(id string, timeout time.Duration, params *agentpb.StartActionRequest_MongoDBExplainParams, tempDir string) Action {
+func NewMongoDBExplainAction(id string, timeout time.Duration, params *agentpb.StartActionRequest_MongoDBExplainParams, tempDir string) (Action, error) {
+	dsn, err := templates.RenderDSN(params.Dsn, params.TextFiles, filepath.Join(tempDir, strings.ToLower(mongoDBExplainActionType), id))
```
BupycHuk, line 47:
```suggestion
	dsn, err := templates.RenderDSN(params.Dsn, params.TextFiles, filepath.Join(tempDir, mongoDBExplainActionType, id))
```

BupycHuk: probably in other places it's not required too

agent/runner/actions/mysql_explain_action_test.go
```diff
@@ -167,13 +173,9 @@ func TestMySQLExplain(t *testing.T) {
...
 			OutputFormat: agentpb.MysqlExplainOutputFormat_MYSQL_EXPLAIN_OUTPUT_FORMAT_DEFAULT,
 		}
-		a := NewMySQLExplainAction("", time.Second, params)
-		ctx, cancel := context.WithTimeout(context.Background(), a.Timeout())
-		defer cancel()
-
-		_, err := a.Run(ctx)
-		require.Error(t, err)
+		a, err := NewMySQLExplainAction("", time.Second, params)
 		assert.Regexp(t, `Query to EXPLAIN is empty`, err.Error())
```
BupycHuk, line 177: assert has `ErrorContains` to check if error message contains substring. in case of no error this test might panic. but in this case it seems like `assert.EqualError` would be enough

agent/runner/actions/mysql_explain_action_test.go
```diff
@@ -205,12 +209,9 @@ func TestMySQLExplain(t *testing.T) {
...
 			OutputFormat: agentpb.MysqlExplainOutputFormat_MYSQL_EXPLAIN_OUTPUT_FORMAT_DEFAULT,
 		}
-		a := NewMySQLExplainAction("", time.Second, params)
-		ctx, cancel := context.WithTimeout(context.Background(), a.Timeout())
-		defer cancel()
-
-		_, err := a.Run(ctx)
-		require.Error(t, err, "EXPLAIN failed because the query was too long and trimmed. Set max-query-length to a larger value.")
+		a, err := NewMySQLExplainAction("", time.Second, params)
+		assert.Error(t, err, "EXPLAIN failed because the query was too long and trimmed. Set max-query-length to a larger value.")
```
BupycHuk, line 213: I'm not sure if `assert.Error` checks if error is equal to the last argument. it's better ot use `assert.EqualError`

agent/runner/runner.go
```diff
@@ -48,30 +53,100 @@ type Runner struct {
...
+		r.lSemsM.Unlock()
+
+		if err := e.sem.Acquire(ctx, 1); err != nil {
+			return err
+		}
+		e.count.Add(1)
+	}
+
+	if err := r.gSem.Acquire(ctx, 1); err != nil {
+		return err
```
BupycHuk, line 130: shouldn't we release from the semaphore above in case of error here?

agent/runner/runner.go
```diff
@@ -124,40 +199,53 @@ func (r *Runner) ActionsResults() <-chan agentpb.AgentRequestPayload {
...
-	// Job removes itself from rCancel map. So here we only invoke cancel.
-	if cancel, ok := r.rCancel[id]; ok {
+	// Job removes itself from cancels map. So here we only invoke cancel.
+	if cancel, ok := r.cancels[id]; ok {
 		cancel()
 	}
 }
 
 // IsRunning returns true if Action or Job with given ID still running.
 func (r *Runner) IsRunning(id string) bool {
```
BupycHuk, line 212: I just realized that message `JobStatusRequest` isn't sent by pmm-managed, so this function never will be called, so I'm not sure if it make sense to keep it here. seems like dead code

agent/runner/runner_test.go
```diff
@@ -94,14 +98,6 @@ func TestCapacityLimit(t *testing.T) {
...
 	time.Sleep(time.Second)
 
-	// After second first job terminated and third job started
-	assert.False(t, cr.IsRunning(j1.ID()))
-	assert.True(t, cr.IsRunning(j2.ID()))
-	assert.True(t, cr.IsRunning(j3.ID()))
-	assert.False(t, cr.IsRunning(j4.ID()))
-
-	time.Sleep(time.Second)
-
```
BupycHuk, line 104: just curious why was this removed?

agent/runner/runner_test.go
```diff
@@ -121,28 +117,127 @@ func TestDefaultCapacityLimit(t *testing.T) {
...
+
+	// Over time all jobs are terminated
+	time.Sleep(time.Second)
+
+	assert.False(t, cr.IsRunning(j1db1.ID()))
+	assert.False(t, cr.IsRunning(j1db2.ID()))
+	assert.False(t, cr.IsRunning(j2db1.ID()))
+	assert.False(t, cr.IsRunning(j2db2.ID()))
+	assert.False(t, cr.IsRunning(j3db1.ID()))
+	assert.False(t, cr.IsRunning(j3db2.ID()))
```
BupycHuk, line 188: does that mean that timeout timer starts from the moment pmm-agent receives the message. previously behaviour was different and cancel timer was starting after acquiring the global semaphore

agent/runner/runner_test.go
```diff
@@ -121,28 +117,127 @@ func TestDefaultCapacityLimit(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go cr.Run(ctx)
+
+	j1db1 := testJob{id: "test-1", timeout: time.Second, dsn: "postgresql://db1"}
+	j2db1 := testJob{id: "test-2", timeout: time.Second, dsn: "postgresql://db1"}
+	j3db1 := testJob{id: "test-3", timeout: time.Second, dsn: "postgresql://db1"}
+	j1db2 := testJob{id: "test-4", timeout: time.Second, dsn: "postgresql://db2"}
+	j2db2 := testJob{id: "test-5", timeout: time.Second, dsn: "postgresql://db2"}
+	j3db2 := testJob{id: "test-6", timeout: time.Second, dsn: "postgresql://db2"}
```
BupycHuk, line 157: I think in these tests it's better to put db in front of job number to make it easier to see that it's another db job. I would say in this context db is more important then job number.
```suggestion
	db1j1 := testJob{id: "test-1", timeout: time.Second, dsn: "postgresql://db1"}
	db1j2 := testJob{id: "test-2", timeout: time.Second, dsn: "postgresql://db1"}
	db1j3 := testJob{id: "test-3", timeout: time.Second, dsn: "postgresql://db1"}
	db2j1 := testJob{id: "test-4", timeout: time.Second, dsn: "postgresql://db2"}
	db2j2 := testJob{id: "test-5", timeout: time.Second, dsn: "postgresql://db2"}
	db2j3 := testJob{id: "test-6", timeout: time.Second, dsn: "postgresql://db2"}
```

BupycHuk (commented):
Do we really need to limit connections for any other DB types except PostgreSQL? Does it affect other DBs the same way as it does for PostgreSQL. I'm affraid of doing more harm then providing advantage with these changes.

agent/runner/actions/pt_mysql_summary_action.go
```diff
@@ -63,6 +64,11 @@ func (a *ptMySQLSummaryAction) Type() string {
 	return a.command
 }
 
+// DSN returns a DSN for the Action.
+func (a *ptMySQLSummaryAction) DSN() string {
+	return net.JoinHostPort(a.params.Host, strconv.FormatUint(uint64(a.params.Port), 10))
```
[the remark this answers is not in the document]
artemgavrilov: Right, fixed

agent/runner/runner.go
artemgavrilov: Yes, good catch. Fixed.

agent/runner/runner_test.go
artemgavrilov: Yes, and it looks like before it wasn't correct. It seems that from user perspective timeout should be like `total time allocated for execution`.

agent/runner/runner_test.go
artemgavrilov: Before there was one job task queue, I mean `j1` always started before `j2`, and `j2` before `j3`. With recent changes we don't have this type of behavior anymore, order depends on resources availability for each job.

agent/runner/actions/mongodb_explain_action.go
artemgavrilov: Yep, removed

agent/runner/actions/mysql_explain_action_test.go
artemgavrilov: Used `ErrorContains`

agent/runner/actions/mysql_explain_action_test.go
artemgavrilov: Yep, this is an error.

artemgavrilov:
> Do we really need to limit connections for any other DB types except PostgreSQL? Does it affect other DBs the same way as it does for PostgreSQL. I'm affraid of doing more harm then providing advantage with these changes.

MySQL and MongoDB don't have such problems with connections as PostgreSQL, but it may be still an issue. For example MySQL by default allows `151` connections. So having some limits on our side sounds like a good idea. Moreover with this change we can say `exactly` how many connections do we need. But I'm agree that we need to test that overall user experience is still good.

agent/runner/runner_test.go
BupycHuk: as far as I see user doesn't provide this timeout, so it's our internal mechanism to not let action to work forever.

artemgavrilov: By `user` I mean `PMM Server` or anything else that is interested in such action.

agent/runner/runner.go
```diff
@@ -137,6 +138,11 @@ func (r *Runner) acquire(ctx context.Context, token string) error {
 func (r *Runner) release(token string) {
 	r.gSem.Release(1)
 
+	r.releaseL(token)
+}
+
+// releaseL releases local semaphore for given token.
+func (r *Runner) releaseL(token string) {
```
BupycHuk, line 145: let's have `acquireL` for simmetry :)

BupycHuk (approved):
Looks good to me

agent/runner/runner.go
artemgavrilov: Done

BupycHuk (approved):

agent/runner/actions/mysql_explain_action.go
```diff
@@ -54,12 +54,27 @@ var errCannotEncodeExplainResponse = errors.New("cannot JSON encode the explain
...
+
+	// You cant run Explain on trimmed queries.
+	if strings.HasSuffix(params.Query, "...") {
+		return nil, errors.New("EXPLAIN failed because the query was too long and trimmed. Set max-query-length to a larger value.") //nolint:revive
+	}
+
+	// Explain is supported only for DML queries.
+	// https://dev.mysql.com/doc/refman/8.0/en/using-explain.html
+	if !isDMLQuery(params.Query) {
+		return nil, errors.New("Functionality EXPLAIN is supported only for DML queries (SELECT, INSERT, UPDATE, DELETE, REPLACE)")
```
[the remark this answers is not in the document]
artemgavrilov: Done

agent/runner/actions/mysql_explain_action.go
```diff
@@ -54,12 +54,27 @@ var errCannotEncodeExplainResponse = errors.New("cannot JSON encode the explain
...
 // This is an Action that can run `EXPLAIN` command on MySQL service with given DSN.
-func NewMySQLExplainAction(id string, timeout time.Duration, params *agentpb.StartActionRequest_MySQLExplainParams) Action {
+func NewMySQLExplainAction(id string, timeout time.Duration, params *agentpb.StartActionRequest_MySQLExplainParams) (Action, error) {
+	if params.Query == "" {
+		return nil, errors.New("Query to EXPLAIN is empty")
+	}
+
+	// You cant run Explain on trimmed queries.
+	if strings.HasSuffix(params.Query, "...") {
+		return nil, errors.New("EXPLAIN failed because the query was too long and trimmed. Set max-query-length to a larger value.") //nolint:revive
```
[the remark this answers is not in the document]
artemgavrilov: Done

ademidoff (approved):

PR #2925, opened 2024-03-22
PMM-12998 Log message on ansible self update.

talhabinrizwan:
PMM-12998

Link to the Feature Build: SUBMODULES-3609


update/ansible/playbook/tasks/update.yml
```diff
@@ -22,6 +22,38 @@
       - vmproxy
       - grafana-db-migrator
   pre_tasks:
+      - name: Check if ansible-core RPM was updated
+      shell: "grep -A 1 'Install ansible-core RPM          | EL9' /srv/logs/pmm-update-perform.log | grep -qE 'changed' && echo 'Updated' || echo 'Not Updated'"
```
BupycHuk, line 26: since pmm-update-perform.log will keep all logs even related to past upgrades we will see this message again and again in future upgrades. can't we check `changed` task within the same run? And we don't have controllable restart, so restart might happen after we made some changes that's not supposed to be done.

ademidoff, line 26:
```suggestion
      shell: "grep -A 1 'Install ansible-core RPM          | EL9' /srv/logs/pmm-update-perform.log | grep -q 'changed' && echo 'Updated' || echo 'Not Updated'"
```

ademidoff (approved):
Once this is approved and merged, may I ask you to remove this change from v3, i.e.:
1/ merge `main` to `v3`
2/ remove the change from v3

update/ansible/playbook/tasks/update.yml
ademidoff: I think we need to reverse the quotes, i.e. the outermost should be single quotes.

talhabinrizwan: Are you certain that "pmm-update-perform.log" will retain all logs, including those pertaining to past upgrades? I initiated an upgrade via the UI, then attempted to force it again by (option + clicking) on the upgrade check button, which resulted in the replacement of the original log file.

If the log indeed preserves records of previous upgrades, then I believe we may be making this process overly complex by extracting information from the log file. That's why initially I attempted to get info from the repository itself whether an update is available for the ansible-core package.

ademidoff (approved):

BupycHuk (approved):

PR #2935, opened 2024-03-29
PMM-13292 Add INITIALIZATION_ERROR statue for process initialization failure 

GuanqunYang193:
PMM-13292
### what
* Added a new status of process, used for showing the process failed when initialization.
* Supervisor will initiate a port selection and restart the process initialization if failure is detected during the process initialization.
### why
If the process fails to start (due to, for instance, a port conflict), the previous design will attempt to retry the process using the same port.
### how
In this new commit, the supervisor will confirm the successful initialization of the process. If the initialization fails, it will make up to three attempts to start the process again.


agent/agents/process/process.go
```diff
@@ -53,12 +54,14 @@ const (
...
-	changes chan inventorypb.AgentStatus
-	backoff *backoff.Backoff
-	ctxDone chan struct{}
+	params      *Params
+	l           *logrus.Entry
+	pl          *processLogger
+	changes     chan inventorypb.AgentStatus
+	backoff     *backoff.Backoff
+	ctxDone     chan struct{}
+	err         chan error
```
BupycHuk, line 63: I think it can be just error, not a channel.

agent/agents/supervisor/supervisor_test.go
```diff
@@ -73,16 +73,16 @@ func TestSupervisor(t *testing.T) {
 
 		assertChanges(t, s,
 			&agentpb.StateChangedRequest{AgentId: "noop3", Status: inventorypb.AgentStatus_STARTING},
```
BupycHuk, line 75: Can I ask you why this test was changed?

agent/agents/supervisor/supervisor_test.go
```diff
@@ -259,6 +259,44 @@ func TestSupervisor(t *testing.T) {
 	})
 }
 
+func TestStartProcessFail(t *testing.T) {
```
BupycHuk, line 262: can we include this test to test group above? Seems like it supposed to be there

api/inventorypb/agent_status.dot
```diff
@@ -9,8 +9,7 @@ digraph {
   UNKNOWN [fillcolor=yellow];
 
   STARTING -> RUNNING;
-  STARTING -> WAITING;
-  STARTING -> UNKNOWN; [style=dotted]
```
BupycHuk, line 13: it still can go to unknown

api/inventorypb/agent_status.proto
```diff
@@ -11,12 +11,14 @@ enum AgentStatus {
   STARTING = 1;
   // Agent is running.
   RUNNING = 2;
-  // Agent encountered error and will be restarted automatically soon.
+  // Agent will be restarted automatically soon.
```
BupycHuk, line 14: it seems for me like it's still a true statement

api/inventorypb/agent_status.proto
```diff
@@ -11,12 +11,14 @@ enum AgentStatus {
...
   WAITING = 3;
   // Agent is stopping.
   STOPPING = 4;
   // Agent finished.
   DONE = 5;
   // Agent is not connected, we don't know anything about it's state.
   UNKNOWN = 6;
+  // Agent encountered error when starting.
+  INITIALIZATION_ERROR = 7;
```
BupycHuk, line 23: can we move it right after starting, so it's visible that it might be caused during starting. number still should be 7

BupycHuk (commented):
Looks good to me, just few comments

api/inventorypb/agent_status.proto
GuanqunYang193: My primary aim here is to differentiate this error from an initialization error. How about "Agent encountered error when running and will be restarted automatically soon."

agent/agents/supervisor/supervisor_test.go
GuanqunYang193: Given that the supervisor confirms whether the process starts successfully, the "RUNNING" state will follow the "STARTING" state before the supervisor initiates another process.

agent/agents/supervisor/supervisor_test.go
GuanqunYang193: TestSupervisor is a function with a group of tests where the states of different processes are highly interconnected, making it challenging to add a new individual test seamlessly. Personally, I find that using a new context would offer a cleaner and more straightforward way to verify this case.

agent/agents/supervisor/supervisor.go
```diff
@@ -473,6 +488,11 @@ func (s *Supervisor) startProcess(agentID string, agentProcess *agentpb.SetState
 		close(done)
 	}()
 
+	if !<-process.IsInitialized() {
+		defer cancel()
+		return process.GetError()
+	}
```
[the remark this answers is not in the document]
GuanqunYang193: The supervisor here is supposed to wait until the process update the initialization status. But I am ok with make it a non-blocking operation with a sleep so that the supervisor will not be blocked forever.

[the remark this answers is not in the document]
BupycHuk: we can revert last changes made here, I just realized that `initialized` will be returned anyway.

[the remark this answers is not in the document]
GuanqunYang193: I am ok to leave it there. Non-blocking operation for supervisord is a better design.

JiriCtvrtka (approved):

PR #2968, opened 2024-04-22
[Proposal] PMM-7 Updade year in license headers

artemgavrilov:
PMM-7

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -19,7 +19,7 @@ header:
     license:
       spdx-id: AGPL-3.0-or-later
       copyright-owner: Percona LLC
-      copyright-year: 2023
+      copyright-year: 2024
 
   - paths:
       - "agent/**/*.go"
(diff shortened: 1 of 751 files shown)


ademidoff:
@artemgavrilov IMO this is not necessary, since the meaning of the year in copyrights is "Since <year> up to now". 

Therefore, the repo files will bear that license until it changes, which is when we should update the year.

ademidoff:
> > @artemgavrilov IMO this is not necessary, since the meaning of the year in copyrights is "Since [year] up to now".
> > Therefore, the repo files will bear that license until it changes, which is when we should update the year.
> 
> Hm, I barely remember that we discussed what to do with the year and I thought that we agreed to update it. I am mistaken and we agreed to not touch year, shouldn't in be 2017 then (year when PMM project started)?

In theory we could put 2017 in there, but I don't see much value added.

Anyway, even if there was a benefit, in my opinion it would just not be worth the effort of:
- touching 750+ files
- dealing with conflicts in v3

artemgavrilov:
@BupycHuk @ademidoff Folks, I'm leaving this to you to decide what to do. I'm ok with any option.

ademidoff:
Closing this for now, will revisit once we stop maintaining v2.

PR #2970, opened 2024-04-23
PMM-13020 Drop identifier prefixes

ademidoff:
PMM-13020

Link to the Feature Build: SUBMODULES-3625

- [x] API Docs updated


BupycHuk (commented):
looks good so far,
what do you think about compatibilty mode?
it's for pmm-agents which still has `/agent_id/` prefix and for users who has automations to communicate to PMM API

JiriCtvrtka (approved):

agent/config/config.go
```diff
@@ -196,6 +197,10 @@ func get(args []string, cfg *Config, l *logrus.Entry) (string, error) { //nolint
 		}
 
 		// set default values
+		if strings.HasPrefix(cfg.ID, agentPrefix) {
+			l.Warnf("The agent ID '%s' contains a legacy prefix '%s'. It will be used without it.", cfg.ID, agentPrefix)
+			cfg.ID, _ = strings.CutPrefix(cfg.ID, agentPrefix)
+		}
```
BupycHuk, line 203: 👍🏼

BupycHuk (approved):
Great job, can we create a task for compatible mode?

ademidoff:
> Great job, can we create a task for compatible mode?

Sure, will do right away!

PR #2974, opened 2024-04-25
PMM-7 Upgrade codecov-action

ademidoff:
**Background**

We lost codecov.io uploads for some reason. Maybe because of `CODECOV_TOKEN`, which wasn't being passed.

This PR is to rectify that and upgrade codecov's action to the latest.

### DIFF
--- a/.github/workflows/admin.yml
+++ b/.github/workflows/admin.yml
@@ -70,6 +70,7 @@ jobs:
           file: cover.out
           flags: admin
           fail_ci_if_error: false
+          token: ${{ secrets.CODECOV_TOKEN }}
 
       - name: Run debug commands on failure
         if: ${{ failure() }}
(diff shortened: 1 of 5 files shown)


JiriCtvrtka (approved):

.github/workflows/admin.yml
```diff
@@ -161,9 +160,6 @@ jobs:
       - name: Run debug commands on failure
         if: ${{ failure() }}
         run: |
-          env
-          go version
-          go env
-          node --version
```
BupycHuk, line 167: why did we drop this line?

BupycHuk (approved):

.github/workflows/admin.yml
ademidoff: Because NodeJS is not used anywhere :)

ademidoff: I can revert, no prob

BupycHuk: all commands above uses npx which is a node package, I would also add info about npm, npx and playwright versions

ademidoff: Done.

PR #2976, opened 2024-04-28
PMM-7 Fix for service accounts tests stability.

JiriCtvrtka:
Fix for service accounts tests stability


JiriCtvrtka:
Tests should not be run in parallel. 

https://github.com/grafana/grafana/issues/77462

Issue should be fixed, but it is still persists for us in case of many concurrent service account permissions.

ademidoff (approved):

managed/services/grafana/auth_server_test.go
```diff
@@ -292,6 +292,7 @@ func TestServerClientConnection(t *testing.T) {
 		assert.Nil(t, authError)
 	})
 
+	// Beware: Five or more wrong tries will lock user with error message: "Invalid user or password".
```
BupycHuk, line 295: let's change username here to `admin1`, error message should be the same.

JiriCtvrtka: This is not happening when tests are triggered on github. It can happen when you run them locally multiple times (like -count=5 etc). I added this to not forget this could happen since it is not clear from the error message. Should I keep this comment or remove it?

PR #2984, opened 2024-05-02
PMM-13095 Remove trimming of slowlog query IDs.

JiriCtvrtka:
PMM-13095

### DIFF
--- a/agent/agents/mysql/slowlog/slowlog.go
+++ b/agent/agents/mysql/slowlog/slowlog.go
@@ -17,6 +17,7 @@ package slowlog
 
 import (
 	"context"
+	"crypto/md5"
 	"database/sql"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


JiriCtvrtka:
I kept MD5 hashing and not switched for SHA etc since it is ideal length (32 chars) without trimming and this corresponds with Perfschema MySQL DIGEST (query ID).

agent/agents/mysql/slowlog/slowlog.go
```diff
@@ -391,6 +393,18 @@ func (s *SlowLog) processFile(ctx context.Context, file string, outlierTime floa
 	}
 }
 
+// HashIntoQueryID returns slowlog query ID hashed by MD5 from given fingerprint.
+func HashIntoQueryID(fingerprint string) string {
+	// MD5 is used only to hash fingerprint into query ID, so there is no risk.
+	// It is ideal due to it's length (32 chars) and it corresponds with Perfschema query ID length.
```
ademidoff, line 399:
```suggestion
	// It is ideal due to its length (32 chars) and it corresponds to Perfschema query ID length.
```

ademidoff (approved):

idoqo (approved):

qan-api2/exporters/slow_log.go
```diff
@@ -35,6 +35,7 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 
+	slowlogAgent "github.com/percona/pmm/agent/agents/mysql/slowlog"
```
BupycHuk, line 38: qan shouldn't import internal packages of pmm-agent. Please extract that function to utils.

BupycHuk: seems like we can drop whole this file without any harm. Please check if we use it anywhere

PR #2987, opened 2024-05-08
PMM-12522 collect chunks metrics in low resolution

BupycHuk:
PMM-12522

Link to the Feature Build: SUBMODULES-3623

- [ ] API Docs updated

https://github.com/percona/mongodb_exporter/pull/849


managed/services/victoriametrics/prometheus.go
```diff
@@ -70,26 +70,29 @@ func AddScrapeConfigs(l *logrus.Entry, cfg *config.Config, q *reform.Querier, s
 		// find Node address where the agent runs
 		var paramsHost string
 		var paramPMMAgentVersion *version.Parsed
+		var pmmAgent *models.Agent
+		if agent.PMMAgentID != nil {
+			// extract node address through pmm-agent
```
JiriCtvrtka, line 75: Is it really extraction of node address not pmm-agent version?

JiriCtvrtka (approved):

managed/services/agents/mongodb.go
```diff
@@ -38,8 +38,8 @@ type collectorArgs struct {
 var (
 	// New MongoDB Exporter will be released with PMM agent v2.10.0.
 	newMongoExporterPMMVersion = version.MustParse("2.9.99")
-	v2_24_99                   = version.MustParse("2.24.99")
-	v2_25_99                   = version.MustParse("2.25.99")
+	v2_25_0                    = version.MustParse("2.25.0-0")
```
ademidoff, line 41: I wonder: will this version work with `-0` suffix? Do we have a unit test just in case?

BupycHuk: Yes, we have a test for it
https://github.com/percona/pmm/blob/main/version/parsed_test.go#L44

ademidoff (approved):

managed/services/victoriametrics/scrape_configs.go
```diff
@@ -418,7 +418,7 @@ func scrapeConfigsForMongoDBExporter(s *models.MetricsResolutions, params *scrap
 			"indexstats",
 			"collstats",
 		}
-		if params.pmmAgentVersion != nil && !params.pmmAgentVersion.Less(version.MustParse("2.41.1-0")) {
+		if !params.pmmAgentVersion.Less(version.MustParse("2.41.1-0")) {
```
ademidoff, line 421: AFAICS, `params.pmmAgentVersion` can be `nil` as per line 390.

BupycHuk: It can be nil in 390, but there is return in that logic branch, so here it’s safe to assume that it's not nil

PR #3002, opened 2024-05-23
PMM-13129 Encryption.

JiriCtvrtka:
PMM-13129

FB: https://github.com/Percona-Lab/pmm-submodules/pull/3645


managed/models/agent_helpers.go
```diff
@@ -229,7 +229,9 @@ func FindAgents(q *reform.Querier, filters AgentFilters) ([]*Agent, error) {
 
 	agents := make([]*Agent, len(structs))
 	for i, s := range structs {
-		agents[i] = s.(*Agent) //nolint:forcetypeassert
+		agent := s.(*Agent) //nolint:forcetypeassert
+		DecryptAgent(agent)
```
ademidoff, line 233: Shouldn't `DecryptAgent` return the same pointer to the agent? I'd much more convenient to use it as follows:

`agents[i] = DecryptAgent(s.(*Agent)) //nolint:forcetypeassert`

BupycHuk: returning the same pointer will make wrong assumption that initial object isn't changed

BupycHuk (changes requested):

managed/models/encryption_helpers.go
```diff
@@ -0,0 +1,336 @@
...
+		awsSecretKey, err := handler(*agent.AWSSecretKey)
+		if err != nil {
+			logrus.Debugf("aws_secret_key:%s", awsSecretKey)
+			logrus.Warning(err)
+		}
+		agent.AWSSecretKey = &awsSecretKey
+	}
+
+	var err error
+	if agent.MySQLOptions != nil {
```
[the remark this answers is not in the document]
JiriCtvrtka: They are used to handle read from specific column in DB (which is not string). For example PostgreSQLOptions column is json. Here it is used to handle agent. So they are working differently.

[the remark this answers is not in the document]
JiriCtvrtka: Methods DecryptMySQLOptionsHandler, DecryptPostgreSQLOptionsHandler, DecryptMongoDBOptionsHandler, DecryptAzureOptionsHandler are not used now (since during migration we using only Encrypt ones). I can remove them if you want. But since they are 3 lines of code I originally defines them due to consistency.

managed/models/encryption_helpers.go
```diff
@@ -0,0 +1,336 @@
...
+func EncryptAgent(agent *Agent) {
+	agentEncryption(agent, encryption.Encrypt)
+}
+
+// DecryptAgent decrypt agent.
+func DecryptAgent(agent *Agent) {
+	agentEncryption(agent, encryption.Decrypt)
+}
+
+func agentEncryption(agent *Agent, handler func(string) (string, error)) {
```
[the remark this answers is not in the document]
JiriCtvrtka: Isn't reflection much heavier on performance? More lines of code are longer, but since it is compiled there is no impact in binary itself.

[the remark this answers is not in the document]
BupycHuk: ok

managed/models/database.go
```diff
@@ -1173,17 +1274,21 @@ func migrateDB(db *reform.DB, params SetupDBParams) error {
...
+		err = EncryptDB(tx, params, itemsToEncrypt)
+		if err != nil {
 			return err
 		}
-		if err = setupFixture2(tx.Querier, params.Username, params.Password); err != nil {
+
+		err = setupPMMServerAgents(tx.Querier, params)
+		if err != nil {
 			return err
 		}
```
BupycHuk, line 1285: why are we doing it after encryption?

JiriCtvrtka: setupPMMServerAgents using CreateAgent methods. Those methods will encrypt agent during adding. Then EncryptDB is called and it will end up with double encryption for PMM server agents, since they are already encrypted once during CreateAgent.

With changed order first all current agents are encrypted, then PMM server ones are added. So non of them is double encrypted.

managed/models/database.go
```diff
@@ -1059,12 +1062,105 @@ func SetupDB(ctx context.Context, sqlDB *sql.DB, params SetupDBParams) (*reform.
...
+			columnsToEncrypt, encryption.Column{Name: "mysql_options", CustomHandler: EncryptMySQLOptionsHandler})
+	}
+	if params.MigrationVersion == nil || *params.MigrationVersion >= 41 {
+		columnsToEncrypt = append(
+			columnsToEncrypt, encryption.Column{Name: "postgresql_options", CustomHandler: EncryptPostgreSQLOptionsHandler})
+	}
+	if params.MigrationVersion == nil || *params.MigrationVersion >= 42 {
+		columnsToEncrypt = append(
+			columnsToEncrypt, encryption.Column{Name: "agent_password"})
+	}
```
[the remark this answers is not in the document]
JiriCtvrtka: Yes, but we have migrations tests from different migration versions. For example migration from version 9 to 10. In this one you have only username, password, aws_access_key, aws_secret_key fields. Migration test will end up with error message of non existing column to encrypt, like agent_password in this case etc. So it is mainly to pass migration tests.

managed/models/database.go
```diff
@@ -1059,12 +1062,105 @@ func SetupDB(ctx context.Context, sqlDB *sql.DB, params SetupDBParams) (*reform.
...
+		return nil
+	}
+
+	settings, err := GetSettings(tx)
+	if err != nil {
+		return err
+	}
+	alreadyEncrypted := make(map[string]bool)
+	for _, v := range settings.EncryptedItems {
+		alreadyEncrypted[v] = true
```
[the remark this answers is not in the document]
JiriCtvrtka: If I am not mistaken then if we introduce new column to be encrypted in new version then old columns will be encrypted again too. So old ones twice and only new one will be ok.

managed/utils/encryption/encryption.go
```diff
@@ -0,0 +1,198 @@
...
+const DefaultEncryptionKeyPath = "/srv/pmm-encryption.key"
+
+var (
+	// ErrEncryptionNotInitialized is error in case of encryption is not initialized.
+	ErrEncryptionNotInitialized = errors.New("encryption is not initialized")
+	// DefaultEncryption is the default implementation of encryption.
+	DefaultEncryption = New(DefaultEncryptionKeyPath)
+)
+
+// New create encryption, if key on path doesnt exists will be generated.
```
ademidoff, line 39:
```suggestion
// New creates an encryption; if key on path doesn't exist, it will be generated.
```

managed/utils/encryption/models.go
```diff
@@ -0,0 +1,56 @@
...
+	SSLMode              string
+	SSLCAPath            string
+	SSLKeyPath           string
+	SSLCertPath          string
+}
+
+// Table represents table name, it's identificators and columns to be encrypted/decrypted.
+type Table struct {
+	Name           string
+	Identificators []string
```
ademidoff, line 41:
```suggestion
	Identifiers []string
```

ademidoff: Check [here](https://translate.google.com/?sl=en&tl=cs&text=identifier&op=translate)

managed/utils/encryption/models.go
```diff
@@ -0,0 +1,56 @@
...
+import "github.com/google/tink/go/tink"
+
+// Encryption contains fields required for encryption.
+type Encryption struct {
+	Path      string
+	Key       string
+	Primitive tink.AEAD
+}
+
+// DatabaseConnection represents DB connection and it's encrypted items.
```
ademidoff, line 27:
```suggestion
// DatabaseConnection represents DB connection and its encrypted items.
```

managed/utils/encryption/helpers.go
```diff
@@ -0,0 +1,177 @@
...
+		q.SetValues = append(q.SetValues, setValues)
+
+		where := []string{}
+		whereValues := []any{}
+		for k, id := range table.Identificators {
+			where = append(where, fmt.Sprintf("%s = $%d", id, i))
+			whereValues = append(whereValues, row[k])
+			i++
+		}
+		whereSQL := fmt.Sprintf("WHERE %s", strings.Join(where, " AND "))
```
ademidoff, line 166:
```suggestion
		whereSQL := "WHERE " + strings.Join(where, " AND ")
```

ademidoff: There is a new lint rule (don't remember exactly which one), that will go red here claiming we can do the same without calling Sprintf.

managed/utils/encryption/encryption.go
```diff
@@ -0,0 +1,198 @@
...
+
+	bytes, err := os.ReadFile(e.Path)
+	switch {
+	case os.IsNotExist(err):
+		err = e.generateKey()
+		if err != nil {
+			logrus.Errorf("Encryption: %v", err)
+		}
+	case err != nil:
+		logrus.Errorf("Encryption: %v", err)
```
ademidoff, line 52: Same here.

managed/utils/encryption/encryption.go
```diff
@@ -0,0 +1,198 @@
...
+	default:
+		e.Key = string(bytes)
+	}
+
+	primitive, err := e.getPrimitive()
+	if err != nil {
+		logrus.Errorf("Encryption: %v", err)
```
ademidoff, line 59: Need to exit here as well.

JiriCtvrtka:
Main / Checks (pull_request) failing because it seems there is no "srv" directory in this test.
"Encryption: open /srv/pmm-encryption.key: permission denied"

Others are OK.

JiriCtvrtka:
I am unable to check snyk results with my @ext.percona.com account.

managed/models/encryption_helpers.go
```diff
@@ -0,0 +1,308 @@
...
+	if !value.Valid {
+		return sql.NullString{}, nil
+	}
+
+	err := json.Unmarshal([]byte(value.String), &o)
+	if err != nil {
+		return nil, err
+	}
+
+	o.SSLCa, err = handler(o.SSLCa)
```
ademidoff, line 199: @JiriCtvrtka @BupycHuk 
Do we need to encrypt the `Certificate Authority` (aka CA) fields, given that they are meant to be public? Or is it not always the case?

managed/models/settings.go
```diff
@@ -100,6 +100,9 @@ type Settings struct {
 		// Enabled is true if access control is enabled.
 		Enabled bool `json:"enabled"`
 	} `json:"access_control"`
+
+	// Contains all already encrypted tables in format 'db.table.column'.
```
ademidoff, line 104:
```suggestion
	// Contains all encrypted tables in format 'db.table.column'.
```

managed/utils/encryption/encryption.go
```diff
@@ -0,0 +1,198 @@
...
+	primitive, err := e.getPrimitive()
+	if err != nil {
+		logrus.Panicf("Encryption: %v", err)
+	}
+	e.Primitive = primitive
+
+	return e
+}
+
+// Encrypt is wrapper around DefaultEncryption.Encrypt.
```
ademidoff, line 66:
```suggestion
// Encrypt is a wrapper around DefaultEncryption.Encrypt.
```

managed/utils/encryption/encryption.go
```diff
@@ -0,0 +1,198 @@
...
+
+	cipherText, err := e.Primitive.Encrypt([]byte(secret), []byte(""))
+	if err != nil {
+		return secret, err
+	}
+
+	return base64.StdEncoding.EncodeToString(cipherText), nil
+}
+
+// EncryptItems is wrapper around DefaultEncryption.EncryptItems.
```
ademidoff, line 85:
```suggestion
// EncryptItems is a wrapper around DefaultEncryption.EncryptItems.
```

managed/models/agent_helpers.go
```diff
@@ -722,14 +737,18 @@ func CreateExternalExporter(q *reform.Querier, params *CreateExternalExporterPar
...
+	if err := encryptedAgent.SetCustomLabels(params.CustomLabels); err != nil {
 		return nil, err
 	}
-	if err := q.Insert(row); err != nil {
+	if err := q.Insert(&encryptedAgent); err != nil {
 		return nil, errors.WithStack(err)
 	}
 
-	return row, nil
+	decryptedAgent := DecryptAgent(encryptedAgent)
```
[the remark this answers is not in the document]
JiriCtvrtka: Now changed and agent DB options fields are dereferenced on encrypt/decrypt so redundancy is removed now.

managed/models/encryption_helpers.go
JiriCtvrtka: It was leftover in handler, but it is removed.

managed/models/database.go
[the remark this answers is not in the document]
BupycHuk: oh, that tests can be dropped with PMM 3

managed/models/agent_helpers.go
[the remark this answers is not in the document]
BupycHuk: Can we copy them in Encrypt and Decrypt to be safe?

managed/models/database.go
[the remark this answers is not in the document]
BupycHuk: ok

managed/models/database.go
BupycHuk: got it, thanks

managed/models/database.go
```diff
@@ -1161,6 +1257,11 @@ func migrateDB(db *reform.DB, params SetupDBParams) error {
 		}
 
 		if params.SetupFixtures == SkipFixtures {
+			err := EncryptDB(tx, params, itemsToEncrypt)
+			if err != nil {
+				return err
+			}
+
```
BupycHuk, line 1264: can we move it up before `if`, so we don't have to have it both here and below

managed/models/encryption_helpers.go
```diff
@@ -0,0 +1,296 @@
...
+		if err != nil {
+			logrus.Warning(err)
+		}
+	}
+
+	return agent
+}
+
+// EncryptMySQLOptionsHandler returns encrypted MySQL Options.
+func EncryptMySQLOptionsHandler(e *encryption.Encryption, val any) (any, error) {
```
BupycHuk, line 135: can we change usage of any with some type to make it easier to read

managed/models/database.go
JiriCtvrtka: Nice catch, moved.

managed/models/encryption_helpers.go
JiriCtvrtka: Discussed on GO BE.

JiriCtvrtka:
After our discussion on GO BE:
1. No decrypt after encrypt agent when adding agent. Instead I removed exposing of username, aws_access_key and aws_secret_key. But it was also necessary to remove all agent DB options for now. Because of field isSSLKeySet etc. So we should remove also fields isSet due to exposing information related to passwords. With that you can easily scan for empty ones. 
2. Added ENV variable to provide custom path for encryption key.

managed/services/management/agent.go
```diff
@@ -137,15 +137,11 @@ func (s *AgentService) agentToAPI(agent *models.Agent) (*agentv1beta1.UniversalA
 	ua := &agentv1beta1.UniversalAgent{
 		AgentId:                        agent.AgentID,
 		AgentType:                      string(agent.AgentType),
-		AwsAccessKey:                   pointer.GetString(agent.AWSAccessKey),
 		CreatedAt:                      timestamppb.New(agent.CreatedAt),
 		CustomLabels:                   labels,
 		Disabled:                       agent.Disabled,
 		DisabledCollectors:             agent.DisabledCollectors,
 		IsConnected:                    s.r.IsConnected(agent.AgentID),
-		IsAgentPasswordSet:             agent.AgentPassword != nil,
```
ademidoff, line 146: @JiriCtvrtka We need to create a task to update frontend accordingly, since Inventory now uses those keys.

JiriCtvrtka:
Responses now doesn't contains username, aws_access_key, aws_secret_key or isSet variations. FE ticket will be created soon based on these changes.

admin/commands/inventory/add_agent_external_exporter.go
```diff
@@ -28,7 +28,6 @@ External Exporter added.
 Agent ID              : {{ .Agent.AgentID }}
 Runs on node ID       : {{ .Agent.RunsOnNodeID }}
 Service ID            : {{ .Agent.ServiceID }}
-Username              : {{ .Agent.Username }}
```
BupycHuk, line 31: oh, I meant excluding username from encryption, not from exposed data

managed/services/management/agent.go
```diff
@@ -168,44 +164,37 @@ func (s *AgentService) agentToAPI(agent *models.Agent) (*agentv1beta1.UniversalA
...
 		TlsSkipVerify:                  agent.TLSSkipVerify,
-		Username:                       pointer.GetString(agent.Username),
 		UpdatedAt:                      timestamppb.New(agent.UpdatedAt),
 		Version:                        pointer.GetString(agent.Version),
 	}
 
 	if agent.AzureOptions != nil {
 		ua.AzureOptions = &agentv1beta1.UniversalAgent_AzureOptions{
-			ClientId:          agent.AzureOptions.ClientID,
-			IsClientSecretSet: agent.AzureOptions.ClientSecret != "",
```
BupycHuk, line 179: we still need this field to be shown

managed/services/management/agent.go
BupycHuk: @JiriCtvrtka you said that we don't encrypt values if they are nil, so I think these 3 ifs should work as it used to, right?

BupycHuk (changes requested):
let's not update API yet.

JiriCtvrtka:
@BupycHuk about https://github.com/percona/pmm/pull/3002#discussion_r1671869844.

These fields are not pointers, so we saving them as empty string. Due to this it is encrypted and then it is not considered as not set password. Applies for

JiriCtvrtka:
Agreed changes:
1. No decrypt after encrypt agent when adding agent. Instead I removed exposing of username, aws_access_key.
2. Added ENV variable to provide custom path for encryption key.
3. Empty values are NOT encrypted.

api/serverpb/server.proto
```diff
@@ -169,6 +169,8 @@ message Settings {
   bool enable_access_control = 21;
   // Default Access Control role ID for new users.
   uint32 default_role_id = 22;
+  // Contains all encrypted tables in format 'db.table.column'.
+  repeated string encrypted_items = 23;
```
BupycHuk, line 173: we don't need to expose this value to public

BupycHuk (approved):

managed/models/database.go
```diff
@@ -1059,12 +1062,85 @@ func SetupDB(ctx context.Context, sqlDB *sql.DB, params SetupDBParams) (*reform.
 		return nil, errCV
 	}
 
-	if err := migrateDB(db, params); err != nil {
+	columnsToEncrypt := []encryption.Column{
```
BupycHuk, line 1065:
```suggestion
	agentColumnsToEncrypt := []encryption.Column{
```

ademidoff (approved):

PR #3014, opened 2024-06-04
PMM-13145 Certs Fix - Backward compatibility for older clients.

JiriCtvrtka:
PMM-13145

Link to the Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/3636

It passed pmm2-upgrade-tests checks now.

### DIFF
--- a/managed/models/agent_model.go
+++ b/managed/models/agent_model.go
@@ -346,7 +346,7 @@ func (s *Agent) DSN(service *Service, dsnParams DSNParams, tdp *DelimiterPair, p
 			// Skip verify for "custom" is handled on pmm-agent side.
 			switch {
 			// Backward compatibility
-			case pmmAgentVersion != nil && s.TLSSkipVerify && pmmAgentVersion.Less(v2_42):
+			case s.TLSSkipVerify && (pmmAgentVersion == nil || pmmAgentVersion.Less(v2_42)):
 				cfg.Params["tls"] = skipVerify
 			case len(s.Files()) != 0:
 				cfg.Params["tls"] = "custom"
--- a/managed/models/agent_model.go
+++ b/managed/models/agent_model.go
@@ -73,7 +73,7 @@ const (
 	VMAgentType                         AgentType = "vmagent"
 )
 
-var v2_42_0 = version.MustParse("2.42.0-0")
+var v2_42 = version.MustParse("2.41.99")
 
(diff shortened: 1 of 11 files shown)


managed/models/agent_model.go
```diff
@@ -73,6 +73,8 @@ const (
 	VMAgentType                         AgentType = "vmagent"
 )
 
+var v2_42 = version.MustParse("2.41.99")
```
ademidoff, line 76: I'm seeing different patterns here and [there](https://github.com/percona/pmm/blob/6aedc981c283edc91b677501a2b20890fa3bf0af/managed/services/agents/mongodb.go#L41-L43).

I wonder if we should stick to one standard, i.e. `var v2_42 = version.MustParse("2.42.0-0")`, which seems to be easier to read?

BupycHuk, line 76:
```suggestion
var v2_42 = version.MustParse("2.42.0-0")
```

managed/models/agent_model.go
```diff
@@ -340,8 +342,12 @@ func (s *Agent) DSN(service *Service, dsnParams DSNParams, tdp *DelimiterPair) s
 		cfg.Params = make(map[string]string)
 		if s.TLS {
 			// It is mandatory to have "custom" as the first case.
+			// Except case for backward compatibility.
 			// Skip verify for "custom" is handled on pmm-agent side.
 			switch {
+			// Backward compatibility
+			case pmmAgentVersion != nil && s.TLSSkipVerify && pmmAgentVersion.Less(v2_42):
```
BupycHuk, line 349: when we can have pmmAgentVersion nil?
```suggestion
			case s.TLSSkipVerify && (pmmAgentVersion == nil || pmmAgentVersion.Less(v2_42)):
```

ademidoff (approved):

managed/models/agent_model.go
JiriCtvrtka: @BupycHuk In tests https://github.com/percona/pmm/pull/3014/files#diff-e8c710b75bb00d5db0460e38961c6d4bd9c63fa549791eb9d091057a387f9b22R281 since there is new param. Or you want to extend them with specific versions?

managed/models/agent_model.go
JiriCtvrtka: Updated as you and Nurlan suggested.

PR #3026, opened 2024-06-12
PMM-12899 Use go mod cache in actions

ademidoff:
PMM-12899

Please check the [motivation](https://perconadev.atlassian.net/browse/PMM-12899) for using a docker image to persist cache before reviewing.


managed/Makefile
```diff
@@ -22,6 +22,13 @@ PMM_LD_FLAGS = -ldflags " \
 			-X 'github.com/percona/pmm/version.Branch=$(PMM_RELEASE_BRANCH)' \
 			"
 
+IS_ARM        := $(filter arm64,$(shell uname -m))
```
idoqo, line 25: Maybe we should include `aarch64` in the filter list too?

idoqo (approved):

managed/Makefile
ademidoff: Good idea!

Do you know what type of hardware we could get it from? I've checked my M2 processor and only got `arm64`.

ademidoff: I think it must be RaspberryPi, right? In such case I doubt it's a good idea to run PMM Server on a Pi machine, it'd not have sufficient resources to be performant.

JiriCtvrtka (approved):

.github/workflows/managed.yml
```diff
@@ -21,82 +21,56 @@ on:
...
-    name: Tests
+    name: Managed tests
     runs-on: ubuntu-22.04
     timeout-minutes: 30
+    permissions:
+      packages: write    
 
     env:
       PMM_SERVER_IMAGE: perconalab/pmm-server:3-dev-latest
+      PMM_CACHE_IMAGE: ghcr.io/percona/pmm:dev-cache
```
BupycHuk, line 32: Do we still use docker image for cache? Didn't Github Action work?

BupycHuk (approved):

PR #3058, opened 2024-07-05
PMM-13197 Update scripts to stop using deprecated repos

talhabinrizwan:
PMM-13197 Update scripts to stop using deprecated repos

Link to the Feature Build: SUBMODULES-3658

### DIFF
--- a/update/.devcontainer/install-dev-tools.sh
+++ b/update/.devcontainer/install-dev-tools.sh
@@ -15,7 +15,7 @@ sed -i '/nodocs/d' /etc/yum.conf
 
 # enable experimental repository with latest development packages
 sed -i'' -e 's^/release/^/experimental/^' /etc/yum.repos.d/pmm2-server.repo
-percona-release enable original testing
+percona-release enable pmm2-client testing
 
 RHEL=$(rpm --eval '%{rhel}')
 if [ "$RHEL" = "7" ]; then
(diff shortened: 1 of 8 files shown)


update/.devcontainer/install-dev-tools.sh
```diff
@@ -15,7 +15,7 @@ sed -i '/nodocs/d' /etc/yum.conf
 
 # enable experimental repository with latest development packages
 sed -i'' -e 's^/release/^/experimental/^' /etc/yum.repos.d/pmm2-server.repo
-percona-release enable original testing
+percona-release enable pmm2-client testing
```
ademidoff, line 18: This one won't work on RHEL 7 without an additional effort to make it work :(

ademidoff (changes requested):

update/.devcontainer/install-dev-tools.sh
BupycHuk: we don't develop on RHEL 7, so let's spend time on things that we might not use at all.

BupycHuk (approved):

update/.devcontainer/install-dev-tools.sh
BupycHuk: ok, now I get you. let's just drop pmm-2.0.0 from CI matrix

ademidoff: I have tested `percona-release enable pmm2-client testing` on all PMM Server versions up to 2.30.0 - I couldn't get it to work. Here is [the fix](percona-release enable pmm2-client testing) I hope.

ademidoff (approved):

PR #3060, opened 2024-07-07
PMM-13197 Fix update.yml workflow

ademidoff:
PMM-13197

**Issue**
Currently, the update workflow fails on some PMM image versions as yum can't reach RHEL repository mirrors.

**Solution**
In this PR we are working around the issue by changing the base URL to point to the archive.

Link to the Feature Build: N/A, since we are fixing the github workflow

### DIFF
--- a/update/.devcontainer/install-dev-tools.sh
+++ b/update/.devcontainer/install-dev-tools.sh
@@ -8,38 +8,54 @@ set -o errexit
 set -o xtrace
 
 # download (in the background) the same verison as used by PMM build process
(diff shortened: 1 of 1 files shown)


BupycHuk (approved):
interesting why 2.16.0 didn't fail?

ademidoff:
> interesting why 2.16.0 didn't fail?

Because it has a different set of mirrors inside the image compared to 2.0.0. (Disclaimer: I'm not quite sure I understood well the question :) )

talhabinrizwan (approved):

PR #3152, opened 2024-08-14
PMM-12451 MongoDB explain.

JiriCtvrtka:
PMM-12451

Link to the Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/3694


JiriCtvrtka:
1. Explain for MongoDB is now independent from Percona Tools.
2. We dropped logic for < 4.0.0. Now there is min version check
3. I included customer query into tests to avoid wrong bump in future.

JiriCtvrtka:
1. Explain for MongoDB is now independent from Percona Tools. So we have full control now.
2. We dropped logic for < 4.0.0. Now there is min version check.
3. I included customer query into tests to avoid problems with parsing date time, timestamp in future.

JiriCtvrtka:
1. Explain for MongoDB is now independent from Percona Tools. So we have full control now.
2. We dropped logic for < 4.0.0. Now there is min version check.
3. I included customer query into tests to avoid problems with parsing date time, timestamp in future.

Once Ivan will be back from vacation I am going to ask him if he can provide more queries to include them in tests.

JiriCtvrtka:
1. Explain for MongoDB is now independent from Percona Tools. So we have full control now.
2. We dropped logic for < 4.0.0. Now there is min version check.
3. I included customer query into tests to avoid problems with parsing date time, timestamp in future.

Once Ivan will be back from vacation I am going to ask him if he can provide more queries to include them in tests.

agent/runner/actions/mongodb_explain_action.go
```diff
@@ -89,28 +88,13 @@ func (a *mongodbExplainAction) Run(ctx context.Context) ([]byte, error) {
 	}
 	defer client.Disconnect(ctx) //nolint:errcheck
 
-	var eq proto.ExampleQuery
-
-	err = bson.UnmarshalExtJSON([]byte(a.params.Query), true, &eq)
-	if err != nil {
-		return nil, errors.Wrapf(err, "Query: %s", a.params.Query)
+	if notMinMongoDBVersion(ctx, client) {
```
BupycHuk, line 91: let's drop this check, it will simplify code. No need for additional check, we can just write in documentation that we don't support EOLed versions.

agent/runner/actions/mongodb_explain_action_test.go
```diff
@@ -33,6 +36,221 @@ import (
...
+	t.Helper()
+
+	big, err := rand.Int(rand.Reader, big.NewInt(27))
+	require.NoError(t, err)
+	id := strconv.FormatUint(big.Uint64(), 10)
+	ex, err := NewMongoDBExplainAction(id, 0, params, os.TempDir())
+	require.NoError(t, err)
+	res, err := ex.Run(ctx)
+	require.NoError(t, err)
+	require.NotEmpty(t, string(res))
```
BupycHuk, line 239: is there any standard thing that we can use to check that explain returns some logical response, not a random value.

JiriCtvrtka: Main problem with this is warning in MongoDB doc:
```
WARNING
MongoDB does not guarantee any specific output format from the [explain] command, even when using the Stable API.
```

https://www.mongodb.com/docs/manual/reference/command/explain/

Right now we are testing the possibility to parse a query and run it without error. This should mean "correct" implementation (for specific query) on our side. Everything else comes from Mongo itself. If we are talking about checking something specific in response I think "winningPlan" should be ideal. But to have "winningPlan" not "EOF" we need to fulfil DB with data corresponding with every query in tests. I will verify this with Ivan, if it makes sense and what we should check. Once he will be back from vacation.

agent/runner/actions/mongodb_explain_action.go
```diff
@@ -89,28 +98,152 @@ func (a *mongodbExplainAction) Run(ctx context.Context) ([]byte, error) {
...
+			return command[:i]
+		}
+
+		return append(command[:i], command[i+1:]...)
+	}
+
+	return command
+}
+
+func fixReduceField(command bson.D) bson.D {
```
BupycHuk, line 208: Could you please add comments to each function why we are doing that?

BupycHuk (approved):
Please add release notes

agent/runner/actions/mongodb_explain_action.go
JiriCtvrtka: RN and comments added.

BupycHuk (approved):

JiriCtvrtka:
@catalinaadam Please check RN.

docs/release-notes/2.43.0.md
```diff
@@ -67,3 +67,7 @@ If you're looking to upgrade, you can easily [install the latest version of Perc
  
  While these vulnerabilities did not directly impact PMM's core functionality, fixing them enhanced the overall security of the PMM Server environment. 
 
+- [PMM-12451](https://perconadev.atlassian.net/browse/PMM-12451) and [PMM-13017](https://perconadev.atlassian.net/browse/PMM-13017)  - Fix issue with parsing JSON objects into correct data types. Aligning `explain` functionality with the official client.
+
+- [PMM-13071](https://perconadev.atlassian.net/browse/PMM-13071) - Fix issue with unsupported explain type - `INSERT`. You will get a message that `INSERT` is
```
catalinaadam, line 72:
```suggestion
- [PMM-13071](https://perconadev.atlassian.net/browse/PMM-13071) -  The **Explain** tab on the Query Analytics (QAN) page now properly handles unsupported MongoDB query types. For operations like `INSERT`, which don’t support explain functionality, you will now see a clear message saying that the operation is not explainable. This replaces the previous, confusing error message about duplicate BSON fields, offering more accurate feedback in QAN.
```

agent/runner/actions/mongodb_explain_action.go
```diff
@@ -89,28 +98,156 @@ func (a *mongodbExplainAction) Run(ctx context.Context) ([]byte, error) {
...
+				filter = command[0].Value
+			} else {
+				filter = command
+			}
+
+			return bson.D{
+				{Key: "find", Value: e.getCollection()},
+				{Key: "filter", Value: filter},
+			}, nil
+		}
```
ademidoff, line 127: I believe, this whole block is difficult to reason about when it comes to boolean logic, so it could be problematic to test it reliably.

Can we refactor it to make the logic a bit more straightforward?

JiriCtvrtka: I already refactored this one from Percona Tools.

Original one:
`	case "query":
		if len(cmd) == 0 {
			cmd = self.Query
		}

		// MongoDB 2.6:
		//
		// "query" : {
		//   "query" : {
		//
		//   },
		//	 "$explain" : true
		// },
		if _, ok := cmd.Map()["$explain"]; ok {
			cmd = bson.D{
				{"explain", ""},
			}
			break
		}

		if len(cmd) == 0 || cmd[0].Key != "find" {
			var filter interface{}
			if len(cmd) > 0 && cmd[0].Key == "query" {
				filter = cmd[0].Value
			} else {
				filter = cmd
			}

			coll := ""
			s := strings.SplitN(self.Ns, ".", 2)
			if len(s) == 2 {
				coll = s[1]
			}

			cmd = bson.D{
				{"find", coll},
				{"filter", filter},
			}
		} else {
			for i := range cmd {
				switch cmd[i].Key {
				// PMM-1905: Drop "ntoreturn" if it's negative.
				case "ntoreturn":
					// If it's non-negative, then we are fine, continue to next param.
					if cmd[i].Value.(int64) >= 0 {
						continue
					}
					fallthrough
				// Drop $db as it is not supported in MongoDB 3.0.
				case "$db":
					if len(cmd)-1 == i {
						cmd = cmd[:i]
					} else {
						cmd = append(cmd[:i], cmd[i+1:]...)
					}
				}
			}
		}`
		
		
I will try think about what else could be simplified/refactored.

ademidoff, line 127:
```suggestion
		if len(command) == 0 || command[0].Key != "find" {
			return bson.D{
				{Key: "find", Value: e.getCollection()},
				{Key: "filter", Value: command},
			}, nil
		}

		if len(command) != 0 && command[0].Key == "query" {
			return bson.D{
				{Key: "find", Value: e.getCollection()},
				{Key: "filter", Value: command[0].Value},
			}, nil
		}
```

ademidoff (commented):
I think this does the same, but is simpler to read.

ademidoff (approved):

PR #3154, opened 2024-08-15
PMM-13146 Fix v3 telemetry issues

ademidoff:
PMM-13146

Link to the Feature Build: SUBMODULES-3692


managed/utils/envvars/parser.go
```diff
@@ -173,7 +173,7 @@ func ParseEnvVars(envs []string) (*models.ChangeSettingsParams, []error, []strin
 		case "PMM_VM_URL":
 			_, err = url.Parse(v)
 			if err != nil {
-				err = fmt.Errorf("invalid value %q for environment variable %q", v, k)
+				errs = append(errs, fmt.Errorf("invalid value %q for environment variable %q", v, k))
```
ademidoff, line 176: @BupycHuk 
The `err` was not being used, hence the fix. Should we add `continue` here as well?

managed/services/telemetry/config.default.yml
```diff
@@ -24,45 +24,45 @@ telemetry:
       - metric_name: "pmm_server_alert_rules_number"
         value: 1
 
-  - id: PMMServerSTTEnabled
+  - id: PMMServerAdvisorsEnabled
     source: PMMDB_SELECT
-    query: (CASE WHEN stt->'stt_disabled' = 'false' THEN '1' ELSE '0' END) AS stt_enabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS stt
-    summary: "PMM Server Security Thread Tool feature enabled/disabled"
+    query: (CASE WHEN advisors->'enabled' = 'true' THEN '1' WHEN advisors->'enabled' = 'null' THEN '1' ELSE '0' END) AS advisors_disabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS advisors
```
BupycHuk, line 29: seems like it can be simplified to
```suggestion
    query: (CASE WHEN advisors->'enabled' = 'false' THEN '0' ELSE '1' END) AS advisors_disabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS advisors
```

managed/services/telemetry/config.default.yml
```diff
@@ -24,45 +24,45 @@ telemetry:
...
-        column: "stt_enabled"
-
-  - id: PMMServerAdvisorsDisabled
-    source: PMMDB_SELECT
-    query: (CASE WHEN advisors->'enabled' = 'true' THEN '0' ELSE '1' END) AS advisors_disabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS advisors
-    summary: "PMM Advisors Disabled"
-    data:
-      - metric_name: "pmm_server_advisors_disabled"
-        column: "advisors_disabled"
+      - metric_name: "pmm_server_advisors_enabled"
```
BupycHuk, line 32: do we really need to rename metric name? Was it discussed with @rnovikovP?

managed/services/telemetry/config.default.yml
BupycHuk: the same everywhere below

managed/services/telemetry/config.default.yml
```diff
@@ -92,14 +92,12 @@ telemetry:
     summary: "Plugins on the MySQL instance"
     data:
       - metric_name: "1"
-        label: "library"
-      - metric_name: "2"
```
BupycHuk, line 96: why library was removed?

BupycHuk: if response is MySQL version based then shouldn't it be more complex query?

managed/services/telemetry/config.default.yml
```diff
@@ -951,12 +948,19 @@ telemetry:
       - metric_name: "postgresql_db_count"
         value: 1
 
-  - id: PMMServerHAEnabled
+  - id: PMMServerTelementryEnabled
```
BupycHuk, line 951: If we receive any metric then telemetry is enabled, isn't it?
```suggestion
  - id: PMMServerTelemetryEnabled
```

managed/services/telemetry/config.default.yml
```diff
@@ -951,12 +948,19 @@ telemetry:
       - metric_name: "postgresql_db_count"
         value: 1
 
-  - id: PMMServerHAEnabled
```
BupycHuk, line 954: What is the reason to remove this metric?

managed/utils/envvars/parser.go
BupycHuk: Continue isn't required, because we don't have any action after

managed/services/telemetry/config.default.yml
ademidoff: > why library was removed?

`library` is no longer available from the metrics.

managed/services/telemetry/config.default.yml
ademidoff: Oh, I didn't realize it was renamed to `PMM_TEST_HA_ENABLE`, so I put it back along with the new name.

docs/process/v2_to_v3_environment_variables.md
```diff
@@ -9,7 +9,7 @@ Below is a list of affected variables and their new names.
 | `DISABLE_UPDATES`                             | `PMM_ENABLE_UPDATES`                       |                                                              |
 | `DISABLE_TELEMETRY`                           | `PMM_ENABLE_TELEMETRY`                     |                                                              |
 | `PERCONA_PLATFORM_API_TIMEOUT`                | `PMM_DEV_PERCONA_PLATFORM_API_TIMEOUT`     |                                                              |
-| `DISABLE_BACKUP_MANAGEMENT`                   | `PMM_DISABLE_BACKUP_MANAGEMENT`            |                                                              |
+| `DISABLE_BACKUP_MANAGEMENT`                   |                                            | Removed in PMM v3, use `PMM_ENABLE_BACKUP_MANAGEMENT`        |
```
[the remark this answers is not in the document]
ademidoff: Done.

managed/services/telemetry/config.default.yml
ademidoff: Makes sense! Done.

managed/services/telemetry/config.default.yml
ademidoff: Roma was Ok'ish with that :)

managed/services/telemetry/config.default.yml
ademidoff: My bad )

I realized there is no harm if the config has a metric that can potentially be absent - it will still handle the payload without faltering.

I put it bask and added a test.

managed/services/telemetry/config.default.yml
```diff
@@ -51,18 +43,26 @@ telemetry:
...
-    summary: "PMM Server Access Control feature enabled/disabled"
+    summary: "PMM Server Access Control feature enabled"
     data:
       - metric_name: "pmm_server_access_control_enabled"
         column: "access_control_enabled"
 
-  - id: PMMServerUpdatesDisabled
+  - id: PMMServerUpdatesEnabled
+    source: PMMDB_SELECT
+    query: (CASE WHEN updates->'enabled' = 'true' THEN '1' ELSE '0' END) AS updates_disabled FROM settings s, jsonb_extract_path(s.settings, 'updates') AS updates
```
BupycHuk, line 53:
```suggestion
    query: (CASE WHEN updates->'enabled' = 'true' THEN '1' ELSE '0' END) AS updates_enabled FROM settings s, jsonb_extract_path(s.settings, 'updates') AS updates
```

managed/services/telemetry/config.default.yml
```diff
@@ -24,25 +24,17 @@ telemetry:
       - metric_name: "pmm_server_alert_rules_number"
         value: 1
 
-  - id: PMMServerSTTEnabled
+  - id: PMMServerAdvisorsEnabled
     source: PMMDB_SELECT
-    query: (CASE WHEN stt->'stt_disabled' = 'false' THEN '1' ELSE '0' END) AS stt_enabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS stt
-    summary: "PMM Server Security Thread Tool feature enabled/disabled"
+    query: (CASE WHEN advisors->'enabled' = 'false' THEN '0' ELSE '1' END) AS advisors_disabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS advisors
```
BupycHuk, line 29:
```suggestion
    query: (CASE WHEN advisors->'enabled' = 'false' THEN '0' ELSE '1' END) AS advisors_enabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS advisors
```
I'm almost sure that column name here should be the same as in a list below

ademidoff: Thanks, fixed.

BupycHuk (approved):

PR #3168, opened 2024-08-21
PMM-13317 Remove clickhouse exporter, update the guide

ademidoff:
PMM-13317

Link to the Feature Build: SUBMODULES-3697

Other PRs:
- https://github.com/Percona-Lab/pmm-submodules/pull/3697


managed/models/database.go
```diff
@@ -1253,12 +1250,6 @@ func setupFixture1(q *reform.Querier, params SetupDBParams) error {
 	return nil
 }
 
-func setupFixture2(q *reform.Querier, username, password string) error { //nolint:revive
```
ademidoff, line 1256: Removing it since it's been there for very long, but we can always restore if needed.

PR #3176, opened 2024-08-23
Migrate to buf schema v2 and move api related tasks to dedicated Makefile

artemgavrilov:
PMM-7

This PR migrates `buf` configuration to v2. It allowed to move all `api` related tasks to dedicated Makefile.

P.S. I was trying to find a way to use some plugins only for part of the modules, but unfortunately it's not possible for now.

https://buf.build/docs/migration-guides/migrate-v2-config-files

Same for PMM v3: https://github.com/percona/pmm/pull/3178


Makefile.include
```diff
@@ -29,9 +28,9 @@ release:              ## Build release versions of all components
 	make -C qan-api2 release
 
 gen: clean            ## Generate files
-	make gen-api
-	make gen-alertmanager
-	make clean_swagger
+	make -C api gen
+	make -C api gen-alertmanager
+	make -C api clean_swagger
```
ademidoff, line 33: Can we rename it to `clean-swagger`? It looks inconsistent :|

ademidoff (commented):
Great stuff, but I'm wondering if it is really worth it to contribute such a change to v2 versus v3? 

After we release 2.43.0, the whole team should completely switch their focus to getting v3 out the door.

artemgavrilov:
> Great stuff, but I'm wondering if it is really worth to contribute such a change to v2 versus v3?
> 
> After we release 2.43.0, the whole team should completely switch their focus to getting v3 out the door.

I can provide same changes for `v3` as well, or just for `v3`.

Makefile.include
artemgavrilov: Aha, done

ademidoff:
> > Great stuff, but I'm wondering if it is really worth to contribute such a change to v2 versus v3?
> > After we release 2.43.0, the whole team should completely switch their focus to getting v3 out the door.
> 
> I can provide same changes for `v3` as well, or just for `v3`.

OK, there is seemingly no harm for v3, so let's do both :)

Thanks for contributing, bro!

ademidoff (approved):

BupycHuk (approved):

JiriCtvrtka (approved):

PR #3200, opened 2024-09-12
PMM-13348 Fix status presentation.

BupycHuk:
PMM-13348

Link to the Feature Build: SUBMODULES-3708

- [ ] API Docs updated

### DIFF
--- a/admin/commands/list.go
+++ b/admin/commands/list.go
@@ -65,7 +65,9 @@ func (a listResultAgent) NiceAgentStatus() string {
 	if res == "" {
 		res = "unknown" //nolint:goconst
 	}
-	res = cases.Title(language.English).String(strings.ToLower(res))
+	res = strings.ToLower(res)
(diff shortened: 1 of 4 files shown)


admin/commands/list.go
```diff
@@ -65,7 +65,9 @@ func (a listResultAgent) NiceAgentStatus() string {
 	if res == "" {
 		res = "unknown" //nolint:goconst
 	}
-	res = cases.Title(language.English).String(strings.ToLower(res))
+	res = strings.ToLower(res)
+	res = cases.Title(language.English).String(res)
+	res = strings.ReplaceAll(res, "_", " ")
```
JiriCtvrtka, line 70: Maybe it would be good to extract this into method since it is used in two places. With that we can ensure always same "formatting" in both methods.

JiriCtvrtka (approved):

idoqo (approved):

PR #3204, opened 2024-09-18
PMM-7 run pmm client for api tests

BupycHuk:
PMM-7

- [ ] API Docs updated


api-tests/docker-compose.yml
```diff
@@ -36,13 +36,16 @@ services:
       --sql-mode="ANSI_QUOTES"
       --performance-schema --innodb_monitor_enable=all
       --slow_query_log --slow_query_log_file=/mysql/slowlogs/slow.log --long_query_time=0
-    ports:
-      - 127.0.0.1:3306:3306
```
ademidoff, line 40: Do we imply that the ports are already exposed in the parent image, right?

BupycHuk: I don't think that we need these ports to be exposed outside of docker network

BupycHuk: yes, pmm-client will be able to connect to these instances even without these ports being exposed.

ademidoff (approved):

JiriCtvrtka (approved):

PR #3221, opened 2024-10-01
PMM-13315 Service account max length.

JiriCtvrtka:
PMM-13315

https://github.com/Percona-Lab/pmm-submodules/pull/3722


api-tests/server/auth_test.go
```diff
@@ -617,3 +620,11 @@ func deleteServiceToken(t *testing.T, serviceAccountID, serviceTokenID int) {
 
 	require.Equalf(t, http.StatusOK, resp.StatusCode, "failed to delete service token, status code: %d, response: %s", resp.StatusCode, b)
 }
+
+func sanitizeSAName(name string) string {
```
ademidoff, line 624: I wonder if we can't keep and then import this method from `github.com/percona/pmm/managed/utils/strings` package?

JiriCtvrtka: It is to avoid dependencies between managed and API tests. To keep them separate. We doing this in more places. It was discussed with Nurlan.

managed/services/grafana/client.go
```diff
@@ -672,15 +673,27 @@ type serviceToken struct {
 	Role string `json:"role"`
 }
 
+// Max length of service account name is 190 chars (limit in Grafana Postgres DB).
+// However, prefix added by grafana is counted too. Prefix is sa-{orgID}-.
+// Bare minimum is 5 chars reserved (orgID is <10, like sa-1-) and could be more depends
+// on orgID number. Let's reserve 10 chars. It will cover almost one million orgIDs.
+// Sanitizing, ensure its length by hashing postfix when length is exceeded.
+// MD5 is used because it has fixed length 32 chars.
+func sanitizeSAName(name string) string {
```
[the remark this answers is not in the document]
JiriCtvrtka: Not sure if it belongs to utils when its related only to service accounts, which are part of Grafana.

idoqo (approved):

BupycHuk (approved):

PR #3230, opened 2024-10-09
PMM-13057 New telemetry endpoint.

JiriCtvrtka:
PMM-13057

Link to the Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/3736


BupycHuk (approved):

managed/services/telemetry/datasource_envvars.go
```diff
@@ -20,7 +20,7 @@ import (
 	"context"
 	"os"
 
-	pmmv1 "github.com/percona-platform/saas/gen/telemetry/events/pmm"
+	genericv1 "github.com/percona/saas/gen/telemetry/generic"
```
ademidoff, line 23: I would call it `telemetryv1`, since `generic` doesn't tell much when you see it in the code.

JiriCtvrtka: Both (genericv1 and reporter) renamed to telemetryv1.

ademidoff (approved):

PR #3234, opened 2024-10-10
PMM-13409 Fix update status failure after upgrade

ademidoff:
PMM-13409

Link to the Feature Build: SUBMODULES-3732


build/ansible/pmm-docker/update.yml
```diff
@@ -170,3 +170,8 @@
       file:
         state: absent
         path: /usr/share/pmm-server/maintenance/maintenance.html
+
+    - name: Remove the file provisioned by 'getStatus'
+      file:
+        state: absent
+        path: /srv/pmm-update.json
```
BupycHuk, line 177: I don't think that we call it after upgrade. seems like we need to cleanup this file from leftovers of old UI upgrade process.

ademidoff: I think we are still running this playbook when building PMM - https://github.com/percona/pmm/blob/v3/build/docker/server/Dockerfile.el9#L32

However, as you pointed out, we may need to get rid of it in view of the new update workflow implemented in v3.

BupycHuk: I mean that we still have file creation/removals, readiness checks that was required for UI update, but not a case anymore

ademidoff: Removed.

idoqo (approved):

ui/src/pages/updates/update-log/UpdateLog.hooks.ts
```diff
@@ -17,7 +17,7 @@ export const useUpdateLog = (authToken: string) => {
 
       return getUpdateStatus({ authToken, logOffset });
     },
-    refetchInterval: 500,
+    refetchInterval: ({ state }) => (state.data?.done ? 0 : 500),
```
ademidoff, line 20: IMO `500ms` is too frequent, I'd recommend we set it to 1s, which should be good enough to not degrade the UX.

ademidoff: @matejkubinec FYI ^

matejkubinec: Makes sense, updated.

YashSartanpara1 (approved):

PR #3245, opened 2024-10-16
PMM-7 Update default RPM build vars

ademidoff:
PMM-7

This PR:
- updates the default RPM build variables after we stopped building RHEL7
- removes the redundant Dockerfile(s) for RHEL7-based `rpmbuild` images


build/scripts/build-submodules
```diff
@@ -1,8 +1,12 @@
...
 set -o errexit
 
-git submodule update --init --jobs 10
-git submodule status
+pmm_source="sources/pmm/src/github.com/percona/pmm"
+if [ -d "$pmm_source" ] && [ -z $(ls -A "$pmm_source") ]; then
+    git submodule update --init --jobs 10
+    git submodule status
+fi
+unset pmm_source
```
BupycHuk, line 9: what's the reason behind it?

ademidoff: The reason behind it is to avoid running `git submodule update...` twice after we merge https://github.com/Percona-Lab/jenkins-pipelines/pull/2830. It is what prevents us fro merging that ticket.

BupycHuk: isn't it better to remove these two lines from that repo?

ademidoff: No, I actually had to add them back, otherwise the pipeline fails because `${PATH_TO_SCRIPTS}/build-submodules` does not resolve before we update the submodules.

In summary, `git submodule update...` has to take place before we pass the control over to the build script.

BupycHuk: checking for existense of that path makes no sense, because this script in that path, so it won't be called at all

ademidoff: True, but the same script will run if you build it locally, so it still make a bit of sense :)

BupycHuk: the same script running locally will skip these two lines, right?

ademidoff: The same script running locally will update the submodules after a `shallow` clone, and skip it otherwise.

talhabinrizwan (approved):

build/scripts/build-submodules
BupycHuk: why doesn't it work on jenkins?

ademidoff: Sorry, not sure I understand the last question.

ademidoff: Can you give a more detailed context as to what doesn't work on Jenkins?

BupycHuk: we usually run this script from the root of submodules, right?
This script is located in sources/pmm/src/github.com/percona/pmm/build/scripts/build-submodules, and if we execute this script it means that sources/pmm/src/github.com/percona/pmm exists. So `if [ -d "$pmm_source" ] && [ -z $(ls -A "$pmm_source") ]; ` check will always be false and these two lines will always be skipped. Which means that on local build we might have outdated submodules

ademidoff: OK, I see your point now. Yeah, building PMM locally needs a totally different approach, and this will probably not work. OK, let's remove these and move forward.

BupycHuk (approved):

PR #3319, opened 2024-11-18
PMM-13553 Revert the upgrade of kong

ademidoff:
PMM-13533

Link to the Feature Build: SUBMODULES-3771

- [ ] API Docs updated

### DIFF
--- a/build/ansible/roles/pmm-images/tasks/main.yml
+++ b/build/ansible/roles/pmm-images/tasks/main.yml
@@ -157,13 +157,6 @@
     option: autostart
     value: "false"
 
-- name: Disable grafana
-  ini_file:
-    path: /etc/supervisord.d/grafana.ini
-    section: program:grafana
-    option: autostart
-    value: "false"
-
 # During build time, this will be the first start of supervisord.
 - name: Start supervisord
   shell: supervisord -c /etc/supervisord.conf &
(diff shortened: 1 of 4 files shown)


JiriCtvrtka (approved):

build/ansible/roles/pmm-images/tasks/main.yml
```diff
@@ -157,6 +157,13 @@
     option: autostart
     value: "false"
 
+- name: Disable grafana
+  ini_file:
+    path: /etc/supervisord.d/grafana.ini
+    section: program:grafana
+    option: autostart
+    value: "false"
```
BupycHuk, line 165: let's enable it after initialization

BupycHuk: it will slowdown first initialization and we will loose time that we win by running initialization during build time

ademidoff: Btw, the block above is not even necessary given the current initial `grafana.ini` [config](https://github.com/percona/pmm/blob/v3/build/ansible/roles/supervisord/files/grafana.ini#L13).

ademidoff: Moreover, setting `autostart:true` does not do anything on its own - the service will not restart without supervisord, which either needs to restart or to reread the config.

PR #3320, opened 2024-11-18
PMM-7 Fix grafana stop/start

ademidoff:
PMM-7

Link to the Feature Build: SUBMODULES-3773

### DIFF
--- a/build/ansible/roles/dashboards/tasks/main.yml
+++ b/build/ansible/roles/dashboards/tasks/main.yml
@@ -1,12 +1,11 @@
 ---
-- name: Stop and remove Grafana from Supervisor
-  shell: "supervisorctl {{ item }} grafana"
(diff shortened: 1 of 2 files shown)


BupycHuk (approved):

BupycHuk (approved):
we can't ensure that grafana stops, because pmm-managed might start it any moment. But it's good enough for now.

ademidoff:
> we can't ensure that grafana stops, because [pmm-managed](https://perconadev.atlassian.net/browse/PMM-managed) might start it any moment. But it's good enough for now.

Does `pmm-managed` launch grafana? I think it doesn't, it only generates `grafana.ini` config. Grafana's lifecycle is managed by supervisord only I believe.

talhabinrizwan (approved):

PR #3344, opened 2024-11-26
PMM-13567 fix ami ovf build

talhabinrizwan:
PMM-13567

Link to the Feature Build: SUBMODULES-3781

### DIFF
--- a/build/packer/ansible/roles/podman-setup/tasks/main.yml
+++ b/build/packer/ansible/roles/podman-setup/tasks/main.yml
@@ -24,13 +24,12 @@
     policy: targeted
     state: permissive
 
-- name: Update podman.conf for ami and ovf
+- name: Update podman.conf
   lineinfile:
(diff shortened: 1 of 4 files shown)


build/packer/ansible/roles/podman-setup/tasks/main.yml
```diff
@@ -24,6 +24,14 @@
     policy: targeted
     state: permissive
 
+- name: Update podman.conf for ami and ovf
+  lineinfile:
+    path: /usr/lib/tmpfiles.d/podman.conf
+    regexp: '^x /tmp/storage-run-\*'
+    line: 'R! /tmp/storage-run-*'
+    backrefs: yes
+  when: pmm_distribution_method in ['ovf', 'ami']
```
ademidoff, line 33: This role can only run for `ovf/ami`, so the `when` statement is not necessary.

BupycHuk (approved):

PR #3370, opened 2024-12-09
PMM-13597 Move `pmm-doc` v3 to monorepo

ademidoff:
PMM-13597


.github/workflows/documentation.yml
```diff
@@ -0,0 +1,64 @@
...
+          git config --global url."https://percona-platform-robot:${ROBOT_TOKEN}@github.com".insteadOf "https://github.com"
+          git config user.name "GitHub Action"
+          git config user.email "github-action@users.noreply.github.com"
+          git config user.password "${ROBOT_TOKEN}"
+          echo "GIT_USER=percona-platform-robot:${ROBOT_TOKEN}" >> $GITHUB_ENV
+
+      - name: Check for unused images
+        run: make doc-check-images
+
+      # Builds Material-themed static web site in 'publish' branch
```
BupycHuk, line 57: it should run only for main and v3 branches, but not for PRs

documentation/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+doc-build-pdf:			## Build documentation in PDF format
+	# @docker buildx build --platform=linux/amd64 --progress=plain -t perconalab/pmm-doc-md:latest -f documentation/resources/Dockerfile.build .
+	docker run --rm --platform=linux/amd64 -v $(DIR):/docs -w /build -e ENABLE_PDF_EXPORT=1 perconalab/pmm-doc-md:latest \
+		bash -c "cp -r /docs/* /build/ && \
+		ls -la /build && \
+		git init && \
+		git config user.email 'doc-team@percona.com' && \
+		git add --all && \
+		git commit -am 'Initial commit' && \
```
BupycHuk, line 31: do we know why all these steps are required?

catalinaadam (approved):

documentation/Makefile
ademidoff: Definitely. Mike is opinionated in that it expects the root directory to be a git repository. Problem is that we are not passing the whole project to mide, so this is a workaround to make it happy.

Nevertheless, if there is a better way to do it, we can certainly iterate.

.github/workflows/documentation.yml
ademidoff: Absolutely! It was necessary to enforce the workflow for this `migration`, but otherwise we don't need it. 

Removed. Also, the `main` branch has its own workflow, so I removed it as well.

PR #3464, opened 2025-01-16
PMM-13671 Update client image labels

talhabinrizwan:
PMM-13671

Link to the Feature Build: SUBMODULES-3819


build/docker/client/Dockerfile
```diff
@@ -12,11 +12,13 @@ FROM redhat/ubi9-micro
...
-LABEL org.opencontainers.image.title Percona Monitoring and Management Client
-LABEL org.opencontainers.image.vendor Percona
-LABEL org.opencontainers.image.version ${VERSION}
+LABEL org.opencontainers.image.created=${BUILD_DATE} \
+      org.opencontainers.image.licenses=Apache-2.0 \
+      org.opencontainers.image.title="Percona Monitoring and Management Client" \
+      org.opencontainers.image.vendor="Percona, LLC" \
+      org.opencontainers.image.version=${VERSION} \
+      vendor="Percona, LLC" \
+      maintainer="Percona, LLC"
```
ademidoff, line 21: IMO we need to update more labels. 

How about those:
- io.k8s.description
- io.k8s.display-name
- name
- summary
- url
- version

I believe we shouldn't leave them as they are now.

talhabinrizwan: How about skipping the `version` field since it will need to be updated with every release?

ademidoff: If we consider the current value meaningful ("version": "9.5"), then we can leave it. 

BTW, what does it refer to? Release version of UBI?

ademidoff: IMO the `version` tag should have the same value as `org.opencontainers.image.version`.

BupycHuk (approved):
We need to backport it to PMM 2 and we need to make similar changes to PMM Server dockerfile.

JiriCtvrtka (approved):

talhabinrizwan:
> Is failure of managed version test expected?

No

ademidoff (approved):

PR #3515, opened 2025-01-28
PMM 13626 install perl into pmm client

talhabinrizwan:
PMM-13626

Link to the Feature Build: SUBMODULES-3831


build/docker/client/Dockerfile
```diff
@@ -1,12 +1,38 @@
-FROM redhat/ubi9-minimal AS users
+FROM redhat/ubi9 AS builder
 
-RUN microdnf install shadow-utils jq tar -y
+RUN dnf install -y shadow-utils wget jq tar https://repo.percona.com/yum/percona-release-latest.noarch.rpm \
```
ademidoff, line 3: Do we really need `wget` given that there is already `curl`, which is much more feature-rich?

ademidoff: How is it supposed to be used?

talhabinrizwan: Good point. I'll remove it.

BupycHuk (approved):

JiriCtvrtka (approved):

PR #3587, opened 2025-02-06
Fix the `easy install` script to point to v2 on 'main'

ademidoff:
PMM-7

### DIFF
--- a/documentation/docs/setting-up/server/easy-install.md
+++ b/documentation/docs/setting-up/server/easy-install.md
@@ -6,12 +6,12 @@
 ## Linux or macOS
 Using `curl`:
 ```sh
-curl -fsSL https://www.percona.com/get/pmm | /bin/bash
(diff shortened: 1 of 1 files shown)


documentation/docs/setting-up/server/easy-install.md
```diff
@@ -6,24 +6,24 @@
 ## Linux or macOS
```
catalinaadam, line 6: Do we also need to change the text in the note above? 
"You can download and check get-pmm2.sh before running it from our [github]"

ademidoff: Maybe not... I believe it's still named [get-[mm.sh](https://github.com/percona/pmm/blob/main/get-pmm.sh), no?

catalinaadam (approved):

BupycHuk (approved):

PR #3595, opened 2025-02-09
HD-26283 Fix removal of default vagrant user for OVA

talhabinrizwan:
HD-26283

### DIFF
--- a/build/ansible/roles/ami-ovf/tasks/main.yml
+++ b/build/ansible/roles/ami-ovf/tasks/main.yml
@@ -68,3 +68,37 @@
 
 - name: PMM                        | Delete Azure user
   shell: cd /tmp; nohup sh -c "trap '/usr/sbin/waagent -force -deprovision+user && sync' EXIT; sleep 600" </dev/null >/dev/null 2>&1 &
+
+- name: Lock vagrant user
+  ansible.builtin.user:
+    name: vagrant
+    password_lock: true
+  when: ansible_virtualization_type == "virtualbox"
+
+- name: Configure systemd service to remove vagrant user
+  block:
+    - name: Create systemd service file
+      copy:
+        dest: /etc/systemd/system/remove-vagrant.service
+        content: |
+          [Unit]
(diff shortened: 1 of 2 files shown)


EvgeniyPatlan (approved):

build/ansible/roles/ami-ovf/tasks/main.yml
```diff
@@ -68,3 +68,37 @@
 
 - name: PMM                        | Delete Azure user
   shell: cd /tmp; nohup sh -c "trap '/usr/sbin/waagent -force -deprovision+user && sync' EXIT; sleep 600" </dev/null >/dev/null 2>&1 &
+
+- name: Lock vagrant user
+  ansible.builtin.user:
+    name: vagrant
+    password_lock: true
+  when: ansible_virtualization_type == "virtualbox"
+
```
ademidoff, line 77:
```suggestion
```

That seems to be redundant now that the user is being deleted.

talhabinrizwan: I believe this should be in place as a failsafe mechanism. It ensures that the Vagrant user's SSH access is already locked, and the other block will only execute the first time the system starts. Additionally, if, for some reason, we're unable to delete the user, it will remain locked as a precaution.

ademidoff: IMO, if for some reason we are unable to delete the user, the pipeline should fail. We probably don't want to test two scenarios, but only one, where the user is not present.

ademidoff: I don't quite understand why would we want to continue shipping the `vagrant` with the image?

talhabinrizwan: We're not deleting the `vagrant` user in our pipeline. Instead, we're creating/enabling a `systemd` service that will remove it on the first boot. The reason for this approach is that we couldn't delete the `vagrant` user while the pipeline was running since the OVA image is created with the `vagrant` user.

PR #3596, opened 2025-02-09
HD-26283 Fix removal of default vagrant user for OVA

talhabinrizwan:
HD-26283

### DIFF
--- a/build/packer/ansible/roles/ami-ovf/tasks/main.yml
+++ b/build/packer/ansible/roles/ami-ovf/tasks/main.yml
@@ -11,6 +11,12 @@
 - name: PMM                        | Delete Azure user
   shell: cd /tmp; nohup sh -c "trap '/usr/sbin/waagent -force -deprovision+user && sync' EXIT; sleep 600" </dev/null >/dev/null 2>&1 &
 
+- name: Lock vagrant user
+  ansible.builtin.user:
+    name: vagrant
+    password_lock: true
+  when: ansible_virtualization_type == "virtualbox"
+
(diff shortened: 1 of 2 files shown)


EvgeniyPatlan (approved):

build/packer/ansible/roles/ami-ovf/tasks/main.yml
```diff
@@ -10,3 +10,37 @@
 
 - name: PMM                        | Delete Azure user
   shell: cd /tmp; nohup sh -c "trap '/usr/sbin/waagent -force -deprovision+user && sync' EXIT; sleep 600" </dev/null >/dev/null 2>&1 &
+
+- name: Lock vagrant user
+  ansible.builtin.user:
+    name: vagrant
+    password_lock: true
+  when: ansible_virtualization_type == "virtualbox"
+
```
ademidoff, line 19:
```suggestion
```

If the user is being deleted, we don't need this task.

talhabinrizwan: I believe this should be in place as a failsafe mechanism. It ensures that the Vagrant user's SSH access is already locked, and the other block will only execute the first time the system starts. Additionally, if, for some reason, we're unable to delete the user, it will remain locked as a precaution.

ademidoff: I believe we need to make sure the user is deleted to avoid shipping it with the image. Otherwise, the pipeline should fail.

PR #3634, opened 2025-02-12
PMM-7 Bump up the runner to latest

ademidoff:
PMM-7

We are bumping up the dependabot's runner due to its forthcoming deprecation:

"The Ubuntu 20.04 runner image will be fully unsupported by April 1, 2025"

After the deprecation deadline, the actions will be blocked.

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -8,7 +8,7 @@ permissions:
 jobs:
   dependabot:
     name: Enable auto-merge
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     if: ${{ github.actor == 'dependabot[bot]' }}
     steps:
       - name: Dependabot metadata
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -8,7 +8,7 @@ permissions:
 jobs:
   dependabot:
     name: Enable auto-merge
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     if: ${{ github.actor == 'dependabot[bot]' }}
     steps:
       - name: Dependabot metadata


BupycHuk (approved):

.github/workflows/dependabot.yml
```diff
@@ -8,7 +8,7 @@ permissions:
 jobs:
   dependabot:
     name: Enable auto-merge
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
BupycHuk, line 11: shouldn't it be 22.04 like in other jobs?

ademidoff: Franky, I'd migrate them all to `latest` and forget about it.

BupycHuk: sometimes it might lead to unexpected failures

ademidoff: Really? Any examples?

ademidoff: I'm sure dependabot will just work, because both the runner and the actor are from the same vendor :)

ademidoff: Btw, even if we pin the version, there is never a guarantee, since runners also get updated regularly without us noticing it... but they still mostly work 🤞

BupycHuk: https://github.com/actions/runner-images/issues/10788
https://github.com/actions/runner-images/issues/2234
https://exercism.org/docs/building/github/gha-best-practices#h-pin-test-runners-to-version
I'm sure that you are right about this case and that action will keep working, but behavior might change when new ubuntu will come with new node version and this action will still be using old node version.

matejkubinec (approved):

.github/workflows/dependabot.yml
ademidoff: Sorry, the examples above are irrelevant IMO. Again, the runner and the bot are maintained by the same company.

On the other hand this problem is so minor, that I don't see why we should sweat over it more than that, so I concur.

PR #3662, opened 2025-02-18
PMM-9288 Extended profiler.

JiriCtvrtka:
PMM-9288

Percona Toolkit: https://github.com/percona/percona-toolkit/pull/924
Grafana Dashboards: https://github.com/percona/grafana-dashboards/pull/1656
FB: https://github.com/Percona-Lab/pmm-submodules/pull/3856


agent/agents/mongodb/internal/profiler/aggregator/aggregator.go
```diff
@@ -272,38 +276,80 @@ func (a *Aggregator) createResult(_ context.Context) *report.Result {
...
 		bucket.Mongodb.MDocsReturnedMax = float32(v.Returned.Max)
 		bucket.Mongodb.MDocsReturnedMin = float32(v.Returned.Min)
 		bucket.Mongodb.MDocsReturnedP99 = float32(v.Returned.Pct99)
 		bucket.Mongodb.MDocsReturnedSum = float32(v.Returned.Total)
 
-		bucket.Mongodb.MDocsScannedCnt = float32(v.Count) // TODO: Check is it right value
-		bucket.Mongodb.MDocsScannedMax = float32(v.Scanned.Max)
-		bucket.Mongodb.MDocsScannedMin = float32(v.Scanned.Min)
-		bucket.Mongodb.MDocsScannedP99 = float32(v.Scanned.Pct99)
-		bucket.Mongodb.MDocsScannedSum = float32(v.Scanned.Total)
```
JiriCtvrtka, line 295: Deprecated

agent/agents/mongodb/internal/profiler/aggregator/aggregator.go
```diff
@@ -272,38 +276,80 @@ func (a *Aggregator) createResult(_ context.Context) *report.Result {
...
-		bucket.Mongodb.MResponseLengthCnt = float32(v.Count) // TODO: Check is it right value
+		bucket.Mongodb.MResponseLengthCnt = float32(v.ResponseLengthCount)
 		bucket.Mongodb.MResponseLengthMax = float32(v.ResponseLength.Max)
 		bucket.Mongodb.MResponseLengthMin = float32(v.ResponseLength.Min)
 		bucket.Mongodb.MResponseLengthP99 = float32(v.ResponseLength.Pct99)
 		bucket.Mongodb.MResponseLengthSum = float32(v.ResponseLength.Total)
 
 		bucket.Mongodb.MFullScanCnt = float32(v.CollScanCount)
-		bucket.Mongodb.MFullScanSum = float32(v.CollScanSum) / 1000
+		bucket.Mongodb.MFullScanSum = float32(v.CollScanCount) // Sum is same like count in this case
```
JiriCtvrtka, line 303: @igroene please check if in case of Full Scan (CollScan in Mongo) are Sum and Count values same.

igroene: what if we count the distinct number of collscans and use sum for the total number? would it be too complicated?

JiriCtvrtka: I'm not sure if I understood you correctly, but:
1. Currently, it is the total sum of the number of collection scans. Is this what you meant?
2. If you're asking about the total number of all stages, then yes, it would be complicated at this point. To achieve that, we would need to change the way data is collected, which could be done in a separate ticket.

Or did you mean something entirely different?

igroene: what I mean is, keep the SUM as the total number of recorded COLSCANS for that particular query shape. Use COUNT for the number of recorded COLSCANS of the particular query shape that have different values for the variables. Assuming we captured the following 4 queries:
```
db.testcol.find({a:101})
db.testcol.find({a:101})
db.testcol.find({a:200})
db.testcol.find({a:140})
```
assuming all the above are collscans, SUM would be 4 and COUNT would be 3. Makes sense?

JiriCtvrtka: I see. Yes, this should be possible. Let me adjust the code to make it work. Thanks for the explanation.

BupycHuk: @JiriCtvrtka current implementation should make it, no adjustment required

JiriCtvrtka: Thank you for verifying.

idoqo (approved):

BupycHuk (approved):

PR #3766, opened 2025-03-11
PMM 13626 Install perl into pmm client

talhabinrizwan:
PMM-13626

Link to the Feature Build: SUBMODULES-3863


BupycHuk (approved):

ademidoff (approved):
I think we can do better to avoid such issues in the future. I suggest we use the `--installroot` parameter of `dnf install` command. This gives at least two benefits:
1. We don't have to know the exact names of the binaries we need to copy from one stage to another
2. The resulting image will have much fewer layers, which are too many right now because of multiple `COPY --from` commands. That will allow to use one `COPY` command only.

idoqo (approved):

PR #3901, opened 2025-04-11
PMM-13700 created help center page

ioana-maniga:
https://perconadev.atlassian.net/browse/PMM-13700

image


ui/src/components/page/Page.tsx
```diff
@@ -19,9 +19,11 @@ export const Page: FC<PageProps> = ({ title, footer, children }) => {
   return (
     <Stack
       sx={{
+        '@media (min-width:1000px)': {
```
matejkubinec, line 22: Any reason not to use the [breakpoint helpers](https://mui.com/material-ui/customization/breakpoints/#default-breakpoints) from MUI?

ui/src/pages/help-center/HelpCenter.constants.ts
```diff
@@ -0,0 +1,110 @@
+export const cardIds = {
```
matejkubinec, line 1: We generally prefer upper snake case (eq. `CARD_IDS`) for constants.

ui/src/pages/help-center/HelpCenter.constants.ts
```diff
@@ -0,0 +1,110 @@
...
+      'From setup to troubleshooting, you’ll find step-by-step instructions, tips, and best practices to get the most out of PMM.',
+    buttons: [
+      {
+        buttonText: 'View docs',
+        target: '_blank',
+        url: 'https://per.co.na/pmm_documentation',
+        startIconName: '',
+      },
+    ],
+    borderColor: '#1486FF',
```
matejkubinec, line 29: Is there a color from the pallete (mui one) that can be used?

ui/src/pages/help-center/HelpCenter.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+    expect(
+      screen.queryByTestId(`help-card-${cardIds.pmmDump}`)
+    ).toBeInTheDocument();
+    expect(
+      screen.queryByTestId(`help-card-${cardIds.pmmLogs}`)
+    ).toBeInTheDocument();
+    expect(screen.queryAllByTestId(/^help-card-/).length).toEqual(6);
+  });
+
+  it('should not show pmm dump and pmm longs if user is not admin', () => {
```
matejkubinec, line 30: Could you add this check both for viewer and editor?

ui/src/pages/help-center/HelpCenter.tsx
```diff
@@ -0,0 +1,131 @@
...
+            xs: '1fr',
+            sm: '1fr',
+            md: 'repeat(2, 1fr)',
+          },
+          gap: 4,
+        }}
+      >
+        {CardsData.map(
+          (item) =>
+            shouldDisplayCard(item.id) && (
```
matejkubinec, line 70: I think it may be good to extract the card into it's own component. Then you could do a check like this:

```tsx
const Card: FC = (props) => {
 if (!shouldDisplayCard(id)) {
    return null;
 }

  return <Card .... />
}
```

ui/src/pages/help-center/HelpCenter.tsx
```diff
@@ -0,0 +1,131 @@
...
+                style={{
+                  borderTop: item.borderColor
+                    ? `solid 12px ${item.borderColor}`
+                    : 'none',
+                }}
+                key={item.id}
+                data-testid={`help-card-${item.id}`}
+              >
+                <CardContent style={{ paddingRight: 16, paddingLeft: 16 }}>
+                  <div
```
matejkubinec, line 81: You can use [Stack](https://mui.com/material-ui/react-stack/) from MUI to use as a flex container.

ui/src/pages/help-center/HelpCenter.tsx
```diff
@@ -0,0 +1,131 @@
...
+            md: 'repeat(2, 1fr)',
+          },
+          gap: 4,
+        }}
+      >
+        {CardsData.map(
+          (item) =>
+            shouldDisplayCard(item.id) && (
+              <Card
+                style={{
```
matejkubinec, line 72: Please prefer to use the [sx](https://mui.com/system/getting-started/the-sx-prop/) prop or `@emotion` library (either className or css prop - css prop might need additional setup)

If possible try to not inline style (directly styling the component through the `style` attribute and not css/css-framework)

ui/src/pages/help-center/HelpCenter.tsx
```diff
@@ -0,0 +1,131 @@
...
+                  <div style={{ display: 'flex', paddingTop: 16 }}>
+                    {item.buttons.map((button) => (
+                      <Button
+                        key={button.url}
+                        variant="outlined"
+                        size="small"
+                        style={{ marginRight: 8 }}
+                        startIcon={getButtonStartIcon(button.startIconName)}
+                        endIcon={button.target && <NorthEast />}
+                        onClick={() => {
```
matejkubinec, line 108: Please extract into a separate function (have a `handleClick` method or something like that)

Also from an accessibility perspective it isn't good practice to use buttons as links (https://www.w3schools.com/accessibility/accessibility_buttons_links.php)

You can use the `component` prop on MUI components (for example `component={Link}` where Link is from react-router - used for navigating within the app) or the `href` attribute for external links (should render an `a` (anchor) on the page)

ui/src/pages/help-center/HelpCenter.tsx
```diff
@@ -0,0 +1,131 @@
...
+                          if (button.target !== '') {
+                            window.open(
+                              button.url,
+                              button.target,
+                              'noopener,noreferrer'
+                            );
+                          } else {
+                            window.location.assign(button.url);
```
matejkubinec, line 116: For navigation within PMM use react router.

ui/src/pages/help-center/HelpCenter.tsx
```diff
@@ -0,0 +1,131 @@
...
+      !(
+        !user?.isPMMAdmin &&
+        (itemId === cardIds.pmmDump || itemId === cardIds.pmmLogs)
+      ),
+    [user]
+  );
+
+  return (
+    <Page title={Messages.pageTitle}>
+      <Box
```
matejkubinec, line 57: MUI has a [Grid](https://mui.com/material-ui/react-grid/) component which may simplify some things here.

ui/src/pages/help-center/HelpCenter.constants.ts
```diff
@@ -0,0 +1,110 @@
...
+  tips: 'tips',
+};
+
+export const startIcon = {
+  download: 'download',
+  map: 'map',
+};
+
+export const CardsData = [
+  {
```
matejkubinec, line 16: One thing you could add here is a boolean prop like `requiredRole` / `adminOnly` / `roles` (list of org roles), which I think would simplify the auth check.

ui/src/pages/help-center/HelpCenter.constants.ts
```diff
@@ -0,0 +1,110 @@
...
+    id: cardIds.pmmDocs,
+    title: 'PMM Documentation',
+    description:
+      'From setup to troubleshooting, you’ll find step-by-step instructions, tips, and best practices to get the most out of PMM.',
+    buttons: [
+      {
+        buttonText: 'View docs',
```
matejkubinec, line 23: Since this prop is is specific to buttons, I don't it needs the `button` prefix (can be either text/content)

ui/src/pages/help-center/HelpCenter.constants.ts
```diff
@@ -0,0 +1,110 @@
...
+  pmmLogs: 'pmm-logs',
+  tips: 'tips',
+};
+
+export const startIcon = {
+  download: 'download',
+  map: 'map',
+};
+
+export const CardsData = [
```
matejkubinec, line 15: Could you create a type for `HelpCard` something like:

```tsx
interface CardButton {
...
}

interface HelpCard {
   id: string;
   title: string;
   ...
   buttons: CardButton[];
}

Comment (ioana-maniga) [reply ui/src/pages/help-center/HelpCenter.constants.ts:15]: Done, in a separate types file.

File: ui/src/pages/help-center/HelpCenter.tsx (hunk 10 above)
Comment (ioana-maniga) [reply ui/src/pages/help-center/HelpCenter.tsx:19]: As we discussed, I kept the implementation with Box component

File: ui/src/pages/help-center/help-center-card/HelpCenterCard.tsx [hunk 14]
```diff
@@ -0,0 +1,99 @@
...
+  SaveAlt,
+  MapOutlined,
+} from '@mui/icons-material';
+import { KnowledgeBaseIcon } from 'icons';
+import { FC, ReactNode, useCallback } from 'react';
+import { CARD_IDS, START_ICON } from '../HelpCenter.constants';
+import { HelpCenterCardProps, CardButton } from './HelpCenterCard.types';
+
+export const HelpCenterCard: FC<HelpCenterCardProps> = (props) => {
+  const { shouldDisplayCard, card } = props;
```
matejkubinec, line 16: You can destruct them directly, e.q.:

```tsx
export const HelpCenterCard: FC<HelpCenterCardProps> = ({ shouldDisplayCard, card }) => {
...
}
```

ui/src/pages/help-center/HelpCenter.constants.ts
```diff
@@ -0,0 +1,116 @@
...
+  pmmLogs: 'pmm-logs',
+  tips: 'tips',
+};
+
+export const START_ICON = {
+  download: 'download',
+  map: 'map',
+};
+
+export const CardsData = [
```
matejkubinec, line 15: Could you also name this `CARD_DATA`

ui/src/pages/help-center/HelpCenter.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { FC, useCallback } from 'react';
+import { Messages } from './HelpCenter.messages';
+import { CardsData } from './HelpCenter.constants';
+import { useUser } from 'contexts/user';
+import { HelpCenterCard } from './help-center-card/HelpCenterCard';
+
+export const HelpCenter: FC = () => {
+  const { user } = useUser();
+
+  const shouldDisplayCard = useCallback(
```
matejkubinec, line 12: You can move this into the `HelpCenterCard` component like so:

```tsx
const HelpCenterCard: FC = ({ ...,card }) => {
  const { user } = useUser();
  const { adminOnly, ... } = card;
  
  if (adminOnly && !user.isPMMAdmin) {
    return null;
  }
}
```
Doing the check this way would prevent the additional passing of the `shouldDisplayCard` function down to the card components.


Another option (preffered) is to filter out the cards like so:

```ts
const { user } = useUser(); 
const cards = CARDS_DATA.filter(card => !card.adminOnly || user.isPMMAdmin);

// render cards as usual
```

This is probably a bit cleaner since it will not render nulls in place of the cards if the permissions don't match.

ui/src/pages/help-center/help-center-card/HelpCenterCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+      default:
+        return null;
+    }
+  }, []);
+
+  const onButtonClick = useCallback((button: CardButton) => {
+    if (button.target !== '') {
+      window.open(button.url, button.target, 'noopener,noreferrer');
+    } else {
+      window.location.assign(button.url);
```
matejkubinec, line 50: Just add a todo here to use react router once the grafana iframe is in place.

ui/src/pages/help-center/help-center-card/HelpCenterCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    >
+      <CardContent sx={{ paddingRight: '16px', paddingLeft: '16px' }}>
+        <Stack
+          flexDirection={'row'}
+          justifyContent={'flex-start'}
+          alignItems={'center'}
+          marginBottom={'8px'}
+        >
+          {getIcon(id)}
+          <Typography variant="h6" sx={{ marginLeft: getIcon(id) ? '8px' : '0px' }}>
```
matejkubinec, line 74: You can use [spacing](https://mui.com/system/spacing/#notation) and [short prop names](https://mui.com/system/properties/#properties-reference-table) from MUI for these values 

so for example `{ mr: 1 }` would get translated to `{ marginLeft: 8px }`

This also applies to other values in the component.

ui/src/pages/help-center/help-center-card/HelpCenterCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    <Card
+      sx={{
+        borderTop: borderColor ? `solid 12px ${borderColor}` : 'none',
+      }}
+      key={id}
+      data-testid={`help-card-${id}`}
+    >
+      <CardContent sx={{ paddingRight: '16px', paddingLeft: '16px' }}>
+        <Stack
+          flexDirection={'row'}
```
matejkubinec, line 68: You can omit the brackets (`{` / `}`) if the attribute/prop value is a string.

ui/src/pages/help-center/help-center-card/HelpCenterCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+          justifyContent={'flex-start'}
+          alignItems={'center'}
+          marginBottom={'8px'}
```
matejkubinec, line 71: I think you can replace this with `marginBottom={2}` mui should pick that up as 8px.

ui/src/pages/help-center/help-center-card/HelpCenterCard.types.ts
```diff
@@ -0,0 +1,20 @@
+export interface CardButton {
+  text: string;
+  target: string;
+  url: string;
+  startIconName: string;
+}
+
+export interface CardType {
```
matejkubinec, line 8: I think a better name would be `HelpCard`.

ui/src/pages/help-center/help-center-card/HelpCenterCard.types.ts
```diff
@@ -0,0 +1,20 @@
+export interface CardButton {
```
matejkubinec, line 1: I think a better name would be `HelpCardButton`.

ui/src/pages/help-center/help-center-card/HelpCenterCard.types.ts
```diff
@@ -0,0 +1,20 @@
...
+  text: string;
+  target: string;
```
matejkubinec, line 3: If we are linking inside the app the `target` prop could be omitted (`target?: string`)

This also applies to other props - `url`, `startIconName` etc.

Having them optional should simplify `CARD_DATA` object (omit props which are not required).

ui/src/pages/help-center/help-center-card/HelpCenterCard.types.ts
```diff
@@ -0,0 +1,20 @@
...
+  url: string;
+  startIconName: string;
+}
+
+export interface CardType {
+  id: string;
+  title: string;
+  description: string;
+  buttons: Array<CardButton>;
```
matejkubinec, line 12: I think the square bracket declaration for arrays is more [preffered](https://graphite.dev/guides/typescript-arrays). eq.

```ts
buttons: HelpCardButton[]
```

ui/src/pages/help-center/HelpCenter.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    expect(
+      screen.queryByTestId(`help-card-${CARD_IDS.pmmDump}`)
+    ).not.toBeInTheDocument();
+    expect(
+      screen.queryByTestId(`help-card-${CARD_IDS.pmmLogs}`)
+    ).not.toBeInTheDocument();
+    expect(screen.queryAllByTestId(/^help-card-/).length).toEqual(4);
+  });
+
+  it('should not show pmm dump and pmm longs if user is viewer', () => {
```
matejkubinec, line 52:
```suggestion
  it('should not show pmm dump and pmm logs if user is viewer', () => {
```

just a small typo (I think it's in other test names too) 😊

ui/src/pages/help-center/HelpCenter.constants.ts
matejkubinec: Could you add a type info here:

```tsx
export const CARD_DATA: HelpCard[] = [ ... ]
```

This should allow typescript to check if the props/prop types match.

ui/src/pages/help-center/help-center-card/HelpCenterCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+      default:
+        return null;
+    }
+  }, []);
+
+  const onButtonClick = useCallback((button: HelpCardButton) => {
+    if (button.target) {
+      window.open(button.url, button.target, 'noopener,noreferrer');
+    } else {
+      //TO DO: use react router once the grafana iframe is in place
```
matejkubinec, line 48: Please write todo comments as `// TODO: ...` to keep consistency.

ui/src/pages/help-center/HelpCenter.tsx
```diff
@@ -0,0 +1,35 @@
...
+          gridTemplateColumns: {
+            xs: '1fr',
+            sm: '1fr',
+            md: 'repeat(2, 1fr)',
+          },
+          gap: 4,
+        }}
+      >
+        {cards.map((item) => (
+          <HelpCenterCard card={item} />
```
matejkubinec, line 30: Please add a key prop (value should be the item id)

ui/src/pages/help-center/HelpCenter.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Page } from 'components/page';
+import { FC } from 'react';
+import { Messages } from './HelpCenter.messages';
+import { CARDS_DATA } from './HelpCenter.constants';
+import { useUser } from 'contexts/user';
+import { HelpCenterCard } from './help-center-card/HelpCenterCard';
+
+export const HelpCenter: FC = () => {
+  const { user } = useUser();
+
```
matejkubinec, line 11: Please remove empty line.

matejkubinec (approved):

YashSartanpara1 (approved):

ui/src/pages/help-center/help-center-card/HelpCenterCard.tsx
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }, []);
+
+  return (
+    <Card key={id} data-testid={`help-card-${id}`} variant="outlined">
+      <CardContent
+        sx={{
+          px: 2,
+          '&:last-child': {
+            paddingBottom: '16px',
```
matejkubinec, line 59: I think you can use `2` in place of `16px`, it will use the spacing system from MUI.

PR #4032, opened 2025-05-19
PMM-13794 fix pmm agent on pmm server

BupycHuk:
PMM-13794

Link to the Feature Build: SUBMODULES-3927

- [ ] API Docs updated


build/ansible/roles/initialization/tasks/main.yml
```diff
@@ -56,7 +56,7 @@
       host: 127.0.0.1
       port: 5432
       timeout: 150
-    when: lookup('env','PMM_POSTGRES_ADDR') == '127.0.0.1:5432' and lookup('env','PMM_DISABLE_BUILTIN_POSTGRES') != '1'
+    when: lookup('env','GF_DATABASE_URL') == '' and lookup('env','GF_DATABASE_HOST') == '' and lookup('env','PMM_DISABLE_BUILTIN_POSTGRES') == ''
```
ademidoff, line 59: What if the user sets `PMM_DISABLE_BUILTIN_POSTGRES` to `0`?

BupycHuk: it will be very weird decission 😊, but this solution is easier to support then multiple variations of True value

docker-compose.ha.yml
```diff
@@ -0,0 +1,330 @@
+---
+# version: '3.7' // Deprecated: see https://docs.docker.com/compose/compose-file/04-version-and-name/
```
ademidoff, line 2:
```suggestion
```

ademidoff (approved):

build/ansible/roles/initialization/tasks/main.yml
ademidoff: Well, there is a better way, as usual. For example, we could solve it with the following syntax:

```
    when: 
      - lookup('env','GF_DATABASE_URL') == '' 
      - lookup('env','GF_DATABASE_HOST') == '' 
      - not (lookup('env','PMM_DISABLE_BUILTIN_POSTGRES') | bool)
```

PR #4046, opened 2025-05-24
Changes in docker commands used.

guriandoro:
The volume path was incorrect for the pmm-client-data container, as it was using /srv/ when it should be using /usr/local/percona/pmm2/.

Also, what was mentioned about -d was incorrect, and using --rm as it was meant that the container will be removed when the process is stopped. It's better to use -d and leave it running in the background, and docker logs will still work in this case, so we don't lose any functionality.

PMM-0

- [ ] API Docs updated

### DIFF
--- a/documentation/docs/setting-up/client/index.md
+++ b/documentation/docs/setting-up/client/index.md
@@ -54,7 +54,7 @@ The [PMM Client Docker image] is a convenient way to run PMM Client as a preconf
 
     ```sh
     docker create \
-    --volume /usr/local/percona/pmm2/ \
+    --volume /usr/local/percona/pmm2/config \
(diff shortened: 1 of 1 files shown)


documentation/docs/setting-up/client/index.md
```diff
@@ -54,17 +54,17 @@ The [PMM Client Docker image] is a convenient way to run PMM Client as a preconf
 
     ```sh
     docker create \
-    --volume /srv \
+    --volume /usr/local/percona/pmm2/ \
```
BupycHuk, line 57:
```suggestion
    --volume /usr/local/percona/pmm2/config \
```

BupycHuk (approved):

PR #4047, opened 2025-05-26
PMM-13978 Empty examples.

JiriCtvrtka:
PMM-13978

FB: https://github.com/Percona-Lab/pmm-submodules/pull/3937


BupycHuk (approved):

agent/agents/mysql/perfschema/perfschema.go
```diff
@@ -204,9 +206,7 @@ func (m *PerfSchema) Run(ctx context.Context) {
 			running = true
 			m.changes <- agents.Change{Status: inventoryv1.AgentStatus_AGENT_STATUS_RUNNING}
 		}
-	}
 
-	if err != nil {
```
JiriCtvrtka: Not related bug:
```	
var running bool
var err error
m.changes <- agents.Change{Status: inventoryv1.AgentStatus_AGENT_STATUS_STARTING}

if s, err := getSummaries(m.q); err == nil {
  if err = m.summaryCache.Set(s); err == nil {
    m.l.Debugf("Got %d initial summaries.", len(s))
    running = true
    m.changes <- agents.Change{Status: inventoryv1.AgentStatus_AGENT_STATUS_RUNNING}
  }
}

if err != nil {
  m.l.Error(err)
  m.changes <- agents.Change{Status: inventoryv1.AgentStatus_AGENT_STATUS_WAITING}
}
```

Above is previous code. But check if err != nil would be always nil. So agent had wrong status in case of error.

BupycHuk: In first check err == nil

JiriCtvrtka: Changed to cover both errors.

agent/agents/mysql/perfschema/perfschema.go
```diff
@@ -258,7 +260,17 @@ func (m *PerfSchema) Run(ctx context.Context) {
 }
 
 func (m *PerfSchema) runHistoryCacheRefresher(ctx context.Context) {
-	t := time.NewTicker(refreshHistory)
+	interval := refreshHistory
+	if value := os.Getenv("PMM_PERFSCHEMA_REFRESH_RATE"); value != "" {
```
BupycHuk: it's interesting, we never before had any env variable set directly to PMM Client except PMM_DEBUG and PMM_TRACE. What do you think if adding additional field to pmm-agent.yml would work better?

JiriCtvrtka: Up to you. So you want to have property in config and also env variable?

BupycHuk: let's have only config for now

idoqo (approved):

PR #4193, opened 2025-07-01
PMM-13748 GSSAPI auth support

idoqo:
PMM-13748

SUBMODULES-3977

- https://github.com/percona/mongodb_exporter/pull/1106

### DIFF
--- a/agent/agents/mysql/perfschema/perfschema_test.go
+++ b/agent/agents/mysql/perfschema/perfschema_test.go
@@ -308,7 +308,7 @@ func TestPerfSchema(t *testing.T) {
 			"SELECT * FROM `city`": "9c799bdb2460f79b3423b77cd10403da",
 		}
 
-	case "8.0-oracle", "8.0-percona", "8.4-oracle", "9.0-oracle", "9.1-oracle", "9.2-oracle", "9.3-oracle":
+	case "8.0-oracle", "8.0-percona", "8.4-oracle", "9.0-oracle", "9.1-oracle", "9.2-oracle", "9.3-oracle", "9.4-oracle":
 		digests = map[string]string{
 			"SELECT `sleep` (?)":   "0b1b1c39d4ee2dda7df2a532d0a23406d86bd34e2cd7f22e3f7e9dedadff9b69",
 			"SELECT * FROM `city`": "950bdc225cf73c9096ba499351ed4376f4526abad3d8ceabc168b6b28cfc9eab",


agent/Makefile
```diff
@@ -30,9 +30,16 @@ release:                        ## Build static pmm-agent release binary (Linux
 	$(PMM_RELEASE_PATH)/pmm-agent --version
 	ldd $(PMM_RELEASE_PATH)/pmm-agent 2>&1 | grep -Fq 'not a dynamic executable'
 
+release-gssapi:                 ## Build dynamic pmm-agent release binary (for GSSAPI support)
+	CGO_ENABLED=1 go build -v -ldflags "$(VERSION_FLAGS)" -tags 'osusergo netgo static_build gssapi' -o $(PMM_RELEASE_PATH)/pmm-agent-gssapi
```
ademidoff: Would it still make sense to have the `static_build` tag in a dynamic build? Is seems to work against the logic.

agent/Makefile
```diff
@@ -30,9 +30,16 @@ release:                        ## Build static pmm-agent release binary (Linux
 	$(PMM_RELEASE_PATH)/pmm-agent --version
 	ldd $(PMM_RELEASE_PATH)/pmm-agent 2>&1 | grep -Fq 'not a dynamic executable'
 
+release-gssapi:                 ## Build dynamic pmm-agent release binary (for GSSAPI support)
+	CGO_ENABLED=1 go build -v -ldflags "$(VERSION_FLAGS)" -tags 'osusergo netgo gssapi' -o $(PMM_RELEASE_PATH)/pmm-agent-gssapi
+	$(PMM_RELEASE_PATH)/pmm-agent-gssapi --version
```
BupycHuk: I believe output file name should stay the same

idoqo: Oh, that'd mean we can't have both static and dynamic variants on the same machine right?

BupycHuk: we can, but not in the same time

build/scripts/build-client-binary
```diff
@@ -116,6 +127,7 @@ copy_component() {
 }
 
 main() {
+    rm -rf "${source_dir}" "${binary_dir}"
```
ademidoff: Do we absolutely have to remove the source directory?

idoqo: cc @talhabinrizwan

talhabinrizwan: @ademidoff, are you absolutely sure that leaving the source directory intact won't affect the next iteration? It's going to be shared across three builds:
1. static - OL8
2. dynamic - OL8
3. dynamic - OL9

Alternatively, should we consider running a make clean before make release-gssapi instead?

ademidoff: I'm not absolutely sure of anything, that's why I prefer to ask.

ademidoff: IMO, the source directory remains the same for any OS flavor. The build pipeline is not changing the source files I believe.

talhabinrizwan: @idoqo What do you think? If we keep the source directory as is, could it affect the other builds? Or should we run `make clean` before `make release-gssapi` instead? I have no issue keeping the source directory otherwise, I just want to be sure we're creating a clean artifact.

idoqo: Yeah the sources should remain unchanged (since there's no code changes needed specifically for dynamic builds), we can keep it as is then (since `make clean` won't have any much effect either).

ademidoff (approved):

JiriCtvrtka (approved):

PR #4257, opened 2025-07-21
PMM-13830 Use fingerprint instead query for parsing comments.

JiriCtvrtka:
PMM-14194

- [ ] API Docs updated

### DIFF
--- a/agent/queryparser/helpers.go
+++ b/agent/queryparser/helpers.go
@@ -36,9 +36,9 @@ var (
 	dashOnce   sync.Once
 	errDash    error
 
-	sharpRegexp *regexp.Regexp
-	sharpOnce   sync.Once
-	errSharp    error
+	hashRegexp *regexp.Regexp
+	hashOnce   sync.Once
+	errHash    error
 
 	keyValueRegexp *regexp.Regexp
 	keyValueOnce   sync.Once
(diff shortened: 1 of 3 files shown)


BupycHuk (approved):

agent/queryparser/helpers.go
```diff
@@ -198,49 +116,39 @@ func parseKeyValueFromComment(s string) (map[string]bool, error) {
 	return res, nil
 }
 
-func prepareMultilineRegexp() error {
+func prepareMySQLComments() error {
 	// to compile regexp only once
-	multilineOnce.Do(func() {
-		multilineRegexp, errMultiline = regexp.Compile(`(?s)\/\*(.*?)\*\/`) //nolint:gocritic
+	mySQLQuotedOnce.Do(func() {
```
ademidoff: I wonder if we can simplify the code by using `regexp.MustCompile` to initialize `mySQLQuotedRegexp` (and other regex variables) as package-level variables. This would eliminate the need for `sync.Once` and error handling during regex compilation, since `MustCompile` will panic if the pattern is invalid.

ademidoff: Then those `prepare...` functions would not be necessary :)

JiriCtvrtka: Great idea. Suggestion applied.

PR #4280, opened 2025-07-27
PMM-14215 sync pmm-dump branches

idoqo:
PMM-14215 

- [ ] API Docs updated

### DIFF
--- a/build/packages/rpm/server/SPECS/pmm-dump.spec
+++ b/build/packages/rpm/server/SPECS/pmm-dump.spec
@@ -2,14 +2,14 @@
 
 %global repo            pmm-dump
 %global provider        github.com/percona/%{repo}
-%global commit          f226dbb3afb62ac4b9b39032935b5694a48d526f
+%global commit          93057635fd4890c5751e5c99f9a51e5246fc1f8a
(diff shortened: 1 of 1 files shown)


managed/services/dump/dump.go
```diff
@@ -113,6 +113,28 @@ func (s *Service) StartDump(params *Params) (string, error) {
...
+		}
+		node, err := models.FindNodeByID(s.db.Querier, service.NodeID)
+		if err != nil {
+			s.running.Store(false)
+			return "", errors.Wrapf(err, "failed to find node for service %s", serviceName)
+		}
+		instances = append(instances, serviceName)
+		nodeNames[node.NodeName] = struct{}{}
+	}
+	for nodeName := range nodeNames {
```
ademidoff, line 134: IMO we don't need this extra `for` loop, it can be done in the loop above.

BupycHuk: Not really, it ensures that there are no duplicates. But I'm not sure how critical is that

ademidoff (approved):

managed/services/dump/dump.go
ademidoff: Ok, not critical )

PR #4305, opened 2025-08-03
PMM-13994 Disconnects between server and clients.

JiriCtvrtka:
PMM-13994

Link to the Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/4003


managed/services/agents/registry.go
```diff
@@ -297,13 +308,13 @@ func (r *Registry) ping(ctx context.Context, agent *pmmAgentInfo) error {
...
 	r.mClockDrift.Observe(clockDrift.Seconds())
-	return nil
+	return true, nil
 }
 
 // addOrRemoveVMAgent - creates vmAgent agentType if pmm-agent's version supports it and agent not exists yet,
 // otherwise ensures that vmAgent not exist for pmm-agent and pmm-agent's agents don't have push_metrics mode,
 // removes it if needed.
-func (r *Registry) addOrRemoveVMAgent(q *reform.Querier, pmmAgentID, runsOnNodeID string, pmmAgentVersion *version.Parsed) error {
+func (r *Registry) addOrRemoveVMAgent(q *reform.Querier, pmmAgentID, runsOnNodeID string) error {
```
JiriCtvrtka, line 317: The parameter was unused.

managed/services/agents/registry.go
```diff
@@ -185,14 +186,23 @@ func (r *Registry) register(stream agentv1.AgentService_ConnectServer) (*pmmAgen
...
+		// so we return AlreadyExists error. Otherwise we kick the previous connection
+		// and proceed with the new one.
+		pong, err := r.ping(ctx, currentAgent)
+		if pong {
+			return nil, status.Errorf(codes.AlreadyExists, "pmm-agent with ID %q is already connected.", agentMD.ID)
+		}
+
+		l.Warningf("Failed to ping pmm-agent with ID %q: %v", agentMD.ID, err)
 		r.Kick(ctx, agentMD.ID)
+		l.Warningf("pmm-agent with ID %q is kicked.", agentMD.ID)
```
ademidoff:
```suggestion
		// If the agent responds with `pong`, we return AlreadyExists error.
		// Otherwise we kick out the previous connection and proceed with the new one.
		pong, err := r.ping(ctx, currentAgent)
		if pong {
			return nil, status.Errorf(codes.AlreadyExists, "pmm-agent with ID %q is already connected.", agentMD.ID)
		}

		l.Warningf("Failed to ping pmm-agent with ID %q: %v", agentMD.ID, err)
		r.Kick(ctx, agentMD.ID)
		l.Warningf("pmm-agent with ID %q is kicked.", agentMD.ID)
```

managed/services/agents/registry.go
```diff
@@ -278,15 +288,16 @@ func (r *Registry) unregister(pmmAgentID, disconnectReason string) *pmmAgentInfo
...
 	l := logger.Get(ctx)
 	start := time.Now()
 	resp, err := agent.channel.SendAndWaitResponse(&agentv1.Ping{})
 	if err != nil {
-		return err
+		return false, err
 	}
 	if resp == nil {
-		return nil
+		return false, errors.New("pong is not received, response is nil")
```
BupycHuk: I'm curious what's the reason to introduce boolean variable? it can be simply replaced by comparing if err != nil

JiriCtvrtka: I forgot that during refactoring I added error to nil response and it is not needed anymore. Thanks for point out.

agent/connectionuptime/service.go
```diff
@@ -141,6 +141,9 @@ func (c *Service) RunCleanupGoroutine(ctx context.Context) {
 //
 //	time_between(f1, now) - total time between first event (f1) and current moment
 func (c *Service) GetConnectedUpTimeUntil(toTime time.Time) float32 {
+	c.mx.Lock()
+	defer c.mx.Unlock()
+
```
BupycHuk: I noticed that we don't change any field here, so we can change mutex type to RWMutex and use RLock here.

BupycHuk (commented):
I realized that actual problem might occure because of backoff logic in https://github.com/percona/pmm/blob/v3/agent/client/client.go#L152. Every reconnect increases backoff delay and it never resets and will always be on max value which is 15 seconds.

PR #4396, opened 2025-08-21
PMM-6659 Disable pg_stat_monitor agent on PMM Server

ademidoff:
PMM-6659

Link to the Feature Build: SUBMODULES-4015

- [ ] API Docs updated

- API tests are 🟢 -> https://github.com/Percona-Lab/pmm-submodules/pull/4015#issuecomment-3220438992


managed/models/agent_model.go
```diff
@@ -320,6 +322,9 @@ func (s *Agent) BeforeInsert() error {
 	if s.Status == "" && s.AgentType != ExternalExporterType && s.AgentType != PMMAgentType {
 		s.Status = AgentStatusUnknown
 	}
+	if s.Disabled {
+		s.Status = agentStatusDone
+	}
```
idoqo, line 346: is there some reason we don't want a separate status for "disabled" agents?

ademidoff: No particular reason, I thought that status "done" was also good for "disabled", but I'm no longer sure this was the best idea. 

We can certainly add the "disabled" state if we want to. @BupycHuk WDYT?

BupycHuk: Hi, disabled might work as well as done, but I think it's provided to user as a separate field. This one presents current status of exporter process.

ademidoff: OK, so in summary: `disabled` is an attribute on its own, it does not reflect the agent's status (i.e. "Running", "Done", "Waiting", etc.). We consider it orthogonal to "status", which is why we keep the logic related to "status" attribute unchanged.

BupycHuk (approved):

PR #4402, opened 2025-08-22
[DOC] Cleartext auth (PMM-13513)

catalinaadam:
PMM-0

- [ ] API Docs updated


documentation/docs/use/commands/pmm-admin.md
```diff
@@ -471,6 +471,30 @@ When you remove a service, collected data remains on PMM Server for the specifie
...
+        Allows passing custom connection parameters as `key=value` pairs to configure 
+        how PMM connects to your MySQL instance.
+    
+            **Common parameters**
+                
+            • charset=utf8mb4 - Set character encoding
+
+            • parseTime=true - Parse time values to time
+
+            • timeout=30s - Set connection timeout
```
idoqo: Even though supporting the others are trivial, we only fully support the `allowCleartextPasswords` parameter - so I think we can drop these, and go with something like:

"This flag can be used to add MySQL instances with cleartext password authentication"

PR #4221, opened 2025-07-10
PMM-13513 mysql cleartext auth

idoqo:
PMM-13513

Link to the Feature Build: SUBMODULES-3987

-  https://github.com/percona/mysqld_exporter/pull/307

PR #4403, opened 2025-08-24
PMM-7 Fix Alerting and Inventory API tests

ademidoff:
This PR addresses the following API test failures:
1. We added 8 more alert templates to the free tier via Portal, and now ListTemplates returns +8 templates.
2. It seems like [adding ExtraDsnParams](https://github.com/percona/pmm/pull/4221) broke some Inventory API tests. Example:
```bash
00:06:08.422          	            	Diff:
00:06:08.422          	            	--- Expected
00:06:08.422          	            	+++ Actual
00:06:08.422          	            	@@ -18,3 +18,4 @@
00:06:08.422          	            	   Version: (string) "",
00:06:08.422          	            	-  ExtraDsnParams: (map[string]string) <nil>
00:06:08.422          	            	+  ExtraDsnParams: (map[string]string) {
00:06:08.422          	            	+  }
```

Link to the Feature Build: SUBMODULES-4017

Test results: https://github.com/Percona-Lab/pmm-submodules/pull/4017#issuecomment-3218424435

### DIFF
--- a/api-tests/alerting/alerting_test.go
+++ b/api-tests/alerting/alerting_test.go
@@ -474,7 +474,7 @@ func TestTemplatesAPI(t *testing.T) {
 
 			// list rules, so they are all on the first page
 			listAllTemplates, err := client.ListTemplates(&alerting.ListTemplatesParams{
-				PageSize:  pointer.ToInt32(30),
+				PageSize:  pointer.ToInt32(38),
 				PageIndex: pointer.ToInt32(0),
 				Context:   pmmapitests.Context,
 			})
--- a/api-tests/inventory/agents_mysqld_exporter_test.go
+++ b/api-tests/inventory/agents_mysqld_exporter_test.go
@@ -66,9 +66,6 @@ func TestMySQLdExporter(t *testing.T) {
 				CustomLabels: map[string]string{
 					"custom_label_mysql_exporter": "mysql_exporter",
 				},
-				ExtraDsnParams: map[string]string{
-					"allowCleartextPasswords": "1",
-				},
 				SkipConnectionCheck:       true,
 				TablestatsGroupTableLimit: 2000,
 			},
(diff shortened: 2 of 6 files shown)


api-tests/alerting/alerting_test.go
```diff
@@ -474,7 +474,7 @@ func TestTemplatesAPI(t *testing.T) {
 
 			// list rules, so they are all on the first page
 			listAllTemplates, err := client.ListTemplates(&alerting.ListTemplatesParams{
-				PageSize:  pointer.ToInt32(30),
+				PageSize:  pointer.ToInt32(38),
```
JiriCtvrtka, line 477: Just from curiosity. Is 38 random number or specific for reason?

ademidoff: No, it's the exact number of alert templates in the free tier :)

ademidoff: That's what the Portal returns )

idoqo (approved):

api-tests/inventory/agents_mysqld_exporter_test.go
```diff
@@ -66,9 +66,6 @@ func TestMySQLdExporter(t *testing.T) {
 				CustomLabels: map[string]string{
 					"custom_label_mysql_exporter": "mysql_exporter",
 				},
-				ExtraDsnParams: map[string]string{
-					"allowCleartextPasswords": "1",
-				},
```
BupycHuk, line 71: I don't get why this was removed? Should we fix the behaviour instead of the test?

api-tests/alerting/alerting_test.go
BupycHuk: Why don't we make it 100 so we don't have to change it later?

api-tests/inventory/agents_mysqld_exporter_test.go
ademidoff: My bad, it was a bad revert "git restore" 🤦‍♂️ 

I'll follow up with another one shortly.

PR #4410, opened 2025-08-25
PMM-7 Fix duplicate Dependabot PRs

ademidoff:
PMM-7

We are observing Dependabot create duplicate PRs. This is probably because the `target-branch` was not specified explicitly for the "main" branch. 

This PR tries to provide an explicit target branch.


BupycHuk (approved):

JiriCtvrtka:
We are using main branch as target one. Branch to merge PR is v3. Is that correct?

JiriCtvrtka (approved):

PR #4412, opened 2025-08-26
PMM-7 Race condition possible fix.

JiriCtvrtka:


ademidoff:
I think we need to pass `service` as a variable to the goroutine to avoid the race. 

```go
for id, service := range s.all {
    if _, ok := s.running[id]; !ok {
        s.wg.Add(1)
        s.running[id] = service
        go func(svc LeaderService) {
            s.l.Infoln("Starting", svc.ID())
            err := svc.Start(ctx)
            if err != nil {
                s.l.Errorln(err)
            }
        }(service)
    }
}
```

PR #4525, opened 2025-09-15
PMM-14253 Update flow

matejkubinec:
PMM-14253

Link to the Feature Build: SUBMODULES-4038

- [ ] API Docs updated

### DIFF
--- a/ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
+++ b/ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
@@ -76,7 +76,7 @@ export const GrafanaProvider: FC<PropsWithChildren> = ({ children }) => {
     messenger.addListener({
       type: 'LOCATION_CHANGE',
       onMessage: ({ payload: location }: LocationChangeMessage) => {
-        if (!location || location.action === 'POP') {
+        if (!location) {
           return;
         }
 
@@ -107,8 +107,13 @@ export const GrafanaProvider: FC<PropsWithChildren> = ({ children }) => {
   useEffect(() => {
     if (!isBrowser() || !isLoaded) return;
 
-    const state = location.state as LocationState;
-    if (!location.pathname.includes('/graph') || state?.fromGrafana) return;
+    const isGrafanaPage = location.pathname.includes('/graph');
+    const isSourceGrafana = (location.state as LocationState)?.fromGrafana;
+    const isBackNavigation = navigationType === 'POP';
+
+    if (!isGrafanaPage || (isSourceGrafana && !isBackNavigation)) {
+      return;
+    }
 
     messenger.sendMessage({
       type: 'LOCATION_CHANGE',
--- a/ui/apps/pmm/src/contexts/grafana/grafana.utils.ts
+++ b/ui/apps/pmm/src/contexts/grafana/grafana.utils.ts
@@ -1,6 +1,7 @@
 import { PMM_NEW_NAV_GRAFANA_PATH, PMM_NEW_NAV_PATH } from 'lib/constants';
+import { Path } from 'react-router-dom';
 
-export const getLocationUrl = (location: Location) => {
+export const getLocationUrl = (location: Path) => {
   const pathname = location.pathname.startsWith('/pmm-ui')
     ? location.pathname.replace('/pmm-ui', PMM_NEW_NAV_PATH)
     : PMM_NEW_NAV_GRAFANA_PATH + location.pathname;


matejkubinec:
@idoqo @ademidoff 

Hello, would like some feedback on the BE part of this.

I've added a separate snooze updates api endpoint which handled the logic related to snoozing:
1. Keeps a timestamp of when was the update last snoozed 
2. Mismatch version - resets snooze counter 
3. Increments snooze counter if current snoozed version & payload one matches

Also I've extended settings with snooze duration (currently 7 days) - time after snooze for which an update modal is not shown to the user (configurable via env variable - maybe just for QAs?)

api/server/v1/server.proto
```diff
@@ -306,6 +325,17 @@ service ServerService {
       description: "Returns PMM Server update status."
     };
   }
+  // SnoozeUpdate snoozes updates for currently logged in user and returns the count and timestamp
```
ademidoff:
```suggestion
  // SnoozeUpdate snoozes updates for an authenticated user and returns the count and timestamp
```

api/server/v1/server.proto
```diff
@@ -235,12 +238,28 @@ message ChangeSettingsRequest {
   optional bool enable_backup_management = 12;
   // Enable Access Control
   optional bool enable_access_control = 13;
+  // A number of full days for which an update is snoozed. Should have a suffix in JSON: 2592000s, 43200m, 720h.
```
ademidoff:
```suggestion
  // A number of full days for which an update is snoozed, i.e. a multiple of 24h: 2592000s, 43200m, 720h.
```
I wonder if it wouldn't be easier to provide an `uint` here? That'd be simple to read and test, no?

api/user/v1/user.proto
```diff
@@ -21,6 +22,8 @@ message GetUserResponse {
   // Snoozed PMM version update
   string snoozed_pmm_version = 4;
   reserved 5;
+  google.protobuf.Timestamp snoozed_at = 6;
+  uint32 snoozed_count = 7;
```
ademidoff:
```suggestion
  uint32 snooze_count = 7;
```

managed/models/database.go
```diff
@@ -1141,6 +1141,17 @@ var databaseSchema = [][]string{
...
 		`UPDATE nodes SET instance_id = address WHERE instance_id = ''`,
 	},
+	111: {
+		`ALTER TABLE user_flags
+			ADD COLUMN snoozed_at TIMESTAMP,
+			ADD COLUMN snoozed_count INTEGER NOT NULL DEFAULT 0`,
+	},
+	112: {
+		`UPDATE settings
+			SET settings = settings || '{"updates": {"snooze_duration": 604800000000000}}'
```
ademidoff: Let's provide a comment here, i.e. how many days it is.

ademidoff: Is it 7?

managed/services/server/server.go
```diff
@@ -422,6 +422,56 @@ func (s *Server) UpdateStatus(ctx context.Context, req *serverv1.UpdateStatusReq
 	}, nil
 }
 
+// Snooze updates for specific pmm version for current user
```
ademidoff:
```suggestion
// SnoozeUpdate snoozes the updates for specific pmm version for current user.
```

managed/services/server/server.go
```diff
@@ -422,6 +422,56 @@ func (s *Server) UpdateStatus(ctx context.Context, req *serverv1.UpdateStatusReq
...
+		return nil, err
+	}
+
+	res := &serverv1.SnoozeUpdateResponse{
+		SnoozedPmmVersion: userInfo.SnoozedPMMVersion,
+		SnoozedAt:         timestamppb.New(pointer.Get(userInfo.SnoozedAt)),
+		SnoozedCount:      uint32(userInfo.SnoozedCount),
+	}
+
+	err = s.db.InTransaction(func(tx *reform.TX) error {
```
ademidoff: It seems like we don't need a transaction here.

managed/services/server/server.go
```diff
@@ -593,6 +644,10 @@ func (s *Server) validateChangeSettingsRequest(ctx context.Context, req *serverv
 		return status.Error(codes.FailedPrecondition, "Data retention for queries is set via PMM_DATA_RETENTION environment variable.")
 	}
 
+	if !canUpdateDurationSetting(req.DataRetention.AsDuration(), s.envSettings.UpdatesSnoozeDuration) {
+		return status.Error(codes.FailedPrecondition, "Updates snooze duration is set via PMM_UPDATES_SNOOZE_DURATION environment variable.")
```
ademidoff:
```suggestion
		return status.Error(codes.FailedPrecondition, "Updates snooze duration is set via PMM_UPDATE_SNOOZE_DURATION environment variable.")
```

I'd rename this variable to models.UpdateUser, is we are dealing with one single update at a time.

managed/services/user/user.go
```diff
@@ -67,11 +70,18 @@ func (s *Service) GetUser(ctx context.Context, _ *userv1.GetUserRequest) (*userv
 		return nil, err
 	}
 
+	var snoozedAt *timestamppb.Timestamp
```
ademidoff: If you put this block past line 85, there will be no need to define the variable.

managed/services/user/user.go
```diff
@@ -99,8 +109,12 @@ func (s *Service) UpdateUser(ctx context.Context, req *userv1.UpdateUserRequest)
 			Tour:         req.ProductTourCompleted,
 			AlertingTour: req.AlertingTourCompleted,
 		}
-		if req.SnoozedPmmVersion != nil {
+
+		// Keep for backwards compatibility, prefer the new snooze updates:endpoint
+		if req.SnoozedPmmVersion != nil && req.SnoozedPmmVersion != &userInfo.SnoozedPMMVersion {
```
ademidoff: I think we should check the values here, not the pointers, since pointers will be different.

managed/services/user/user.go
```diff
@@ -114,11 +128,18 @@ func (s *Service) UpdateUser(ctx context.Context, req *userv1.UpdateUserRequest)
 		return nil, e
 	}
 
+	var snoozedAt *timestamppb.Timestamp
```
ademidoff: Same - let's put this block after line 143.

managed/utils/envvars/parser.go
```diff
@@ -116,6 +116,11 @@ func ParseEnvVars(envs []string) (*models.ChangeSettingsParams, []error, []strin
 				continue
 			}
 			envSettings.EnableUpdates = &b
+		case "PMM_UPDATES_SNOOZE_DURATION":
```
ademidoff:
```suggestion
		case "PMM_UPDATE_SNOOZE_DURATION":
```

managed/utils/envvars/parser.go
```diff
@@ -116,6 +116,11 @@ func ParseEnvVars(envs []string) (*models.ChangeSettingsParams, []error, []strin
...
+			if envSettings.UpdatesSnoozeDuration, err = parseStringDuration(v); err != nil {
```
ademidoff:
```suggestion
			if envSettings.UpdateSnoozeDuration, err = parseStringDuration(v); err != nil {
```

api/user/v1/user.proto
```diff
@@ -43,6 +46,8 @@ message UpdateUserResponse {
   // Snooze update alert for a PMM version
   string snoozed_pmm_version = 4;
   reserved 5;
+  google.protobuf.Timestamp snoozed_at = 6;
+  uint32 snoozed_count = 7;
```
ademidoff:
```suggestion
  uint32 snooze_count = 7;
```

ui/apps/pmm/src/components/main/update-modal/UpdateModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+import Card from '@mui/material/Card';
+import IconButton from '@mui/material/IconButton';
+import { useSnooze } from 'hooks/updates';
+
+const UpdateModal: FC = () => {
+  const { isLoading, versionInfo } = useUpdates();
+  const [open, setIsOpen] = useState(true);
+  const { snoozeUpdate, snoozeActive, snoozeCount } = useSnooze();
+  const location = useLocation();
+
```
dmitri-saricev-3pillargloball, line 28: `  // Derived values
  const latestVersion = versionInfo?.latest?.version ?? '';
  const releaseNotesUrl = versionInfo?.latest?.releaseNotesUrl ?? '';
  const updateAvailable = Boolean(versionInfo?.updateAvailable);

  // Robust "are we on updates page?" check
  const onUpdatesPage =
    location.pathname === PMM_NEW_NAV_UPDATES_PATH ||
    location.pathname.startsWith(`${PMM_NEW_NAV_UPDATES_PATH}/`);

  // 1) Reopen UI when a new version becomes available (and snooze isn't active)
  useEffect(() => {
    if (updateAvailable && !snoozeActive) {
      setOpen(true);
    }
  }, [latestVersion, updateAvailable, snoozeActive]);`


Nice work!
Pulling the derived values out like latestVersion, releaseNotesUrl, and updateAvailable makes the logic way clearer — much easier to read at a glance.
Also really like the updated onUpdatesPage check — it’s safer for nested paths.
The new useEffect is the key part here: it makes sure the modal reopens automatically when a new version becomes available instead of staying closed forever after the user snoozes it once. That’s a great UX improvement!

matejkubinec: Updated

api-tests/user/snooze_test.go
```diff
@@ -0,0 +1,83 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package user
+
+import (
+	"testing"
+	"time"
+
+	"github.com/go-openapi/strfmt"
```
ademidoff: We don't use this package directly.

api/user/v1/user.proto
```diff
@@ -21,6 +23,8 @@ message GetUserResponse {
   // Snoozed PMM version update
   string snoozed_pmm_version = 4;
   reserved 5;
+  google.protobuf.Timestamp snoozed_at = 6;
```
ademidoff, line 26: Let's provide comments here as well.

api/user/v1/user.proto
```diff
@@ -43,6 +47,8 @@ message UpdateUserResponse {
   // Snooze update alert for a PMM version
   string snoozed_pmm_version = 4;
   reserved 5;
+  google.protobuf.Timestamp snoozed_at = 6;
```
ademidoff, line 52: Let's provide comments for consistency.

api/user/v1/user.proto
```diff
@@ -83,4 +103,15 @@ service UserService {
       description: "Retrieve user details for all users from PMM server"
     };
   }
+  // SnoozeUpdate snoozes updates for an authenticated user and returns the count and timestamp
+  rpc SnoozeUpdate(SnoozeUpdateRequest) returns (SnoozeUpdateResponse) {
```
ademidoff: I think we can reuse the `UpdateUser` endpoint (i.e. `PUT /v1/users/me`), I don't see a strong reason to have a dedicated one.

managed/services/user/user.go
```diff
@@ -99,8 +108,15 @@ func (s *Service) UpdateUser(ctx context.Context, req *userv1.UpdateUserRequest)
 			Tour:         req.ProductTourCompleted,
 			AlertingTour: req.AlertingTourCompleted,
 		}
-		if req.SnoozedPmmVersion != nil {
+
+		// Keep for backwards compatibility, prefer the snoozeUpdate endpoint
+		if req.SnoozedPmmVersion != nil && *req.SnoozedPmmVersion != userInfo.SnoozedPMMVersion {
```
ademidoff: Even though nested `if` statements are evil, this is the case where a nested `if` would be much more readable and maintainable, i.e.:

```
if req.SnoozedPmmVersion != nil {
  if *req.SnoozedPmmVersion != userInfo.SnoozedPMMVersion ...
}
```

ademidoff:
I also noticed this PR needs a `make gen` to be run before pushing.

managed/models/database.go
```diff
@@ -1150,7 +1150,19 @@ var databaseSchema = [][]string{
...
 	113: {
-		// Reset product tour for new navigation
+		`ALTER TABLE user_flags
+			ADD COLUMN snoozed_at TIMESTAMP,
+			ADD COLUMN snooze_count INTEGER NOT NULL DEFAULT 0`,
+	},
+	// Default snooze duration - 7 days
+	114: {
+		`UPDATE settings
+			SET settings = settings || '{"updates": {"snooze_duration": 604800000000000}}'
```
JiriCtvrtka: Maybe could you assign duration value into variable? Like weekDuration or something?

managed/models/settings.go
```diff
@@ -231,4 +232,9 @@ func (s *Settings) fillDefaults() {
 	if s.SaaS.AdvisorRunIntervals.FrequentInterval == 0 {
 		s.SaaS.AdvisorRunIntervals.FrequentInterval = 4 * time.Hour
 	}
+
+	if s.Updates.SnoozeDuration == 0 {
+		// 7 days
+		s.Updates.SnoozeDuration = 7 * 24 * time.Hour
```
JiriCtvrtka: Since it is used multiple times could you create const like defaultSnoozeDuration?

managed/models/database.go
```diff
@@ -1150,7 +1153,18 @@ var databaseSchema = [][]string{
 		`UPDATE agents SET disabled = true WHERE agent_type = 'qan-postgresql-pgstatements-agent' AND service_id = (SELECT service_id FROM services WHERE service_name = 'pmm-server-postgresql' LIMIT 1);`,
 	},
 	113: {
-		// Reset product tour for new navigation
+		`ALTER TABLE user_flags
```
ademidoff, line 1160: I'm afraid, we can't afford inserting migrations - they won't work like that. We can only append them.

ademidoff: I mean, No. 113 hat to be left where it was, and those two new should come right after.

This holds true if No. 113 has been already released, otherwise it's good (but still unclear to the reviewer).

PR #4542, opened 2025-09-19
PMM-14272 Migrate Advisor checks to v2 format

ademidoff:
PMM-14272


managed/CONTRIBUTING.md
```diff
@@ -94,24 +94,26 @@ If there is a need to change the logic of Advisor checks (actual logic executed
 
 Changes to Advisors will be most visible in the list of all advisors by categories, such as https://pmmdemo.percona.com/graph/advisors/configuration.
 
-<img width="1024" alt="image" src="https://github.com/percona/checked/assets/25495422/6b369cd6-f080-49aa-a132-72db5fbdf046">
+![Advisors interface](../docs/assets/advisors/pmm-advisor-interface.png)
 
-``advisors.summary`` = https://github.com/percona/checked/blob/223ae162ced83793bc00e5e6c29edfbf1bf5e27e/data/advisors/example.yml.example#L4
+``advisors.summary`` = https://github.com/percona/checked/blob/223ae162ced83793bc00e5e6c29edfbf1bf5e27e/data/advisors/example.yml.example#L5
```
idoqo: we can even replace the entire line

```suggestion
``advisors.summary`` = https://github.com/percona/pmm/blob/b951d3c14eb1d5e4d716a61811da599af869054b/managed/data/advisors/example.yml.example#L5
```

managed/data/checks/mongodb_auth.yml
```diff
@@ -1,34 +1,27 @@
...
 
-          if len(docs) != 1:
+          rows = docs[0]
+          if len(rows) != 1:
               return "Unexpected number of documents"
 
           results = []
 
-          parsed = docs[0]["parsed"]
+          parsed = rows[0]["parsed"]
```
idoqo, line 24: i see rows is already assigned to docs[0] though. Is this because we now expect multiple results from `queries`?

managed/CONTRIBUTING.md
ademidoff: Absolutely, thanks!

managed/data/checks/mongodb_auth.yml
ademidoff: Yes, query 1 results would be stored in `docs[0]`, query 2 - in `docs[1]`, etc.

ademidoff: I believe it's correct, since in v1 `docs` used to be an array of db rows, now it's an array of an array of db rows. So the change seems to be OK. And it's tested :)

PR #4575, opened 2025-09-27
PMM-14118 Generate SSL certificates for ClickHouse

ademidoff:
PMM-14118

### DIFF
--- a/build/ansible/roles/clickhouse/tasks/main.yml
+++ b/build/ansible/roles/clickhouse/tasks/main.yml
@@ -19,6 +19,16 @@
     enablerepo: clickhouse
   ignore_errors: "{{ ansible_check_mode }}" # We don't have clickhouse repo when we run ansible with --check
 
+- name: Generate DH params
+  command: openssl dhparam -out /etc/clickhouse-server/dhparam.pem 4096
+  args:
+    creates: /etc/clickhouse-server/dhparam.pem
+
+- name: Generate SSL certificates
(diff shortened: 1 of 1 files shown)


build/ansible/roles/nginx/files/ssl/generate-ssl-certificate
```diff
@@ -55,5 +55,5 @@ if [ ! -f /srv/nginx/certificate.key ] || [ ! -f /srv/nginx/certificate.crt ]; t
     openssl req -x509 -nodes -days 365 -newkey rsa:2048 \
         -keyout /srv/nginx/certificate.key \
         -out    /srv/nginx/certificate.crt \
-        -config /etc/nginx/ssl/certificate.conf
+        -config /srv/nginx/certificate.conf
```
ademidoff, line 59: @BupycHuk Are we OK with this fix?

BupycHuk: Have you tested it?

ademidoff: Yes, no issues )

ademidoff: I wonder, wasn't it meant to be that way? I thought it was an oversight, no? If we don't take the `certificate.conf` from `/srv/nginx/`, then there is no way for the user to override ours, which comes from `/etc/nginx/`.

PR #4576, opened 2025-09-27
PMM-10461 No absolute Nginx redirects

ademidoff:
PMM-10461

### DIFF
--- a/build/ansible/roles/nginx/files/conf.d/pmm.conf
+++ b/build/ansible/roles/nginx/files/conf.d/pmm.conf
@@ -46,6 +46,8 @@
     server_name _;
     server_tokens off;
 
+    absolute_redirect off;
+
     # allow huge requests
     large_client_header_buffers 128 64k;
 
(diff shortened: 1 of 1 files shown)


build/ansible/roles/nginx/files/conf.d/pmm.conf
```diff
@@ -133,7 +135,7 @@
     }
 
     # Grafana
-    rewrite ^/$ $scheme://$http_host/graph/;
+    rewrite ^/$ /graph/ permanent;
```
BupycHuk: Let's not make it permanent, it might change in future

ademidoff: "permanent" stands for 301 and gets cached by browsers for speed. But whenever we change the `/graph/` part, the cache will be busted immediately. So that should be fine, I beleieve.

BupycHuk: https://stackoverflow.com/questions/9130422/how-long-do-browsers-cache-http-301s

ademidoff: I don't think it's a problem since we have the "Cache-control" and "Pragma" directives set to "no-cache". 

On the other hand I don't mind dropping that attribute either :)

Image

ademidoff: Dropped now )

BupycHuk (approved):

PR #4620, opened 2025-10-07
PMM-14348: QAN settings for internal Postgres DB

idoqo:
PMM-14348


api-tests/inventory/agents_pg_stat_statement_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+			// Verify the setting was persisted
+			getRes, err = client.Default.AgentsService.GetAgent(&agents.GetAgentParams{
+				AgentID: agentID,
+				Context: pmmapitests.Context,
+			})
+			require.NoError(t, err)
+			assert.True(t, getRes.Payload.QANPostgresqlPgstatementsAgent.Disabled)
+		})
+	}
```
BupycHuk: it shouldn't be API test, because ENV vars in PMM Server docker container and env vars in API tests might be different

api-tests/inventory/agents_pg_stat_statement_test.go
```diff
@@ -0,0 +1,150 @@
...
+	defer func() {
+		_, _ = client.Default.AgentsService.ChangeAgent(&agents.ChangeAgentParams{
+			AgentID: agentID,
+			Body: agents.ChangeAgentBody{
+				QANPostgresqlPgstatementsAgent: &agents.ChangeAgentParamsBodyQANPostgresqlPgstatementsAgent{
+					Enable: pointer.ToBool(!originalDisabledState),
+				},
+			},
+			Context: pmmapitests.Context,
+		})
```
BupycHuk: if we have this API endpoint why just don't use it in UI?

idoqo: We could (and frontend already used it actually). But it means it won't be a "server setting" anymore, and frontend also needs to track which pgss QAN agent is attached to the pmm-server.

The server endpoint allows us move the responsibility to BE...

api-tests/inventory/agents_pg_stat_statement_test.go
idoqo: yeah was trying to see if running the different tests based on env var works, but agreed, better to drop it completely then

managed/services/server/server.go
```diff
@@ -699,6 +705,37 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverv1.ChangeSetting
 		}
 	}
 
+	// if QAN for internal PostgreSQL is toggled, we need to update the agent's disabled status
+	if oldSettings.IsInternalPgQANEnabled() != newSettings.IsInternalPgQANEnabled() {
+		if err := s.db.InTransaction(func(tx *reform.TX) error {
```
ademidoff:
```suggestion
		if err := s.db.InTransactionContext(ctx, nil, func(tx *reform.TX) error {
```

ademidoff: Let's make sure we always pass the context to transactions.

api-tests/inventory/agents_pg_stat_statement_test.go
ademidoff: I think FE mentioned they prefer calling the Settings endpoint, this is why do it that way.

managed/models/database.go
```diff
@@ -1518,7 +1518,17 @@ func setupPMMServerAgents(q *reform.Querier, params SetupDBParams) error {
...
+	// It can be enabled by setting PMM_ENABLE_INTERNAL_PG_QAN=1
+	// We rely on just the environment variable here since we run this set up before loading the server settings.
 	ap.Disabled = true
+	envVar, exists := os.LookupEnv("PMM_ENABLE_INTERNAL_PG_QAN")
+	if exists {
+		if enable, err := strconv.ParseBool(envVar); err != nil {
+			return fmt.Errorf("cannot parse PMM_ENABLE_INTERNAL_PG_QAN value %q: %w", envVar, err)
+		} else if enable {
+			ap.Disabled = false
+		}
```
ademidoff:
```suggestion
		enable, err := strconv.ParseBool(envVar)
		if err != nil {
			return fmt.Errorf("cannot parse PMM_ENABLE_INTERNAL_PG_QAN value %q: %w", envVar, err)
		}
		if enable {
			ap.Disabled = false
		}
```

api-tests/server/settings_test.go
```diff
@@ -666,6 +668,85 @@ func TestSettings(t *testing.T) {
...
+						}
+					}
+
+					// Restore original state
+					_, _ = serverClient.Default.ServerService.ChangeSettings(&server.ChangeSettingsParams{
+						Body: server.ChangeSettingsBody{
+							EnableInternalPgQAN: pointer.ToBool(!originalDisabledState),
+						},
+						Context: pmmapitests.Context,
+					})
```
BupycHuk: Shouldn't it be part of restoreSettingsDefaults(t)?

api-tests/server/settings_test.go
```diff
@@ -666,6 +668,85 @@ func TestSettings(t *testing.T) {
...
+					if agent.PMMAgentID == "pmm-server" {
+						internalPgQANAgent = agent
+						break
+					}
+				}
+				require.NotNil(t, internalPgQANAgent, "Internal PostgreSQL QAN agent should exist")
+
+				originalDisabledState := internalPgQANAgent.Disabled
+
+				t.Run("DisableAndEnableInternalPgQAN", func(t *testing.T) {
```
BupycHuk: Do we really need another test inside this test if it's the only one?

BupycHuk: Would be good to call GetSettings to make sure that it returns correct values.

managed/utils/envvars/parser.go
```diff
@@ -123,6 +125,11 @@ func ParseEnvVars(envs []string) (*models.ChangeSettingsParams, []error, []strin
 				continue
 			}
 			envSettings.EnableTelemetry = &b
+		case "PMM_ENABLE_INTERNAL_PG_QAN":
```
BupycHuk: EnvEnableInternalPgQAN?

managed/services/server/server_test.go
```diff
@@ -217,6 +218,15 @@ func TestServer(t *testing.T) {
 			EnableTelemetry: pointer.ToBool(true),
 		}))
 
+		s.envSettings.EnableInternalPgQAN = pointer.ToBool(true)
+		expected = status.New(codes.FailedPrecondition, "QAN for internal PostgreSQL is configured via PMM_ENABLE_INTERNAL_PG_QAN environment variable.")
```
BupycHuk: I think we better to use envvars.EnvEnableInternalPgQAN everywhere or remove it from error message.

managed/services/server/server.go
```diff
@@ -699,6 +704,35 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverv1.ChangeSetting
...
+				}
+
+				s.agentsState.RequestStateUpdate(ctx, agent.AgentID)
+			}
+
+			return nil
+		}); err != nil {
+			s.l.Error(errors.WithStack(err))
+		}
+	}
```
BupycHuk: put it before UpgradeConfiguration otherwise pmm-agent won't get a signal

api-tests/server/settings_test.go
idoqo: yeah could be - though idea here was to restore to original state (either enabled or disabled) - `restoreSettingsDefaults` doesn't keep previous state and only sets it to the defaults

managed/services/server/server_test.go
idoqo: yeah - but leads to cyclic dependency in some cases.

BupycHuk (approved):

PR #4653, opened 2025-10-12
PMM-7: fix api tests

idoqo:
PMM-7


api-tests/inventory/agents_valkey_exporter_test.go
```diff
@@ -114,7 +114,7 @@ func TestValkeyExporter(t *testing.T) {
 					Username:           "default",
 					PMMAgentID:         pmmAgentID,
 					Disabled:           true,
-					Status:             &AgentStatusUnknown,
+					Status:             &AgentStatusDone,
```
BupycHuk, line 117: I don't get why it should be Done

idoqo: We updated "Disabled" agents to return "Done" instead of "Unknown" in https://github.com/percona/pmm/pull/4396

PR #4682, opened 2025-10-17
PMM 13702 theme toggle button compatibility dark theme

dmitri-saricev-3pillargloball:
[PMM-0
](https://perconadev.atlassian.net/browse/PMM-13702)

- [ ] API Docs updated

- FB
-https://github.com/Percona-Lab/pmm-submodules/pull/4083


ui/apps/pmm-compat/src/compat.ts
```diff
@@ -108,10 +102,63 @@ export const initialize = () => {
...
+  // Helper to apply our scheme attribute and notify parent
+  const apply = (incoming: 'light' | 'dark' | string) => {
+    // normalize to 'light' | 'dark'
+    const mode: 'light' | 'dark' = String(incoming).toLowerCase() === 'dark' ? 'dark' : 'light';
+
+    const html = document.documentElement;
+    const scheme = mode === 'dark' ? 'percona-dark' : 'percona-light';
+
+    // Set attributes our CSS reads
+    html.setAttribute('data-md-color-scheme', scheme);
```
matejkubinec: Why do we need to modify the HTML attributes?

ui/apps/pmm-compat/src/compat.ts
```diff
@@ -108,10 +102,63 @@ export const initialize = () => {
...
+    html.setAttribute('data-theme', mode);
+    (html.style as CSSStyleDeclaration & { colorScheme: string }).colorScheme = mode;
+    // Notify outer PMM UI (new nav) to sync immediately
+    try {
+      const target = window.top && window.top !== window ? window.top : window.parent || window;
+      target?.postMessage({ type: 'grafana.theme.changed', payload: { mode } }, targetOrigin);
```
matejkubinec: Use messenger to handle the communication between the `iframe` and `pmm-ui` (no need to handle the target, etc.). We also use predefined message types (see `ui/packages/shared/src/types.ts`)

ui/apps/pmm-compat/src/contexts/theme/theme.provider.tsx
```diff
@@ -1,15 +1,62 @@
...
+// ------------------------------------------------------
+
+import React, { FC, PropsWithChildren, useEffect, useRef, useState } from 'react';
 import { ThemeContext } from '@grafana/data';
 import { config, getAppEvents, ThemeChangedEvent } from '@grafana/runtime';
-import React, { FC, PropsWithChildren, useEffect, useState } from 'react';
+
+type Mode = 'light' | 'dark';
 
 export const ThemeProvider: FC<PropsWithChildren> = ({ children }) => {
```
matejkubinec, line 5: I don't think you need to touch this provider at all, I think I'll add a comment there, but this specifically is used just for two toolbar buttons used inside the grafana iframe where we extend the functionality. See `ui/apps/pmm-compat/src/compat/toolbar.tsx` for usage, but in short we append two divs which we then take as root for two small react apps (just a button for each)

ui/apps/pmm-compat/src/theme.ts
```diff
@@ -20,19 +18,154 @@ export const changeTheme = async (themeId: 'light' | 'dark'): Promise<void> => {
...
     document.head.insertBefore(newCssLink, document.head.firstChild);
   }
 };
+
+/* ---------------------------
+ * Right → left theme wiring
+ * --------------------------*/
+
+// Normalize and apply <html> attributes so CSS-based nav updates immediately
+function applyHtmlTheme(modeRaw: unknown) {
```
matejkubinec: Why is there a need to apply the theme to html? The `changeTheme` handles the theme change on the grafana side.

ui/apps/pmm-compat/src/theme.ts
```diff
@@ -20,19 +18,154 @@ export const changeTheme = async (themeId: 'light' | 'dark'): Promise<void> => {
...
+function sendToParent(msg: unknown) {
+  const w = resolveTargetWindow();
+  if (!w) {
+    return;
+  }
+  w.postMessage(msg, targetOriginRef.current);
+}
+
+/** Dev-only handshake: lock '*' to the real origin after the first ACK. */
+(function setupOriginHandshake() {
```
matejkubinec: Handshake is pretty good idea to not send information where we do not want.

ui/apps/pmm-compat/src/theme.ts
```diff
@@ -20,19 +18,154 @@ export const changeTheme = async (themeId: 'light' | 'dark'): Promise<void> => {
...
+      sendToParent({ type: 'grafana.theme.changed', payload: { mode } });
+      lastSentMode = mode;
+    }
+  } catch (err) {
+    console.warn('[pmm-compat] failed to post initial grafana.theme.changed:', err);
+  }
+})();
+
+// React to Grafana ThemeChangedEvent (Preferences change/changeTheme())
+getAppEvents().subscribe(ThemeChangedEvent, (evt: unknown) => {
```
matejkubinec: You're subscribing to theme changes on multiple places each with a different handling why is that?

ui/apps/pmm-compat/src/theme.ts
```diff
@@ -20,19 +18,154 @@ export const changeTheme = async (themeId: 'light' | 'dark'): Promise<void> => {
...
+  }
+  if (html.getAttribute('data-md-color-scheme') !== scheme) {
+    html.setAttribute('data-md-color-scheme', scheme);
+  }
+  (html.style as CSSStyleDeclaration).colorScheme = mode;
+
+  return mode;
+}
+
+const isIp = (h: string) => /^\d+\.\d+\.\d+\.\d+$/.test(h);
```
matejkubinec: The communication between `iframe` and `pmm-ui` has it's place in the `CrossFrameMessenger` (`ui/packages/shared/src/messenger.ts`)

ui/apps/pmm-compat/src/theme.ts
```diff
@@ -20,19 +18,154 @@ export const changeTheme = async (themeId: 'light' | 'dark'): Promise<void> => {
...
+    }
+    // Lock to the explicit origin provided by the parent
+    targetOriginRef.current = e.origin || targetOriginRef.current;
+    window.removeEventListener('message', onAck);
+  };
+  window.addEventListener('message', onAck);
+})();
+
+// Initial apply from current Grafana theme and notify parent once
+(function initThemeBridge() {
```
matejkubinec: Declaring a function and then immediately calling it like this might lead to unexpected results - this will be called when the `theme.ts` file is imported and the communication between `iframe <> pmm-ui` might not be initialized (or the plugin will not be inside an iframe at all)

ui/apps/pmm/src/components/sidebar/nav-item/NavItem.tsx
```diff
@@ -32,6 +34,33 @@ const NavItem: FC<NavItemProps> = ({ item, drawerOpen, level = 0 }) => {
   const dataTestid = `navitem-${item.id}`;
   const navigate = useNavigate();
 
+  // Detect "theme-toggle" item and compute dynamic label/icon from current palette mode.
+  const isThemeToggle = item.id === 'theme-toggle';
```
matejkubinec: The handling of the theme toggle navitem shouldn't be in a generic NavItem component see `ui/apps/pmm/src/contexts/navigation/navigation.utils.ts` and the `addAcount` function how the toggle is currently implemented.

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,45 +1,195 @@
...
 import { updateDocumentTitle } from 'lib/utils/document.utils';
-import { useColorMode } from 'hooks/theme';
 import { useKioskMode } from 'hooks/utils/useKioskMode';
+import { useColorMode } from 'hooks/theme';
+import { useSetTheme } from 'themes/setTheme';
+
+type Mode = 'light' | 'dark';
+
+/** Minimal runtime shape of our messenger to avoid `any`. */
+type MessengerLike = {
```
matejkubinec: Why is this needed at all?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,45 +1,195 @@
...
   }, [isGrafanaPage]);
 
+  // Lazily import and register messenger when iframe area is ready (browser only).
+  useEffect(() => {
+    if (!isLoaded || !isBrowser()) return;
+
+    let mounted = true;
+    (async () => {
+      try {
+        const mod = await import('lib/messenger');
```
matejkubinec: What's the reason behind lazily importing the messenger?

ui/apps/pmm/src/hooks/useGrafanaThemeSyncOnce.ts
```diff
@@ -0,0 +1,90 @@
...
+    __pmm_has_theme_listener__?: string;
+  }
+}
+
+/**
+ * One-time sync with Grafana preferences and live postMessage events.
+ * In dev we accept any origin to support split hosts/ports (vite + docker).
+ * In prod we only accept messages from allowed origins.
+ */
+export function useGrafanaThemeSyncOnce(
```
matejkubinec: why?

ui/apps/pmm/src/hooks/useGrafanaThemeSyncOnce.ts
```diff
@@ -0,0 +1,90 @@
...
+      });
+      colorModeRef.current = desired;
+      try {
+        localStorage.setItem('colorMode', desired);
+      } catch (err) {
+        console.warn('[useGrafanaThemeSyncOnce] localStorage set failed:', err);
+      }
+    };
+
+    fetch('/graph/api/user/preferences', { credentials: 'include' })
```
matejkubinec: why fetch instead of using react query?

ui/apps/pmm/src/themes/setTheme.ts
```diff
@@ -0,0 +1,124 @@
...
+/** Idempotently applies theme attributes to <html>. */
+function applyDocumentTheme(mode: Mode) {
+  const html = document.documentElement as HTMLElement & {
+    style: CSSStyleDeclaration & { colorScheme?: string };
+  };
+  const scheme = mode === 'dark' ? 'percona-dark' : 'percona-light';
+  const wantDark = mode === 'dark';
+  const hasDarkClass = html.classList.contains('dark');
+
+  // If everything (including the Tailwind 'dark' class) is already correct, skip.
```
matejkubinec: We don't use tailwind, why is it here?

ui/apps/pmm-compat/src/compat.ts
```diff
@@ -108,10 +102,63 @@ export const initialize = () => {
...
   });
 };
+
+/**
+ * Wires Grafana theme changes (ThemeChangedEvent) to Percona CSS scheme.
+ * Ensures <html> has the attribute our CSS reads and informs the parent frame.
+ */
+function setupThemeWiring() {
+  // Resolve parent origin robustly (dev vs prod)
+  const getParentOrigin = (): string => {
```
matejkubinec: Origin handling shouldn't be needed.

ui/apps/pmm-compat/src/compat.ts
```diff
@@ -108,10 +102,63 @@ export const initialize = () => {
...
+
+  const isDev =
+    location.hostname === 'localhost' ||
+    location.hostname === '127.0.0.1' ||
+    /^\d+\.\d+\.\d+\.\d+$/.test(location.hostname);
+
+  const targetOrigin = isDev ? '*' : getParentOrigin();
+
+  // Helper to apply our scheme attribute and notify parent
+  const apply = (incoming: 'light' | 'dark' | string) => {
```
matejkubinec: Is there a possibility of the theme being other than light/dark at the moment? (In Grafana 12 they have additional themes - https://grafana.com/whats-new/2025-04-11-introducing-experimental-themes/, but we're still on 11)

ui/apps/pmm-compat/src/compat.ts
```diff
@@ -108,10 +102,63 @@ export const initialize = () => {
...
+      const target = window.top && window.top !== window ? window.top : window.parent || window;
+      target?.postMessage({ type: 'grafana.theme.changed', payload: { mode } }, targetOrigin);
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.warn('[pmm-compat] failed to post grafana.theme.changed:', err);
+    }
+  };
+
+  // Initial apply from current Grafana theme config
+  const initialMode = (config?.theme2?.colors?.mode === 'dark' ? 'dark' : 'light') as 'light' | 'dark';
```
matejkubinec: The initial sync shouldn't happen in the body of the file (since it would get called outside the iframe aswell)

ui/apps/pmm-compat/src/compat.ts
```diff
@@ -108,10 +102,63 @@ export const initialize = () => {
...
+  apply(initialMode);
+
+  // React to Grafana theme changes (Preferences change/changeTheme())
+  getAppEvents().subscribe(ThemeChangedEvent, (evt: any) => {
```
matejkubinec: Please move to the body of initialize method.

ui/apps/pmm-compat/src/theme.ts
```diff
@@ -20,19 +18,154 @@ export const changeTheme = async (themeId: 'light' | 'dark'): Promise<void> => {
     newCssLink.rel = 'stylesheet';
     newCssLink.href = config.bootData.assets[newTheme.colors.mode];
     newCssLink.onload = () => {
-      // Remove old css file
```
matejkubinec, line 23: What's the reason for this change to code and comments?

ui/apps/pmm-compat/src/theme.ts
```diff
@@ -20,19 +18,154 @@ export const changeTheme = async (themeId: 'light' | 'dark'): Promise<void> => {
...
+  }
+  return window;
+}
+
+/** Runtime-locked origin (handshake will tighten '*' in dev). */
+const targetOriginRef = { current: resolveInitialTargetOrigin() };
+let lastSentMode: 'light' | 'dark' | null = null;
+
+/** Send helper that always uses the current locked origin. */
+function sendToParent(msg: unknown) {
```
matejkubinec: Posting messages between `iframe` and `pmm-ui` should be handled by CrossFrameMessenger.

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,45 +1,195 @@
...
 import { DocumentTitleUpdateMessage, LocationChangeMessage } from '@pmm/shared';
-import messenger from 'lib/messenger';
 import { getLocationUrl } from './grafana.utils';
 import { updateDocumentTitle } from 'lib/utils/document.utils';
-import { useColorMode } from 'hooks/theme';
 import { useKioskMode } from 'hooks/utils/useKioskMode';
+import { useColorMode } from 'hooks/theme';
+import { useSetTheme } from 'themes/setTheme';
+
+type Mode = 'light' | 'dark';
```
matejkubinec: What's the reason for declaring a `Mode` type when there's `ColorMode` in `ui/packages/shared/src/types.ts`?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,45 +1,195 @@
...
+  }) => void;
+};
+
+type NavState = { fromGrafana?: boolean } | null;
+
+const isBrowser = (): boolean =>
+  typeof window !== 'undefined' && typeof window.addEventListener === 'function';
+
+/** Reads canonical mode from <html> attributes set by our theme hook. */
+const readHtmlMode = (): Mode => {
```
matejkubinec: What's the reason behind adding a color scheme to the html node?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,45 +1,195 @@
...
+  if (!isBrowser()) return 'light';
+  return document.documentElement
+    .getAttribute('data-md-color-scheme')
+    ?.includes('dark')
+    ? 'dark'
+    : 'light';
+};
+
+/** Normalizes any incoming value to 'light' | 'dark'. */
+const normalizeMode = (v: unknown): Mode =>
```
matejkubinec: Do we expect mode to be something other than light/dark?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,45 +1,195 @@
...
+  typeof v === 'string' && v.toLowerCase() === 'dark'
+    ? 'dark'
+    : v === true
+      ? 'dark'
+      : 'light';
+
+/** Resolve optional Grafana origin provided via env (e.g. https://pmm.example.com). */
+const resolveGrafanaOrigin = (): string | undefined => {
```
matejkubinec: Why do we need to resolve the grafana origin and use a VITE_GRAFANA_ORIGIN env variable?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,45 +1,195 @@
...
+  // Ensure our theme context is mounted (also mounts the global theme sync hook elsewhere)
+  useColorMode();
+
+  const { setFromGrafana } = useSetTheme();
+
+  // Remember last theme we sent to avoid resending the same value.
+  const lastSentThemeRef = useRef<Mode>('light');
+
+  // Keep messenger instance lazily loaded and scoped to browser only.
+  const messengerRef = useRef<MessengerLike | null>(null);
```
matejkubinec: Messenger (without any changes) is outside of the react scope, we do we need to keep a ref to it?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -49,64 +199,46 @@ export const GrafanaProvider: FC<PropsWithChildren> = ({ children }) => {
...
+    };
 
-    messenger.sendMessage({
-      type: 'CHANGE_THEME',
-      payload: {
-        theme: colorMode,
-      },
-    });
-  }, [isLoaded, colorMode]);
+    window.addEventListener('message', onMsg);
```
matejkubinec: We shouldn't use window directly to listen for messages from iframe.

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -49,64 +199,46 @@ export const GrafanaProvider: FC<PropsWithChildren> = ({ children }) => {
...
+    ) => {
+      // Security: ignore unexpected origins in production
+      if (!isTrustedOriginRef.current(e.origin)) return;
 
-        updateDocumentTitle(payload.title);
-      },
-    });
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isLoaded]);
+      if (!e?.data || e.data.type !== 'grafana.theme.changed') return;
```
matejkubinec: Use an already predefined `MessageType` or add a new one if needed `ui/packages/shared/src/types.ts`.

ui/apps/pmm/src/hooks/theme.ts
```diff
@@ -1,17 +1,45 @@
...
+
+type Mode = 'light' | 'dark';
 
 export const useColorMode = () => {
   const { colorMode, toggleColorMode } = useContext(ColorModeContext);
-  const { mutate } = useUpdatePreferences();
+  const { mutateAsync } = useUpdatePreferences();
+  const colorModeRef = useRef<Mode>(colorMode);
+
+  useEffect(() => {
```
matejkubinec: if we want to keep `colorModeRef` to be always in sync with `colorMode` you can omit the `useEffect` since it would update the ref after the render.

ui/apps/pmm/src/hooks/theme.ts
```diff
@@ -1,17 +1,45 @@
...
+    // Optimistic UI update
     toggleColorMode();
-    mutate({
-      theme: colorMode === 'light' ? 'dark' : 'light',
-    });
-  };
+    colorModeRef.current = next;
+
+    try {
+      await mutateAsync({ theme: next });
```
matejkubinec: If you need to add handling on error you can use the second parameter of mutate:

```tsx
mutate(
      {
        theme: colorMode === 'light' ? 'dark' : 'light',
      },
      {
        onError: () => {},
      }
    );
```

ui/apps/pmm/src/setupTests.ts
```diff
@@ -1 +1,58 @@
+/**
```
matejkubinec: What is this used for?

ui/apps/pmm/src/themes/setTheme.ts
```diff
@@ -0,0 +1,124 @@
+import { useContext, useRef } from 'react';
+import { ColorModeContext } from '@percona/design';
+import messenger from 'lib/messenger';
+import { grafanaApi } from 'api/api';
+
+type Mode = 'light' | 'dark';
```
matejkubinec: Additional declaration of `Mode` type, it would be better to use `ColorMode` that we have in the shared package.

ui/apps/pmm/src/themes/setTheme.ts
```diff
@@ -0,0 +1,124 @@
...
+
+/** Normalizes any incoming value to 'light' | 'dark'. */
+function normalizeMode(v: unknown): Mode {
```
matejkubinec: Do we expect mode to be something other than light/dark? If we need normalization why are there like 3 different implementations?

ui/apps/pmm/src/themes/setTheme.ts
```diff
@@ -0,0 +1,124 @@
...
+  html.style.colorScheme = mode;
+}
+
+/**
+ * useSetTheme centralizes theme changes for:
+ * - local PMM UI (left)
+ * - persistence to Grafana preferences
+ * - broadcasting to Grafana iframe (right)
+ */
+export function useSetTheme() {
```
matejkubinec: It would be good to centralize the theme handling/changes into a separate provider.

Should load the current theme based on the localStorage/userPreferences and store changes in user preferences when theme changes.

ui/apps/pmm/src/themes/setTheme.ts
```diff
@@ -0,0 +1,124 @@
...
+  ) => {
+    const next = normalizeMode(nextRaw);
+
+    // 1) local apply (instant, idempotent)
+    applyLocal(next);
+
+    // 2) persist to Grafana (only for left-initiated actions)
+    if (opts.persist) {
+      try {
+        await grafanaApi.put('/user/preferences', { theme: next });
```
matejkubinec: what's the reason behind using api directly instead of through react query?

ui/apps/pmm/tsconfig.node.json
```diff
@@ -6,5 +6,7 @@
     "moduleResolution": "bundler",
     "allowSyntheticDefaultImports": true
   },
-  "include": ["vite.config.ts"]
+  "include": [
+    "./vite.config.ts"
```
matejkubinec: Do we need this change?

ui/apps/pmm/vitest.config.ts
```diff
@@ -0,0 +1,47 @@
+/**
```
matejkubinec: Why do we need a specific vitest config?

ui/apps/pmm-compat/src/compat.ts
```diff
@@ -25,44 +26,66 @@ export const initialize = () => {
...
-    type: 'GRAFANA_READY',
+    type: 'GRAFANA_THEME_CHANGED',
+    payload: { theme: initial },
+  });
+
+  // Forward future Grafana ThemeChangedEvent to PMM UI
+  getAppEvents().subscribe(ThemeChangedEvent, (evt: unknown) => {
+    // Grafana 11 emits ThemeChangedEvent with different shapes; normalize robustly.
+    // Try known fields, then fall back to 'light'.
+    const raw =
```
matejkubinec: Could you extract this event parsing into an util function?

ui/apps/pmm-compat/src/compat.ts
```diff
@@ -25,44 +26,66 @@ export const initialize = () => {
...
+  // Initial emit from Grafana current config
+  const initial: ColorMode = config?.theme2?.colors?.mode === 'dark' ? 'dark' : 'light';
   messenger.sendMessage({
-    type: 'GRAFANA_READY',
+    type: 'GRAFANA_THEME_CHANGED',
+    payload: { theme: initial },
+  });
+
+  // Forward future Grafana ThemeChangedEvent to PMM UI
+  getAppEvents().subscribe(ThemeChangedEvent, (evt: unknown) => {
```
matejkubinec: Does evt/event have a default type available (is the unknown necessary here?)

ui/apps/pmm-compat/src/compat.ts
```diff
@@ -80,7 +104,6 @@ export const initialize = () => {
 
   let prevLocation: Location | undefined;
   locationService.getHistory().listen((location: Location, action: HistoryAction) => {
-    // re-add custom toolbar buttons after closing kiosk mode
```
matejkubinec, line 83: Why was the comment removed?

ui/apps/pmm-compat/src/theme.ts
```diff
@@ -1,38 +1,99 @@
...
+};
+
+/**
+ * Initialize theme sync inside the Grafana iframe.
+ * - Single subscription to Grafana ThemeChangedEvent => notify PMM UI (left).
+ * - Listen to CHANGE_THEME from PMM UI => apply locally via changeTheme().
+ * - Perform initial one-shot sync after listeners are in place.
+ * - No IIFEs, no window.postMessage, no origin handshake, no HTML attributes.
+ */
+export const initialize = (messenger: CrossFrameMessenger): void => {
```
matejkubinec: Where's this called? Is it used?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,123 +1,163 @@
...
+import { ColorMode } from '@pmm/shared';
 import { getLocationUrl } from './grafana.utils';
 import { updateDocumentTitle } from 'lib/utils/document.utils';
-import { useColorMode } from 'hooks/theme';
 import { useKioskMode } from 'hooks/utils/useKioskMode';
+import { useColorMode } from 'hooks/theme';
+import { useSetTheme } from 'themes/setTheme';
+import messenger from 'lib/messenger';
+
+type NavState = { fromGrafana?: boolean } | null;
```
matejkubinec: What's the reason for defining those types here? If you need payload types extract and define those types in the shared package.

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,123 +1,163 @@
...
+                // Defensive read: normalize to 'light' | 'dark'
+                const next: ColorMode = message.payload?.theme === 'dark' ? 'dark' : 'light';
+                // Apply locally without re-broadcast/persist
+                setFromGrafana(next).catch((err: unknown) => {
+                    console.warn('[GrafanaProvider] setFromGrafana failed:', err);
+                });
+            },
+        });
+
+        // Grafana -> PMM: route changes (except POP)
```
matejkubinec: Why is there a need to change code that is not related to theme handling?

ui/apps/pmm/src/hooks/theme.ts
```diff
@@ -1,17 +1,28 @@
 import { ColorModeContext } from '@percona/design';
 import { useContext } from 'react';
 import { useUpdatePreferences } from './api/useUser';
+import messenger from 'lib/messenger';
+import type { MessageType } from '@pmm/shared';
 
 export const useColorMode = () => {
-  const { colorMode, toggleColorMode } = useContext(ColorModeContext);
-  const { mutate } = useUpdatePreferences();
+    const { colorMode, toggleColorMode } = useContext(ColorModeContext);
```
matejkubinec: I see multiple changes related to tabWidth the `@percona/prettier-config` tab width is set to 2.

ui/apps/pmm/src/themes/theme.context.ts
```diff
@@ -0,0 +1,16 @@
+import { createContext, useContext } from 'react';
```
matejkubinec: Is this even used?

ui/apps/pmm/src/themes/setTheme.ts
matejkubinec: @dmitri-saricev-3pillargloball I was thinking in realms of something like this below (keep in mind I haven't tested the code):

```tsx
import { ChangeThemeMessage, ColorMode } from '@pmm/shared';
import {
  createContext,
  FC,
  PropsWithChildren,
  useCallback,
  useContext,
  useEffect,
  useMemo,
  useRef,
  useState,
} from 'react';
import { ThemeProvider as EmotionThemeProvider } from '@emotion/react';
import { CssBaseline, createTheme } from '@mui/material';
import { useUpdatePreferences } from 'hooks/api/useUser';
import pmmThemeOptions from 'themes/PmmTheme';
import { ColorModeContext } from '@percona/design';
import messenger from 'lib/messenger';

interface ThemeProviderProps {
  colorMode: ColorMode;
  setColorMode: (mode: ColorMode) => void;
}

export const ThemeContext = createContext<ThemeProviderProps>({
  colorMode: 'light',
  setColorMode: () => {},
});

export const ThemeProvider: FC<PropsWithChildren> = ({ children }) => {
  const [colorMode, setColorMode] = useState<ColorMode>('light');

  const theme = useMemo(
    () => createTheme(pmmThemeOptions(colorMode)),
    [colorMode]
  );

  const toggleColorMode = useCallback(
    () => setColorMode(colorMode === 'light' ? 'dark' : 'light'),
    [colorMode, setColorMode]
  );

  return (
    <ThemeContext.Provider
      value={{
        colorMode,
        setColorMode,
      }}
    >
      {/* reimplement providers for ColorMode and Emotion/Mui theme */}
      <ColorModeContext.Provider value={{ colorMode, toggleColorMode }}>
        <EmotionThemeProvider theme={theme}>
          <CssBaseline />
          {children}
        </EmotionThemeProvider>
      </ColorModeContext.Provider>
    </ThemeContext.Provider>
  );
};

// UNTESTED
export const UserProvider: FC = () => {
  // ...
  const userPreferencesQuery = useUserPreferences();
  const { setColorMode } = useTheme();

  useEffect(() => {
    // update the color mode to match what is in the user preferences
    setColorMode(userPreferencesQuery.data.theme);
  }, [userPreferencesQuery.data.theme]);

  return null;
};

export const useTheme = () => useContext(ThemeContext);

// UNTESTED
const GrafanaProvider: FC<PropsWithChildren> = () => {
  const { colorMode } = useTheme();
  const { mutate: updatePrefrences } = useUpdatePreferences();
  const colorModeGrafana = useRef<ColorMode>();
  const colorModePmm = useRef<ColorMode>(colorMode);

  colorModePmm.current = colorMode;

  useEffect(() => {
    messenger.addListener({
      type: 'CHANGE_THEME',
      onMessage: (msg: ChangeThemeMessage) => {
        if (msg.payload?.theme) {
          colorModeGrafana.current = msg.payload.theme;

          // don't update prefrences if it's the same as the current theme
          if (colorModePmm.current !== colorModeGrafana.current) {
            // mutation should be updated to update the data of userPreferencesQuery
            updatePrefrences({ theme: msg.payload.theme });
          }
        }
      },
    });
  }, []);

  useEffect(() => {
    // check if the theme is the same as the one from Grafana
    if (colorModeGrafana.current !== colorMode) {
      messenger.sendMessage({
        type: 'CHANGE_THEME',
        payload: {
          theme: colorMode,
        },
      });
    }
  }, [colorMode]);

  return null;
};

```

ui/apps/pmm-compat/src/compat.ts
```diff
@@ -25,44 +27,57 @@ export const initialize = () => {
...
   applyCustomStyles();
-
   adjustToolbar();
 
-  // sync with PMM UI theme
-  changeTheme('light');
-
+  // -------- Theme relay: Grafana (right) → PMM UI (left) --------
+  // Initial emit from Grafana current config
+  const initial: ColorMode = parseThemeChangedEvent({
```
matejkubinec: Do we need to send the theme to pmm-ui once the grafana frame is loaded? It cane be loaded at a later time that the rest of ui (loads when a grafana page is needed) PMM-UI should be in sync with theme through the user/preferences.

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,111 +1,142 @@
...
-        payload: {
-          theme: colorMode,
-        },
-      });
+    // Theme: apply without re-broadcast/persist (avoid ping-pong)
+    messenger.addListener({
+      type: 'GRAFANA_THEME_CHANGED',
+      onMessage: (message: { payload?: { theme?: ColorMode } }) => {
+        const next: ColorMode =
+          message.payload?.theme === 'dark' ? 'dark' : 'light';
```
matejkubinec: Is there a possibility of the theme in payload to be different then dark/light?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,111 +1,142 @@
...
+        setFromGrafana(next).catch((err: unknown) => {
+          console.warn('[GrafanaProvider] setFromGrafana failed:', err);
+        });
+      },
     });
 
+    // Location: navigate PMM when Grafana pushes/replace (skip POP/back)
     messenger.addListener({
```
matejkubinec, line 76: What's the reason to modify this listener?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,111 +1,142 @@
...
 
-        navigate(getLocationUrl(location), {
+        navigate(getLocationUrl(adapted), {
           state: { fromGrafana: true },
           replace: true,
         });
       },
     });
 
-    messenger.addListener({
```
matejkubinec, line 85: Where did the listener for `DOCUMENT_TITLE_CHANGE` disappear?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,111 +1,142 @@
...
+        action: navigationType,
       },
     });
+  }, [location, navigationType]);
+
+  // PMM -> Grafana: propagate theme when left-side theme changes
+  useEffect(() => {
+    if (!isLoaded || !isBrowser()) return;
+
+    const mode: ColorMode = colorMode === 'dark' ? 'dark' : 'light';
```
matejkubinec: colorMode shouldn't be anything other the dark/light, why do need to do this check?

ui/apps/pmm/src/themes/setTheme.ts
```diff
@@ -0,0 +1,93 @@
...
+  const { colorMode, toggleColorMode } = useContext(ColorModeContext);
+  const modeRef = useRef<ColorMode>(normalizeMode(colorMode));
+
+  const { mutateAsync: updatePreferences } = useUpdatePreferences();
+
+  // Keep the reference always up-to-date with current color mode
+  modeRef.current = normalizeMode(colorMode);
+
+  /** Apply theme locally via design system only (no direct DOM mutations). */
+  const applyLocal = (nextRaw: unknown) => {
```
matejkubinec: Looking at the implementation of the `useColorMode` it doesn't just update the theme locally but also does send a message to grafana + updates preferences.

ui/apps/pmm/src/themes/setTheme.ts
matejkubinec: What's the reason behind having to api calls here (through react query + directly through grafana api)

ui/packages/shared/src/types.ts
```diff
@@ -8,8 +8,11 @@ export type MessageType =
   | 'DASHBOARD_VARIABLES'
   | 'GRAFANA_READY'
   | 'DOCUMENT_TITLE_CHANGE'
+  | 'GRAFANA_THEME_CHANGED'
   | 'CHANGE_THEME';
 
+export type NavState = { fromGrafana?: boolean } | null;
```
matejkubinec: It would be better named `LocationState` (since it's part of location in react router/navigation)

ui/apps/pmm-compat/src/utils/themeEvent.ts
```diff
@@ -0,0 +1,59 @@
+import type { GrafanaTheme2 } from '@grafana/data';
+import { config } from '@grafana/runtime';
+import type { ColorMode } from '@pmm/shared';
+
+/** Grafana 11+ often passes the whole theme object on the event. */
+type ThemeChangedEventV11 = {
```
matejkubinec: We support only Grafana version that matches our fork (at the moment 11.6.4), we don't need to take into account older versions.

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,112 +1,143 @@
...
-          theme: colorMode,
-        },
-      });
+    // -------- INCOMING FROM GRAFANA --------
+
+    // Theme: apply without re-broadcast/persist (avoid ping-pong)
+    messenger.addListener({
+      type: 'GRAFANA_THEME_CHANGED',
+      onMessage: (message: { payload?: { theme?: ColorMode } }) => {
+        const next: ColorMode =
```
matejkubinec: Do we need to normalize the theme here? (I see there's already a normalization in `setFromGrafana/setBaseTheme`.

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,112 +1,143 @@
...
+        setFromGrafana(next).catch((err: unknown) => {
+          console.warn('[GrafanaProvider] setFromGrafana failed:', err);
+        });
+      },
     });
 
+    // Location: navigate PMM when Grafana pushes/replace (skip POP/back)
     messenger.addListener({
       type: 'LOCATION_CHANGE',
-      onMessage: ({ payload: location }: LocationChangeMessage) => {
```
matejkubinec, line 73: Do we need to modify this method since it isn't related to theme handling?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,112 +1,143 @@
...
           state: { fromGrafana: true },
           replace: true,
         });
       },
     });
 
+    // Grafana -> PMM: document title
     messenger.addListener({
       type: 'DOCUMENT_TITLE_CHANGE',
-      onMessage: ({ payload }: DocumentTitleUpdateMessage) => {
```
matejkubinec, line 87: What's the reason for changing the implementation here? `updateDocumentTitle` does a bit more than `document.title = payload.title;`

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,112 +1,143 @@
...
+const isBrowser = () =>
+  typeof window !== 'undefined' &&
+  typeof window.addEventListener === 'function';
 
+/**
+ * GrafanaProvider wires two bridges:
+ * 1) THEME: PMM (left) ↔ Grafana iframe (right)
+ * 2) ROUTING: PMM (left) ↔ Grafana iframe (right)
+ *
+ * Scope limited to the above; document title and other concerns are excluded.
```
matejkubinec: It does handle the document title change.

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -1,112 +1,143 @@
...
+    };
+  }, [isLoaded, navigate, setFromGrafana]);
+
+  // -------- OUTGOING TO GRAFANA --------
+  // PMM -> Grafana: propagate PMM location (except if it came from Grafana)
   useEffect(() => {
-    if (!isLoaded) {
-      return;
-    }
+    if (!isBrowser()) return;
```
matejkubinec: It probably is missing the isLoaded condition (altough, it will not propagate when not loaded since the pathname wont be `/graph`, it would be good to have the guard here)

ui/apps/pmm-compat/src/utils/themeEvent.ts
dmitri-saricev-3pillargloball: About this, I said in the call that if I try to resolve this, a small bug appears in the first render, or I need to write a lot of code to resolve this issue. Much more than there is now.

PR #4689, opened 2025-10-20
PMM-13812 Clickhouse HA fixes.

JiriCtvrtka:
PMM-13812

Link to the Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/4047

- [ ] API Docs updated


qan-api2/migrations/migrations.go
```diff
@@ -94,33 +96,37 @@ func GetEngine(dsn string) string {
 }
 
 func Run(dsn string, templateData map[string]any, isCluster bool, clusterName string) error {
```
ademidoff, line 95: So the variable `isCluster` is no longer needed, right?

qan-api2/db.go
```diff
@@ -67,7 +68,8 @@ func NewDB(dsn string, maxIdleConns, maxOpenConns int, isCluster bool, clusterNa
 	db, err := sqlx.Connect("clickhouse", dsn)
 	if err != nil {
 		log.Printf("error connecting to clickhouse: %v", err)
-		if exception, ok := err.(*clickhouse.Exception); ok && exception.Code == databaseNotExistErrorCode { //nolint:errorlint
+		var exception *clickhouse.Exception
+		if errors.As(err, &exception) && exception.Code == databaseNotExistErrorCode {
```
ademidoff, line 72: 👍

qan-api2/migrations/migrations.go
JiriCtvrtka: It’s still needed, because only when isCluster is true we can call isClickhouseClusterReady. Otherwise, it will get stuck waiting.

PR #4705, opened 2025-10-24
Encryption - fix typos and improve clarity

ademidoff:
This pull request makes minor improvements to the documentation for PMM's data encryption configuration. The changes clarify instructions and correct some wording to improve readability.

- Documentation clarity:
  * Corrected the wording to refer to a "custom key" instead of "custom keys" and clarified that the environment variable should be set before upgrading to PMM 3 or starting a new PMM 3.x instance.
  * Improved phrasing to state that PMM uses "the custom key" for encryption and decryption, rather than "custom keys".

PR #4706, opened 2025-10-24
PMM-7 Encryption key format in doc.

JiriCtvrtka:
PMM-7


ademidoff:
I suggest we keep the changes in one PR - https://github.com/percona/pmm/pull/4705.

PR #4717, opened 2025-10-30
PMM-14375: Propagate environment variables to agents

idoqo:
PMM-14375

Link to the Feature Build: SUBMODULES-4092


managed/models/agent_model.go
```diff
@@ -379,6 +389,16 @@ func (s *Agent) SetCustomLabels(m map[string]string) error {
 	return setLabels(m, &s.CustomLabels)
 }
 
+// GetEnvironmentVariables decodes environment variables.
+func (s *Agent) GetEnvironmentVariables() (map[string]string, error) {
```
BupycHuk: if error isn't used anywhere why do we return it?

BupycHuk (changes requested):
I don't see how this env variable is passed to QAN agent.

idoqo:
> I don't see how this env variable is passed to QAN agent.

Based on the discussion [here](https://perconadev.atlassian.net/browse/PMM-14375?focusedCommentId=431126), I don't think we need to do anything special to share env variables with qan agents - they are still part of the original pmm-agent process and have access to the same env variables - will remove them from the proto definitions for qan agents as well

managed/models/agent_model.go
idoqo: it was because of the underlying getLabel() function - will update the callers to use the errors.

BupycHuk:
So you have to set that env variable for pmm-agent and also pass it in pmm-admin command? isn't it too complex?

idoqo:
> So you have to set that env variable for [pmm-agent](https://perconadev.atlassian.net/browse/PMM-agent) and also pass it in [pmm-admin](https://perconadev.atlassian.net/browse/PMM-admin) command? isn't it too complex?

yeahh opened to a different flow really, one option will be to automatically pass pmm-agent environment variables to exporters. Sharing all of it might not be a good idea, hence, we might still need some flag to specify what variables should be shared (only difference is that this time, user only specifies the variable names, not values).

admin/commands/management/add_mongodb.go
```diff
@@ -179,16 +184,17 @@ func (cmd *AddMongoDBCommand) RunCmd() (commands.Result, error) {
...
-				MaxQueryLength:                cmd.MaxQueryLength,
-				TLS:                           cmd.TLS,
-				TLSSkipVerify:                 cmd.TLSSkipVerify,
-				TLSCertificateKey:             tlsCertificateKey,
-				TLSCertificateKeyFilePassword: cmd.TLSCertificateKeyFilePassword,
-				TLSCa:                         tlsCa,
-				AuthenticationMechanism:       cmd.AuthenticationMechanism,
-				AuthenticationDatabase:        cmd.AuthenticationDatabase,
+				CustomLabels:                   customLabels,
+				SharedEnvironmentVariableNames: agentVarNames,
```
ademidoff:
```suggestion
				EnvironmentVariableNames: agentVarNames,
```
It can be made shorter, envars are mostly shared by definition.

admin/commands/base.go
```diff
@@ -184,6 +184,31 @@ func ParseDisableCollectors(collectors []string) []string {
...
+	if len(varNames) == 0 {
+		return nil, nil
+	}
+
+	result := make([]string, 0, len(varNames))
+	validNamePattern := regexp.MustCompile(`^[A-Z_][A-Z0-9_]*$`)
+
+	for _, name := range varNames {
+		name = strings.TrimSpace(name)
+		if name == "" {
```
ademidoff, line 198: If the variable name is empty, we may have to signal an error to the user, who has probably no intention to pass empty names. WDYT?

api-tests/management/mongodb_test.go
```diff
@@ -597,9 +597,9 @@ func TestAddMongoDB(t *testing.T) {
 				},
 			},
 		}
-		addProxySQLOK, err := client.Default.ManagementService.AddService(params)
+		addMongoDBOK, err := client.Default.ManagementService.AddService(params)
 		pmmapitests.AssertAPIErrorf(t, err, 400, codes.InvalidArgument, "Socket and address cannot be specified together.")
-		assert.Nil(t, addProxySQLOK)
+		assert.Nil(t, addMongoDBOK)
```
ademidoff, line 602: 👍

api/agent/v1/agent.proto
```diff
@@ -80,6 +82,8 @@ message SetStateRequest {
     bool tls = 8;
     // TLS certificate wont be verified.
     bool tls_skip_verify = 9;
+    // environment variables to be passed to the built-in agent.
```
ademidoff:
```suggestion
    // Environment variables to be passed to the built-in agent.
```

api/management/v1/mongodb.proto
```diff
@@ -92,6 +92,9 @@ message AddMongoDBServiceParams {
   inventory.v1.LogLevel log_level = 33;
   // Optionally expose the exporter process on all public interfaces
   bool expose_exporter = 34;
+  // Environment variable names to pass to the exporter.
+  // Values will be resolved from pmm-agent's environment when starting the exporter.
+  repeated string shared_environment_variable_names = 36;
```
ademidoff:
```suggestion
  repeated string environment_variable_names = 36;
```
Let's shorten it a bit :)

managed/models/agent_helpers.go
```diff
@@ -759,25 +759,26 @@ func CreateExternalExporter(q *reform.Querier, params *CreateExternalExporterPar
...
-	PostgreSQLOptions PostgreSQLOptions
-	ValkeyOptions     ValkeyOptions
+	PMMAgentID                     string
+	NodeID                         string
+	ServiceID                      string
+	Username                       string
+	Password                       string
+	AgentPassword                  string
+	CustomLabels                   map[string]string
+	SharedEnvironmentVariableNames []string
```
ademidoff:
```suggestion
	EnvironmentVariableNames []string
```

managed/models/database.go
```diff
@@ -1149,6 +1149,9 @@ var databaseSchema = [][]string{
 	112: {
 		`UPDATE agents SET disabled = true WHERE agent_type = 'qan-postgresql-pgstatements-agent' AND service_id = (SELECT service_id FROM services WHERE service_name = 'pmm-server-postgresql' LIMIT 1);`,
 	},
+	113: {
+		`ALTER TABLE agents ADD COLUMN shared_environment_variables TEXT`,
```
ademidoff:
```suggestion
		`ALTER TABLE agents ADD COLUMN environment_variables TEXT`,
```

ademidoff (approved):

JiriCtvrtka (approved):

BupycHuk (approved):

PR #4722, opened 2025-10-31
PMM-7 Fix unit tests

idoqo:
PMM-0

- [ ] API Docs updated

### DIFF
--- a/agent/runner/actions/mongodb_explain_action_test.go
+++ b/agent/runner/actions/mongodb_explain_action_test.go
@@ -304,6 +304,8 @@ func TestMongoDBExplain(t *testing.T) {
 			want["optimizationTimeMillis"] = map[string]interface{}{"$numberInt": "0"}
 			want["winningPlan"] = map[string]interface{}{"stage": "EOF", "isCached": false}
 			want["prunedSimilarIndexes"] = false
+		case mongoDBVersion.Major == 8 && mongoDBVersion.Minor >= 2:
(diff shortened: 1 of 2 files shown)


ademidoff (approved):

agent/runner/actions/mongodb_explain_action_test.go
```diff
@@ -304,6 +304,9 @@ func TestMongoDBExplain(t *testing.T) {
 			want["optimizationTimeMillis"] = map[string]interface{}{"$numberInt": "0"}
 			want["winningPlan"] = map[string]interface{}{"stage": "EOF", "isCached": false}
 			want["prunedSimilarIndexes"] = false
+			if mongoDBVersion.Minor >= 2 {
```
ademidoff, line 307: @idoqo If this is applicable to 8.2+, then we should probably check the major version as well. Otherwise it may fail for 7.2, for example.

JiriCtvrtka (approved):

agent/runner/actions/mongodb_explain_action_test.go
idoqo: yeahh though this is already wrapped in the `case mongoDBVersion.Major == 8:` block above.

But issue I see is when 9.x gets released with the same field in the map, it will likely fail again.

ademidoff: Ah, Ok then, we will come back to it once 9.x comes out.

PR #4785, opened 2025-11-27
PMM-14325 HA Badge

matejkubinec:
PMM-14325

Link to the Feature Build: SUBMODULES-4150

**Screenshots**:

Screenshot 2025-12-18 at 13 14 00


ui/apps/pmm/src/hooks/api/useHA.ts
```diff
@@ -0,0 +1,50 @@
...
+    ...options,
+  });
+
+export const useHaInfo = () => {
+  const statusQuery = useHAStatus();
+  const nodesQuery = useHANodes({
+    enabled: statusQuery.data?.status === 'Enabled',
+    refetchInterval: 5000,
+  });
+
```
dmitri-saricev-3pillargloball, line 34: It is only a suggestion)))
With 100 active users, this creates 20 requests/sec to /ha/nodes. For HA monitoring, 15-30 seconds should be sufficient and more backend-friendly.

 const nodesQuery = useHANodes({                                                                                                                  
    enabled: statusQuery.data?.status === 'Enabled',                                                                                               
    refetchInterval: 15000, // 15s is enough for HA status updates                                                                                 
    refetchIntervalInBackground: false, // Stop polling when tab is inactive                                                                       
  });

ui/apps/pmm/src/utils/ha.utils.ts
```diff
@@ -0,0 +1,19 @@
+import { GetHANodeResponse, HAHealth } from 'types/ha.types';
+
+export const getHAHealth = (nodes: GetHANodeResponse[]): HAHealth => {
+  const nonAliveCount = nodes.filter((node) => node.status !== 'alive').length;
+
```
dmitri-saricev-3pillargloball, line 5: suggestion

Currently, getHAHealth([]) returns 'healthy' because all checks fail through. Should an empty cluster be considered healthy?                      

  export const getHAHealth = (nodes: GetHANodeResponse[]): HAHealth => {                                                                           
    if (nodes.length === 0) {                                                                                                                      
      return 'down'; // or 'unknown' depending on business logic                                                                                   
    }                                                                                                                                              
                                                                                                                                                   
    const nonAliveCount = nodes.filter((node) => node.status !== 'alive').length;                                                                  
    // ... rest of logic                                                                                                                           
  };

matejkubinec: There shouldn't be an empty cluster. (then it wouldn't be HA)

ui/apps/pmm/src/hooks/api/useHA.ts
matejkubinec: Interval decreased.

dmitri-saricev-3pillargloball (approved):

PR #4801, opened 2025-12-02
PMM-14442 Increasing PG connections limit.

JiriCtvrtka:
PMM-14442

FB: https://github.com/Percona-Lab/pmm-submodules/pull/4133

All other changes are already merged.


ademidoff (approved):

idoqo (approved):

ademidoff:
@JiriCtvrtka What is the default limit? Can we leave a note wrt the default?

JiriCtvrtka:
> @JiriCtvrtka What is the default limit? Can we leave a note wrt the default?

The default value for max_connections is 100. I can leave a comment with the default, but for example, in the case of pg_stat_statements.max https://github.com/percona/pmm/pull/4801/changes#diff-e1084fce306f7ddce3eb67a9c69572dd9d39d6922b22c19e8fcd6e26c3aee2f5L34, we also don’t have a comment with the default value. So I’m not sure whether it makes sense to add one here.

PR #4804, opened 2025-12-04
PMM-14587 Add custom header for RT QAN

matejkubinec:
- Adds tab setup for switching between Historical (TBD) and Real-Time QAN
- Provides placeholders for running agents modal & RTA selection page
- Handles copy for both rta and legacy qan pages

PMM-14587 

Link to the Feature Build: SUBMODULES-4136

**Screenshots**:

screenshot-2025-12-04-15-42-07

screenshot-2025-12-04-15-42-11

screenshot-2025-12-04-15-42-14


ui/apps/pmm/src/components/main/header/qan-header/qan-header-actions/QanHeaderActions.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const QanHeaderActions: FC = () => {
+  const { mutateAsync: createShortUrl } = useCreateShortUrl();
+  const [openRunningAgentsModal, setOpenRunningAgentsModal] = useState(false);
+
+  const handleCopy = async () => {
+    const path = constructUrl(location).replace('/pmm-ui/next/graph/', '');
+    const res = location.pathname.includes('/graph')
+      ? await createShortUrl(path)
+      : { url: window.location.href };
+
```
dmitri-saricev-3pillargloball, line 22: Maybe it will be a good solution to add error handling!
Example:
 const handleCopy = async () => {
    try {
      const path = constructUrl(location).replace('/pmm-ui/next/graph/', '');
      const res = location.pathname.includes('/graph')
        ? await createShortUrl(path)
        : { url: window.location.href };

      if (navigator.clipboard && window.isSecureContext) {
        await navigator.clipboard.writeText(res.url);
        enqueueSnackbar('Link copied to clipboard', { variant: 'success' });
      } else {
        enqueueSnackbar('Clipboard is not available', { variant: 'error' });
      }
    } catch (error) {
      enqueueSnackbar('Failed to create short URL', { variant: 'error' });
    }
  };

ui/apps/pmm-compat/src/lib/utils/document.ts
```diff
@@ -6,3 +6,12 @@ export const documentTitleObserver = {
...
     }),
 };
+
+export const updateBodyClassByLocation = (location: Location) => {
+  const previous = Array.from(document.body.classList).find((className) => className.startsWith('grafana-compat-page'));
+  if (previous) {
+    document.body.classList.remove(previous);
+  }
+  const className = `grafana-compat-page${location.pathname.replace('/graph', '').replaceAll('/', '-')}`;
+  document.body.classList.add(className);
```
dmitri-saricev-3pillargloball, line 20: Quick question about `document.ts:15` - should we add character filtering when generating class names from `location.pathname`?

  Current:

`  const className = `grafana-compat-page${location.pathname.replace('/graph', '').replaceAll('/', '-')}`;`


  Since this goes directly to document.body.classList.add() can cause issues with special characters in the pathname. Would something like this be worth adding?

  ```
const sanitized = location.pathname
    .replace(/[^a-zA-Z0-9/-]/g, '')  // Allow only safe chars
    .replace('/graph', '')
    .replaceAll('/', '-');
```


  Would something like this be worth adding, or do you think it's overkill for our use case?

dmitri-saricev-3pillargloball (approved):

PR #4813, opened 2025-12-08
PMM-14615: Tour should not be dismissible when clicking on backdrop

dmitri-saricev-3pillargloball:
[[PMM-0
](https://perconadev.atlassian.net/browse/PMM-14615)](https://perconadev.atlassian.net/browse/PMM-14615)

result:

https://github.com/user-attachments/assets/6d688955-eef9-40e5-bcb4-410306a7602d

FB
https://github.com/Percona-Lab/pmm-submodules/pull/4142
https://pmm.cd.percona.com/blue/organizations/jenkins/pmm3-submodules/detail/pmm3-submodules/2033/pipeline/


ui/apps/pmm/src/contexts/tour/tour.provider.tsx
```diff
@@ -96,7 +96,9 @@ export const TourProvider: FC<PropsWithChildren> = ({ children }) => {
           currentStep={currentStep}
           setCurrentStep={setCurrentStep}
           onClickClose={endTour}
-          onClickMask={endTour}
+          onClickMask={() => {
```
matejkubinec, line 99: Can we just omit the property, instead of putting a blank function there? (or use `false` if omitting is not enough)

dmitri-saricev-3pillargloball: I initially tried omitting the property, but that caused issues: the tour would close on backdrop click and wouldn't reopen properly. The library internally handles the backdrop click and changes state, so we need to explicitly provide a no-op handler to prevent the default closing behavior. Using an empty function is the most straightforward way to disable this without interfering with the library's internal state management.

matejkubinec (approved):

YashSartanpara1 (approved):

PR #4816, opened 2025-12-08
PMM-14619 Use meta key only on Apple platforms

matejkubinec:
PMM-14619

Link to the Feature Build: SUBMODULES-4144


dmitri-saricev-3pillargloball (approved):

ui/apps/pmm-compat/src/lib/utils/navigator.ts
```diff
@@ -0,0 +1 @@
+export const isUserAgentApple = () => /Mac|iPod|iPhone|iPad/.test(navigator.userAgent);
```
ademidoff:
```suggestion
export const isUserAgentApple = /Mac|iPod|iPhone|iPad/.test(navigator.userAgent);
```
Does it need to be a function?

ademidoff (approved):

PR #4822, opened 2025-12-10
PMM-14625 Add PMM HA endpoints

ademidoff:
PMM-14625

Link to the Feature Build: SUBMODULES-4148

- [ ] API Docs updated


managed/services/ha/ha.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return &hav1beta1.ListNodesResponse{Nodes: nodes}, nil
+}
+
+// memberlistStateToString converts memberlist state to a string representation.
+func memberlistStateToString(state memberlist.NodeStateType) string {
+	switch state {
+	case memberlist.StateAlive:
+		return "alive"
```
idoqo, line 89: maybe we need enums/aliased types for these statuses?

idoqo (approved):

managed/services/ha/ha.go
ademidoff: Well, `memberlist.NodeStateType` is already an enum. Maybe it's worth creating a map, but frankly we are not using it anywhere for now, the only value that is used in FE is "alive". 

Let's revisit it once we need it on BE.

JiriCtvrtka (approved):

PR #4848, opened 2025-12-16
PMM-14577 Fix CVEs.

JiriCtvrtka:
PMM-14577

PMM Dump PR: https://github.com/percona/pmm-dump/pull/343
Spec: https://github.com/Percona-Lab/spec/pull/7


go.mod
```diff
@@ -11,7 +11,7 @@ replace github.com/go-openapi/spec => github.com/Percona-Lab/spec v0.21.0-percon
 
 replace github.com/alecthomas/kingpin/v2 => github.com/Percona-Lab/kingpin/v2 v2.4.0-percona
```
JiriCtvrtka, line 12: No newer versions for those.

go.mod
```diff
@@ -5,13 +5,11 @@ go 1.25.5
...
-replace github.com/grpc-ecosystem/go-grpc-prometheus => github.com/Percona-Lab/go-grpc-prometheus v0.0.0-20230116133345-3487748d4592
+replace github.com/go-openapi/spec => github.com/JiriCtvrtka/spec v0.0.0-20251222101011-485c12f10ee3
 
-replace github.com/go-openapi/spec => github.com/Percona-Lab/spec v0.21.0-percona
+replace github.com/alecthomas/kingpin/v2 => github.com/Percona-Lab/kingpin/v2 v2.0.0-20251219125400-95ff74a47c0a
 
-replace github.com/alecthomas/kingpin/v2 => github.com/Percona-Lab/kingpin/v2 v2.4.0-percona
-
-replace golang.org/x/crypto => github.com/percona-lab/crypto v0.0.0-20231108144114-756dfb24eaf2
+replace golang.org/x/crypto => github.com/percona-lab/crypto v0.0.0-20251219125031-82d8ba2cda22
```
JiriCtvrtka: Before merge of PR all will be replaced with proper commit (after those are merged).

qan-api2/main.go
```diff
@@ -106,7 +106,6 @@ func runGRPCServer(ctx context.Context, db *sqlx.DB, mbm *models.MetricsBucket,
 		l.Debug("RPC response latency histogram enabled.")
 		grpc_prometheus.EnableHandlingTimeHistogram()
 	}
-	grpc_prometheus.Register(grpcServer)
```
ademidoff, line 110: I wonder, is this an intended removal?

.golangci.yml
```diff
@@ -110,6 +110,7 @@ linters:
           - recvcheck
         path: managed/models/
     paths:
+      - managed/utils/interceptors/go-grpc-prometheus.go
```
ademidoff, line 114: Do we to leave this disabled or is it a temp workaround?

JiriCtvrtka: I would leave it disabled (a lot of lint issues there) until go-grpc-prometheus is replaced by go-grpc-middleware. Code in this file will be deleted/rewritten.

qan-api2/main.go
JiriCtvrtka: Yes it is. Should be registered with another one. Two different registrations for same metrics are forbidden with new version. Everything should work properly, but testing will give us confirmation. With middleware usage will be probably different anyway.

go.mod
```diff
@@ -5,13 +5,11 @@ go 1.25.5
 // Update saas with
 // go get -v github.com/percona/saas@latest
 
-replace github.com/grpc-ecosystem/go-grpc-prometheus => github.com/Percona-Lab/go-grpc-prometheus v0.0.0-20230116133345-3487748d4592
+replace github.com/go-openapi/spec => github.com/JiriCtvrtka/spec v0.0.0-20251222101011-485c12f10ee3
```
JiriCtvrtka: I dont have permissions to this repo. This is only for now. Once finished it will be replaced with proper commit in Percona spec repo.

managed/utils/interceptors/go-grpc-prometheus.go
```diff
@@ -0,0 +1,389 @@
+// Copyright (C) 2023 Percona LLC
```
idoqo, line 1: if this is copied - should we keep their license?

idoqo (approved):

managed/utils/interceptors/go-grpc-prometheus.go
JiriCtvrtka: This file contains mainly code from Percona fork itself. It is bare minimum to make PMM work with upstream one. So I think it this case it is fine.

go.mod
ademidoff: I believe we have a new version now.

ademidoff (approved):

PR #4857, opened 2025-12-18
PMM-14651 Improve & fix dashboard menu variables sharing

matejkubinec:
PMM-14651

Link to the Feature Build: SUBMODULES-4165


ui/apps/pmm-compat/src/lib/utils/variables.ts
```diff
@@ -42,14 +42,31 @@ export const getLinkWithVariables = (url?: string): string => {
...
+    .replace('/d/', '')
+    .split('/')[0];
+
+  if (dashboardUid.includes('-')) {
+    return dashboardUid.split('-')[0];
+  }
+
+  return dashboardUid;
+};
+
```
dmitri-saricev-3pillargloball, line 79: // Line 7-11
  export const shouldIncludeVars = (url: string): boolean => {
    const currentDB = getDbType(window.location.pathname);
    const targetDB = getDbType(url);
    return (currentDB !== undefined && currentDB === targetDB) || targetDB === 'node';
    //      ^^^^^^^^^^^^^^^^^^^^^^^^ This check doesn't work as intended
  };

  // Line 13-23
  const getDbType = (url: string): string => {
    //                              ^^^^^^ Always returns string, never undefined
    // ...
    return dashboardUid.includes('-') ? dashboardUid.split('-')[0] : dashboardUid;
  };

  The thing is: getDbType() always returns a string (never undefined), but on line 10 shouldIncludeVars() checks currentDB !== undefined. This check will always be true, so it doesn't really do anything.

  Maybe what you meant was:
  // Line 13
  const getDbType = (url: string): string | undefined => {
    const pathname = new URL(url, window.location.origin).pathname;
    const dashboardUid = pathname
      .replace('/pmm-ui', '')
      .replace('/next', '')
      .replace('/graph', '')
      .replace('/d/', '')
      .split('/')[0];

    // Return undefined if we couldn't extract a valid DB type
    if (!dashboardUid || dashboardUid === '') {
      return undefined;
    }

    return dashboardUid.includes('-') ? dashboardUid.split('-')[0] : dashboardUid;
  };

ui/apps/pmm-compat/src/lib/utils/variables.ts
```diff
@@ -42,14 +42,31 @@ export const getLinkWithVariables = (url?: string): string => {
...
 
+const getDbType = (url: string): string => {
+  const pathname = new URL(url, window.location.origin).pathname;
+  // normalize to the dashboard uid
+  const dashboardUid = pathname
+    .replace('/pmm-ui', '')
+    .replace('/next', '')
+    .replace('/graph', '')
+    .replace('/d/', '')
+    .split('/')[0];
```
dmitri-saricev-3pillargloball: .split('/')[0]; -  Could potentially fail if split returns empty array

Not likely in practice since you're working with paths, but it might be worth adding a safety check or a comment explaining why it's safe.

ui/apps/pmm-compat/src/lib/utils/variables.ts
matejkubinec: Updated.

ui/apps/pmm-compat/src/lib/utils/variables.ts
matejkubinec: Updated.

dmitri-saricev-3pillargloball (approved):

fabio-silva (approved):

YashSartanpara1 (approved):

PR #4858, opened 2025-12-19
PMM-14655 Add listeners for settings change and service added

matejkubinec:
PMM-14655

FB: SUBMODULES-4167

Related PRs:
- https://github.com/percona/grafana/pull/864


ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -34,6 +36,13 @@ export const GrafanaProvider: FC<PropsWithChildren> = ({ children }) => {
   const location = useLocation();
   const navigate = useNavigate();
 
+  const { refetch: refetchSettings } = useSettings({
+    enabled: false,
+  });
+  const { refetch: refetchServiceTypes } = useServiceTypes({
+    enabled: false,
+  });
+
```
dmitri-saricev-3pillargloball, line 45: The main thing I noticed - listener cleanup, the listeners are being added, but I don't see them getting removed:

  messenger.addListener({
    type: 'SETTINGS_CHANGED',
    onMessage: () => refetchSettings(),
  });

  This could cause a memory leak if the component remounts (like during navigation) - you'd keep stacking up listeners. The fix is pretty simple, though, just need to return a cleanup function:


  useEffect(() => {
    const settingsListener = messenger.addListener({
      type: 'SETTINGS_CHANGED',
      onMessage: () => refetchSettings(),
    });

    const serviceListener = messenger.addListener({
      type: 'SERVICE_ADDED',
      onMessage: () => refetchServiceTypes(),
    });

    return () => {
      messenger.removeListener(settingsListener);
      messenger.removeListener(serviceListener);
    };
  }, [refetchSettings, refetchServiceTypes]);

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -34,6 +36,13 @@ export const GrafanaProvider: FC<PropsWithChildren> = ({ children }) => {
   const location = useLocation();
   const navigate = useNavigate();
 
+  const { refetch: refetchSettings } = useSettings({
+    enabled: false,
+  });
```
dmitri-saricev-3pillargloball, line 41: Quick question about initial data loading

  I see you're creating these queries with enabled: false:

  const { refetch: refetchSettings } = useSettings({
    enabled: false,
  });

  Just want to make sure I understand - are these queries being run somewhere else on initial load? If not, the UI might not have this data until the first event comes in. Might be worth a quick double-check, or if it's intentional, maybe add a comment so future devs know what's up ))))

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -95,10 +104,22 @@ export const GrafanaProvider: FC<PropsWithChildren> = ({ children }) => {
       },
     });
 
+    messenger.addListener({
+      type: 'SETTINGS_CHANGED',
+      onMessage: () => refetchSettings(),
+    });
+
```
dmitri-saricev-3pillargloball, line 111: **A few other things to consider (not blockers, just ideas):**

Error handling on refetch:
Right now if the refetch fails, it'll be silent. Maybe worth catching errors and showing a toast or logging them? Totally up to you though:

  onMessage: async () => {
    try {
      await refetchSettings();
    } catch (error) {
      console.error('Failed to refetch settings:', error);
    }
  }

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
matejkubinec: The initial fetch of settings is in `SettingsProvider`. This is just used for refetching when triggered from the Grafana side.

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
matejkubinec: Good catch, I've updated the unregister method in messenger to clear the listeners.

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
matejkubinec: I don't think there's a need to notify the user about it failing.

dmitri-saricev-3pillargloball (approved):

fabio-silva (approved):

YashSartanpara1 (approved):

PR #4900, opened 2026-01-08
PMM-14694 Reduce saas package dependency, drop Portal API

idoqo:
PMM-14694
PMM-14012

https://github.com/Percona-Lab/pmm-submodules/pull/4182


managed/pi/logger/global.go
maxkondr, line 1: looks like this is yet another instance of application logger (just copied-pasted from saas repo?). Is it really needed in addition to already existing logger?

managed/utils/signatures/signatures.go
maxkondr, line 1: as I remember signatures were used in order to make sure that received data (advisor, alert template, etc) are really originated by Percona (signed by Percona's private key). in case the templates are moved to pmm repo itself - is it still required to check the signature?

ademidoff: You are raising a good question! Let's prepare for a discussion on the next BE call.

managed/pi/logger/global.go
idoqo: Yeah can be dropped

managed/utils/signatures/signatures.go
ademidoff: FYI: we will be asked to source the advisors and alert templates from a GitHub repo. This is why I'm a bit reluctant to drop them now. It's never late to do that.

managed/pi/check/check.go
```diff
@@ -0,0 +1,587 @@
...
+
+// Validate validates check family.
+func (f Family) Validate() error {
+	switch f {
+	case MySQL:
+		fallthrough
+	case PostgreSQL:
+		fallthrough
+	case MongoDB:
+		return nil
```
ademidoff:
```suggestion
	case MySQL, PostgreSQL, MongoDB:
		return nil
```

managed/pi/check/check.go
```diff
@@ -0,0 +1,587 @@
...
+
+// Validate validates check interval.
+func (i Interval) Validate() error {
+	switch i {
+	case Standard:
+		fallthrough
+	case Frequent:
+		fallthrough
+	case Rare:
+		fallthrough
```
ademidoff: Let's get rid of `fallthrough` in favor of comma-separated cases.

ademidoff (approved):
@idoqo Can we remove checks and alerts from that saas repo?

maxkondr (approved):

PR #4906, opened 2026-01-09
PMM-14556 RTA Mongo Source.

JiriCtvrtka:
PMM-14556


agent/agents/mongodb/realtime/realtime.go
```diff
@@ -0,0 +1,183 @@
...
+
+
+import (
+	agentv1 "github.com/percona/pmm/api/agent/v1"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+
+// parseCurrentOp extracts fields from a currentOp document and returns proto struct.
+func parseCurrentOp(raw bson.M) *agentv1.MongoDBRealtimeQueryData {
```
JiriCtvrtka, line 26: try to improve those

api/agent/v1/mongodb_realtime.proto
```diff
@@ -0,0 +1,40 @@
...
+  int64 rows_sent = 6;
+  google.protobuf.Timestamp timestamp = 7;
+  string raw_query_json = 8;
+  // MongoDB-specific fields
+  string opid = 9;
+  int64 secs_running = 10;
+  string client = 11;
+  bool waiting_for_lock = 12;
+  bool index_utilized = 13;
+}
```
maxkondr, line 23: I am thinking of the following implementation:

```suggestion
message MongoDBRealtimeQueryData {
  // MongoDB-specific fields
  string opid = 1;
  int64 secs_running = 2;
  string client = 3;
  bool waiting_for_lock = 4;
  bool index_utilized = 5;
}

message RealtimeQueryData {
  string query_id = 1;
  string query_text = 2;
  string state = 3;
  double current_execution_time = 4;
  int64 rows_examined = 5;
  int64 rows_sent = 6;
  google.protobuf.Timestamp timestamp = 7;
  string raw_query_json = 8;
  oneof payload {
      MongoDBRealtimeQueryData mongoDB_payload = 9;
      <add in future the rest of DB-specific messages>
  }
}
```

managed/services/agents/grpc/mongodb_realtime_server.go
```diff
@@ -0,0 +1,51 @@
...
+	agentv1.UnimplementedMongoDBRealtimeServer
+}
+
+// NewMongoDBRealtimeServer creates a new MongoDBRealtime gRPC server.
+func NewMongoDBRealtimeServer(handler *agents.Handler) agentv1.MongoDBRealtimeServer {
+	return &mongoDBRealtimeServer{handler: handler}
+}
+
+// StreamRealtimeQueries handles bidirectional streaming of MongoDB real-time query data.
+func (s *mongoDBRealtimeServer) StreamRealtimeQueries(stream agentv1.MongoDBRealtime_StreamRealtimeQueriesServer) error {
```
maxkondr: Why is it implemented as a separate stream reader?
Can it be done in the same way as QAN is handled in https://github.com/percona/pmm/blob/0527f6582462bf49aeef75c3adbf9ea1a1e62c03/managed/services/agents/handler.go#L135

I mean RTA message with collected queries is send via a general GRPC stream?

api/agent/v1/mongodb_realtime.proto
```diff
@@ -0,0 +1,40 @@
...
+
+// MongoDBRealtimeRequest is sent from agent to server with a batch of queries.
+message MongoDBRealtimeRequest {
+  repeated MongoDBRealtimeQueryData queries = 1;
+}
+
+// MongoDBRealtimeResponse is sent from server to agent (ack or config).
+message MongoDBRealtimeResponse {
+  // Optionally, server can send new polling interval or config.
+  int32 polling_interval_seconds = 1;
```
maxkondr, line 33: as I understand we already have a mechanism for passing a configuration (connection parameters, intervals, etc) from server to agent. May be re-use already existing one and keep this `service MongoDBRealtime` for receiving RTA data from agent only?

managed/cmd/pmm-managed/main.go
```diff
@@ -262,10 +262,14 @@ func runGRPCServer(ctx context.Context, deps *gRPCServerDeps) {
 		l.Debug("RPC response latency histogram enabled.")
 		grpcMetrics.EnableHandlingTimeHistogram()
 	}
+
 	serverv1.RegisterServerServiceServer(gRPCServer, deps.server)
 	agentv1.RegisterAgentServiceServer(gRPCServer, agentgrpc.NewAgentServer(deps.handler))
 	agentpb.RegisterAgentServer(gRPCServer, agentgrpc.NewAgentPBServer(deps.handler))
 
+	// Register MongoDBRealtime gRPC server
+	agentv1.RegisterMongoDBRealtimeServer(gRPCServer, agentgrpc.NewMongoDBRealtimeServer(deps.handler))
```
maxkondr, line 271: in spec we have 
```
4.1. Agent to Server Communication
The real-time analytics agent on the monitored node will collect and send query data to the real-time analytics server.
...
- **Endpoint**: To avoid overwhelming the existing communication channel used for metrics, a separate, dedicated gRPC endpoint will be established for real-time query data. This will ensure that the flow of regular monitoring data is not interrupted.

```

The current implementation re-uses the same TCP connection anyway but a separate HTTP/2 stream. Is it intended?

PR #4912, opened 2026-01-12
PMM-14012 drop portal connections

idoqo:
PMM-14012

Link to the Feature Build: SUBMODULES-4183


api/server/v1/json/client/server_service/change_settings_responses.go
```diff
@@ -959,7 +959,7 @@ type ChangeSettingsOKBodySettings struct {
 	// True if Advisor is enabled.
 	AdvisorEnabled bool `json:"advisor_enabled,omitempty"`
 
-	// Percona Platform user's email, if this PMM instance is linked to the Platform.
+	// platform email
 	PlatformEmail string `json:"platform_email,omitempty"`
```
maxkondr, line 963: why is it kept?

managed/services/server/server.go
```diff
@@ -471,7 +471,7 @@ func (s *Server) readUpdateAuthToken() (string, error) {
 }
 
 // convertSettings merges database settings and settings from environment variables into API response.
-// Checking if PMM is connected to Platform is separated from settings for security and concurrency reasons.
+// While platform is no longer supported, the connectedToPlatform parameter is kept for compatibility.
```
maxkondr, line 476: what compatibility is referred here? Between what components?

managed/services/server/server.go
```diff
@@ -471,7 +471,7 @@ func (s *Server) readUpdateAuthToken() (string, error) {
...
 func (s *Server) convertSettings(settings *models.Settings, disableInternalPgQan bool, connectedToPlatform bool) *serverv1.Settings {
```
maxkondr, line 477: is `connectedToPlatform` still required here?

maxkondr (commented):
I remember there were some ENV vars related to Percona Platform -are they still relevant?

api/server/v1/json/client/server_service/change_settings_responses.go
idoqo: Yeah I should leave comment about it. But it's to avoid any break in compatibility.

Removing it will require us to remove it from proto definitions and generate new descriptor - since the deleted fields means it'll no longer be backward compatible.

managed/services/server/server.go
idoqo: Previous API versions

managed/services/server/server.go
idoqo: Not required. Just thought to keep it for same reason above

idoqo:
> I remember there were some ENV vars related to Percona Platform -are they still relevant?

Yeah - usually the percona platform address - we use those for PMM updates checks so we still need them.

api/server/v1/server.proto
```diff
@@ -159,7 +159,6 @@ message Settings {
   repeated string aws_partitions = 6;
   // True if Advisor is enabled.
   bool advisor_enabled = 7;
-  // Percona Platform user's email, if this PMM instance is linked to the Platform.
   string platform_email = 8;
```
ademidoff:
```suggestion
  string platform_email = 8 [deprecated=true];
```

ademidoff (changes requested):

api/server/v1/json/client/server_service/change_settings_responses.go
maxkondr: such fields shall be marked as `deprecated` and completely dropped in several releases.

ademidoff (approved):

PR #4919, opened 2026-01-13
[DOCS] Improve Doc Contributing Guide

catalinaadam:
PMM-0

- [ ] API Docs updated


alina-derkach-oaza (approved):

travagliad (approved):
LGTM

documentation/docs-contributing.md
```diff
@@ -0,0 +1,239 @@
...
+7. Stage your changes:
+
+    ```shell
+    git add documentation/docs/example.md
+    ```
+
+8. Commit with a descriptive message:
+
+    ```shell
+    git commit -m 'Fixed typo in setting-up.md'
```
ademidoff: That doesn't match with the previous step 7. It'd better be something like this:
`git commit -m 'Document feature X'`, since step 7 adds a new Markdown page.

documentation/docs-contributing.md
```diff
@@ -0,0 +1,239 @@
...
+
+    ```shell
+    git push -u origin <my_changes>
+    ```
+
+10. GitHub will show a **Compare & pull request** button - click it to open your PR. Or navigate to your fork and click **Create pull request**.
+
+### Building the documentation
+
+Want to see how your changes will look on the live site? You can build and preview the docs locally using MkDocs.
```
ademidoff: Honestly, I wouldn't even consider suggesting the user install Mkdocs, sounds like a long journey. I'd also drop Python. 

Why not go ahead with just Docker or Podman? Our OCI image has all it takes to build the docs, ain't it?

documentation/docs-contributing.md
```diff
@@ -0,0 +1,239 @@
...
+
+You'll need Python 3.x and Docker. Don't have them? Grab [Python](https://www.python.org/downloads/) and [Docker](https://docs.docker.com/get-docker/) first.
```
ademidoff: Not really. Again, Python is there already.

documentation/docs-contributing.md
```diff
@@ -0,0 +1,239 @@
...
+
+    ```shell
+    docker run --rm -v $(pwd):/docs perconalab/pmm-doc-md mkdocs build -f documentation/mkdocs.yml
+    ```
+
+2. Open `site/index.html` in your browser to view the docs.
+
+#### Live preview with Docker
+
+Want to see changes as you type? Use the built-in live preview server:
```
ademidoff: Does it really work that way, i.e. "as you type"? Do we have proof?

ademidoff (approved):
IMO, it's a good start, but I would make it even simpler, reducing the whole big manual to just one or two `make` commands.

I think we are making it too complicated. 

Also, I propose the following:

* change the file name to CONTRIBUTING.md
* get it rendered by Mkdocs, similar to other pages
* have a link on the front page and a call to action (CTA)

documentation/docs-contributing.md
catalinaadam: @alina-derkach-oaza — About Alex's feedback above: 

1. Visibility of the contributing guide
We previously agreed to keep the contributing-to-docs guide only in GitHub and not include it in the ToC. However, if it’s not visible on the docs portal at all, does the **Rate and comment on documentation pages** section remain relevant for users?
Alex thinks that this should be more visible, so he also suggested adding a link to the front page. Do we want users to find this topic easier or is the **Edit this page** button on the topics intuitive enough and Github info is only for more advanced users?

2. Workflow changes (Docker / Podman)
What do you think about the proposed workflow changes? If we move forward with Docker and Podman only, PMM would no longer be consistent with the other guides. That said, if this approach is better overall, should we consider updating the other guides to match?

alina-derkach-oaza: @catalinaadam 1. We agreed during the TW meeting that we would add a link to the contributing guide under the “Edit this page” link. I’ve already created a PR for you with the changes that include the link to your new contributing guide. This change will be implemented across all our projects.

2. Since every project has its own peculiarities, it's ok that your contrib guide has additional info or different steps. So you can add any info you think works best for PMM.

alina-derkach-oaza: @catalinaadam https://github.com/percona/pmm/pull/4920

catalinaadam: @ademidoff I’ve simplified the document. Could you please take another look?
Also, the file is called docs-contributing to keep consistency with other Percona doc contributing guides and distinguish it from[CONTRIBUTING.md](https://github.com/percona/pmm/blob/v3/CONTRIBUTING.md) which explains how to contribute to the code.

catalinaadam: > @catalinaadam #4920

@ademidoff, could you please approve this too? We need a 2nd review to merge. Thanks!

ademidoff (approved):

PR #4934, opened 2026-01-15
PMM-14588: Implement Cluster Service Selection Page

dmitri-saricev-3pillargloball:
https://perconadev.atlassian.net/browse/PMM-14588

Result:

https://github.com/user-attachments/assets/e93bc9f6-1a7a-4aef-8ec7-0f7b75a3064c

FB:


ui/apps/pmm/src/api/realtime.ts
```diff
@@ -0,0 +1,23 @@
...
+  ListRunningRealtimeAgentsRequest,
+  ListRunningRealtimeAgentsResponse,
+} from 'types/realtime.types';
+import { api } from './api';
+
+export const listRunningRealtimeAgents = async (
+  params?: ListRunningRealtimeAgentsRequest
+): Promise<ListRunningRealtimeAgentsResponse> => {
+  const res = await api.get('/realtime/agents', { params });
+
```
matejkubinec: Please remove empty lines here.

ui/apps/pmm/src/api/services.ts
```diff
@@ -13,6 +13,6 @@ export const getServiceTypes = async (): Promise<ListTypesResponse> => {
 export const listServices = async (
   params: ListServicesParams
 ): Promise<ListServicesResponse> => {
-  const res = await api.get('/inventory/services', { params });
+  const res = await api.get('/management/services', { params });
```
matejkubinec, line 24: Why the change here?

ui/apps/pmm/src/icons/real-time-database-off.svg
```diff
@@ -0,0 +1,10 @@
+<svg width="192" height="192" viewBox="0 0 192 192" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M75 133.5C75.8284 133.5 76.5 134.172 76.5 135V141C76.5 141.828 75.8284 142.5 75 142.5C74.1716 142.5 73.5 141.828 73.5 141V135C73.5 134.172 74.1716 133.5 75 133.5Z" fill="#F9A98A"/>
```
matejkubinec: Could you replace the fill value with `"currentColor"`, it should allow us to customize the color later.

ui/apps/pmm/src/pages/help-center/welcome-card/WelcomeCard.utils.ts
```diff
@@ -8,6 +8,7 @@ export const shouldShowAddService = (
   }
 
   const minRequiredByService: Record<keyof ListServicesResponse, number> = {
+    services: 0,
```
matejkubinec: What's this?

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.constants.ts
```diff
@@ -0,0 +1,63 @@
+import { SxProps, Theme } from '@mui/material';
+
+/**
+ * External documentation and feedback URLs
+ */
+export const DOCS_URL =
```
matejkubinec: Have you checked in with @catalinaadam about the doc urls? Also we should be using the short links in place of direct urls.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.constants.ts
```diff
@@ -0,0 +1,63 @@
...
+  'https://docs.percona.com/percona-monitoring-and-management/3/get-started/query-analytics.html';
+export const FORUM_URL =
+  'https://forums.percona.com/c/percona-monitoring-and-management-pmm/percona-monitoring-and-management-pmm-v3';
+
+/**
+ * Reusable link styles for documentation and feedback links
+ */
+export const linkStyles: SxProps<Theme> = (theme) => ({
```
matejkubinec: Does the theme from mui/percona-design not provide the appropriate styling? (Question also applies to styles below)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.messages.ts
```diff
@@ -1,12 +1,13 @@
...
-  featureAvailable:
-    'Feature available for MongoDB database technology only, more to come soon.',
+    'Select a service to monitor queries and performance metrics in real time.',
+  mongoOnly: 'Currently available for MongoDB only. More databases coming soon.',
+  searchPlaceholder: 'Search cluster/service...',
+  startButton: 'Start session',
   documentation: 'Documentation',
-  provideFeedback: 'Provide Feedback',
+  feedback: 'Provide feedback',
+  noMongoServices: 'No MongoDB services found',
```
matejkubinec, line 11: Just a note, please check in with @catalinaadam about the wordings, so we are in sync with the actual wording used.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.mocks.ts
```diff
@@ -0,0 +1,172 @@
...
+    customLabels: {},
+    externalGroup: '',
+    address: '127.0.0.1',
+    port: 27017,
+    socket: '',
+    version: '7.0.5',
+    status: 'STATUS_UP',
+  },
+  {
+    serviceId: '/service_id/billing-service',
```
matejkubinec: I don't think we are using `/service_id/` prefixes in IDs anymore.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.mocks.ts
```diff
@@ -0,0 +1,172 @@
...
+    port: 27018,
+    socket: '',
+    version: '7.0.5',
+    status: 'STATUS_UP',
+  },
+
+  // Cluster: mc-ga (matching Figma design)
+  {
+    serviceId: '/service_id/mc-ga-config-01',
+    serviceType: 'SERVICE_TYPE_MONGODB_SERVICE',
```
matejkubinec: Use actual `ServiceType` enum here.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+vi.mock('react-router-dom', async () => {
+  const actual = await vi.importActual('react-router-dom');
+
+  return {
+    ...actual,
+    useNavigate: () => mockNavigate,
+  };
+});
+
+const createWrapper = () => {
```
matejkubinec: You can use the following helpers from `utils/testUtils`:
1. wrapWithQueryProvider
2. wrapWithRouter

You could also create a helper for the `SnackbarProvider`.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+      </BrowserRouter>
+    </QueryClientProvider>
+  );
+};
+
+describe('RealTimeSelection', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+
+    vi.mocked(useUser).mockReturnValue({
```
matejkubinec: You could use `TEST_USER_ADMIN` (or other) from `utils/testStubs`.

Also instead of mock the user differently (see `renderWelcomeCard` in WelcomeCard test)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+        isPMMAdmin: false,
+      },
+    } as ReturnType<typeof useUser>);
+  });
+
+  describe('Rendering', () => {
+    it('renders title and description', () => {
+      render(<RealTimeSelection />, { wrapper: createWrapper() });
+
+      expect(screen.getByText('Real-Time Query Analytics')).toBeInTheDocument();
```
matejkubinec: Should probably use `Messages` instead of strings.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+      render(<RealTimeSelection />, { wrapper: createWrapper() });
+
+      expect(
+        screen.getByText('Currently available for MongoDB only. More databases coming soon.')
+      ).toBeInTheDocument();
+    });
+  });
+
+  describe('Permissions', () => {
+    it('disables controls when user is not editor or admin', () => {
```
matejkubinec: Viewers shouldn't see this page at all (will be redirected to the running agents list)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+      render(<RealTimeSelection />, { wrapper: createWrapper() });
+
+      const autocomplete = screen.getByRole('combobox');
+
+      expect(autocomplete).toHaveAttribute('aria-expanded', 'false');
+    });
+  });
+
+  describe('Form Submission', () => {
+    it('shows error when starting with no services selected', async () => {
```
matejkubinec: This test check if the button is disabled by default, doesn't show an error.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+      render(<RealTimeSelection />, { wrapper: createWrapper() });
+
+      const button = screen.getByRole('button', { name: /start session/i });
+
+      expect(button).toBeDisabled();
+    });
+
+    it('shows success message on successful start', async () => {
```
matejkubinec: The test description isn't matching what the test checks (doesnt redirect by default)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+      render(<RealTimeSelection />, { wrapper: createWrapper() });
+
+      await waitFor(() => {
+        expect(mockNavigate).not.toHaveBeenCalled();
+      });
+    });
+  });
+
+  describe('Loading States', () => {
+    it('shows loading indicator while fetching services', () => {
```
matejkubinec, line 192: Same as with the above ones, doesn't really check what the description says.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+      await waitFor(() => {
+        const autocomplete = screen.getByRole('combobox');
+
+        expect(autocomplete).toBeInTheDocument();
+      });
+    });
+  });
+
+  describe('Accessibility', () => {
+    it('has proper ARIA roles', () => {
```
matejkubinec: Is this needed since those components already provide it (it isn't something that we've added)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+    it('has proper labels', () => {
+      render(<RealTimeSelection />, { wrapper: createWrapper() });
+
+      expect(screen.getByPlaceholderText('Search cluster/service...')).toBeInTheDocument();
+    });
+  });
+
+  describe('Success Handling', () => {
+    it('clears selection on successful start', async () => {
```
matejkubinec: If the test is incomplete, it should report success but rather be skipped (`it.skip`).

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
-import { Autocomplete, Button, Link, TextField } from '@mui/material';
-import Stack from '@mui/material/Stack';
-import Typography from '@mui/material/Typography';
+import { FC, useState } from 'react';
+import {
```
matejkubinec: I'll add a eslint rule for this in another PR but it is better to import the component directly rather than use `@mui/material` (see https://mui.com/material-ui/guides/minimizing-bundle-size/ - Avoid barrel imports)

There are probably places where we've used it previously, so just focus on your changes.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
...
+import { DOCS_URL, FORUM_URL, linkStyles } from './RealTimeSelection.constants';
 
-// Placeholder for the real-time selection page
-const RealTimeSelectionPage: FC = () => {
-  return (
-    <Page footer={null}>
-      <Stack flex={1} alignItems="center" justifyContent="center">
+// Set to true to use mock data for development/testing
+// Set to false to use real MongoDB services from PMM
+const USE_MOCK_DATA = false;
```
matejkubinec: This shouldn't be in a "production ready" code.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
...
+
+export const RealTimeSelection: FC = () => {
+  const { user } = useUser();
+  const canManageRTA = user?.isEditor || user?.isPMMAdmin;
+  const [modalOpen, setModalOpen] = useState(false);
+
+  // Fetch running agents
+  const { data: runningAgentsData, isLoading: isLoadingAgents, refetch: refetchAgents } = useQuery({
```
matejkubinec: Please create `useRunningAgents` (see - https://github.com/percona/pmm/blob/6193f25ada2a65dac14434469d43b1fd074c0cbc/ui/apps/pmm/src/hooks/api/useRealTime.ts).

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
...
+    queryKey: ['runningRealtimeAgents'],
+    queryFn: () => listRunningRealtimeAgents(),
+  });
+
+  // Fetch services to check if all are running
+  const { data: servicesData, isLoading: isLoadingServices } = useQuery({
```
matejkubinec: Same as the above agents call for `useServices` (why not use the current one? Or if we want to list data from the `management` path instead of `inventory` why not add a new hook?)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
...
+    queryKey: ['services'],
+    queryFn: () =>
+      listServices({
+        serviceType: ServiceType.mongodb,
+      }),
+  });
+
+  // Mutation to stop all RTA sessions
+  const stopAllMutation = useMutation({
```
matejkubinec: Same as with running agents, should be in hooks.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
...
+      refetchAgents();
+    },
+    onError: (error) => {
+      const message = error instanceof Error ? error.message : 'Failed to stop sessions';
+      enqueueSnackbar(message, { variant: 'error' });
+    },
+  });
+
+  // Check if all services are running (no available services to start)
+  const runningServiceIds = new Set(
```
matejkubinec: There shouldn't be duplicates right? (one agent == one service) So the Set here isn't needed.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
...
+  // Viewer with no running agents - show viewer empty state
+  const showViewerEmptyState =
+    !canManageRTA &&
+    (!runningAgentsData?.agents || runningAgentsData.agents.length === 0) &&
+    !isLoadingAgents;
+
+  // Viewer with no running agents - show special empty state
+  if (showViewerEmptyState) {
+    return (
+      <Page footer={<></>}>
```
matejkubinec: You can use `null` for the footer not to show.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
...
+    servicesData.services.length > 0;
+
+  // Viewer with no running agents - show viewer empty state
+  const showViewerEmptyState =
+    !canManageRTA &&
+    (!runningAgentsData?.agents || runningAgentsData.agents.length === 0) &&
+    !isLoadingAgents;
+
+  // Viewer with no running agents - show special empty state
+  if (showViewerEmptyState) {
```
matejkubinec, line 39: Please extract viewer empty state into it's own component.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
...
+        sx={{
+          maxWidth: 392,
+          mx: 'auto',
+          py: 6,
+          px: 2,
+          alignItems: 'center',
+          textAlign: 'center',
+        }}
+      >
+        {allServicesRunning ? (
```
matejkubinec, line 76: Not sure if this is needed, the selection page will not be presented to user if there are running sessions (the will be presented the running agents list)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
...
+          /* All services running - show empty state */
+          <RealTimeSelectionEmptyState />
+        ) : (
+          <>
+            {/* Intro section */}
+            <Stack gap={1} sx={{ width: '100%' }}>
+              <Typography
+                variant="h5"
+                sx={{
+                  fontFamily: 'Poppins, sans-serif',
```
matejkubinec: Are the styles needed here? Shouldn't they already be part of the theme?

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
...
+                  textAlign: 'center',
+                  fontVariationSettings: "'wdth' 100",
+                }}
+              >
+                {Messages.description}
+              </Typography>
+            </Stack>
+
+            {/* Form section - reusable component */}
+            <RealTimeSelectionForm useMockData={USE_MOCK_DATA} />
```
matejkubinec: Please `useMockData` and related, shouldn't be in "production ready" code.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,383 @@
...
-                  {Messages.provideFeedback}
+                <Link href={FORUM_URL} target="_blank" sx={linkStyles}>
+                  {Messages.feedback}
                 </Link>
               </Stack>
             </Stack>
+          </>
+        )}
+
+        {/* ========================================
```
matejkubinec: I think you can remove this section altogether, shouldn't not be necessary.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionEmptyState.tsx
```diff
@@ -0,0 +1,42 @@
+import { FC } from 'react';
+import { Stack, Typography, Link, Box } from '@mui/material';
+import { Icon } from 'components/icon';
+import { Messages } from './RealTimeSelection.messages';
+import { DOCS_URL, linkStyles, titleStyles, descriptionStyles } from './RealTimeSelection.constants';
+
+export const RealTimeSelectionEmptyState: FC = () => {
+  return (
```
matejkubinec: You can use direct return here:

```tsx
const RealTimeSelectionEmptyState: FC = () => (
<Stack ...>
...
</Stack>
);
```

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionEmptyState.tsx
```diff
@@ -0,0 +1,42 @@
...
+          sx={{
+            width: 192,
+            height: 192,
+            opacity: 0.5,
+          }}
+        />
+      </Box>
+      <Stack gap={1} sx={{ width: '100%' }}>
+        <Typography variant="h6" sx={titleStyles}>
+          No active sessions now...
```
matejkubinec: Please put messages into it's own file. Also helper components (or parts, etc.) should be in it's own subfolder (e.q. `rta/selection/empty-state/RealTimeSelectionEmptyState`)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,493 @@
...
+}) => {
+  const { user } = useUser();
+  const queryClient = useQueryClient();
+
+  const [selectedServices, setSelectedServices] = useState<ServiceOption[]>([]);
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+
+  const canManageRTA = user?.isEditor || user?.isPMMAdmin;
+
+  const { data: servicesData, isLoading: isLoadingServices } = useQuery<ListServicesResponse>({
```
matejkubinec: Should be in hooks folder. Since we are passing the service type as query key, why aren't we using it when calling the api?

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,493 @@
...
+              service.serviceType === 'mongodb' ||
+              service.serviceType === 'SERVICE_TYPE_MONGODB_SERVICE'
+          );
+
+          return { services: mongodbServices };
+        },
+  });
+
+  // Fetch running agents to filter out services with active RTA
+  const { data: runningAgentsData, isLoading: isLoadingAgents } = useQuery({
```
matejkubinec: Should be in hooks (see previous comment), that way there won't be an issue with wrong query key passed.

Also, since this is a form, wouldn't it make sense to pass the services/running agents/etc. from the parent component?

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,493 @@
...
+    queryKey: ['runningRealtimeAgents'],
+    queryFn: () => listRunningRealtimeAgents(),
+    // Always fetch to filter out running services from dropdown
+  });
+
+  const serviceOptions = useMemo<ServiceOption[]>(() => {
```
matejkubinec: You could probably extract this into an util function to not polute the body of the component so much e.q.

RealTimeSelectionForm.utils.ts
```tsx
export const getOptions = (...) => {...}
```

RealTimeSelectionForm.tsx
```tsx
...
const serviceOptions = useMemo(() => getOptions(...), [...])
...
```

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,493 @@
...
+              serviceId: service.serviceId,
+              cluster: clusterName,
+            });
+          });
+      });
+
+    return options;
+  }, [servicesData, runningAgentsData]);
+
+  const startMutation = useMutation({
```
matejkubinec: The mutation shouldn't be defined here (should be in hooks), when calling a `mutateAsync` function you can pass in the onSuccess/etc. there:

See: https://tanstack.com/query/v5/docs/framework/react/reference/useMutation
```tsx
mutateAsync: (variables: TVariables, { onSuccess, onSettled, onError }) => Promise<TData>
```

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,493 @@
...
+        value={selectedServices}
+        onChange={handleServiceChange}
+        getOptionLabel={(option) => option.label}
+        isOptionEqualToValue={(option, value) => option.id === value.id}
+        disableCloseOnSelect
+        limitTags={2}
+        renderInput={(params) => (
+          <TextField
+            {...params}
+            label={selectedServices.length > 0 || isOpen ? 'Cluster/Service' : undefined}
```
matejkubinec: Label should be defined in messages file.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,493 @@
...
+            placeholder={selectedServices.length === 0 ? Messages.searchPlaceholder : ''}
+            variant="outlined"
+            InputProps={{
+              ...params.InputProps,
+              sx: {
+                fontFamily: 'Roboto, sans-serif',
```
matejkubinec: Same here with the styling, doesn't the theme already provide those? (**This is a question for all the styling in this file**)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,493 @@
...
+              sx: (theme) => ({
+                color: theme.palette.text.primary,
+                fontFamily: 'Roboto',
+                fontSize: '16px',
+                fontStyle: 'normal',
+                fontWeight: 500,
+                lineHeight: '100%',
+                letterSpacing: '0.12px',
+                fontFeatureSettings: "'liga' off, 'clig' off",
+                '&.Mui-focused': {
```
matejkubinec: If you need the mui classes you could use `inputBaseClasses.focused` (etc. same for other components labels etc.)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,493 @@
...
+                },
+                '&.Mui-focused fieldset': {
+                  borderColor: theme.palette.primary.main,
+                  borderWidth: '2px',
+                },
+              },
+            })}
+          />
+        )}
+        renderTags={(value, getTagProps) =>
```
matejkubinec, line 125: Could you extract this into it's own component.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,493 @@
...
+                  '& .MuiChip-deleteIcon': {
+                    m: 0,
+                    p: 0,
+                  },
+                })}
+              />
+            );
+          })
+        }
+        renderOption={(props, option, { selected }) => {
```
matejkubinec: Same here, extract into it's own component, I think it gets messy if we have logic in the `renderOption` (etc.), I think it would be cleaner to have the logic encompassed in a component (such as `<ServiceOption ... />`) then have a call like:

```tsx
renderOption{(props, option, { selected }) => <ServiceOption ... />}
```

ui/apps/pmm/src/types/services.types.ts
```diff
@@ -52,14 +52,37 @@ export interface ExternalService extends BaseService {
...
+  environment: string;
+  cluster: string;
+  replicationSet: string;
+  customLabels: Record<string, string>;
+  externalGroup: string;
+  address: string;
+  port: number;
+  socket: string;
+  version: string;
+  status?: 'STATUS_UNSPECIFIED' | 'STATUS_UP' | 'STATUS_DOWN' | 'STATUS_UNKNOWN';
```
matejkubinec: Should be an enum.

ui/apps/pmm/src/types/services.types.ts
```diff
@@ -52,14 +52,37 @@ export interface ExternalService extends BaseService {
 
 export interface ValkeyService extends VersionedService {}
 
+// Universal service from /v1/management/services API
+export interface UniversalService {
```
matejkubinec: Should be probably a `ManagedService` to different from the other service call (`inventory`)

ui/apps/pmm/src/types/services.types.ts
```diff
@@ -52,14 +52,37 @@ export interface ExternalService extends BaseService {
...
+
 export interface ListServicesResponse {
-  mysql: MySqlService[];
-  mongodb: VersionedService[];
-  postgresql: PostgreSqlService[];
-  proxysql: ProxySqlService[];
-  haproxy: HaProxyService[];
-  external: ExternalService[];
-  valkey: ValkeyService[];
+  // New API returns flat services array
```
matejkubinec: Why mix those two apis? As you can see it creates so much conflict, either use one or the other but do not combine them. (responses etc. should match with api calls)

If you want to create a new structure that is better used for you component, it should be done at the Component/Service level rather an api.

ui/apps/pmm/src/types/services.types.ts
```diff
@@ -52,14 +52,37 @@ export interface ExternalService extends BaseService {
...
+  services?: UniversalService[];
+  // Legacy format (kept for backward compatibility)
```
matejkubinec: Why legacy?

ui/apps/pmm/src/api/services.ts
dmitri-saricev-3pillargloball: Hey! Good question 😊

The new `/management/services` endpoint returns a flat `services` array with `UniversalService` objects that include extra fields we need for RTA — like `serviceType`, `nodeName`, `status`, etc.

The old `/inventory/services` endpoint returns services grouped by type (mysql, mongodb, postgresql...) which doesn't have all the fields we need.

Let me know if you'd prefer a different approach!

ui/apps/pmm/src/pages/help-center/welcome-card/WelcomeCard.utils.ts
dmitri-saricev-3pillargloball: Good catch! 👀

Since we updated `ListServicesResponse` to support the new API format (which has an optional `services?: UniversalService[]` field), we needed to:

1. Add `services: 0` to the config so TypeScript is happy with the type
2. Change `services[key].length` → `services[key]?.length ?? 0` to safely handle cases where a field might be undefined

Without this, the function would crash when using the new API response format. It's basically just making the existing code compatible with both old and new API shapes!

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.constants.ts
catalinaadam: - https://per.co.na/PMM3_forums
- https://per.co.na/QAN

ui/apps/pmm/src/api/services.ts
matejkubinec: @dmitri-saricev-3pillargloball Let's add another endpoint here rather then to modify the existing one.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+      await waitFor(() => {
+        const autocomplete = screen.getByRole('combobox');
+
+        expect(autocomplete).toBeInTheDocument();
+      });
+    });
+  });
+
+  describe('Success Handling', () => {
+    it.skip('clears selection on successful start', async () => {
```
matejkubinec, line 225: It should redirect the user, but since this test will be updated anyways it's ok to keep it like this.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,98 @@
...
+  const { data: runningAgentsData, isLoading: isLoadingAgents } = useRunningRealtimeAgents();
+
+  const { data: servicesData, isLoading: isLoadingServices } = useServices({
+    serviceType: ServiceType.mongodb,
+  });
+
+  const runningServiceIds =
+    runningAgentsData?.agents?.map((agent) => agent.serviceId) ?? [];
+
+  const availableServices =
```
matejkubinec: Could you memoize this so it isn't running on every render? (I think you can combine it with runningServiceIds like so:

```tsx
const availableServices = useMemo(() => {
  if (!servicesData?.services) {
    return [];
  }
  
  // might need to check if it works with an empty array
  if (!runningAgentsData?.agents) {
    return [];
  }

  const runningServiceIds =
    runningAgentsData.agents.map((agent) => agent.serviceId);
  return servicesData.services.filter(
      (service) => !runningServiceIds.includes(service.serviceId)
    );
}
```

ui/apps/pmm/src/pages/rta/selection/RealTimeSelection.tsx
```diff
@@ -1,74 +1,98 @@
...
+    servicesData?.services?.filter(
+      (service) => !runningServiceIds.includes(service.serviceId)
+    ) ?? [];
+
+  const allServicesRunning =
+    !isLoadingServices &&
```
matejkubinec: We should probably show loading indicator on the page / input if the services/agents are initally loading (so you may omit it from this check)

ui/apps/pmm/src/pages/help-center/welcome-card/WelcomeCard.utils.ts
```diff
@@ -1,13 +1,23 @@
-import { ListServicesResponse } from 'types/services.types';
+import { ManagedServicesResponse } from 'types/services.types';
```
matejkubinec: Let's keep using here the inventory endpoint (not managed), we can update it later if needed (changing the code here would also require additional testing for the welcome cards / help page)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,471 @@
...
+}) => {
+  const { user } = useUser();
+  const queryClient = useQueryClient();
+
+  const [selectedServices, setSelectedServices] = useState<ServiceOption[]>([]);
+  const [isOpen, setIsOpen] = useState(false);
+
+  const canManageRTA = user?.isEditor || user?.isPMMAdmin;
+
+  const { data: servicesData, isLoading: isLoadingServices } = useQuery<ManagedServicesResponse>({
```
matejkubinec: I think it would make sense to use here the same `useRunningRealtimeAgents` / `useServices/useManagedServices` instead.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,471 @@
...
+    queryKey: ['services', ServiceType.mongodb],
+    queryFn: async (): Promise<ManagedServicesResponse> => {
+      const response = await listServices({});
+
+      // Filter MongoDB services from the services array
+      // API returns serviceType as lowercase "mongodb", not "SERVICE_TYPE_MONGODB_SERVICE"
+      const mongodbServices = (response.services || []).filter(
```
matejkubinec: If the api returns different service type, it would be good to have another enum for these.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,471 @@
...
+
+  const serviceOptions = useMemo<ServiceOption[]>(() => {
+    const services: ManagedService[] = servicesData?.services ?? [];
+
+    if (services.length === 0) {
+      return [];
+    }
+
+    // Filter out services that already have running RTA agents
+    const runningServiceIds = new Set(
```
matejkubinec: I think this is the same filtering of available services as in the parent component, could you create an util function? (or hook - example below - if it makes sense)

```tsx
export const useNonRTAServices = () => {
  const { data: runningAgentsData } = useRunningRealtimeAgents();
  const { data: servicesData } = useManagedServices();
  // see other comment
  const availableServices = useMemo(() => ..., [])
  
  return availableServices;
}
```

Then you can use that in components that you need.

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,471 @@
...
+      );
+
+      return { services: mongodbServices };
+    },
+  });
+
+  // Fetch running agents to filter out services with active RTA
+  const { data: runningAgentsData, isLoading: isLoadingAgents } = useRunningRealtimeAgents();
+
+  const serviceOptions = useMemo<ServiceOption[]>(() => {
```
matejkubinec: Could you extract this into an util function?

```tsx
// RealTimeSelectionForm.utils.ts
export const getServiceOptions = (...) => {...}

// RealTimeSelectionForm.tsx
const serviceOptions = useMemo(() => getServiceOptions(availableServices), [availableServices]);
```

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,471 @@
...
+        if (!newSelections.some((selected) => selected.id === service.id)) {
+          newSelections.push(service);
+        }
+      });
+      setSelectedServices(newSelections);
+    }
+  };
+
+  // Check if all services in a cluster are selected
+  const isClusterFullySelected = (clusterName: string): boolean => {
```
matejkubinec: Can be reused in `handleClusterToggle` (make an util?)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,471 @@
...
+    const servicesInCluster = serviceOptions.filter(
+      (option) => option.type === 'service' && option.cluster === clusterName
+    );
+    return servicesInCluster.every((service) =>
+      selectedServices.some((selected) => selected.id === service.id)
+    );
+  };
+
+  // Check if some (but not all) services in a cluster are selected
+  const isClusterPartiallySelected = (clusterName: string): boolean => {
```
matejkubinec: Maybe an util function that will return cluster selection state?

```tsx
const getClusterSelectionState = (services, clusterName): 'all' | 'partial' | 'none'  => {...}

File: ui/apps/pmm/src/hooks/api/useRealtime.ts [hunk 56]
```diff
@@ -0,0 +1,70 @@
...
+    ...options,
+  });
+};
+
+/**
+ * Hook to get MongoDB services that don't have running RTA agents
+ */
+export const useAvailableServices = () => {
+  const { data: runningAgentsData, isLoading: isLoadingAgents } = useRunningRealtimeAgents();
+  const { data: servicesData, isLoading: isLoadingServices } = useManagedServices();
```
matejkubinec: Shouldn't it be possible to just filter mongo services here - SERVICE_TYPE_MONGODB_SERVICE?

image

ui/apps/pmm/src/pages/help-center/welcome-card/WelcomeCard.utils.ts
```diff
@@ -23,5 +23,5 @@ export const shouldShowAddService = (
       keyof ListServicesResponse,
       number,
     ][]
-  ).every(([key, min]) => services[key].length < min);
+  ).every(([key, min]) => (services[key]?.length ?? 0) < min);
```
matejkubinec, line 26: can it be null/undefined?

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.tsx
```diff
@@ -0,0 +1,176 @@
...
+    try {
+      await Promise.all(
+        realServices.map((service) =>
+          changeRealtimeAnalytics.mutateAsync(
+            {
+              enable: true,
+              serviceId: service.serviceId!,
+            },
+            {
+              onSuccess: undefined, // Prevent individual success callbacks
```
matejkubinec: Add more info here (`to prevent the running agents query invalidation`)

ui/apps/pmm/src/pages/rta/selection/RealTimeSelectionForm.utils.ts
```diff
@@ -0,0 +1,145 @@
+import { ManagedService } from 'types/services.types';
+
+export interface ServiceOption {
```
matejkubinec: Could you put types into `RealTimeSelectionForm.types` if they are used across multiple files?

ui/apps/pmm/src/pages/rta/selection/components/ServiceOption.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const handleClick = isCluster
+    ? (e: React.MouseEvent) => {
+        e.stopPropagation();
+        onClusterToggle?.(option.label);
+      }
+    : otherProps.onClick;
+
+  return (
+    <li
```
matejkubinec: Could you use `<Box component='li'>`, using style attribute directly on the html component is discouraged (separation of concerts, overrides, etc.)

ui/apps/pmm/src/pages/rta/selection/components/ServiceOptionTag.tsx
```diff
@@ -0,0 +1,76 @@
+import { FC } from 'react';
+import Box from '@mui/material/Box';
+import Chip from '@mui/material/Chip';
+import CloseIcon from '@mui/icons-material/Close';
+import { ServiceOption } from '../RealTimeSelectionForm.utils';
+
+interface ServiceOptionTagProps {
+  option: ServiceOption;
+  tagProps: {
```
matejkubinec: I think you could reuse `ChipProps` (from mui) for the tag props, if it's the same. If it isn't (haven't checked) please disregard :)

ui/apps/pmm/src/pages/rta/selection/components/ServiceOptionTag.tsx
```diff
@@ -0,0 +1,76 @@
...
+          sx={(theme) => ({
+            display: 'flex',
+            alignItems: 'center',
+            justifyContent: 'center',
+            width: 16,
+            height: 16,
+            borderRadius: '50%',
+            backgroundColor:
+              theme.palette.mode === 'dark'
+                ? 'rgba(255, 255, 255, 0.2)'
```
matejkubinec: Is there something available in peak design (there should be design tokens available), that match this (you should be able to check in figma the specific ones)

ui/apps/pmm/src/pages/rta/selection/components/ServiceOptionTag.tsx
```diff
@@ -0,0 +1,76 @@
...
+      sx={(theme) => ({
+        height: 24,
+        borderRadius: 12,
+        backgroundColor:
+          theme.palette.mode === 'dark'
+            ? 'rgba(255, 255, 255, 0.16)'
+            : 'rgba(0, 0, 0, 0.08)',
+        px: 1,
+        gap: 0.5,
+        '& .MuiChip-label': {
```
matejkubinec: You can use the `chipClasess` here instead of strings.

matejkubinec (approved):

PR #4986, opened 2026-01-30
PMM-14786 Harden PG security

ademidoff:
PMM-14786

Link to the Feature Build: SUBMODULES-4210


JiriCtvrtka (approved):

build/ansible/pmm-docker/post-build.yml
```diff
@@ -24,11 +24,15 @@
         name: pmm-managed
         force: true
         state: absent
+      environment:
+        PGPASSWORD: "{{ lookup('file', '/srv/.postgres_password') }}"
```
maxkondr, line 28: we have `/srv/postgres14/` directory, can this file be placed there?

build/ansible/roles/postgres/files/postgres-migration
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+set -o errexit
+set -o pipefail
+
+declare POSTGRES_DATA_DIR="${POSTGRES_DATA_DIR:-/srv/postgres14}"
+declare POSTGRES_PASSWORD_FILE="${POSTGRES_PASSWORD_FILE:-/srv/.postgres_password}"
+
+ensure_postgres_password() {
+    if [ ! -f "$POSTGRES_DATA_DIR/PG_VERSION" ]; then
```
maxkondr, line 10: 1. what is the reason to check for this file if it is not used later in function?
2. Exiting silently may introduce an additional problems during troubleshooting

build/ansible/roles/postgres/files/postgres-migration
```diff
@@ -0,0 +1,60 @@
...
+    fi
+
+    echo "Generating postgres superuser password for existing data directory..."
+    declare POSTGRES_PASSWORD
+    POSTGRES_PASSWORD=$(openssl rand -hex 16)
+    echo -n "$POSTGRES_PASSWORD" > "$POSTGRES_PASSWORD_FILE"
+    chmod 600 "$POSTGRES_PASSWORD_FILE"
+
+    local started=0
+    if ! /usr/pgsql-14/bin/pg_ctl status -D "$POSTGRES_DATA_DIR" > /dev/null 2>&1; then
```
maxkondr, line 24: it is recommended do not mix command execution and checking for status at once in bash

```suggestion
    $(/usr/pgsql-14/bin/pg_ctl status -D "$POSTGRES_DATA_DIR" > /dev/null 2>&1)
    pg_status=$?
    if [[ $pg_status -ne 0 ]] ; then
        pg_start_output=$(/usr/pgsql-14/bin/pg_ctl start -D "$POSTGRES_DATA_DIR" -o "-c logging_collector=off" 2>&1)
        pg_start_status=$?
        [[ $pg_start_status -ne 0 ]] && echo "Failed to start PostgreSQL: $pg_start_output" && exit 1
        started=1
    fi
```

build/ansible/roles/postgres/files/postgres-migration
```diff
@@ -0,0 +1,60 @@
...
+        /usr/pgsql-14/bin/pg_ctl start -D "$POSTGRES_DATA_DIR" -o "-c logging_collector=off"
+        started=1
+    fi
+
+    PGPASSWORD="$POSTGRES_PASSWORD" /usr/bin/psql -U postgres -h /run/postgresql -d postgres \
+        -c "ALTER USER postgres WITH PASSWORD '${POSTGRES_PASSWORD}'"
```
maxkondr, line 30:
```suggestion
    PGPASSWORD="$(cat $POSTGRES_PASSWORD_FILE)" /usr/bin/psql -U postgres -h /run/postgresql -d postgres \
        -c "ALTER USER postgres WITH PASSWORD '${POSTGRES_PASSWORD}'"
```

or even (but need to check explicitly)

```suggestion
    /usr/bin/psql -U postgres -h /run/postgresql -d postgres \
        -c "ALTER USER postgres WITH PASSWORD '$(cat $POSTGRES_PASSWORD_FILE)'"
```

build/ansible/roles/postgres/files/postgres-migration
```diff
@@ -0,0 +1,60 @@
...
+
+    if [ "$started" -eq 1 ]; then
+        /usr/pgsql-14/bin/pg_ctl stop -D "$POSTGRES_DATA_DIR"
+    fi
```
maxkondr, line 34: this check has no much sense anymore because in case PG failed to start execution won't get here
```suggestion
    pg_stop_output=$(/usr/pgsql-14/bin/pg_ctl stop -D "$POSTGRES_DATA_DIR" 2>&1)
    pg_stop_status=$?
    [[ $pg_stop_status -ne 0 ]] && echo "Failed to stop PostgreSQL: $pg_stop_output" && exit 1
```

build/ansible/roles/postgres/files/postgres-migration
```diff
@@ -0,0 +1,60 @@
...
+
+    unset POSTGRES_PASSWORD
+}
+
+update_pg_hba_auth() {
+    local hba="$POSTGRES_DATA_DIR/pg_hba.conf"
+    if [ ! -f "$hba" ]; then
+        return 1
```
maxkondr, line 43: Exiting silently may introduce an additional problems during troubleshooting

build/ansible/roles/postgres/files/postgres-migration
```diff
@@ -0,0 +1,60 @@
...
+    fi
+
+    if ! grep -Eq '^[[:space:]]*host[[:space:]].*[[:space:]]trust[[:space:]]*$' "$hba"; then
+        return 1
+    fi
```
maxkondr, line 48:
```suggestion
    $(grep -Eq '^[[:space:]]*host[[:space:]].*[[:space:]]trust[[:space:]]*$' "$hba")
    cmd_status=$?
    if [[ $cmd_status -ne 0 ]]; then
        echo "<some meaningful message>"
        return 1
    fi
```

build/ansible/roles/postgres/files/postgres-migration
```diff
@@ -0,0 +1,60 @@
...
+
+    echo "Updating pg_hba.conf to require scram-sha-256 for host connections..."
```
maxkondr:
```suggestion
    echo "Updating $hba to require scram-sha-256 for host connections..."
```

build/ansible/roles/postgres/files/postgres-migration
```diff
@@ -0,0 +1,60 @@
...
+    local tmp
+    tmp=$(mktemp)
+    sed -E 's/^([[:space:]]*host[[:space:]].*[[:space:]])trust([[:space:]]*)$/\1scram-sha-256\2/' "$hba" > "$tmp"
+    mv "$tmp" "$hba"
+
+    if /usr/pgsql-14/bin/pg_ctl status -D "$POSTGRES_DATA_DIR" > /dev/null 2>&1; then
+        /usr/pgsql-14/bin/pg_ctl reload -D "$POSTGRES_DATA_DIR"
+    fi
```
maxkondr, line 58: 1.
```suggestion
    status_output=$(/usr/pgsql-14/bin/pg_ctl status -D "$POSTGRES_DATA_DIR" > /dev/null 2>&1)
    cmd_status=$?
    if [[ $cmd_status -eq 0 ]]; then
        reload_output=$(/usr/pgsql-14/bin/pg_ctl reload -D "$POSTGRES_DATA_DIR")
        cmd_status=$?
        [[ $cmd_status -ne 0 ]] && echo "Failed to reload PostgreSQL: $reload_output" && exit 1
    fi
```

2. isn't it better to check the status at the beginning of function before executing all commands above?
3. what if status != 0 ?

build/docker/server/entrypoint.sh
```diff
@@ -94,20 +96,40 @@ if [ ! -f "$DIST_FILE" ]; then
...
-    /usr/pgsql-14/bin/initdb -D /srv/postgres14 --auth=trust --username=postgres
-
+    install -d -m 750 "$POSTGRES_DATA_DIR"
+    
+    # Generate a random password for postgres superuser
+    declare POSTGRES_PASSWORD
+    POSTGRES_PASSWORD=$(openssl rand -hex 16)
+    
+    # Store the password securely with restricted permissions
+    declare POSTGRES_PASSWORD_FILE="/srv/.postgres_password"
```
maxkondr: already declared in line 7

build/docker/server/entrypoint.sh
```diff
@@ -94,20 +96,40 @@ if [ ! -f "$DIST_FILE" ]; then
...
+    echo -n "$POSTGRES_PASSWORD" > "$POSTGRES_PASSWORD_FILE"
+    chmod 600 "$POSTGRES_PASSWORD_FILE"
+    
+    # Initialize database with password authentication
+    /usr/pgsql-14/bin/initdb -D "$POSTGRES_DATA_DIR" --auth-host=scram-sha-256 --auth-local=trust --username=postgres --pwfile="$POSTGRES_PASSWORD_FILE"
```
maxkondr: status check?

build/docker/server/entrypoint.sh
```diff
@@ -94,20 +96,40 @@ if [ ! -f "$DIST_FILE" ]; then
...
+    
     echo "Enabling pg_stat_statements extension for PostgreSQL..."
-    /usr/pgsql-14/bin/pg_ctl start -D /srv/postgres14
-    /usr/bin/psql postgres postgres -c 'CREATE EXTENSION pg_stat_statements SCHEMA public'
-    /usr/pgsql-14/bin/pg_ctl stop -D /srv/postgres14
+    /usr/pgsql-14/bin/pg_ctl start -D "$POSTGRES_DATA_DIR" -o "-c logging_collector=off"
```
maxkondr: status check?

build/docker/server/entrypoint.sh
```diff
@@ -94,20 +96,40 @@ if [ ! -f "$DIST_FILE" ]; then
...
+    PGPASSWORD="$POSTGRES_PASSWORD" /usr/bin/psql -U postgres -h /run/postgresql -d postgres -c 'CREATE EXTENSION pg_stat_statements SCHEMA public'
+    /usr/pgsql-14/bin/pg_ctl stop -D "$POSTGRES_DATA_DIR"
+    
+    # Clean up password from environment
+    unset POSTGRES_PASSWORD
+fi
+
+if [ -f "/opt/ansible/roles/postgres/files/postgres-migration" ]; then
+    bash /opt/ansible/roles/postgres/files/postgres-migration
```
maxkondr: the logic from this script is partially executed in lines 102-108, is it intentional?

managed/models/database.go
```diff
@@ -1360,46 +1360,58 @@ func checkVersion(ctx context.Context, db reform.DBTXContext) error {
...
 		params.Logf("Creating database %s and role %s", params.Name, params.Username)
 	}
-	// we use empty password/db and postgres user for creating database
-	db, err := OpenDB(SetupDBParams{Address: params.Address, Username: "postgres"})
+
+	// Read postgres password from the secure file
+	passwordFile := "/srv/.postgres_password" //nolint:gosec
+	passwordBytes, err := os.ReadFile(passwordFile)
+	if err != nil {
+		return errors.Wrapf(err, "failed to read postgres password from %s", passwordFile)
```
maxkondr: `errors.Wrap` function is from `github.com/pkg/errors` package that is deprecated for usage

build/ansible/pmm-docker/post-build.yml
ademidoff: Yes, we can put it there, but I see two issues with doing it that way:
1. It's a pure data directory fully under control by the db engines, and most probably we don't want to put anything non-native there
2. We will migrate from PG 14 to PG 17 this or next year, there is no point in moving this file from one directory to another if it doesn't change.

build/ansible/roles/postgres/files/postgres-migration
ademidoff: This check is meant to verify if the data directory is empty. If it is, we can't continue. 
Maybe I don't quite get the question, but I see no other use for this file. It's a rail guard.

build/ansible/roles/postgres/files/postgres-migration
ademidoff: Honestly, I don't see the reason for such verbosity, since both variants achieve the same goal, the second one is just being more difficult to read and maintain.

For example, there is not need to capture the output of `pg_ctl start ...` and then echo it, since the user will see the same output even without such complication. Similarly, `exit 1` does not bring anything to the table, since the script will exit on failure anyway.

build/ansible/roles/postgres/files/postgres-migration
ademidoff: > Exiting silently may introduce an additional problems during troubleshooting

Agree, I added an echo statement.

build/ansible/roles/postgres/files/postgres-migration
ademidoff: Declaring a variable and assigning a value to it by running a `cat` command is bad practice in bash, it doesn't pass the shellcheck linter (see [SC2155](https://www.shellcheck.net/wiki/SC2155)).

It may seem redundant, but that's the best practice.

build/ansible/roles/postgres/files/postgres-migration
ademidoff: The suggested change is more verbose and does the same.

build/ansible/roles/postgres/files/postgres-migration
ademidoff: 2. No, it's not - it's totally acceptable. 
3. If the status is not equal to 0, the script exits with the error message. Benefit: no verbosity, simple to read, understand and maintain.

build/docker/server/entrypoint.sh
ademidoff: Good catch, fixed 👍

build/docker/server/entrypoint.sh
ademidoff: No need, the script output an error to `/dev/stdout` and exit with errorlevel > 0.

build/docker/server/entrypoint.sh
ademidoff: Yes, it's intentional. The lines 102-108 are part of the initialization of an empty `/srv` directory, while the script performs the migration. The migration script has a guard rail, so it will do nothing if the password is found on disk.

managed/models/database.go
ademidoff: Nice catch, fixed 👍

build/docker/server/entrypoint.sh
ademidoff: Same as per our discussion above - I see no need to complicate things.

build/ansible/roles/postgres/files/postgres-migration
ademidoff: Done.

build/ansible/roles/postgres/files/postgres-migration
ademidoff: Yes, it's an intentional silent exit. The reason is that this migration script is running on every start/restart of PMM Server, so it would be too annoying and irrelevant to see such messages in case the migration has already taken place.

build/ansible/roles/postgres/files/postgres-migration
ademidoff: Same reason for a silent exit as above.

build/ansible/pmm-docker/post-build.yml
maxkondr:
```
It's a pure data directory fully under control by the db engines, and most probably we don't want to put anything non-native there
```

not really:
```
[pmm@pmm-server postgres14] # ls /srv/postgres14/
base	pg_commit_ts  pg_hba.conf    pg_logical    pg_notify	pg_serial     pg_stat	   pg_subtrans	pg_twophase  pg_wal   postgresql.auto.conf  postmaster.opts
global	pg_dynshmem   pg_ident.conf  pg_multixact  pg_replslot	pg_snapshots  pg_stat_tmp  pg_tblspc	PG_VERSION   pg_xact  postgresql.conf	    postmaster.pid
```
it contains:
- data
- config files
- pid file

that is actually mixing different concepts together from Linux POV.

build/ansible/roles/postgres/files/postgres-migration
maxkondr:
```
This check is meant to verify if the data directory is empty.
```
Placing some comment near this check would answer this question that any script ready may have here

maxkondr (approved):

build/ansible/roles/postgres/files/postgres-migration
ademidoff: Done.

PR #4991, opened 2026-02-02
PM-14557, PMM-14553 Implement  RTA Agent-to-Server Data Transmission

maxkondr:
PMM-14557, PMM-14553

Main changes:
- now underlying TCP connection from pmm-agent to pmm-managed is created separately from gRPC streams.
- additional  gRPC stream (for `/realtimeanalytics.v1.CollectorService/Collect` endpoint) from pmm-agent to pmm-managed is created always during establishing connection to server. We do not setup/teardown this stream depending on RTA feature usage, it is created always and ready for use.
- gRPC streams (for Agents and RTA services) share the same underlying TCP connection. In case smth happens with connection - both streams are affected.

Link to the Feature Build: [SUBMODULES-14557](https://github.com/Percona-Lab/pmm-submodules/pull/4213)


agent/client/client.go
```diff
@@ -850,17 +904,50 @@ func dial(dialCtx context.Context, cfg *config.Config, l *logrus.Entry) (*dialRe
...
+		md:      md,
 	}, nil
 }
 
+// createChannelToRealTimeAnalyticsService tries to create a uni-directional communication channel to the Real-Time Analytics service using the provided gRPC connection.
+func createChannelToRealTimeAnalyticsService(dialCtx context.Context, conn *grpc.ClientConn, cfg *config.Config, l *logrus.Entry) (*channel.RTAChannel, error) {
+	// gRPC stream is created without lifetime timeout.
+	// However, we need to cancel it if communication channel can't be established
+	// when pmm-managed is down. A separate timer is used for that.
+	streamCtx, streamCancel := context.WithCancel(context.Background())
```
ademidoff, line 921: Doesn't this new context need to be derived from the upstream one, i.e. `dialCtx`? I mean, if the upstream is cancelled, this one needs to be cancelled in any case.

go.mod
```diff
@@ -68,7 +68,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	github.com/tink-crypto/tink-go v0.0.0-20230613075026-d6de17e3f164
 	go.mongodb.org/mongo-driver v1.17.6
-	go.starlark.net v0.0.0-20230717150657-8a3343210976
+	go.starlark.net v0.0.0-20260102030733-3fee463870c9
```
ademidoff: Let's revert this change. We usually need a ticket to bump it.

maxkondr: hm, I haven't updated it explicitly... 
reverted

agent/client/client.go
maxkondr: this was done explicitly even before me: in case smth happens with stream or underlying connection -> the whole connection is closed and re-connection (`client.Run`) is performed again

ademidoff (approved):

JiriCtvrtka (approved):

PR #4999, opened 2026-02-03
PMM-14561, PMM-14575 Implement RTA Agent Configuration and Lifecycle management

maxkondr:
PMM-14561, PMM-14575

Handle new RTA MongoDB Agent type on pmm-agent side - create built-in agent and run it in goroutine.


managed/services/agents/state.go
```diff
@@ -165,6 +165,7 @@ func (u *StateUpdater) sendSetStateRequest(ctx context.Context, agent *pmmAgentI
 	filters := models.AgentFilters{
 		PMMAgentID:  agent.id,
 		IgnoreNomad: !settings.IsNomadEnabled(),
+		Disabled:    pointer.To(false), // fetch enabled only
```
ademidoff:
```suggestion
 		// fetch enabled only
		Disabled:    pointer.To(false),
```

ademidoff: Better put the comment above.

managed/services/agents/state.go
```diff
@@ -180,10 +181,6 @@ func (u *StateUpdater) sendSetStateRequest(ctx context.Context, agent *pmmAgentI
 	agentProcesses := make(map[string]*agentv1.SetStateRequest_AgentProcess)
 	builtinAgents := make(map[string]*agentv1.SetStateRequest_BuiltinAgent)
 	for _, row := range agents {
-		if row.Disabled {
```
ademidoff, line 183: I'm curious if this was in intentional removal or not. Do we no longer need to skip disabled agents?

maxkondr: yes this is intentional, because now we filter disabled agents on DB level (`Disabled:    pointer.To(false),`)

ademidoff (approved):

JiriCtvrtka (approved):

PR #5031, opened 2026-02-12
PMM-14737 Improvements for QAN dev env.

JiriCtvrtka:
PMM-14737

The current bash script used to apply migrations has been removed and replaced with a Go implementation that handles migrations properly using a migration library (like in PMM itself), instead of simply running them through the ClickHouse client. With this approach, you should no longer encounter dirty migrations when making changes.


ademidoff (approved):

qan-api2/clickhouse_migrate/main.go
```diff
@@ -0,0 +1,59 @@
...
+	flag.Parse()
+
+	user := *userFlag
+	password := *passwordFlag
+	lastMigration := *lastMigrationFlag
+	if lastMigration == 0 {
+		log.Println("Usage: go run main.go --last-migration <number> [--user <user>] [--password <password>]")
+		os.Exit(1)
+	}
+	migrationDir := "file:///root/go/src/github.com/percona/pmm/qan-api2/migrations/sql"
```
maxkondr: better to read GOPATH from ENV

JiriCtvrtka: Since all others are flags I set migrations dir also as flag

maxkondr (approved):

PR #5036, opened 2026-02-13
PMM-14556 Collect data from MongoDB

maxkondr:
PMM-14556

Summary:
- built-in RTA agent connects to MongoDB using provided credentials
- runs aggregation 
```
/*
The matchStage below equal to the following aggregation pipeline in MongoDB shell,
which filters out operations from RTA agent itself and internal MongoDB tools, 
and gets only active operations/commands with execution time => 0.01s:
*/

db.aggregate([
  { $currentOp : { allUsers: true, idleSessions: false,  idleCursors:false, idleConnections:false} },
  { $match : {
      $and: [
          { active: true},
          {"microsecs_running": {$gte: 10_000}},
          { "desc": {$nin: ["Checkpointer", "JournalFlusher"]}},
          {"appName": {$not: {$regex: "^(rta-mongodb-.*$)"}}},
     ]}}]);
``` 

- parses DB output
-- new parser of `currentOp()` output has been implemented.
-- parser does its best to parse the input. It uses `mongodb-driver` that may call `panic()` - in such a case parser returns just `nil` and this particular currentOperation is skipped. 
-- it makes an attempt to re-construct the original query/command that is being executed by DB and set it to `query_text` field in QueryData. NOTE: The re-construction process can't guarantee that the result is 100% equivalent to the original query.
- prepares protocol message with all parsed quires and sends it to PMM Server.

Additional changes/improvements:
- RTA Stream compression via `gzip` is added
- Some sort of protection from unauthenticated access during RTA stream initialisation is added (via client's metadata)
- QueryData proto message is adjusted accordingly to latest UI requirements (some fields were removed, some were renamed and added)


Makefile.devcontainer
```diff
@@ -2,57 +2,78 @@
 
 include Makefile.include
 
-release-dev-managed:              ## Build pmm-managed
+.PHONY: release-dev-managed
```
ademidoff: What problem does adding .PHONY everywhere solves? Don't we pollute the Makefile?

Makefile.devcontainer
```diff
@@ -2,57 +2,78 @@
...
+release-dev-managed:	## Build pmm-managed
```
ademidoff: Have you run "make" in CLI? I like the output before this change much better than after. Those spaces served a purpose to produce a nice well-spaced output of help messages.

managed/models/agent_helpers.go
```diff
@@ -272,8 +273,9 @@ func FindAgents(q *reform.Querier, filters AgentFilters) ([]*Agent, error) {
 		args = append(args, NomadAgentType)
 		idx++
 	}
+
 	if filters.Disabled != nil {
-		conditions = append(conditions, fmt.Sprintf("disabled = %s", q.Placeholder(idx)))
+		conditions = append(conditions, "disabled = "+q.Placeholder(idx))
```
ademidoff: It's a good fix to satisfy the linter, but it makes it more difficult to review the PR. I suggest we address such fixes in a separate PR.

api/agent/v1/agent.proto
```diff
@@ -86,6 +86,12 @@ message SetStateRequest {
     map<string, string> env = 10;
     // Real-Time Analytics options.
     inventory.v1.RTAOptions rta_options = 11;
+    // Service identifier of the service where the agent connects to.
+    // Currently used by Real-Time Analytics built-in agent only.
+    string service_id = 12;
```
ademidoff, line 91: Does it really belong to the agent message? Don't we mix concerns? Let's discuss.

agent/client/channel/rta_channel.go
```diff
@@ -146,11 +152,14 @@ func (c *RTAChannel) Send(msg *rtav1.CollectRequest) {
 
 	err := c.s.Send(msg)
 	c.sendM.Unlock()
+
 	if err != nil {
 		c.l.Errorf("Failed to send message: %+v", status.Code(err))
-		c.close(errors.Wrap(err, "failed to send message"))
+		c.close(errors.Join(err, errors.New("failed to send message")))
```
ademidoff: I think `errors.Wrap` is more appropriate in this case for two reasons:
* we are not joining multiple errors, which would be a good use case for Join
* we need the stack trace (Join doesn't provide it)

agent/agents/mongodb/realtimeanalytics/mongodb.go
```diff
@@ -106,6 +229,58 @@ func (m *MongoDBRTA) Collect(_ chan<- prometheus.Metric) {
...
+	matchStage := bson.D{{
+		Key: "$match", Value: bson.D{
+			{
+				Key: "$and", Value: bson.A{
+					// Get operations/commands that are active.
+					bson.D{{Key: "active", Value: true}},
+					// Exclude operations from internal MongoDB tools.
+					bson.D{{Key: "desc", Value: bson.D{{Key: "$nin", Value: bson.A{"Checkpointer", "JournalFlusher"}}}}},
+					// Exclude operations from RTA agent itself.
+					bson.D{{Key: "appName", Value: bson.D{{Key: "$not", Value: bson.D{{Key: "$regex", Value: "^(RTA-mongodb-.*$)"}}}}}},
```
ademidoff: Can we consider making the value lowercase, i.e. `rta-mongodb-.*$`?

ademidoff:
Let's also make sure the tests pass.

Makefile.devcontainer
maxkondr: .PHONY - this is a directive for Make that;
- this target does not produce file and
- improves performance

https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

Makefile.devcontainer
maxkondr: it has no differences

agent/client/channel/rta_channel.go
maxkondr: `errors.Wrap` this is method from package `github.com/pkg/errors` but our linter forbids its usage. 
package `errors` has `Join` only

agent/agents/mongodb/realtimeanalytics/mongodb.go
maxkondr: for what purpose?
our official feature name is `Real-Time Analytics` but not `real-time analytics`. So I used such abbreviation here.

api/agent/v1/agent.proto
maxkondr: these are parameters required for agent's logic

managed/models/agent_helpers.go
maxkondr: otherwise linter will complain in next PR.

Makefile.devcontainer
ademidoff: Can you confirm we generate a file called `release-dev-managed`? If not, then we don't have a problem. 

Regarding performance: can you provide evidence, or measure, of the performance impact of this?

managed/models/agent_helpers.go
ademidoff: That's not true - the linter never complains about things which have not been introduced by the current PR. That's how it's configured as of now.

api/agent/v1/agent.proto
ademidoff: I don't see a strong reason to mix concerns - we should keep service attributes outside of agent.proto

agent/client/channel/rta_channel.go
ademidoff: That's true, we are migrating away from `github.com/pkg/errors`, and this is why we should use fmt.Errorf.

agent/agents/mongodb/realtimeanalytics/mongodb.go
ademidoff: MongoDB is also a product name with all those capital letters, however it's written as `mongodb`. I think we should follow the same logic and be consistent with naming.

api/agent/v1/agent.proto
maxkondr: pmm-agent needs some data from pmm-server, for example we pass parameters how to connect to DB (credentials, TLS, socket, etc). 
In this case rta-agent needs some data from pmm-server in order to fully complete its routine. By coincidense this data relates to PMM's service (service ID and Service name).
What the problem with this? 

And if this  is still a problem - what alternative do you suggest?

ademidoff: One option is to extend the `AgentConnectMetadata`, i.e.:

```
// Add this
message AgentServiceContext {
    string service_id = 1;
    string service_name = 2;
}

// ...to the existing AgentConnectMetadata
message AgentConnectMetadata {
    string id = 1;
    string version = 2;
    AgentServiceContext service_context = 3; // there it goes
}
```

maxkondr: `AgentConnectMetadata` serves a completely different purpose - it is handled on connection/stream level (once the connection/stream to server is being established) and aim to exchange info about pmm-agent general info and pmm-server general info, but not a particular built-in agent (or subprocess) parameters required for its business logic. 

Service_id and Service_Name (that are required for rta-agent) related to business logic and used on different level.

agent/agents/mongodb/realtimeanalytics/mongodb.go
```diff
@@ -78,17 +99,145 @@ func (m *MongoDBRTA) Run(ctx context.Context) {
...
+
+	// Parallel parsing of currentOp results using WaitGroup and channel to store
+	// results from goroutines safely to speed up processing of large number of current operations.
+	wg := sync.WaitGroup{}
+
+	// Buffered channel to store results from goroutines(parsers)
+	// and avoid blocking when sending results to the channel.
+	resultsChan := make(chan *rtav1.QueryData, 10) //nolint:mnd
+
+	for cur.Next(ctx) {
```
JiriCtvrtka, line 197: Based on the ticket implementation details, shouldn’t we use recovery here?

JiriCtvrtka (approved):

agent/agents/mongodb/realtimeanalytics/mongodb.go
maxkondr: `recover` is used in `parser.ParseCurrentOp()` (top level parsing function).
As for this place - the method documentation says nothing about `panic`, instead it says that in case any error appears it just return false

PR #5043, opened 2026-02-16
PMM-7 New PG version.

JiriCtvrtka:
PMM-7

### DIFF
--- a/agent/agents/postgres/pgstatmonitor/pgstatmonitor.go
+++ b/agent/agents/postgres/pgstatmonitor/pgstatmonitor.go
@@ -172,9 +172,9 @@ func getPGVersion(q *reform.Querier) (pgVersion, error) {
 	if err != nil {
 		return pgVersion(0), err
 	}
-	v = version.ParsePostgreSQLVersion(v)
+	major, minor := version.ParsePostgreSQLVersion(v)
 
-	parsed, err := strconv.ParseFloat(v, 64)
+	parsed, err := strconv.ParseFloat(fmt.Sprintf("%s.%s", major, minor), 64)
 
 	return pgVersion(parsed), err
 }
--- a/agent/agents/postgres/pgstatmonitor/pgstatmonitor.go
+++ b/agent/agents/postgres/pgstatmonitor/pgstatmonitor.go
@@ -173,7 +173,9 @@ func getPGVersion(q *reform.Querier) (pgVersion, error) {
 		return pgVersion(0), err
 	}
 	major, minor := version.ParsePostgreSQLVersion(v)
-
+	if minor == "" {
+		minor = "0"
+	}
 	parsed, err := strconv.ParseFloat(fmt.Sprintf("%s.%s", major, minor), 64)
 
 	return pgVersion(parsed), err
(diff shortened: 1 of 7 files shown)


JiriCtvrtka:
This time they changed queryid generation in scope of minor version. So more changes are included, not just new case as usual.

agent/agents/postgres/pgstatstatements/pgstatstatements_test.go
```diff
@@ -163,11 +163,19 @@ func TestPGStatStatementsQAN(t *testing.T) {
 	case "18":
 		selectAllCitiesLong = "SELECT /* AllCitiesTruncated:pgstatstatements controller='test' */ * FROM city WHERE id IN ($1 /*, ... */)"
 		truncatedMSharedBlksHitSum = float32(8)
-		digests = map[string]string{
-			selectAllCities:     "2398197226709363629",
-			selectAllCitiesLong: "-1570108445478818403",
-		}
 		isTruncated = false
+		switch minorVersion {
```
JiriCtvrtka, line 167: Checking also minor version for PG18.

agent/utils/version/postgresql.go
```diff
@@ -25,21 +25,16 @@ var (
...
-	switch len(parts) {
-	case 1: // major only
-		return parts[0]
-	case 2: // major and patch
-		return parts[0]
-	case 3: // major, minor, and patch
-		return parts[0] + "." + parts[1]
-	default:
-		return ""
+	if len(parts) == 1 {
```
ademidoff, line 42: Does it mean we no longer have the patch version?

ademidoff (approved):

agent/utils/version/postgresql.go
JiriCtvrtka: I checked the old code, and it was incorrect (patch not returned) and not used at all. Also, we are not checking for the patch anywhere at the moment.

agent/agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -172,9 +172,11 @@ func getPGVersion(q *reform.Querier) (pgVersion, error) {
...
 		return pgVersion(0), err
 	}
-	v = version.ParsePostgreSQLVersion(v)
-
-	parsed, err := strconv.ParseFloat(v, 64)
+	major, minor := version.ParsePostgreSQLVersion(v)
+	if minor == "" {
+		minor = "0"
+	}
+	parsed, err := strconv.ParseFloat(fmt.Sprintf("%s.%s", major, minor), 64)
```
maxkondr, line 179: can't we use lib for version parsing instead of `strconv` ?

JiriCtvrtka: I am not aware of such func in lib/pq. 

In semver there is something, but I am not sure if it makes sense to use library since it is only one ParseFloat
```
v, err := semver.NewVersion("18.2")
if err != nil {
    // handle error
}
v.Major(), v.Minor(), v.Patch() etc.
```

maxkondr (approved):

PR #5064, opened 2026-02-23
Pmm-14550: preferred timezone when displaying dates

fabio-silva:
PMM-14550

Listen for timezone changed events and apply it right away, thus allowing users to see dates on their set timezone

Works together with https://github.com/percona/grafana/pull/878


ui/docker-compose.yml
```diff
@@ -14,14 +14,16 @@ services:
       # - pmm-data:/srv
 
       # Uncomment to use custom (FE) grafana code
-      #- '../../grafana/public:/usr/share/grafana/public'
```
matejkubinec, line 17: Are changes to this file intentional?

fabio-silva: not at all. Added the comment back, thanks!

matejkubinec (approved):

PR #5074, opened 2026-02-24
PMM-14376 Enrich managed metrics with more labels

ademidoff:
PMM-14376

Link to the Feature Build: SUBMODULES-4234

### DIFF
--- a/managed/services/inventory/inventory_metrics.go
+++ b/managed/services/inventory/inventory_metrics.go
@@ -71,7 +71,7 @@ func NewInventoryMetricsCollector(metrics inventoryMetrics) *InventoryMetricsCol
 		mAgentsDesc: prom.NewDesc(
 			prom.BuildFQName(prometheusNamespace, prometheusSubsystem, "agents"),
 			"Inventory Agent",
-			[]string{"agent_id", "agent_type", "service_id", "node_id", "node_name", "pmm_agent_id", "disabled", "version"},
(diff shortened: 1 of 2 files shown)


JiriCtvrtka (approved):

managed/services/inventory/inventory_metrics.go
maxkondr, line 104: is transaction really required here?

managed/services/inventory/inventory_metrics.go
```diff
@@ -119,11 +119,21 @@ func (i *InventoryMetrics) GetAgentMetrics(ctx context.Context) ([]Metric, error
 			return err
 		}
 
+		dbServices, err := models.FindServices(tx.Querier, models.ServiceFilters{})
+		if err != nil {
+			return err
+		}
+
```
maxkondr, line 126: These 3 SQLs can be replaced with 1 with JOINs

maxkondr (approved):

managed/services/inventory/inventory_metrics.go
ademidoff: Yes, we can do it using a raw SQL query. The problem is that we want to use the model as much as we can to access the data and stay away from raw queries keeping in mind the maintenance effort.

managed/services/inventory/inventory_metrics.go
ademidoff: No, not really required. I think we should treat it as tech debt, since there are too many places like that to refactor.

PR #5075, opened 2026-02-24
PMM-14635 Add adding RTA agent to pmm-admin CLI

maxkondr:
PMM-14635

Extend pmm-admin CLI with possibility to create RTA agent. 
Removing RTA agent is already covered by general functionality.


JiriCtvrtka (approved):

ademidoff (approved):
LGTM

Can we add the `remove` agent as well? They should always go together.

PR #5109, opened 2026-03-05
fix: bad stored QAN short URL

fabio-silva:
- Short URL for QAN was being created as absolute, not allowing the clipboard function;
- QAN URLs have to be sent without `pmm-ui/graph` prefix, or else the resulting URL won't work


matejkubinec (approved):

matejkubinec:
@fabio-silva could you also change the tooltip text when on stored metrics?

image

Not sure if we should use the same for stored/real-time.

fabio-silva:
> @fabio-silva could you also change the tooltip text when on stored metrics?
> 
> image
> Not sure if we should use the same for stored/real-time.

Just talked to Pedro, we're keeping it 👍

PR #5116, opened 2026-03-06
PMM-14900 Add protection to RTA session against unsupported pmm-agent version

maxkondr:
PMM-14900

New endpoint is introduced `/v1/realtimeanalytics/services`:
- It supports service_type  filter as incoming parameter. 
- In case filter is passed - returns service that support RTA of specified type only. (taking into account pmm-agent version as well). For the time being it is MongoDB only, but will be extended in future.
- in case filter is not defined-  returns service that support RTA (taking into account pmm-agent version as well)
it has return interface equal to `/v1/inventory/services`

In case user requested start RTA session for the MongoDB service that is registered in PMM via pmm-agent <3.7.0 version - API returns an error.

Link to the Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/4256


managed/services/realtimeanalytics/service.go
```diff
@@ -215,6 +216,33 @@ func (s *Service) StartSession(ctx context.Context, req *rtav1.StartSessionReque
 				service.ServiceType, service.ServiceID)
 		}
 
+		// Check that a corresponding pmm-agent has version supporting RTA.
+		pmmAgent, err := models.FindAgentByID(tx.Querier, *existingAgent.PMMAgentID)
+		if err != nil {
+			return err
+		}
+
+		// Get agent version
```
ademidoff:
```suggestion
```

managed/services/realtimeanalytics/service.go
```diff
@@ -215,6 +216,33 @@ func (s *Service) StartSession(ctx context.Context, req *rtav1.StartSessionReque
 				service.ServiceType, service.ServiceID)
 		}
 
+		// Check that a corresponding pmm-agent has version supporting RTA.
```
ademidoff:
```suggestion
		// Retrieve the agent
```

ademidoff: 1. That's not what it does 😉 
2. Do we need such trivial comments at all?

managed/services/realtimeanalytics/service.go
```diff
@@ -215,6 +216,33 @@ func (s *Service) StartSession(ctx context.Context, req *rtav1.StartSessionReque
...
+		if err != nil {
+			return err
+		}
+
+		// Get agent version
+		if pmmAgent.Version == nil {
+			return status.Errorf(codes.FailedPrecondition, "pmm-agent with ID %s has no version specified.", *existingAgent.PMMAgentID)
+		}
+
+		var (
```
ademidoff: Do we need to declare these variables at all? IMO it seems to be verbose and redundant as long as they can be declared by assignment (:=) on line 235.

managed/services/realtimeanalytics/service.go
```diff
@@ -215,6 +216,33 @@ func (s *Service) StartSession(ctx context.Context, req *rtav1.StartSessionReque
...
+		)
+
+		pmmAgentVersion, versionParseErr = version.Parse(*pmmAgent.Version)
+		if versionParseErr != nil {
+			return status.Errorf(codes.InvalidArgument, "Can't parse 'version' for pmm-agent with ID %q.", *existingAgent.PMMAgentID)
+		}
+
+		if !pmmAgentVersion.IsFeatureSupported(version.MongoDBRtaAgentSupportVersion) {
+			return status.Errorf(codes.FailedPrecondition,
+				"Service %s has pmm-agent with version not supporting Real-Time Analytics. Minimum required version is %s",
```
ademidoff: @catalinaadam Could you please help with this error message? It doesn't read well IMO.

managed/services/realtimeanalytics/service.go
ademidoff: That's not what it does 😉

managed/services/realtimeanalytics/service.go
maxkondr: This comment is not for the next line only but for next block of logic - and yes, it does exactly what it states.

managed/services/realtimeanalytics/service.go
maxkondr: Removed

managed/services/realtimeanalytics/service.go
```diff
@@ -66,6 +66,80 @@ func NewService(db *reform.DB, registry agentsRegistry, stateUpdater agentsState
...
+
+			serviceList = append(serviceList, tmpServiceList...)
+		}
+	}
+
+	res := &rtav1.ListServicesResponse{}
+
+	for _, svc := range serviceList {
+		// Check that service has pmm-agent with version supporting RTA.
+		pmmAgent, err := models.FindPMMAgentsForService(s.db.Querier, svc.ServiceID)
```
ademidoff: This line makes the `ListServices` endpoint very expensive - it will make O(n) calls to the database. Let's consider the following to make it faster:

1. Fetch all agents
2. Make an in-memory data structure that maps Agents to a ServiceId, i.e. of type map[string]*models.Agent
3. Use the map instead of the call to the database

maxkondr: the problem with the proposed approach is that Service is not directly linked to PMM Agent (doesn't have PMM Agent ID filled and visa versa). 

In order to make such mapping it is required to repeat the logic from `agent_helpers.FindPMMAgentsForService` that results into the same -> no benefits.

JiriCtvrtka (approved):

managed/services/realtimeanalytics/service.go
ademidoff: I checked the implementation of `FindPMMAgentsForService`, it's even worse - we are dealing with O(n*3) requests, which is way more expensive than it should be.

Let's discuss a better way.

fabio-silva (approved):

ademidoff (approved):

matejkubinec (approved):

managed/services/realtimeanalytics/service.go
```diff
@@ -106,29 +114,73 @@ func (s *Service) ListServices(_ context.Context, req *rtav1.ListServicesRequest
...
+	results := make([]*inventoryv1.Service, len(serviceList))
+	for i, svc := range serviceList {
+		g.Go(func() error {
+			// Check that service has pmm-agent with version supporting RTA.
+			pmmAgent, err := models.FindPMMAgentsForService(s.db.Querier, svc.ServiceID)
+			if err != nil {
+				return fmt.Errorf("failed to find pmm-agent for service with ID %s: %w", svc.ServiceID, err)
+			}
+
+			if !isPmmAgentSupportRta(pointer.Get(pointer.Get(pmmAgent[0]).Version)) {
```
ademidoff: What if line 151 returns an empty array, which is not impossible? Seems like a potential out of bounds exception.

maxkondr: in such a case `pointer.Get()` will return empty structure with default values for fields -> in result there will be empty string

ademidoff: You didn't get my point - `pmmAgent[0]` will not resolve to a value and break the execution.

ademidoff: @copilot Is this a valid concern?

maxkondr: Handled

ademidoff (approved):

PR #5125, opened 2026-03-09
PMM-14900: fetch only available services for RTA

fabio-silva:
PMM-14900

### DIFF
--- a/ui/apps/pmm/src/api/rta.ts
+++ b/ui/apps/pmm/src/api/rta.ts
@@ -54,5 +54,5 @@ export const getAvailableServices = async (): Promise<
   const res = await api.get<AvailableServicesResponse>(
     '/realtimeanalytics/services'
   );
-  return (res.data || {}).mongodb || [];
+  return res.data.mongodb;
 };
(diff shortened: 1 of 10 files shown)


ui/apps/pmm/src/hooks/api/useRealtime.ts
```diff
@@ -122,31 +123,28 @@ export const useAvailableServices = () => {
   const { user } = useUser();
   const { data: sessions, isLoading: isLoadingSessions } =
     useRealtimeSessions();
-  const { data: services, isLoading: isLoadingServices } = useManagedServices({
-    serviceType: ServiceType.mongodb,
-  }, {
+  const { data: services = [], isLoading: isLoadingServices } = useQuery({
```
matejkubinec: Shouldn't the api endpoint be available only to admins (RTA is admin only at the moment).

To add to it, the query doesn't run for non admins (see `enabled`)

ui/apps/pmm/src/api/rta.ts
```diff
@@ -46,3 +47,12 @@ export const searchQueries = async (
   );
   return res.data;
 };
+
+export const getAvailableServices = async (): Promise<
```
matejkubinec:
```suggestion
export const getAvailableMongoDBServices = async (): Promise<
```

Or pass a parameter (haven't checked, but didn't Maxim mention a parameter to pass to get only specific services types?)

matejkubinec (approved):

PR #5130, opened 2026-03-10
Pmm 14843 alternative renderer sidebar hide

fabio-silva:
PMM-14843

Use a query param to detect Grafana is visited by the Renderer, instead of relying on User Agent


matejkubinec (approved):
@fabio-silva looks good to me, just add the test case for sidebar which copilot mentions.

PR #5170, opened 2026-03-24
PMM-14880 anonymous role

fabio-silva:
PMM-14880

Allow org roles to anonymous users

Tied to https://github.com/percona/grafana/pull/886

Note: we fallback to "Viewer" if any other org_role is used on anon mode, due to this deprecation notice: 
https://github.com/grafana/grafana/pull/101411/changes


managed/services/grafana/client.go
```diff
@@ -302,6 +342,161 @@ func (c *Client) convertRole(role string) role {
...
+
+type frontendSettingsFull struct {
+	AnonymousEnabled bool                     `json:"anonymousEnabled"`
+	AnonymousOrgRole string                   `json:"anonymousOrgRole"`
+	User             frontendUserSettingsFull `json:"user"`
+}
+
+func (c *Client) getAnonymousRoleFromSettings(ctx context.Context, l *logrus.Entry) (bool, role) {
+	var settings frontendSettings
+	if err := c.do(ctx, http.MethodGet, "/api/frontend/settings", "", nil, nil, &settings); err != nil {
```
maxkondr, line 369: there is a separate method introduced `getFrontendSettings` for this

managed/services/grafana/client.go
```diff
@@ -168,6 +168,30 @@ type authUser struct {
 	userID int
 }
 
+// CurrentUser represents Grafana user payload.
+type CurrentUser struct {
```
maxkondr, line 172: are all fields needed in these structs

managed/services/grafana/client.go
```diff
@@ -302,6 +342,161 @@ func (c *Client) convertRole(role string) role {
...
+	if parsedRole == none {
+		parsedRole = c.convertRole(settings.AnonymousOrgRole)
+	}
+	l.Debugf("Grafana anonymous mode is enabled with role %q.", parsedRole.String())
+	return true, parsedRole
+}
+
+func (c *Client) getFrontendSettings(ctx context.Context) (frontendSettingsFull, error) {
+	var settings frontendSettingsFull
+	if err := c.do(ctx, http.MethodGet, "/api/frontend/settings", "", nil, nil, &settings); err != nil {
```
maxkondr: the same - without proper authHeaders this call returns 401

managed/services/grafana/client.go
```diff
@@ -302,6 +342,161 @@ func (c *Client) convertRole(role string) role {
...
+}
+
+type frontendSettingsFull struct {
+	AnonymousEnabled bool                     `json:"anonymousEnabled"`
+	AnonymousOrgRole string                   `json:"anonymousOrgRole"`
+	User             frontendUserSettingsFull `json:"user"`
+}
+
+func (c *Client) getAnonymousRoleFromSettings(ctx context.Context, l *logrus.Entry) (bool, role) {
+	var settings frontendSettings
```
maxkondr, line 368: in this method the result from API endpoint `/api/frontend/settings` is parsed into `frontendSettings` struct, but the same endpoint call in `getFrontendSettings` is parsed into `frontendSettingsFull` struct. What is the purpose?

managed/services/grafana/client.go
```diff
@@ -302,6 +342,161 @@ func (c *Client) convertRole(role string) role {
...
+		Login:                          "anonymous",
+		CreatedAt:                      "",
+		OrgID:                          orgID,
+		IsAnonymous:                    true,
+		IsDisabled:                     false,
+		IsExternal:                     false,
+		IsExtarnallySynced:             false,
+		IsGrafanaAdmin:                 false,
+		IsGrafanaAdminExternallySynced: false,
+		Theme:                          "",
```
maxkondr, line 452: not sure these fields are required here

managed/services/grafana/client.go
```diff
@@ -302,6 +342,161 @@ func (c *Client) convertRole(role string) role {
...
+// a synthetic anonymous user when /api/user responds with 401.
+func (c *Client) GetCurrentUser(ctx context.Context, authHeaders http.Header) (CurrentUser, error) {
+	var user CurrentUser
+	err := c.do(ctx, http.MethodGet, "/api/user", "", authHeaders, nil, &user)
+	if err == nil {
+		return user, nil
+	}
+
+	var cErr *clientError
+	if !errors.As(errors.Cause(err), &cErr) || cErr.Code != http.StatusUnauthorized || hasAuthHeaders(authHeaders) {
```
maxkondr, line 420: 1. looks like no need to call `Cause()` because `errors.As()` iterates though errors chain
2. seems the condition shall be
```
if (errors.As(err, &cErr) && cErr.Code != http.StatusUnauthorized) || hasAuthHeaders(authHeaders) {
```
because `cErr.Code` has value only in case `errors.As()` call was successful

managed/services/grafana/client.go
```diff
@@ -302,6 +342,161 @@ func (c *Client) convertRole(role string) role {
...
+// a synthetic org list when /api/user/orgs responds with 401.
+func (c *Client) GetCurrentUserOrgs(ctx context.Context, authHeaders http.Header) ([]CurrentUserOrg, error) {
+	var orgs []CurrentUserOrg
+	err := c.do(ctx, http.MethodGet, "/api/user/orgs", "", authHeaders, nil, &orgs)
+	if err == nil {
+		return orgs, nil
+	}
+
+	var cErr *clientError
+	if !errors.As(errors.Cause(err), &cErr) || cErr.Code != http.StatusUnauthorized || hasAuthHeaders(authHeaders) {
```
maxkondr, line 467: the same with `Cause()` and condition as above

managed/services/user/current_http.go
```diff
@@ -0,0 +1,68 @@
...
+	switch req.URL.Path {
+	case "/v1/users/current":
+		user, err := h.c.GetCurrentUser(req.Context(), authHeaders)
+		if err != nil {
+			h.l.WithError(err).Warn("failed to get current user")
+			rw.WriteHeader(http.StatusUnauthorized)
+			_ = json.NewEncoder(rw).Encode(map[string]string{"message": "Unauthorized"})
+			return
+		}
+		if user == (grafana.CurrentUser{}) {
```
maxkondr, line 51: what does it mean when user is empty? is there any difference with the case when user is not authorised?

managed/services/user/current_http.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}
+
+	rw.Header().Set("Content-Type", "application/json")
+
+	switch req.URL.Path {
+	case "/v1/users/current":
+		user, err := h.c.GetCurrentUser(req.Context(), authHeaders)
+		if err != nil {
+			h.l.WithError(err).Warn("failed to get current user")
```
maxkondr, line 46: in pmm-managed the following way for printing errors is used:

```suggestion
			h.l.Errorf("failed to get current user: %v", err)
```

managed/services/user/current_http.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+		if user == (grafana.CurrentUser{}) {
+			_ = json.NewEncoder(rw).Encode(map[string]any{})
+			return
+		}
+		_ = json.NewEncoder(rw).Encode(user)
+	case "/v1/users/current/orgs":
+		orgs, err := h.c.GetCurrentUserOrgs(req.Context(), authHeaders)
+		if err != nil {
+			h.l.WithError(err).Warn("failed to get current user orgs")
```
maxkondr, line 59:
```suggestion
			h.l.Errorf("failed to get current user orgs: %v", err)
```

managed/services/grafana/client.go
```diff
@@ -242,10 +266,26 @@ func (c *Client) getAuthUser(ctx context.Context, authHeaders http.Header, l *lo
...
+	if authHeaders.Get("Authorization") == "" && authHeaders.Get("Cookie") == "" {
+		anonymousEnabled, anonymousRole = c.getAnonymousRoleFromSettings(ctx, l)
+	}
+
 	// https://grafana.com/docs/http_api/user/#actual-user - works only with Basic Auth
 	var m map[string]interface{}
 	err := c.do(ctx, http.MethodGet, "/api/user", "", authHeaders, nil, &m)
 	if err != nil {
+		var cErr *clientError
+		if anonymousEnabled && errors.As(errors.Cause(err), &cErr) && cErr.Code == http.StatusUnauthorized {
```
maxkondr, line 282: looks like `errors.Cause()` is not needed here

managed/services/grafana/client.go
```diff
@@ -302,6 +342,161 @@ func (c *Client) convertRole(role string) role {
...
+	}
+
+	settings, settingsErr := c.getFrontendSettings(ctx)
+	if settingsErr != nil || !settings.AnonymousEnabled {
+		return CurrentUser{}, err
+	}
+	if c.resolveAnonymousRole(settings) == none.String() {
+		// Anonymous mode is enabled but role is not configured.
+		// Return empty payload instead of Unauthorized.
+		return CurrentUser{}, nil
```
maxkondr, line 431: I think it is better to return an `Unauthorized`  error here and decrease the number of variations (empty user w/ error vs empty user w/o error, etc). Anything (wrong credentials, other errors, not properly configured anonymous mode) - returns error

managed/services/grafana/client.go
```diff
@@ -302,6 +342,161 @@ func (c *Client) convertRole(role string) role {
...
+
+	settings, settingsErr := c.getFrontendSettings(ctx)
+	if settingsErr != nil || !settings.AnonymousEnabled {
+		return nil, err
+	}
+	role := c.resolveAnonymousRole(settings)
+	if role == none.String() {
+		// Anonymous mode is enabled but role is not configured.
+		// Return empty payload instead of Unauthorized.
+		return []CurrentUserOrg{}, nil
```
maxkondr, line 479: the same regarding error as in `GetCurrentUser`

fabio-silva:
We need profiling on this one.
CC @maxkondr

managed/services/grafana/client.go
fabio-silva: done!

managed/services/grafana/client.go
fabio-silva: Kept response compatibility, but reduced unnecessary explicit assignments

managed/services/grafana/client.go
fabio-silva: Addressed

managed/services/grafana/client.go
fabio-silva: now setting only required fields

managed/services/grafana/client.go
```diff
@@ -302,6 +368,154 @@ func (c *Client) convertRole(role string) role {
...
+}
+
+// resolveAnonymousRole maps Grafana frontend settings to an effective anonymous org role string.
+//
+// Grafana maps [auth.anonymous] org_role into anonymousOrgRole (and may omit user.orgRole). We honor
+// anonymousOrgRole Viewer so org_role = Viewer in Grafana configuration is applied. Deprecated elevated
+// roles (Editor/Admin/GrafanaAdmin) in either field are clamped to Viewer.
+func (c *Client) resolveAnonymousRole(settings frontendSettingsFull) string {
+	userOrg := strings.TrimSpace(settings.User.OrgRole)
+	anonOrg := strings.TrimSpace(settings.AnonymousOrgRole)
```
maxkondr:
```suggestion
	userOrgRole := strings.TrimSpace(settings.User.OrgRole)
	anonOrgRole := strings.TrimSpace(settings.AnonymousOrgRole)
```

managed/services/grafana/client.go
```diff
@@ -302,6 +368,154 @@ func (c *Client) convertRole(role string) role {
...
+func hasAuthorizationHeader(authHeaders http.Header) bool {
+	return authHeaders.Get("Authorization") != ""
+}
+
+// resolveAnonymousRole maps Grafana frontend settings to an effective anonymous org role string.
+//
+// Grafana maps [auth.anonymous] org_role into anonymousOrgRole (and may omit user.orgRole). We honor
+// anonymousOrgRole Viewer so org_role = Viewer in Grafana configuration is applied. Deprecated elevated
+// roles (Editor/Admin/GrafanaAdmin) in either field are clamped to Viewer.
+func (c *Client) resolveAnonymousRole(settings frontendSettingsFull) string {
```
maxkondr: 1. No need to make it struct's func
2. pass not whole `settings` but an exact fields from it - it will be easier to maintain and test this func

managed/services/grafana/client.go
```diff
@@ -302,6 +368,154 @@ func (c *Client) convertRole(role string) role {
...
+	userOrg := strings.TrimSpace(settings.User.OrgRole)
```
maxkondr: how is it possible to have some value here? On case user is authenticated - it will be handled by another logic. If we reach here, it means user has no role at all. Is it really needed to check his role at all in such a case?

managed/services/grafana/client.go
```diff
@@ -302,6 +368,154 @@ func (c *Client) convertRole(role string) role {
...
+	anonOrg := strings.TrimSpace(settings.AnonymousOrgRole)
+
+	switch {
+	case userOrg == viewer.String():
+		return viewer.String()
+	case userOrg == editor.String() || userOrg == admin.String() || userOrg == grafanaAdmin.String():
+		return viewer.String()
```
maxkondr: replace all such calls `viewer.String()` with prepared beforehand strings.

managed/services/grafana/client.go
fabio-silva: Yeah, good point. In the beginning I was in believe Grafana was setting orgRole regardless, but just confirmed we just rely on anon org role.

maxkondr (approved):

PR #5171, opened 2026-03-24
PMM-14856 Make multiline output of telemetry metrics

maxkondr:
PMM-14856

Link to the Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/4281


managed/services/telemetry/telemetry.go
```diff
@@ -114,32 +136,29 @@ func (s *Service) Run(ctx context.Context) {
...
 		if !settings.IsTelemetryEnabled() {
-			s.l.Info("Disabled via settings.")
+			s.l.Info("Telemetry is disabled via settings.")
 			return
 		}
 
 		report := s.prepareReport(ctx)
 
-		s.l.Debugf("\nTelemetry captured:\n%s\n", s.Format(report))
+		if s.l.Logger.IsLevelEnabled(logrus.DebugLevel) {
```
ademidoff, line 152: I wonder if the logger is not aware of its debug level already? This `if` seems to be redundant, no?

docker-compose.yml
```diff
@@ -19,28 +19,30 @@ services:
...
       # - PMM_ENABLE_INTERNAL_PG_QAN=${PMM_ENABLE_INTERNAL_PG_QAN:-1}
       # - PMM_DISTRIBUTION_METHOD=${PMM_DISTRIBUTION_METHOD:-docker}
       # - PMM_DEV_UPDATE_DOCKER_IMAGE=${PMM_DEV_UPDATE_DOCKER_IMAGE:-}
       # - PMM_DEV_PERCONA_PLATFORM_ADDRESS=https://check.localhost
       # - PMM_DEV_PERCONA_PLATFORM_INSECURE=1
       # - PMM_DEV_PERCONA_PLATFORM_PUBLIC_KEY=<public key>
-      # - PMM_DEV_TELEMETRY_INTERVAL=10s
-      # - PMM_DEV_TELEMETRY_DISABLE_START_DELAY=1
-      # - PMM_DEV_TELEMETRY_RETRY_BACKOFF=10s
+      - PMM_ENABLE_TELEMETRY=${PMM_ENABLE_TELEMETRY:-0}
```
ademidoff: Ideally, we should clean up this file and leave it for community. I suggest we create a copy of this file in `/dev`, where it'd make more sense to set them as default values for development & testing purposes.

It's much better to address it within https://perconadev.atlassian.net/browse/PMM-14887, where its already in the scope.

managed/services/telemetry/telemetry.go
maxkondr: no, this `if` is necessary here because:
1. All function params are evaluated during function call
2. Log level is checked inside `Debugf` function - it means that in case debug level is turned off, the function call `s.Format(report)` was useless wasting of resources

docker-compose.yml
maxkondr: reverted

ademidoff (approved):

JiriCtvrtka (approved):

PR #5205, opened 2026-04-02
PMM-14962 Avoid printing sensitive info into log stream

maxkondr:
PMM-14962

Link to the Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/4290

1. New Protobuf extension is introduced `sensitive ` that contains several values:
- REDACT_TYPE_FULL - fully replace string value with redact mask
- REDACT_TYPE_MASK - partially replace string value with redact mask
- REDACT_TYPE_DSN - replace credentials only in DSN with redact mask

2. Our protobuf message fields were annotated accordingly like the following:
```
message BuiltinAgent {
    inventory.v1.AgentType type = 1;
    string dsn = 2 [(extensions.v1.sensitive) = REDACT_TYPE_DSN];
    ....
}
```

3. New logger helper function is introduced: `logger.RedactMessage` that shall be used in case proto message shall be printing into log stream. It does:
- lookups sensitive annotation for fields in proto message
- applies value redaction accordingly to define sensitivity (REDACT_TYPE_FULL, REDACT_TYPE_MASK, REDACT_TYPE_DSN)
- returns a copy of the original message with redacted values - so that it Is safe to use it without risk to change original message.
- it uses the internal cache in order to speedup reflection process - the first message processing builds the cache for this message type, the rest messages with the same message type will be processed much faster.
- this function shall be used like the following :
```
c.l.Debugf("Sending message (%d bytes): %s.", size, logger.RedactMessage(msg))
or
c.l.Debugf("Sending message (%d bytes):\n%s\n", size, prototext.Format(logger.RedactMessage(msg)))
```

4. The log output will look like the following:
```
time="2026-04-02T11:00:17.048+00:00" level=debug msg="Starting: 
/usr/local/percona/pmm/exporters/postgres_exporter ... (environment: 
DATA_SOURCE_NAME=postgres://***REDACTED***:***REDACTED***@127.0.0.1:5432/postgres?
connect_timeout=1&sslmode=disable)." ...
```


JiriCtvrtka (approved):

utils/logger/protobuf.go
```diff
@@ -0,0 +1,218 @@
...
+	case extensionsv1.RedactType_REDACT_TYPE_DSN:
+		return maskDSN(s)
+	default:
+		return s
+	}
+}
+
+func maskString(s string) string {
+	n := len(s)
+	if n == 0 {
```
ademidoff: A switch statement would work better here.

utils/logger/protobuf.go
```diff
@@ -0,0 +1,218 @@
...
+		return s
+	}
+	if n <= 4 { //nolint:mnd
+		return maskedString
+	}
+	return s[:2] + maskedString + s[n-2:]
+}
+
+func maskDSN(s string) string {
```
ademidoff, line 195: Can we add a unit test for `maskDSN` please?

go.mod
```diff
@@ -58,6 +59,7 @@ require (
 	github.com/prometheus/prometheus v0.310.0
 	github.com/ramr/go-reaper v0.3.1
 	github.com/robfig/cron/v3 v3.0.1
+	github.com/samber/lo v1.53.0
```
ademidoff: I'm not a big fan of adding more dependencies in cases where we can easily get away without them. We are already using too many, which creates additionial maintenance workload.

This is a utility package, IMO it's not difficult to use plain Go.

ademidoff (changes requested):
Good job! What is the performance implication of this?

go.mod
maxkondr: this is a good utility package to start using actually - there are many places in code where it will be handy

utils/logger/protobuf.go
maxkondr: https://github.com/percona/pmm/pull/5205/changes/BASE..f4435f43d8423220b53e115d364af90e9d14de01#diff-abe00a97b09a5992dc58d666e0adb1a13479321d705f1e84bd93e22adcabc25dR250

utils/logger/protobuf.go
maxkondr: done

go.mod
ademidoff: Not really, there is a performance tax to pay. It only looks nice, but I'd definitely stay away from iterating through slices by calling a method on each value, that's too expensive if we want PMM to be performant. 

Let's keep in mind these methods are used in logs, and PMM produces a lot of logs, so the concern is quite valid.

utils/logger/protobuf.go
ademidoff: Sorry, I missed it :)

go.mod
maxkondr: 1. This is exactly what is required here - call function for each item in slice. The suggested lib just make it easier/handy.
2. this lib contains a lot of functions that are very useful with different containers/structs handling and are absent in stdlib unfortunately.

maxkondr:
> Good job! What is the performance implication of this?

Results from benchmark tests:
```
BenchmarkFormatWithRedactionSetStateRequest-12       26769  43920 ns/op  69680 B/op  683 allocs/op
BenchmarkFormatWithoutRedactionSetStateRequest-12    41848  28942 ns/op  55237 B/op  376 allocs/op
```

But there are points:
1. Performance penalty is applied with enabled debug logging only
2. The redaction function is cached-based: the slowest execution will be when it processes the message type first. On the next iterations with the same message type it uses local cache and avoid reflection of the whole message.

managed/services/agents/state.go
```diff
@@ -321,7 +321,12 @@ func (u *StateUpdater) sendSetStateRequest(ctx context.Context, agent *pmmAgentI
 		AgentProcesses: agentProcesses,
 		BuiltinAgents:  builtinAgents,
 	}
-	l.Debugf("sendSetStateRequest:\n%s", prototext.Format(state))
+
+	// Check log level before calling formatting function.
+	// Do not waste resources in case debug level is not enabled.
```
ademidoff, line 326:
```suggestion
```
That's an obvious thing that derives from the `if` statement.

ademidoff (approved):

PR #5213, opened 2026-04-04
PMM-14825 Fix tests related to disabling telemetry in FB

ademidoff:
PMM-14825

Link to the Feature Build: SUBMODULES-4292


api-tests/inventory/agents_test.go
```diff
@@ -92,13 +92,14 @@ func TestAgents(t *testing.T) {
 		require.NotEmpty(t, resByAgent.Payload.MysqldExporter, "There should be at least one service")
 		assertMySQLExporterExists(t, resByAgent, mySqldExporterID)
 
-		resByNode, err := client.Default.AgentsService.ListAgents(&agents.ListAgentsParams{
-			NodeID:  pointer.ToString(nodeID),
+		resAgent, err := client.Default.AgentsService.GetAgent(&agents.GetAgentParams{
```
maxkondr: what is the purpose of changing this? The prev version was quite logical

managed/services/telemetry/telemetry.go
```diff
@@ -103,13 +104,20 @@ func (s *Service) LocateTelemetryDataSource(name string) (DataSource, error) {
...
+var releaseVersionRe = regexp.MustCompile(`^\d+\.\d+(\.\d+)?$`)
+
+// Run sends telemetry.
 func (s *Service) Run(ctx context.Context) {
 	if !s.config.Enabled {
 		s.l.Warn("service is disabled, skip Run")
 		return
 	}
 
+	if !releaseVersionRe.MatchString(s.pmmVersion) {
```
maxkondr: we have `pmm/version` package for such purposes

managed/services/telemetry/telemetry.go
```diff
@@ -103,13 +104,20 @@ func (s *Service) LocateTelemetryDataSource(name string) (DataSource, error) {
 	return s.dsRegistry.LocateTelemetryDataSource(name)
 }
 
-// Run start sending telemetry to SaaS.
+// releaseVersionRe matches only clean release versions like "3.7.0" or "3.7".
+var releaseVersionRe = regexp.MustCompile(`^\d+\.\d+(\.\d+)?$`)
```
maxkondr: will it catch the upcoming `3.7.0-1` ?

ademidoff: Good question: will it be `3.7.0-1` or `3.7.1`?

api-tests/inventory/agents_test.go
ademidoff: I was trying to fix a flaky test that was failing at this step:

```
00:05:20.653  === NAME  TestAgents/List
00:05:20.653      agents_test.go:83: 
00:05:20.653          	Error Trace:	/go/pmm/api-tests/inventory/agents_test.go:83
00:05:20.653          	Error:      	Received unexpected error:
00:05:20.653          	            	[GET /v1/inventory/agents][404] ListAgents default {"code":5,"message":"cannot find pmm_agent by id: 50032140-81f1-4a47-bde6-6618d9afee03, for external_exporter id: fdb7b784-4379-4002-941c-b65ce2235af8 without node_id: rpc error: code = NotFound desc = Agent with ID 50032140-81f1-4a47-bde6-6618d9afee03 not found.","details":[]}
```

If I put it short, `ListAgents` is exposed to a race condition.

Longer version: `ListAgents` fetches all agents in one DB query, then converts them one by one in a separate pass. If a parallel test deletes a `pmmAgent` between those two steps, the `FindAgentByID` lookup in `ToAPIAgent` for an external exporter (which has `PMMAgentID` but no `RunsOnNodeID`) fails with 404.

TLDR: you are right that the test was focusing on ListAgents, not GetAgent, so I'll try to propose a different fix.

maxkondr: Replacing `ListAgents` with `GetAgent` doesn't solve the issue at all, it only decreases the chance of issue appearance (in fact just postponing it).
The proper testing design would look like - using isolated (idempotent) testing (unique IDs across tests), so that test cases do not interfere with each other

ademidoff: Sure, I don't argue, I'm just trying to fix the failing test with minimal effort, while fixing the flaky test design would probably warrant a separate ticket.

I'll follow up...

managed/services/telemetry/telemetry.go
ademidoff: Nice catch, done!

api-tests/inventory/agents_test.go
ademidoff: OK, I think it should be OK now. Thanks!

managed/services/telemetry/telemetry.go
```diff
@@ -125,13 +126,19 @@ func (s *Service) LocateTelemetryDataSource(name string) (DataSource, error) {
...
 		s.l.Warn("service is disabled, skip Run")
 		return
 	}
 
+	p, err := version.Parse(s.pmmVersion)
+	// match only clean release versions like "3.7.1" and skip feature builds
+	if err != nil || p.Rest != "" {
+		return
+	}
+
```
maxkondr: with this change it will be impossible to check the gathered telemetry during verifying functionality in FB, there will be no metrics dumped into log stream at all.

managed/services/telemetry/telemetry_test.go
```diff
@@ -185,6 +185,46 @@ func TestRunTelemetryService(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.version, func(t *testing.T) {
+			t.Parallel()
+
+			var mockSender mockSender
+			mockSender.Test(t)
+			t.Cleanup(func() {
+				mockSender.AssertNotCalled(t, "SendTelemetry")
```
maxkondr, line 215: this test checks are limited:
1. it checks that `SendTelemetry` is not called that literally means "check there was no attempt to send telemetry to Percona"
2. but what is missed here "check that telemetry was gathered at all"

maxkondr (changes requested):

managed/services/telemetry/telemetry.go
ademidoff: Done.

managed/services/telemetry/telemetry_test.go
ademidoff: Too much async there, it's be difficult to implement it. Instead, we are covering with tests the scenario, where telemetry is not getting sent. 

I will manually test the FB to be sure everything works as expected.

maxkondr (approved):

JiriCtvrtka (approved):

PR #5236, opened 2026-04-10
PMM-14751: fix API interceptors

mattiasimonato:
PMM-14751

Adds a shared Axios response interceptor that shows error snackbars for failed API calls, and wires it in App so it is registered on mount and removed on unmount.


fabio-silva (approved):

ui/apps/pmm/src/App.tsx
```diff
@@ -19,33 +22,45 @@ const queryClient = new QueryClient({
...
-        </QueryClientProvider>
-      </SnackbarProvider>
-    </LocalizationProvider>
-  </ThemeContextProvider>
-);
+
+
+const App = () => {
+  
+  useEffect(() => {
```
matejkubinec, line 29: Just thinking, do we need to add the interceptor during react render? (if we apply it globally like with the `applyCaseMiddleware` it should be enough)

Because App should be mounted/unmounted only when the user opens/closes the page, so is the cleanup necessary?

mattiasimonato: Thanks for the question,  it made me think it through again.

This runs after the UI is in place (in an effect), not while React is still painting the screen.

In my understanding `applyCaseMiddleware` only adjusts how data flows through **axios**. Here we also surface errors to the user via the **snackbar**, so we want that wired up once the snackbar provider is actually mounted. Doing it from `App` keeps that order obvious.

On cleanup: you are totally right that for someone just closing the tab it is not doing much. I mostly kept the teardown for StrictMode in dev and for tests that mount the app more than once. If the team would rather keep this minimal, I am happy to drop the cleanup and lean on the guard in `api.ts`, whatever fits our style best.

matejkubinec: Yeah, that makes sense to me, let's keep it.

matejkubinec (approved):

PR #5250, opened 2026-04-14
PMM-14993 Add one-step PMM client installer script and UI integration

theTibi:
- Introduced a new script `install-pmm-client.sh` for streamlined installation of the PMM client.
- Updated Ansible tasks to copy the installer script to the appropriate directory.
- Enhanced documentation with a new section for one-step installation from the UI.
- Added a new Install Client page in the UI for generating installation commands.
- Integrated navigation for the new Install Client page within the application.

This update simplifies the PMM client installation process and improves user experience through the UI.

PMM-14993

SUBMODULES-4330

- https://github.com/percona/grafana/pull/894


mattiasimonato (approved):

mattiasimonato (approved):
After reviewing the frontend implementation, I have one small architectural note: the hook is colocated within the page/component, while the PMM codebase usually extracts hooks into separate modules. This is low priority and can be deferred, it does not need to block the **dev preview**

build/ansible/pmm-docker/post-build.yml
```diff
@@ -83,6 +83,28 @@
...
+    - name: Restore nginx /pmm-static files after pmm-server cleanup
+      copy:
+        src: "/opt/ansible/roles/nginx/files/{{ item.name }}"
+        dest: "/usr/share/pmm-server/static/{{ item.name }}"
+        owner: pmm
+        group: root
+        mode: "{{ item.mode }}"
+      loop:
+        - { name: local-rss.xml, mode: "0644" }
+        - { name: install-pmm-client.sh, mode: "0755" }
```
ademidoff: This should not be necessary, we do have `local-rss.xml` present in the Docker container.

ademidoff:
@theTibi This feature does not need to touch the backend at all, since PMM already exposes endpoints to generate the service account tokens. I suggest we refactor the code, it's going to be much less to review and maintain going forward.

build/ansible/pmm-docker/post-build.yml
theTibi: @ademidoff That’s true earlier in the build: roles/nginx copies local-rss.xml (and the installer script) into /usr/share/pmm-server/static/. Right after that, post-build runs Clean pmm-server dir, which removes all top-level entries under /usr/share/pmm-server (see PMM-14132 — reset that tree after the OpenShift/socket layout work). That wipes static/ too, so without the restore step those files would not be in the final image.

but correct me if I miss something.

theTibi:
@ademidoff I did some refactoring, we do not touch the backend now.

build/ansible/pmm-docker/post-build.yml
ademidoff: OK, that was a real bug, so I fixed it.

build/ansible/roles/nginx/files/install-pmm-client.sh
```diff
@@ -0,0 +1,680 @@
...
+
+  local config_cmd=(pmm-admin config "--server-url=${PMM_SERVER_URL}")
+  if [ "${PMM_SERVER_INSECURE_TLS}" = "1" ] || [ "${PMM_SERVER_INSECURE_TLS}" = "true" ]; then
+    config_cmd+=(--server-insecure-tls)
+  fi
+  if [ -n "${NODE_ADDRESS}" ]; then
+    config_cmd+=("${NODE_ADDRESS}")
+  fi
+  if [ -n "${NODE_NAME}" ]; then
+    config_cmd+=("generic" "${NODE_NAME}")
```
ademidoff: `pmm-admin config` positional args are `[<node-address>] [<node-type>] [<node-name>]`. If `NODE_NAME` is set without `NODE_ADDRESS`, this emits `pmm-admin config --server-url=... generic <name>` and the literal `generic` ends up in the node-address slot.

Two options:
- Require `NODE_ADDRESS` whenever `NODE_NAME` is set (validate early and error out), or
- Fall back to a sentinel address when only the name is provided (e.g. `$(detect_node_hostname)` or `127.0.0.1`).

The second option keeps the wizard UX simple — a user filling in only "Node name" still gets a working command.

build/ansible/roles/nginx/files/install-pmm-client.sh
```diff
@@ -0,0 +1,680 @@
...
+    . /etc/os-release
+    case "${ID:-}" in
+      debian|ubuntu)
+        echo "debian"
+        return
+        ;;
+      rhel|ol|amzn)
+        echo "el"
+        return
+        ;;
```
ademidoff, line 247: Common EL derivatives aren't listed here: Rocky (`rocky`), AlmaLinux (`almalinux`), CentOS Stream (`centos`), Fedora (`fedora`). Rocky/Alma/CentOS happen to ship `/etc/redhat-release` so the fallback below catches them, but Fedora minimal images ship `/etc/fedora-release` and slip through.

Adding `rocky|almalinux|centos|fedora` to the `el` arm of this `case` is cheap and avoids relying on the file-existence fallback for the common cases.

build/ansible/roles/nginx/files/install-pmm-client.sh
```diff
@@ -0,0 +1,680 @@
...
+
+apply_generic_inputs() {
+  DB_USER="${DB_USER:-}"
+  DB_PASSWORD="${DB_PASSWORD:-}"
+  DB_HOST="${DB_HOST:-}"
+  DB_PORT="${DB_PORT:-}"
+  DB_ADDRESS="${DB_ADDRESS:-}"
+  DB_SERVICE_NAME="${DB_SERVICE_NAME:-}"
+  DB_AUTH_DB="${DB_AUTH_DB:-}"
+  DB_SOCKET="${DB_SOCKET:-}"
```
ademidoff: These eight reassignments are no-ops — `DB_USER` through `DB_SOCKET` are already defaulted at script init (lines 60–68). Safe to delete the whole block; the per-tech derivations below them are the actual work this function does.

build/ansible/roles/nginx/files/install-pmm-client.sh
```diff
@@ -0,0 +1,680 @@
...
+DB_PASSWORD="${DB_PASSWORD:-}"
+DB_HOST="${DB_HOST:-}"
+DB_PORT="${DB_PORT:-}"
+DB_NAME="${DB_NAME:-}"
+DB_ADDRESS="${DB_ADDRESS:-}"
+DB_SERVICE_NAME="${DB_SERVICE_NAME:-}"
+DB_AUTH_DB="${DB_AUTH_DB:-}"
+DB_SOCKET="${DB_SOCKET:-}"
+
+MYSQL_USERNAME="${MYSQL_USERNAME:-}"
```
ademidoff, line 80: Small naming inconsistency that's likely to bite users: the generic env var is `DB_USER`, but the per-tech overrides use `_USERNAME` (`MYSQL_USERNAME`, `POSTGRESQL_USERNAME`, …). A user who reads `DB_USER` in the docs and tries the per-tech form will reasonably guess `MYSQL_USER` and find nothing.

Either alias `MYSQL_USER` → `MYSQL_USERNAME` (and friends) inside `apply_generic_inputs`, or rename to match. The flag side already uses `--db-user` consistently, so the inconsistency is env-only.

build/ansible/roles/nginx/files/install-pmm-client.sh
```diff
@@ -0,0 +1,680 @@
...
+    cmd+=("--authentication-database=${MONGODB_AUTH_DB}")
+  fi
+  if [ -n "${MONGODB_SOCKET}" ]; then
+    cmd+=("--socket=${MONGODB_SOCKET}")
+  fi
+  log "Running pmm-admin add mongodb..."
+  "${cmd[@]}"
+}
+
+add_valkey() {
```
ademidoff, line 799: Worth confirming `pmm-admin add valkey` is a real subcommand in the `pmm-client` version that the `pmm3-client release` repo serves at the time this script runs. If Valkey support was added recently to pmm-admin but the released client is older, the script installs fine, registers the node, and then fails at the last step with a confusing "unknown command" error — exactly the failure mode that `report_add_service_failure` doesn't help with (it assumes auth/conflict).

ui/apps/pmm/src/api/installToken.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  let saId = await findServiceAccountIdByName(saName);
+  if (saId === null) {
+    saId = await createServiceAccount(saName);
+  }
+
+  // UUID-suffixed token name keeps concurrent calls from colliding on Grafana's
+  // per-SA unique-name constraint (Grafana returns 409 otherwise).
+  const tokenName = `${TOKEN_NAME_PREFIX}-${technology}-${crypto.randomUUID()}`;
+  const key = await mintToken(saId, tokenName, ttl);
```
ademidoff: Each `Generate token` click creates a new token on the shared per-tech SA (`pmm-install-sa-mysql` etc.). Tokens auto-expire in Grafana, but they stay in the SA's token list until manually deleted. After heavy use the SA's tokens table grows unbounded.

Options, in increasing order of effort:
- Delete expired tokens for this SA right before minting the new one (`GET /api/serviceaccounts/{id}/tokens` → filter → `DELETE …/tokens/{id}`).
- A scheduled cleanup elsewhere.
- At minimum, a comment here pointing operators at where to clean up if it ever becomes a problem.

None of these are blockers, but the unbounded growth will surface eventually.

ui/apps/pmm/src/api/installToken.ts
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Mints a short-lived Grafana service-account token for a PMM Client install command.
+ *
+ * Implementation note: this calls Grafana's serviceaccounts API directly through the
+ * `/graph/api/` reverse proxy. The user must already be authenticated as a Grafana
+ * Admin (Grafana rejects the create/mint requests with 403 otherwise) — that's the
+ * same trust boundary the old backend endpoint had, just one hop shorter.
+ */
+export async function createNodeInstallToken(
```
ademidoff, line 80: The doc comment says Grafana rejects with 403 for non-Admin callers — worth a quick manual confirmation that hitting `/api/serviceaccounts` as a Viewer or Editor really returns 403. The entire security model of this page rests on that check, so it's the one thing worth verifying explicitly rather than trusting by inheritance from the old backend endpoint.

ui/apps/pmm/src/pages/install-client/InstallClientPage.utils.ts
```diff
@@ -0,0 +1,234 @@
...
+
+/**
+ * Builds PMM_SERVER_URL for install scripts. Token is percent-encoded in the userinfo.
+ * `pmmHost` is hostname or hostname:port (defaults to current page host when empty).
+ */
+export function buildPmmServerURL(pmmHost: string, token: string): string {
+  const authority =
+    pmmHost.trim() ||
+    (typeof window !== 'undefined' ? window.location.host : 'localhost');
+  const t = token.trim();
```
ademidoff, line 93: If a user types a full URL into the wizard's "PMM host" field (e.g. `https://pmm.example.com`), this produces `https://service_token:<TOKEN>@https://pmm.example.com`, which curl will reject. A light input sanitization (strip leading `http(s)://`, or validate with a regex in the form and show a helper) would prevent the foot-gun.

ademidoff (commented):
A few comments from me, not blockers — feel free to take or leave them per your judgment.

build/ansible/roles/nginx/files/install-pmm-client.sh
ondrejpatocka: @mattiasimonato could you please address this one as the most critical issue preventing us to ship this?

mattiasimonato: Solved [HERE](https://github.com/percona/pmm/pull/5250/changes/3a05235caf1f1e1e7a6881103adcee6b3531aaab)

build/ansible/roles/nginx/files/install-pmm-client.sh
ademidoff: Done.

build/ansible/roles/nginx/files/install-pmm-client.sh
ademidoff: Done.

ademidoff (approved):

ui/apps/pmm/src/pages/install-client/InstallClientPage.tsx
```diff
@@ -0,0 +1,403 @@
...
+    nodeAddress,
+    nodeName,
+    registerForce,
+    serverURL,
+    technology,
+    ]
+  );
+
+  const handleCopy = async () => {
+    await navigator.clipboard.writeText(command);
```
fabio-silva: Clipboard might not be available if PMM isn't running in a secure context (aka https, localhost, ...).
We should disable the copy button and add a tooltip stating that when it's unavailable

ui/apps/pmm/src/pages/install-client/InstallClientPage.tsx
```diff
@@ -0,0 +1,403 @@
...
+      // installToken.ts always returns expiresAt; the fallback is just defensive
+      // belt-and-braces in case of a future refactor.
+      const expires = res.expiresAt
+        ? new Date(res.expiresAt)
+        : new Date(Date.now() + 15 * 60 * 1000);
+      setTokenExpiresAt(expires);
+      setNow(Date.now());
+    } catch (e: unknown) {
+      let msg = 'Failed to create token';
+      if (axios.isAxiosError(e)) {
```
fabio-silva: we have a global interceptor running that notifices users via toast. This can be safely removed, along with the associated state var

fabio-silva (changes requested):

fabio-silva:
@copilot adress my comments and fix them.
Also, we have frequent re-renders (due to setInterval) and lots of inline callbacks. Please refactor those into `useCallback`.
Fix each one in different commits.

fabio-silva:
@copilot just realized something:
- `addApiErrorInterceptor` -> only applies to our own API. We need to apply it to `grafanaApi` as well, please refactor accordingly;
- actually we need to re-add the genError from the catch block (no need to check for Axios errors as those are caught already by the interceptor. Just to do it generically).

fabio-silva:
@copilot `isExpired ` used before assignment, move it to the top.

fabio-silva (approved):

ui/apps/pmm/src/api/api.ts
```diff
@@ -15,47 +15,62 @@ export const grafanaApi = axios.create({
...
-        }
-
-        return Promise.reject(error);
-      }
+      onApiError
+    );
+  }
+
+  if (grafanaApiErrorInterceptor === null) {
+    grafanaApiErrorInterceptor = grafanaApi.interceptors.response.use(
```
ademidoff: This adds a global response interceptor to `grafanaApi`, which previously had none (only `api` did). `grafanaApi` is used by `settings.ts`, `folders.ts`, `short-urls.ts`, `user.ts`, and `auth.ts` (`/user/auth-tokens/rotate`, likely a background call). Every 4xx/5xx from those will now pop an error snackbar where it was previously silent — a regression risk outside this feature's scope.

It also reports token-generation errors twice: this interceptor shows a snackbar, and `handleGenerateToken` in `InstallClientPage.tsx` already sets `genError`, rendered as an inline Alert.

Consider scoping notification suppression via the existing `disableNotifications` config on the install-token requests, or confirm the broader toast behavior is intended for all grafanaApi callers.

ui/apps/pmm/src/api/installToken.ts
```diff
@@ -0,0 +1,111 @@
...
+  const saName = `${SA_NAME_PREFIX}-${technology}`;
+
+  let saId = await findServiceAccountIdByName(saName);
+  if (saId === null) {
+    saId = await createServiceAccount(saName);
+  }
+
+  // UUID-suffixed token name keeps concurrent calls from colliding on Grafana's
+  // per-SA unique-name constraint (Grafana returns 409 otherwise).
+  const tokenName = `${TOKEN_NAME_PREFIX}-${technology}-${crypto.randomUUID()}`;
```
ademidoff: `crypto.randomUUID()` is `undefined` outside a secure context, so token generation throws over plain HTTP. It's caught and surfaced as `genError` (no crash), but the feature is then silently unusable on HTTP while the rest of the page works. The clipboard path already gates on `window.isSecureContext` in `InstallClientPage.tsx`; token generation has no comparable guard or "requires HTTPS" hint.

ui/apps/pmm/src/api/installToken.ts
```diff
@@ -0,0 +1,111 @@
...
+ * Mints a short-lived Grafana service-account token for a PMM Client install command.
+ *
+ * Implementation note: this calls Grafana's serviceaccounts API directly through the
+ * `/graph/api/` reverse proxy. The user must already be authenticated as a Grafana
+ * Admin (Grafana rejects the create/mint requests with 403 otherwise) — that's the
+ * same trust boundary the old backend endpoint had, just one hop shorter.
+ */
+export async function createNodeInstallToken(
+  technology: string,
+  ttlSeconds = 0
```
ademidoff: `ttlSeconds` defaults to `0` and the only caller passes `0`, so the `ttlSeconds > 0 ? ttlSeconds : DEFAULT_TTL_SECONDS` branch never triggers. Per the project's simplicity bias, consider dropping the parameter until something actually varies it.

build/ansible/roles/nginx/files/install-pmm-client.sh
```diff
@@ -0,0 +1,814 @@
...
+  fi
+
+  log "pmm-agent is up on ${PMM_AGENT_LISTEN_HOST}:${PMM_AGENT_LISTEN_PORT}."
+}
+
+# When stdin is not a terminal (e.g. curl ... | bash), prompts cannot be used for DB
+# credentials. Fail before pmm-admin config so we do not register the node and then fail on add.
+# Caller must have already invoked apply_generic_inputs.
+#
+# Contract with the PMM UI's "Prompt on node" credentials mode:
```
ademidoff: Nit (also applies to `ui/apps/pmm/src/pages/install-client/InstallClientPage.utils.ts`): the prose comments here are unusually long relative to the rest of the codebase. Worth trimming to the non-obvious rationale.

ademidoff (commented):
A few review comments on the one-step installer feature.

ademidoff (approved):

PR #5269, opened 2026-04-19
PMM-15000 Migrate AWS agent configuration to AlmaLinux

ademidoff:
PMM-15000


build/packer/aws.pkr.hcl
```diff
@@ -14,18 +14,18 @@ packer {
 source "amazon-ebs" "amd-agent" {
   ami_name              = "Jenkins Agent v3"
   instance_type         = "t3.xlarge"
-  force_deregister      = true
+  # force_deregister      = true
```
ademidoff: This is intentional so that we can continue using the old image should we face an issue with the new one.

nogueiraanderson (approved):

build/ansible/roles/initialization/tasks/main.yml
```diff
@@ -97,10 +97,6 @@
       mode: 0775
     when: need_upgrade
 
-  # Note: we want to leave this for some time until we achieve stable builds
-  - name: Output pmm-managed logs
-    shell: sleep 10 && tail -n 300 /srv/logs/pmm-managed.log
-
```
ademidoff, line 103: This will speed up upgrades by 10 seconds 🎉

PR #5283, opened 2026-04-22
PMM-14441 Update pmm-dump API to support encryption

maxkondr:
PMM-14441

Link to the Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/4309


build/packages/rpm/server/SPECS/pmm-dump.spec
```diff
@@ -37,6 +37,9 @@ install -p -m 0755 pmm-dump %{buildroot}%{_sbindir}/pmm-dump
 
 
 %changelog
+* Wed Apr 22 2026 Maxim Kondratenko <maxim.kondratenko@percona.com> - 3.8.0
```
ademidoff:
```suggestion
* Wed Apr 22 2026 Maxim Kondratenko <maxim.kondratenko@percona.com> - 3.7.1
```

managed/services/dump/dump.go
```diff
@@ -319,8 +329,12 @@ func (s *Service) StopDump() {
 	s.cancel()
 }
 
-func getDumpFilePath(id string) string {
-	return fmt.Sprintf("%s/%s.tar.gz", dumpsDir, id)
+func getDumpFilePath(id string, encrypted bool) string {
```
ademidoff, line 332: @maxkondr Can we add a few unit tests, for example for non-receiver methods like this one and the one below?

ademidoff: The rest LGTM.

JiriCtvrtka (approved):

managed/models/database.go
```diff
@@ -1394,31 +1399,31 @@ func initWithRoot(params SetupDBParams) error {
 	defer db.Close() //nolint:errcheck
 
 	var countDatabases int
-	err = db.QueryRow(`SELECT COUNT(*) FROM pg_database WHERE datname = $1`, params.Name).Scan(&countDatabases)
+	err = db.QueryRowContext(ctx, `SELECT COUNT(*) FROM pg_database WHERE datname = $1`, params.Name).Scan(&countDatabases)
```
maxkondr: Linter started to complain on this - shall be context aware

ademidoff: While in general it is definitely part of best practices to use the context in such case, I think it was intentional, but unfortunately, undocumented.

The problem is that the container, and consequently, `pmm-managed` may get restarted quite often. If the context cancels one these db queries in the middle, it may leave PMM in an inconsistent state without giving `pmm-managed` a chance to recover.

Therefore, the absence of context is actually working in favor of a more resilient app, not against.

One possible solution I can think of is to wrap the whole set of queries into a transaction to make it more reliable, but that is probably outside of the scope of this ticket as it needs QA.

maxkondr: reverted this change

ademidoff (approved):

PR #5295, opened 2026-04-27
PMM 15014 Upgrade internal PostgreSQL version from 14 to 18

talhabinrizwan:
PMM-15014

Link to the Feature Build: SUBMODULES-4324

---

> [!NOTE]
> **`minPGVersion` stays at `14`.** The embedded cluster moves to PostgreSQL 18, so a standalone PMM Server is on 18 either way. The floor only applies to externally hosted databases (HA and BYO-database setups), where the client remains compatible with older servers. Raising it would make PMM refuse to start against an external PostgreSQL 14 — so anyone who has not been able to migrate their database would be locked out of PMM the moment they upgrade to 3.10.0. That is not an acceptable upgrade experience, and there is no technical need for it.
>
> This was previously listed here as a temporary value to bump to `15` before merging. It is now the intended one, and the `TODO` in `managed/models/database.go` has been replaced with the rationale. See the [comment below](https://github.com/percona/pmm/pull/5295#issuecomment-5133414831) for the measured compatibility data behind the original proposal.

> [!IMPORTANT]
> **`Managed / Unit tests` stays red until `3-dev-container` is rebuilt from this branch.**
>
> That job runs the managed unit tests inside `ghcr.io/percona/pmm:3-dev-container`, which is built from `main` and still ships a supervisord config that writes `/srv/logs/postgresql14.log`. `TestFiles` in `managed/services/server/logs_test.go` compares against `filepath.Glob("/srv/logs/*.log")`, so it now expects `postgresql.log` — the name this branch's supervisord writes.
>
> The branch no longer carries temporary values to paper over this; the expectation is the final one. To make the job pass, dispatch the `Devcontainer` workflow with `branch: PMM-15014-upgrade-to-pg18`. Note that this republishes the shared `3-dev-container` tag used by every PR, so it is a merge-time action rather than something to do early.


JiriCtvrtka:
@talhabinrizwan It still needs some changes in managed to pass tests there properly.

build/ansible/roles/postgres/files/postgres-migration
```diff
@@ -20,8 +116,8 @@ ensure_postgres_password() {
     echo "Generating postgres superuser password..."
     POSTGRES_PASSWORD=$(openssl rand -hex 16)
 
-    if ! /usr/pgsql-14/bin/pg_ctl status -D "$POSTGRES_DATA_DIR" > /dev/null 2>&1; then
-        /usr/pgsql-14/bin/pg_ctl start -D "$POSTGRES_DATA_DIR" -o "-c logging_collector=off"
+    if ! /usr/pgsql-18/bin/pg_ctl status -D "$POSTGRES_DATA_DIR" > /dev/null 2>&1; then
```
maxkondr: it is better to extract `pg_ctl` into global var -> avoid massive path changes in future

build/ansible/roles/postgres/files/postgres-migration
```diff
@@ -53,10 +149,11 @@ update_pg_hba_auth() {
...
     mv "$tmp" "$hba"
 
-    if /usr/pgsql-14/bin/pg_ctl status -D "$POSTGRES_DATA_DIR" > /dev/null 2>&1; then
-        /usr/pgsql-14/bin/pg_ctl reload -D "$POSTGRES_DATA_DIR"
+    if /usr/pgsql-18/bin/pg_ctl status -D "$POSTGRES_DATA_DIR" > /dev/null 2>&1; then
+        /usr/pgsql-18/bin/pg_ctl reload -D "$POSTGRES_DATA_DIR"
     fi
 }
 
+upgrade_pg14_to_pg18
```
maxkondr, line 207: this func uses `/srv/.pgpassword` file but is called before `ensure_postgres_password` that creates this file in case it is absent

managed/models/database.go
```diff
@@ -1282,13 +1282,53 @@ func SetupDB(ctx context.Context, sqlDB *sql.DB, params SetupDBParams) (*reform.
...
 	}
 
 	return db, nil
 }
 
+// ensureSchemaGrant grants CREATE on the public schema to the database user.
+// PostgreSQL 15+ revoked the default CREATE privilege on the public schema from PUBLIC,
+// so it must be granted explicitly before migrations can create tables.
+// Skipped for HA mode and when the embedded postgres password file is absent (external PostgreSQL).
+func ensureSchemaGrant(params SetupDBParams) error {
```
maxkondr: why is this logic located here but not in docker entrypoint?

managed/models/database.go
```diff
@@ -1418,7 +1458,7 @@ func initWithRoot(params SetupDBParams) error {
 			return fmt.Errorf("failed to create user %s: %w", params.Username, err)
 		}
 
-		_, err = db.Exec(`GRANT ALL PRIVILEGES ON DATABASE $1 TO $2`, params.Name, params.Username)
+		_, err = db.Exec(fmt.Sprintf(`GRANT ALL PRIVILEGES ON DATABASE "%s" TO "%s"`, params.Name, params.Username))
```
maxkondr: No, this is exactly the way to SQL-injections

managed/models/database.go
```diff
@@ -1282,13 +1282,53 @@ func SetupDB(ctx context.Context, sqlDB *sql.DB, params SetupDBParams) (*reform.
...
+		return fmt.Errorf("failed to read postgres password from %s: %w", passwordFile, err)
+	}
+
+	db, err := OpenDB(SetupDBParams{Address: params.Address, Name: params.Name, Username: "postgres", Password: string(passwordBytes)})
+	if err != nil {
+		return fmt.Errorf("failed to open the database: %w", err)
+	}
+	defer db.Close() //nolint:errcheck
+
+	_, err = db.Exec(fmt.Sprintf(`GRANT CREATE ON SCHEMA public TO "%s"`, params.Username))
```
maxkondr:
```suggestion
	_, err = db.Exec(`GRANT CREATE ON SCHEMA public TO $1`, params.Username)
```

managed/models/database.go
ademidoff: This is gone now.

ademidoff:
### Note on `minPGVersion`

**Superseding my earlier version of this comment:** the pre-merge action is the one in the PR description — restore the constant to **18**. It is temporarily `14` so the `Unit tests` job can run against the prebuilt dev container, which is built from `main` and still embeds PostgreSQL 14.23. With the floor at 18 that job fails on the first database-backed test and the remaining ~147 failures are all cascade from the leftover connection (`pq: database "pmm-managed-dev" is being accessed by other users`).

The test image only gets PostgreSQL 18 once this branch's server image is published as `3-dev-latest`, which is circular until merge — hence the temporary floor rather than a workflow change.

### Worth deciding separately: 18 or 15?

Leaving this here as data, not as an instruction. `minPGVersion` conflates the version PMM *bundles* (18 either way) with the minimum *external* server PMM accepts. At 18 we reject externally hosted PostgreSQL 15, 16 and 17, which affects HA and BYO-database deployments.

Measured by running the full migration set (118 migrations) as the `pmm-managed` role against real servers:

| `minPGVersion` | PG 14.22 | PG 16.14 | PG 18.3 |
| --- | --- | --- | --- |
| 18 | rejected | rejected | 118/118 |
| 15 | rejected | 118/118 | 118/118 |
| 14 | 118/118 | 118/118 | 118/118 |

Nothing in the schema requires 15+, so the floor is purely a support-policy choice. 15 is a defensible line because it is where `public` stopped being writable by `PUBLIC` and became owned by `pg_database_owner` — the ownership model this PR adopts for provisioning. PostgreSQL 14 reaches end of life this autumn, so 14 is not a viable final value either way.

build/docker/server/entrypoint.sh
```diff
@@ -6,8 +6,7 @@ declare CURRENT_GID CURRENT_UID CURRENT_USER
 
 # Returns 0 (true) if the given variable is set to "1" or "true".
 is_enabled() { [ "$1" = "1" ] || [ "$1" = "true" ]; }
-declare POSTGRES_DATA_DIR="/srv/postgres14"
-declare POSTGRES_PASSWORD_FILE="/srv/.postgres_password"
+declare POSTGRES_DATA_DIR="/srv/postgres18"
```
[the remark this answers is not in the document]
ademidoff: Declined - `POSTGRES_DATA_DIR` is not supposed to be passed from the user to the entrypoint, it's a hard-coded path.

managed/services/supervisord/pmm_config.go
```diff
@@ -132,7 +132,7 @@ stopsignal = INT  ; Fast Shutdown mode
 stopwaitsecs = 300
 ; postgresql.conf contains settings to log to stdout,
 ; so we delegate logfile management to supervisord
-stdout_logfile = /srv/logs/postgresql14.log
+stdout_logfile = /srv/logs/postgresql.log
```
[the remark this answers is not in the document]
ademidoff: Dude, that was intentional to make the tests pass, I left a TODO item to action on before merging.

ademidoff (approved):

ademidoff:
@coderabbitai configuration

managed/models/database.go
```diff
@@ -46,6 +46,8 @@ const (
 	// PMMServerPostgreSQLServiceName is a special Service Name representing PMM Server's PostgreSQL Service.
 	PMMServerPostgreSQLServiceName = "pmm-server-postgresql"
 	// - minPGVersion stands for minimal required PostgreSQL server version for PMM Server.
+	// TODO temporarily lowered to 14 so that the unit tests pass against the prebuilt dev
+	// container, which still embeds PostgreSQL 14. Bump to 15 before merging.
```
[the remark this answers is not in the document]
ademidoff: We decided to continue using the PG 14 client so we don't break the external databases who's owners didn't have the time to upgrade.

build/docs/MIGRATION.md
```diff
@@ -85,7 +85,7 @@ Remember to pass the data volume to the instance so it can bootstrap the databas
 ```
   # Read the postgres password from the secure file
   PGPASSWORD=$(cat /srv/.postgres_password)
-  PGPASSWORD="$PGPASSWORD" /usr/pgsql-14/bin/pg_restore --host=/run/postgresql --username=postgres --file=/srv/backup/postgres.sql -S postgres
+  PGPASSWORD="$PGPASSWORD" /usr/pgsql-14/bin/pg_restore --host=/run/postgresql --username=postgres --file=/srv/backup/pmm-managed.sql -S postgres
```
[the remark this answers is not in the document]
ademidoff: Fixed in 1bb05db75, applied to both lines.

Confirmed the reasoning: the dumps at `build/docs/MIGRATION.md:60-61` are written by plain `pg_dump --file=… --dbname …` with no `--format`, so they are plain-text SQL that `pg_restore` cannot read at all. On top of that `--file` is an output path for `pg_restore`, so these commands had no input archive and pointed their output at the dump files themselves. Dropped the inert `-S postgres` along with them.

For context, this is a pre-existing problem in the v11→v14 section rather than something this PR introduced — the branch only renamed `postgres.sql` to `pmm-managed.sql` on the first line. The v14→v18 section added by this PR already restores with `psql -f`.

PR #5296, opened 2026-04-27
PMM-15015 Remove conflicting package declaration in pmm-client

ademidoff:
PMM-15015

Ref: reported on the forum - https://forums.percona.com/t/pmm-client-v3-7-0-conflicting-with-itself/40715.

Link to the Feature Build: SUBMODULES-4326

This pull request makes a small change to the `pmm-client.spec` file, removing the `Conflicts: pmm-client` line. This simplifies the package specification and may help avoid unnecessary package conflict errors during installation.


build/packages/rpm/client/pmm-client.spec
```diff
@@ -21,7 +21,6 @@ Requires(preun):  systemd
 Requires(postun): systemd
 
 AutoReq:        no
-Conflicts:      pmm-client
```
EvgeniyPatlan, line 24: It should contain version match <3.0.0

EvgeniyPatlan (changes requested):

build/packages/rpm/client/pmm-client.spec
ademidoff: This line was removed. What are you referring to?

EvgeniyPatlan: I suggest not removing it but just add `Conflicts:      pmm-client < 3.0.0`

ademidoff: That's exactly what we want to avoid. We actually want any newer version  of the package called `pmm-client`, which is being installed, to replace any previous version, and not to conflict with it.

Unless I'm unaware of other benefits this may bring.

EvgeniyPatlan: This is exactly RPM pattern that is used for the behavior you want. Yes it will work with the only `Obsoletes` but it will fail in case enduser will execute `rpm -i pmm-client....` So that is why Obsoletes and Conflicts are used together.

ademidoff: Hm, I don't really understand why the user can't install PMM Client by running `rpm -i pmm-client`.

OK, thank you, we will QA this solution anyway.

EvgeniyPatlan: user can download package from repo or can copy it to a server from let's say USB drive as there are a lot of users that doesn't use repo to install packages.

talhabinrizwan (approved):

PR #5299, opened 2026-04-28
PMM-7 API tests rework

### LINKED ISSUE #2342: Enable the `paralleltest` linter rule
### Description

**Why**
We consider parallel test a great speed booster when our tests run in CI environment. However, this is not always possible due to a number of constraint we have.

**What**
- evaluate if we can afford to enable the rule
- if yes, fix all linter error to make tests pass

### Suggested solution

**What**
- evaluate if we can afford to enable the rule
- if yes, fix all linter error to make tests pass

### Additional context

The rule can be enabled by simply removing [this line](https://github.com/percona/pmm/blob/26895fa270fb28db2dd985228ebfd714eca2c70f/.golangci.yml#L115).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

maxkondr:
PMM-7

Link to the Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/4329

The tests flakiness was caused by several reasons:
- tests didn't wait while Grafana inside PMM Server is fully up and running after all initializations and migrations
- not all tests were idempotent
- api-tests/server/settings tests update PMM Server configuration that triggers Grafana restart -> other tests fail while Grafata is unavailable.
- user-related tests used default admin user, but their logic depend on user first signing action that on dev container results to failures. Reworked by creating a separate user for this test

Improvements:
- Added test for RTA Mongodb Agent.
- api-tests/server/settings has been separated from the rest of tests and is running at the end to avoid impact on the rest of tests
- User-related tests now create a dedicated user.
- Make almost all tests running in parallel -> speedup test execution
- Reworked Makefile - now there are targets to prepare and run tests.
- All `pmmapitests.Add*` and `pmmapitests.Register*` automatically cleanup the created/registered resources during test cleanup. No need to call `Remove*/Unregister*` functions directly in test body.

Closes #2342.


api-tests/management/rds_test.go
```diff
@@ -120,18 +128,20 @@ func TestAddRds(t *testing.T) {
...
+
 		params := &mservice.AddServiceParams{
 			Body: mservice.AddServiceBody{
 				RDS: &mservice.AddServiceParamsBodyRDS{
 					Region:                    "region",
 					Az:                        "az",
 					InstanceID:                "d752f1a9-31c9-4b8c-bb2d-d26bc000009",
 					NodeModel:                 "some-model",
-					Address:                   "some.example.rds",
+					Address:                   "some.example-1.rds",
```
ademidoff: How come this field now takes a different value? Was it a buggy test?

api-tests/Dockerfile
```diff
@@ -1,11 +1,25 @@
...
+ARG GIT_COMMIT=unknown
+ARG BUILD_TIME=unknown
+ARG GO_VERSION="1.25"
+ARG GO_BASE_IMAGE="golang:${GO_VERSION}"
 
-FROM golang:1.25
+FROM $GO_BASE_IMAGE AS builder
+ENV GONOPROXY='github.com/percona,github.com/Percona-Lab'
+ENV GONOSUMDB='github.com/percona,github.com/Percona-Lab'
+ENV GOPRIVATE='github.com/percona,github.com/Percona-Lab'
```
ademidoff, line 11: Do we need all of these? It worked without it before, looks like bloat.

api-tests/Makefile
```diff
@@ -1,28 +1,93 @@
...
+PMM_SERVER_INSECURE_TLS ?= 1
+PMM_RUN_UPDATE_TEST ?= 0
+PMM_RUN_ADVISOR_TESTS ?= 0
+
+.PHONY: help guard-% init $(TEST_BINARY_TARGETS) $(RUN_TEST_TARGETS)
+.PHONY: test-server-settings test test-report clean docker-build-image docker-run-tests
+
+help: ## Show available targets
+	@echo "Available targets: $(TARGETS)"
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "  \033[36m%-20s\033[0m %s\n", $$1, $$2}'
```
ademidoff: Let's leave out colors please, don't work well in all environments, especially in the logs (not human-readable).

api-tests/Makefile
```diff
@@ -1,28 +1,93 @@
...
+	echo "🚀 Run $$binary tests"; \
+	$(BIN_DIR)/$$binary $(TEST_FLAGS) 2>&1 | tee $(TEST_OUTPUT_PREFIX)-$@.txt
+
+test: guard-PMM_SERVER_URL $(RUN_TEST_TARGETS) ## Run tests
+	$(MAKE) test-server-settings
+	@echo "✅ All tests completed."
+
+test-report: test ## Run tests and generate JUnit report
+	cat $(CURDIR)/$(TEST_OUTPUT_PREFIX)-*.txt | $(BIN_DIR)/go-junit-report > $(REPORT_OUT)
+	@echo "✅ Report is generated: $(REPORT_OUT)."
```
ademidoff:
```suggestion
	@echo "✅ Report generated: $(REPORT_OUT)."
```
Just to make it consistent with the rest.

api-tests/README.md
```diff
@@ -8,28 +8,33 @@ Make sure you have the latest Go version installed on your systems, execute the
 to set up API-tests in your local systems.
 
 1. Run PMM Server. This can be done by running `make env-up` in the root (`pmm`) directory.
-2. Replace `$PMM_SERVER_URL` with a URL in format `http://USERNAME:PASSWORD@HOST`. For local development it's usually `http://admin:admin@127.0.0.1`.
+2. Replace `$PMM_SERVER_URL` with a URL in format `https://USERNAME:PASSWORD@HOST`. For local development it's usually `http://admin:admin@127.0.0.1`.
```
ademidoff:
```suggestion
2. Replace `$PMM_SERVER_URL` with a URL in format `https://USERNAME:PASSWORD@HOST`. For local development it's usually `https://admin:admin@127.0.0.1`.
```

I think we don't really want to test http.

api-tests/README.md
```diff
@@ -8,28 +8,33 @@ Make sure you have the latest Go version installed on your systems, execute the
...
 
 # Usage
 
+Precompile tests using the following command:
+```
+make -j4 init
```
ademidoff: `-j4` should probably be part of the target so the user doesn't have to remember the syntax.

ademidoff:
@maxkondr The trouble is that the snooze tests are not passing. PTAL

maxkondr:
> @maxkondr The trouble is that the snooze tests are not passing. PTAL

Now passing

api-tests/management/rds_test.go
maxkondr: there shall be unique value, fixed

api-tests/Dockerfile
maxkondr: yes

api-tests/Makefile
maxkondr: done

api-tests/inventory/agents_postgres_exporter_test.go
```diff
@@ -227,12 +209,7 @@ func TestPostgresExporter(t *testing.T) {
 		t.Parallel()
 
 		genericNodeID := pmmapitests.AddGenericNode(t, pmmapitests.TestString(t, "")).NodeID
-		require.NotEmpty(t, genericNodeID)
-		defer pmmapitests.RemoveNodes(t, genericNodeID)
```
JiriCtvrtka, line 231: Are we sure we don’t need this check anymore? Here and other places.

JiriCtvrtka (approved):

api-tests/inventory/agents_postgres_exporter_test.go
maxkondr: Yes, everything that is created via `pmmapitests.AddGenericNode` is scheduled for removal during test cleanup inside `pmmapitests.AddGenericNode` func

ademidoff (approved):
I've triggered the feature build one last time, let's see if the tests pass, then we can merge it.

PR #5352, opened 2026-05-12
PMM-15058 Seamless PMM UI redirect

matejkubinec:
PMM-15058

FB: SUBMODULES-4354


JiriCtvrtka (approved):

ademidoff:
### Code review

Found 1 issue:

1. The new nginx exemptions for Grafana's auth/account routes get undone client-side, so those paths may still end up under `/pmm-ui`.

`ui/apps/pmm-compat/src/compat.ts` redirects whenever the page is not in an iframe and the path does not start with `GRAFANA_LOGIN_PATH`, and that constant is `/graph/login` and nothing else. This PR doesn't touch either file, so nginx correctly lets `/graph/signup`, `/graph/invite/<code>`, `/graph/verify` and `/graph/user/password/reset` through to Grafana, and then the preloaded compat plugin `window.location.replace()`s them into `/pmm-ui/graph/...` anyway — exactly the paths PMM-15058 says should keep the `/graph` prefix.

The fact that `/graph/login` needs an exclusion at all suggests the plugin does execute on unauthenticated auth pages. `/graph/logout` is likely fine, since Grafana handles it server-side and 302s to login before the SPA boots. Worth checking on the feature build; if it reproduces, the two exclusion lists should be kept in sync.

https://github.com/percona/pmm/blob/10790049619c91b48ad9673f612b99098587d3a3/ui/apps/pmm-compat/src/compat.ts#L41-L44

https://github.com/percona/pmm/blob/10790049619c91b48ad9673f612b99098587d3a3/ui/apps/pmm-compat/src/lib/constants.ts#L18-L20

https://github.com/percona/pmm/blob/10790049619c91b48ad9673f612b99098587d3a3/build/ansible/roles/nginx/files/conf.d/pmm.conf#L168-L171

4nte:
There seems to be a small regression with unauthenticated deep links silently lost across login.

I had Grafana generate a real alert email (an always-firing rule annotated with `__dashboardUid__`/`__panelId__`), then clicked the link it produced:

```
https://<pmm>/graph/d/node-cpu?from=1787049360000&orgId=1&to=1787052965025&viewPanel=22
```

| | logged-out click lands on | return path | after login the user sees |
|---|---|---|---|
| `main` | `302 /graph/login?redirectTo=%2Fd%2Fnode-cpu%3F...` | preserved | the requested dashboard |
| **this PR** | `302 /pmm-ui/graph/d/node-cpu?...` | lost | `/pmm-ui/graph/` (home) |

The dashboard, the `viewPanel=22` and the `from`/`to` range are all discarded, with no error and nothing but a `302` and a `200` in the access log.

mattiasimonato (approved):

matejkubinec:
> There seems to be a small regression with unauthenticated deep links silently lost across login.
> 
> I had Grafana generate a real alert email (an always-firing rule annotated with `__dashboardUid__`/`__panelId__`), then clicked the link it produced:
> 
> ```
> https://<pmm>/graph/d/node-cpu?from=1787049360000&orgId=1&to=1787052965025&viewPanel=22
> ```
> 
> logged-out click lands on	return path	after login the user sees
> `main`	`302 /graph/login?redirectTo=%2Fd%2Fnode-cpu%3F...`	preserved	the requested dashboard
> **this PR**	`302 /pmm-ui/graph/d/node-cpu?...`	lost	`/pmm-ui/graph/` (home)
> The dashboard, the `viewPanel=22` and the `from`/`to` range are all discarded, with no error and nothing but a `302` and a `200` in the access log.

Should be fixed now, deep links now should link correctly into the app even when user needs to login beforehand.

ui/apps/pmm/src/contexts/auth/auth.returnTo.ts
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Remember where the user was before we bounce them to Grafana's login page. sessionStorage is
+ * per-tab and survives the cross-document hop to /graph/login and back, which is exactly the
+ * lifetime we need. It also stays clear of the localStorage keys that
+ * ensureClientSessionListener() watches, so writing here cannot kick the client-session store.
+ */
+export const saveReturnTo = (
+  location: Pick<Location, 'pathname' | 'search' | 'hash'> = window.location
+) => {
+  dropGrafanaReturnTo();
```
ademidoff, line 79: **[1/14 - highest severity] Unguarded `sessionStorage` on the login-bounce path can replace the login redirect with a blank page.**

`saveReturnTo()` touches `sessionStorage` on its very first statement, and it is reached from `redirectToLogin()`, which `AuthProvider` calls **during render** (`ui/apps/pmm/src/contexts/auth/auth.provider.tsx:113`).

In Safari private browsing, or any browser configured to block all cookies/site data, `sessionStorage` access throws (SecurityError on the accessor, QuotaExceededError on `setItem` in older Safari private mode). The order is then:

1. `auth.provider.tsx:113` -> `redirectToLogin()`
2. `auth.utils.ts:43` -> `saveReturnTo()`
3. `auth.returnTo.ts:54` `dropGrafanaReturnTo()` **throws**
4. `window.location.replace(PMM_LOGIN_URL)` at `auth.utils.ts:44` never runs

React unwinds the render, there is no error boundary above `Providers`, and the user gets a white screen with no way to reach the login form.

`consumeReturnTo()` has the mirror-image exposure: if it throws inside the effect at `auth.provider.tsx:63`, `returnToChecked` stays `false` and the gate at `auth.provider.tsx:123` renders `null` forever.

Suggest wrapping every `sessionStorage` read/write here in `try/catch` and treating a failure as "nothing stored". `auth.clientSession.ts` gets away with unguarded `localStorage` only because it never runs ahead of a navigation.

ui/apps/pmm/src/contexts/auth/auth.returnTo.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  const target = toShellRelative(location);
+
+  if (!isRestorableReturnTo(target)) {
+    return;
+  }
+
+  // Loop guard: a target we just restored and immediately bounced off of is not worth retrying.
+  if (sessionStorage.getItem(RETURN_TO_LAST_KEY) === target) {
```
ademidoff, line 93: **[2/14] The loop guard is not scoped to the restore that set it, so the first genuine re-login at a restored URL silently drops the deep link.**

Walk the lifecycle:

1. User opens `/pmm-ui/graph/d/node-cpu` logged out -> `saveReturnTo` stores it.
2. User logs in -> `consumeReturnTo()` returns the target **and** sets `RETURN_TO_LAST_KEY = '/graph/d/node-cpu'` (line 134).
3. The restore is a client-side `<Navigate>`, so no further page load happens. `consumeReturnTo()` is the only thing that clears `RETURN_TO_LAST_KEY`, and it is gated by `returnToHandledRef` to once per page load - so the key survives for the rest of the tab's life.
4. Two hours later the session expires while the user is still on that dashboard. `redirectToLogin` -> `saveReturnTo('/graph/d/node-cpu')` matches `RETURN_TO_LAST_KEY` here, so it clears the guard and returns **without storing**.
5. After re-login the user lands on the Grafana home instead of the dashboard they were reading.

That is exactly the bug this PR exists to fix, reintroduced on the second login of every session.

The guard can't currently distinguish "restored and immediately bounced back off" from "restored, worked for two hours, then the session expired". A timestamp (refuse only within a few seconds of the restore) or an attempt counter would separate the two; a bare equality check that outlives the tab won't.

ui/apps/pmm/src/hooks/useFirstLoginRedirect.ts
```diff
@@ -0,0 +1,50 @@
...
+  const location = useLocation();
+  const navigate = useNavigate();
+  const handledRef = useRef(false);
+
+  useEffect(() => {
+    if (handledRef.current || !user || user.isAnonymous) {
+      return;
+    }
+
+    if (!HOME_PATHS.includes(location.pathname)) {
```
ademidoff, line 35: **[3/14] The first-login redirect now fires on any client-side navigation to `/graph`, not just on a fresh boot - so it can yank a user out of a page mid-session.**

When the user does *not* land on a home path, `handledRef` stays `false` and the effect keeps re-running on every location change (dep list on line 49). Concretely:

1. A first-time user is deep-linked - or restored via the new return-to - to `/pmm-ui/graph/d/node-cpu`. `handledRef` stays `false`.
2. Inside the iframe they click Grafana's home breadcrumb.
3. compat relays `LOCATION_CHANGE` with `pathname: '/'`.
4. `getLocationUrl()` (`ui/apps/pmm/src/contexts/grafana/grafana.utils.ts:5-10`) maps that to `/graph/`.
5. `GrafanaProvider` navigates the shell there, `HOME_PATHS.includes('/graph/')` is now true, the flag is burned and `navigate('/help', { replace: true })` throws the user onto the Help page out of a live dashboard session.

The compat implementation this replaces could only ever trigger on a top-level document load, so this is a new failure mode rather than a port of the old one. The doc comment above (lines 20-22) reasons about "landing on the Grafana home path" as a boot-time signal, but the hook has no way to tell a boot from the 40th in-session navigation.

Gating on first render / `useNavigationType() !== 'POP'`, or moving the decision into `AuthProvider`'s existing pre-children gate next to the return-to restore, would restore the original semantics.

ui/apps/pmm/src/contexts/auth/auth.returnTo.ts
```diff
@@ -0,0 +1,137 @@
...
+  if (NOTHING_TO_RESTORE.includes(path)) {
+    return false;
+  }
+
+  // Never store a target that itself bounces back out of the shell.
+  return !isGrafanaDirectPath(path);
+};
+
+const dropGrafanaReturnTo = () => {
+  sessionStorage.removeItem(GRAFANA_REDIRECT_TO_KEY);
```
ademidoff, line 54: **[4/14 - PLAUSIBLE, not verified] `dropGrafanaReturnTo()` may be clearing a key Grafana never writes.**

Flagging this as unverified: I could not confirm from the Grafana source that a `sessionStorage` key named `redirectTo` exists. What I could find points at two *other* mechanisms:

- a `redirect_to` **cookie**, written server-side by `pkg/middleware` when an unauthenticated page request is bounced to `/login` (this is the one behind the `/grafana` -> `/grafana/login` redirect-loop reports);
- a `redirectTo` **query parameter**, e.g. on `/user/auth-tokens/rotate?redirectTo=/d/...`.

If the cookie is the real mechanism, this `removeItem` is a no-op and the desync the doc comment describes is still live. With `serve_from_sub_path = true` (`build/ansible/roles/grafana/files/grafana.ini:41`), the shell's own iframe hitting Grafana unauthenticated is what sets `redirect_to=/graph/d/A`. If the shell has meanwhile remembered `/graph/d/B`, then after login Grafana honours the cookie and sends the top-level window to `/graph/d/A` -> nginx 302 -> `/pmm-ui/graph/d/A`, and only *then* does the shell re-navigate to B - a visible double navigation the comment claims is eliminated.

**What would confirm it:** on a live PMM server, log out, open a deep-linked dashboard, and check `document.cookie` for `redirect_to` at the `/graph/login` page, plus whether `sessionStorage.redirectTo` is ever populated. If it's the cookie, clear that instead (or drop the claim from the comment).

ui/apps/pmm/src/contexts/auth/auth.utils.ts
```diff
@@ -35,5 +36,10 @@ export const getSessionExpiryCookie = () =>
     .find((row) => row.startsWith('grafana_session_expiry='));
 
 export const redirectToLogin = () => {
+  // Single choke point for every bounce to login, so the requested URL is always remembered.
+  // Grafana's own ?redirectTo= is deliberately not used: it is replayed client-side on a later
+  // page load, which under the /pmm-ui redirect happens inside the Grafana iframe and would leave
+  // the shell's URL behind.
+  saveReturnTo();
```
ademidoff, line 43: **[5/14] Storage write as a render-phase side effect makes the one-shot loop guard render-count dependent.**

`redirectToLogin()` is called from `AuthProvider`'s render body (`ui/apps/pmm/src/contexts/auth/auth.provider.tsx:112-115`), and `shouldRedirectToLogin` stays true, so any additional render before the browser processes `window.location.replace` calls `saveReturnTo()` again.

Under StrictMode (dev) React double-invokes render, and the two passes cancel each other out:

- pass 1 hits the guard at `auth.returnTo.ts:93`, clears `RETURN_TO_LAST_KEY`, refuses to store;
- pass 2 sees the now-cleared key and stores anyway.

So the guard meant to break a restore-401-restore loop does nothing in dev and is timing-dependent in prod (a concurrent re-render, or the settings query resolving, is enough).

Moving the save into an effect - or into a module-level one-shot next to the `replace` - would make it fire exactly once per bounce and make the guard's semantics deterministic.

build/ansible/roles/nginx/files/conf.d/pmm.conf
```diff
@@ -159,6 +159,38 @@
...
+      # duplicate slashes first, because window.location.pathname keeps both verbatim while $uri
+      # here does not — otherwise /graph/%61pi/... would be exempt below yet unmatched there.
+      # Match on $uri: it has the query string stripped and is percent-decoded,
+      # so /graph/%61pi/... cannot slip past the exclusions. The redirect itself
+      # must keep $request_uri to preserve the raw query string and encoding.
+      set $redirect_to_pmm_ui 0;
+      if ($http_sec_fetch_dest = "document") {
+        set $redirect_to_pmm_ui 1;
+      }
+      if ($uri ~ "^/graph/(api|render|login|logout|signup|invite|verify|user/password/(send-reset-email|reset))(/|$)") {
```
ademidoff, line 181: **[6/14] Missing exclusion: `/graph/user/auth-tokens/rotate`.**

The regex exempts `user/password/(send-reset-email|reset)` but not Grafana's own session-rotation route, which is a top-level **document** navigation carrying `?redirectTo=`. Grafana sends the browser to `/user/auth-tokens/rotate?redirectTo=/d/foo` when `grafana_session_expiry` has lapsed on a page load.

A bookmark, a Grafana-generated link, or any top-level hit on `/graph/user/auth-tokens/rotate?redirectTo=/d/foo` now matches `Sec-Fetch-Dest: document` and gets 302'd to `/pmm-ui/graph/user/auth-tokens/rotate?redirectTo=/d/foo`. The shell has no such route - `ui/apps/pmm/src/router.tsx` falls through to `graph/*` - so the rotation happens inside the iframe while the address bar stays pinned to the rotate URL until compat's `LOCATION_CHANGE` echo repairs it.

Every other Grafana auth/account route is exempted here; this one looks like an oversight. Adding `user/auth-tokens/rotate` to this regex and to `GRAFANA_DIRECT_PATH_PATTERN` in `ui/packages/shared/src/constants.ts:10` would close it.

(The shell's own `rotateToken()` is fine - it POSTs to `/graph/api/user/auth-tokens/rotate`, which is already covered by the `api` branch.)

build/ansible/roles/nginx/files/conf.d/pmm.conf
```diff
@@ -159,6 +159,38 @@
...
+        set $redirect_to_pmm_ui 0;
+      }
+      # grafana-image-renderer is headless Chrome navigating to the panel URL
+      # itself, so it arrives as Sec-Fetch-Dest: document on an ordinary
+      # dashboard path. Mirrors isRenderingServer() in ui/packages/shared/src/utils.ts.
+      if ($arg_render = "1") {
+        set $redirect_to_pmm_ui 0;
+      }
+      if ($redirect_to_pmm_ui = 1) {
```
ademidoff, line 190: **[7/14] The 302 is emitted before `auth_request`, so `/graph/*` document navigations are no longer authenticated or authorized at the proxy.**

`return` inside `if` executes in `NGX_HTTP_REWRITE_PHASE`, which runs before `NGX_HTTP_ACCESS_PHASE` where the server-level `auth_request /auth_request` (line 100) lives. So this redirect is issued without ever consulting pmm-managed.

That's benign for the redirect itself - it leaks nothing, and RBAC is still enforced when the iframe requests the same URL - but the observable behaviour changes:

- an unauthenticated `GET /graph/d/foo` used to produce pmm-managed's 401 body; it now produces a 302 and a 200 shell;
- a role-denied `GET /graph/d/foo` used to produce the `@access_denied` JSON (lines 140-144); the 403 now surfaces as raw JSON rendered *inside the iframe*, with no PMM chrome around it.

Worth confirming that the access-control tests in `api-tests/` and the pmm-qa e2e suites don't assert on `/graph` status codes, since any browser-driven check that GETs `/graph/...` now gets a 302 first.

(For what it's worth, the multi-`if` flag idiom itself is safe here: `ngx_http_update_location_config` only overwrites `r->content_handler` when the nested location defines one, so `proxy_pass` survives the implicit if-location switch. This isn't the classic "if is evil" trap.)

ui/packages/shared/src/constants.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Grafana routes that must keep the /graph prefix and never be pulled into the PMM UI shell: the
+ * REST API, the image renderer and the auth/account pages. Kept in sync with the redirect
+ * exclusion regex in build/ansible/roles/nginx/files/conf.d/pmm.conf (location /graph) — nginx
+ * exempts these server-side, and neither the compat plugin nor the shell may undo that.
+ *
+ * Test it against a path only, never a path plus query string — nginx matches on $uri for the
+ * same reason.
+ */
+export const GRAFANA_DIRECT_PATH_PATTERN =
```
ademidoff, line 10: **[8/14] The exclusion list is now written out four times with nothing but comments keeping it in sync.**

The same alternation appears in:

- `build/ansible/roles/nginx/files/conf.d/pmm.conf:181`
- this file, line 10-11
- `ui/packages/shared/src/constants.test.ts:6-20`
- `ui/apps/pmm-compat/src/compat.test.ts:74-88`

Both test tables are hand-copied from whichever regex the author edited, so a one-sided change passes CI. And a one-sided change is not a cosmetic problem: if nginx lets a route through to Grafana but `compat.ts` no longer recognises it as a direct path, the plugin bounces it into the shell, the shell's iframe requests it back, and you get a redirect ping-pong. The missing `user/auth-tokens/rotate` above is already an instance of the list drifting.

Options, in rough order of preference: generate the nginx snippet from this constant at build time; or export a single fixture array of paths that both test files import, so adding a route to one place and not the other fails the build.

ui/apps/pmm/src/components/first-login-redirect/FirstLoginRedirect.tsx
```diff
@@ -0,0 +1,9 @@
+import { useFirstLoginRedirect } from 'hooks/useFirstLoginRedirect';
+
+/** Side effect only: navigates first-time users to the welcome page. Renders nothing. */
+const FirstLoginRedirect = () => {
+  useFirstLoginRedirect();
```
ademidoff, line 5: **[10/14] On a first login the Grafana home dashboard is fetched purely to be thrown away.**

nginx sends a first-time user to `/pmm-ui/graph/`. In that first commit, `GrafanaProvider`'s effect (`ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx:70-73`) sets `isLoaded = true` because `isGrafanaPage && isLoggedIn`, so `GrafanaPage` mounts `<iframe src="/graph/">` and Grafana starts booting the home dashboard with all of its panel queries.

`FirstLoginRedirect`'s `navigate('/help')` only lands on the next render, at which point `isGrafanaPage` flips false, `isLoaded` flips false, and the iframe is unmounted. Effects flush children-before-parents and left-to-right, so this component's effect does run first - but `navigate` is a state update, and `GrafanaProvider`'s effect in the same commit still sees the old location.

Net result: the welcome path pays for a full Grafana boot plus dashboard queries that are never shown. Deciding the redirect before the iframe is armed - e.g. inside `AuthProvider`'s existing pre-children gate, alongside the return-to restore - would avoid it, and would also fix the mid-session trigger flagged on `useFirstLoginRedirect.ts:35`.

build/ansible/roles/nginx/files/conf.d/pmm.conf
```diff
@@ -159,6 +159,38 @@
...
+      # Mirrored client-side by isGrafanaDirectPath() in ui/packages/shared/src/constants.ts: the
+      # compat plugin runs inside Grafana and the shell remembers where to return after login, and
+      # neither may pull these routes into the shell. That helper percent-decodes and collapses
+      # duplicate slashes first, because window.location.pathname keeps both verbatim while $uri
+      # here does not — otherwise /graph/%61pi/... would be exempt below yet unmatched there.
+      # Match on $uri: it has the query string stripped and is percent-decoded,
+      # so /graph/%61pi/... cannot slip past the exclusions. The redirect itself
+      # must keep $request_uri to preserve the raw query string and encoding.
+      set $redirect_to_pmm_ui 0;
+      if ($http_sec_fetch_dest = "document") {
```
ademidoff, line 178: **[12/14 - PLAUSIBLE, not verified] The response varies on a request header, but no `Vary: Sec-Fetch-Dest` is sent.**

Two requests for the same `/graph/d/foo` now differ only by `Sec-Fetch-Dest`: `document` gets a 302 into the shell, `iframe` gets proxied Grafana HTML. Nothing in the response signals that dependency, so a shared cache keyed on URL alone could store the 302 and later serve it to the shell's own iframe - which would load the shell, which mounts another iframe, which loads the shell, unbounded, inside the page. PMM is commonly fronted by a customer reverse proxy or CDN, which is where this would bite.

Marking it plausible rather than confirmed, because two things argue against it in practice: 302 is not heuristically cacheable per RFC 9111 (it's absent from the heuristically-cacheable status list), and the http-level `add_header Cache-control "no-cache"` in `build/ansible/roles/nginx/files/nginx.conf:61` is inherited here and forces revalidation on compliant caches.

**What would confirm it:** put PMM behind a CDN or caching proxy that is configured to cache redirects (Cloudflare with a "cache everything" page rule is the common real-world case) and check whether an iframe load ever receives the 302. Cheap insurance either way: `add_header Vary Sec-Fetch-Dest always;` in this location.

ademidoff (commented):
Review of the `/graph` -> `/pmm-ui/graph` redirect work. 14 findings, ranked most severe first; the numbering in each comment is the rank, so it survives GitHub reordering them by file.

The three that I think need to be resolved before merge:

1. **`ui/apps/pmm/src/contexts/auth/auth.returnTo.ts:79`** - unguarded `sessionStorage` on the login-bounce path, reached from render. With site data blocked (Safari private mode), "go to login" becomes an unrecoverable white page.
2. **`ui/apps/pmm/src/contexts/auth/auth.returnTo.ts:93`** - the `RETURN_TO_LAST_KEY` loop guard outlives the restore that set it, so the first re-login at a URL you were already restored to loses the deep link. That's the bug the PR exists to fix, reintroduced on the second login of each session.
3. **`ui/apps/pmm/src/hooks/useFirstLoginRedirect.ts:35`** - the first-login redirect fires on any client-side navigation whose router path is `/graph`, including the one the Grafana iframe echoes back when the user clicks Grafana's home. It can pull a user out of a live dashboard onto the Help page.

Two findings are marked **PLAUSIBLE** rather than asserted - the Grafana `redirect_to` storage mechanism and the `Vary: Sec-Fetch-Dest` cache risk. Each says what would confirm it; please don't treat either as a defect until someone checks.

Findings 9 and 11 have no valid inline anchor (an unchanged line, and a file outside the diff) - they're in a follow-up comment on the PR.

Things I checked and cleared, so nobody re-derives them:

- The multi-`if` flag idiom in nginx is **safe** here. `ngx_http_update_location_config` only overwrites `r->content_handler` when the nested location defines one, so `proxy_pass`, `proxy_set_header`, `proxy_cookie_path` and the inherited `add_header`s all survive the implicit if-location switch. This is not the classic "if is evil" trap.
- The `$uri`-vs-`window.location.pathname` mirroring is genuinely consistent. nginx decodes and merges slashes during URI parsing; `isGrafanaDirectPath` decodes then collapses, in the same order. `/graph/%61pi/...`, `/graph//login` and `/graph%2Fapi` all agree on both sides.
- The first-login localStorage key migrates correctly. `useUser().user.id` comes from Grafana's `/api/user` via `getPerconaUser` (`user.utils.ts:47`), which is the same id the compat plugin read from `grafanaBootData.user.id`, so users who already dismissed the welcome page won't see it again.
- `AuthProvider`'s new `useLocation`/`Navigate` dependency on a Router context is fine - `Providers` is its only call site and it's a route element.
- `ui/packages/shared/src/constants.test.ts` will actually run; that package has a jest config and existing sibling tests.
- Sign in / Sign out render as real anchors with `target: '_self'` (`SidebarNavItem.utils.ts:9-14`), so they perform genuine top-level navigations and hit the nginx exemptions correctly.

I did not run `cd ui && make test` or exercise the nginx config against a live server. Both are worth doing before merge, particularly the redirect chains for `/`, `/graph`, `/graph/login` and a `?render=1` panel URL.

ui/apps/pmm/src/contexts/auth/auth.returnTo.ts
```diff
@@ -0,0 +1,137 @@
...
+const NOTHING_TO_RESTORE = ['/', GRAFANA_SUB_PATH, `${GRAFANA_SUB_PATH}/`];
+
+/**
+ * Targets are shell-relative (no /pmm-ui), so they can be handed straight to react-router's
+ * navigate() under basename /pmm-ui. Both Grafana routes (/graph/d/...) and native PMM UI pages
+ * (/settings/advanced, /help) qualify.
+ */
+export const isRestorableReturnTo = (target: string) => {
+  // Exactly one leading slash keeps this a same-origin path: rejects //evil.com and absolute URLs.
+  if (!target.startsWith('/') || target.startsWith('//')) {
```
ademidoff, line 32: **[13/14] The same-origin gate inspects the raw string, while the route check below normalises first - so the two halves of the validator disagree about what the target is.**

Line 32 rejects a literal `//` prefix and line 41 rejects `..`/`\`, both **pre-decode**. But `isGrafanaDirectPath(path)` on line 50 percent-decodes and collapses duplicate slashes before matching. So a stored value like `/%2F%2Fevil.com`, `/%09/evil.com` or `/%2e%2e/x` sails through the checks that are supposed to establish "this is a same-origin path" and only the route check sees its real shape.

Not exploitable today - the value only ever originates from `window.location.pathname`, and react-router's `pushState` keeps the result same-origin regardless - so this is hardening, not a live vulnerability. But the comment on line 31 asserts a security property the code doesn't actually enforce, and that's the kind of comment someone later relies on when they add a second writer to this key.

Normalising once at the top and running *every* check (leading slash, backslash, `..`, `NOTHING_TO_RESTORE`, `isGrafanaDirectPath`) against that one normalised value would make the guard mean what it says. Note the `..` check has to stay on the path portion only, per the existing `?var-version=1..2` case.

ui/packages/shared/src/constants.ts
```diff
@@ -0,0 +1,31 @@
...
+  /^\/graph\/(api|render|login|logout|signup|invite|verify|user\/password\/(send-reset-email|reset))(\/|$)/;
+
+/**
+ * Match the way nginx normalises $uri before testing its exclusions: percent-decode, then collapse
+ * duplicate slashes. window.location.pathname does neither, so without this /graph/%6Cogin is
+ * exempted server-side yet unrecognised here — and the shell would pull a Grafana auth page in.
+ * ".." needs no handling: the browser resolves it before it reaches location.pathname.
+ */
+export const isGrafanaDirectPath = (pathname: string) => {
```
ademidoff, line 19: **[14/14] A predicate in `constants.ts`, while its sibling predicate lives in `utils.ts`.**

`isGrafanaDirectPath` and `isRenderingServer` are two halves of the same idea - "does this URL bypass the shell?" - and `ui/apps/pmm/src/contexts/auth/auth.returnTo.ts:1` imports both from `@pmm/shared` as though they were co-located. But they aren't, and the nginx comment block has to send readers to two different files as a result (`pmm.conf:169` -> `constants.ts`, `pmm.conf:186` -> `utils.ts`).

Separately, `/graph` is now hard-coded a third time inside the pattern on line 11. It already exists as `GRAFANA_SUB_PATH` in `ui/apps/pmm/src/lib/constants.ts:7` and again in `ui/apps/pmm-compat/src/lib/constants.ts:20`, so a future subpath change needs three coordinated edits with no compiler help.

Suggest moving `isGrafanaDirectPath` next to `isRenderingServer` in `utils.ts` and promoting `GRAFANA_SUB_PATH` into the shared package, building the pattern from it.

ademidoff:
Two findings from the review above that have no valid inline anchor — one lands on an unchanged line, the other on a file that isn't in the diff. Numbering continues the same ranking.

---

### [9/14] `build/ansible/roles/nginx/files/conf.d/pmm.conf:156` — the root URL now costs two redirects

`location = /` still points at `/graph/`, which the new rule then bounces again:

```
GET /                 -> 302 /graph/
GET /graph/           -> 302 /pmm-ui/graph/      (Sec-Fetch-Dest: document)
GET /pmm-ui/graph/    -> 200
```

Three round trips before a byte of HTML, on the single most common entry point into PMM, plus an extra hop in the redirect chain for every corporate proxy, uptime monitor and link checker in front of the server.

The `/graph/` hop only exists for pre-shell PMM. Now that the shell is the destination, `location = /` can point straight at it:

```nginx
location = / {
  auth_request off;
  return 302 /pmm-ui/graph/;
}
```

Worth checking `location = /dashboard/snapshots` (line 211) at the same time — it currently 301s to `/graph/dashboard/snapshots`, which then takes the new 302 as well, for a three-hop chain.

---

### [11/14] `ui/apps/pmm/src/contexts/navigation/navigation.constants.ts:780` — the explicit "Sign in" link drops the user's current page

The return-to mechanism only covers *involuntary* bounces, so the seamless-redirect promise is half-implemented.

With `[auth.anonymous]` enabled, an anonymous user browsing `/pmm-ui/graph/d/node-cpu` clicks **Sign in**. `NAV_SIGN_IN` renders a plain `<a href="/graph/login" target="_self">` (`ui/apps/pmm/src/components/sidebar/nav-item/SidebarNavItem.utils.ts:9-14`), which bypasses `redirectToLogin()` entirely — nothing calls `saveReturnTo()`. After authenticating they land on the Grafana home instead of the dashboard they were reading.

The anchor has to stay a real top-level navigation, since `/graph/login` is deliberately exempted from the nginx redirect and must not be pulled into the shell's router. But an `onClick` that calls `saveReturnTo()` before the default navigation proceeds — or routing the click through `redirectToLogin()` — would make voluntary and involuntary logins behave the same.

Same applies to **Sign out** (`NAV_SIGN_OUT`, line 766) if you want a post-logout re-login to return to where the user was, though there's a reasonable argument that an explicit sign-out *should* forget the location. Worth a deliberate decision either way rather than falling out of the implementation.

ui/packages/shared/src/grafanaDirectPaths.fixtures.ts
```diff
@@ -0,0 +1,40 @@
+/** Sample paths for the redirect contract, shared by every suite that asserts on it. */
+export const GRAFANA_DIRECT_PATHS = [
```
fabio-silva, line 2: `constants.ts` has other Grafana paths as well. Question:
- should these be separate?
- if yes, why is one just "constants.ts" and the other "grafanaDirectPaths.fixtures.ts"?

PR #5355, opened 2026-05-12
PMM-7 - Redesign RTA telemetry - flat metrics, enabled + disabled signals

ondrejpatocka:
**TL;DR**
- Replaces 3 JSON-wrapped RTA telemetry items with 3 flat scalars + 1 flat per-agent running block. 
- Adds adoption and usage signals; keeps DB version + collect interval as flat per-agent columns. 
- No code, no schema - config-only change.

**Reasoning**
- Some feature adoption markers missing
- Inner JSON data are very hard to efficiently aggregate over a large scale data set


managed/services/telemetry/config.default.yml
```diff
@@ -991,69 +991,96 @@ telemetry:
...
-      metric: "rta_db_versions"
+      EXTRACT(EPOCH FROM y.created_at::timestamp)
+        - EXTRACT(EPOCH FROM x.created_at::timestamp) AS value_difference_sec
+      FROM services x
+      JOIN agents y ON y.agent_type = 'rta-mongodb-agent'
+      WHERE x.service_name = 'pmm-server-postgresql'
+        AND y.created_at = (
+          SELECT MIN(created_at) FROM agents WHERE agent_type = 'rta-mongodb-agent'
+        );
+    summary: "Time in seconds from PMM installation to first Real-Time Analytics agent created"
```
maxkondr: the description doesn't match the query logic

managed/services/telemetry/config.default.yml
```diff
@@ -991,69 +991,96 @@ telemetry:
...
-        agent_type, db_version;
-    summary: "Real-Time Analytics agents used for last day grouped by DB version"
-    transform:
-      type: JSON
-      metric: "rta_db_versions"
+      EXTRACT(EPOCH FROM y.created_at::timestamp)
+        - EXTRACT(EPOCH FROM x.created_at::timestamp) AS value_difference_sec
+      FROM services x
+      JOIN agents y ON y.agent_type = 'rta-mongodb-agent'
+      WHERE x.service_name = 'pmm-server-postgresql'
```
maxkondr: how is this WHERE condition related to MongoDB RTA?

managed/services/telemetry/config.default.yml
```diff
@@ -991,69 +991,96 @@ telemetry:
...
+      - metric_name: "rta_mongo_first_agent_sec"
+        column: "value_difference_sec"
+
+  - id: RTAMongoAgentsRunning24h
+    source: PMMDB_SELECT
+    query: >
+      count(*) AS agent_count
+      FROM agents
+      WHERE agent_type = 'rta-mongodb-agent'
+        AND (disabled = false OR created_at >= now() - interval '1 day');
```
maxkondr: this is wrong logic. 
`created_at` column will be filled only once during first RTA agent creation. So that you will get metric only once during first 24h period.

All subsequent RTA agent activations update `updated_at` column only.

managed/services/telemetry/config.default.yml
```diff
@@ -991,69 +991,96 @@ telemetry:
...
-      type: JSON
-      metric: "rta_session_durations"
+      WHERE agent_type = 'rta-mongodb-agent';
+    summary: "Total count of Real-Time Analytics MongoDB agents (running now + in the past)"
     data:
-      - metric_name: "agent_type"
-        column: "agent_type"
-      - metric_name: "duration_sec"
-        column: "duration_sec"
+      - metric_name: "rta_mongo_agents_lifetime_total"
```
maxkondr: metric name doesn't correlate with query logic. agent lifetime equal to agent running duration. Previously it was `RTASessionDuration` that has been removed

managed/services/telemetry/config.default.yml
```diff
@@ -991,69 +991,96 @@ telemetry:
...
-        AND rta_options->'collect_interval' != 'null'
-        AND created_at >= now() -  interval '1 day';
-    summary: "Real-Time Analytics collect interval (sec)"
-    transform:
-      type: JSON
-      metric: "rta_agent_collect_interval"
+      WHERE agent_type = 'rta-mongodb-agent'
+        AND disabled = true
+        AND updated_at >= now() - interval '1 day';
+    summary: "Count of Real-Time Analytics MongoDB agents stopped within past 24 hours"
```
maxkondr: what is a business value from this info?

managed/services/telemetry/config.default.yml
```diff
@@ -991,69 +991,96 @@ telemetry:
...
+        column: "agent_count"
+
+  - id: RTAMongoAgentsStartedThenStopped24h
+    source: PMMDB_SELECT
+    query: >
+      count(*) AS agent_count
+      FROM agents
+      WHERE agent_type = 'rta-mongodb-agent'
+        AND disabled = true
+        AND created_at >= now() - interval '1 day';
```
maxkondr: this logic is wrong. It will produce output only once. the next day it will be empty even if RTA agent is started and stopped. Reason: `created_at` is filled only once during first RTA agent creation. All subsequent RTA agent activations update `updated_at` column only.

managed/services/telemetry/config.default.yml
```diff
@@ -991,69 +991,96 @@ telemetry:
...
+  # Per-agent rows; agent_type is the row marker (consumer assumes 4 lines per agent).
+  # COALESCE keeps row alignment when version/collect_interval are NULL (empty values would be dropped by removeEmpty).
+  # 1_000_000_000 divisor converts the stored nanosecond CollectInterval to seconds. LIKE 'rta-%' is forward-compatible with future PG/MySQL RTA agents.
+  - id: RTAAgentsRunning24hDetail
+    source: PMMDB_SELECT
+    query: >
+      a.agent_type AS agent_type,
+      COALESCE(s.version, 'unknown') AS db_version,
+      EXTRACT(EPOCH FROM a.updated_at - a.created_at)::int AS duration_sec,
+      COALESCE(((a.rta_options->>'collect_interval')::bigint / 1000000000)::int, 0) AS collect_interval_sec
```
maxkondr, line 1043: there shouldn't be situations when `rta_options->collect_interval` is empty.

managed/services/telemetry/config.default.yml
```diff
@@ -991,69 +991,96 @@ telemetry:
...
+      FROM agents a
+      LEFT JOIN services s ON a.service_id = s.service_id
+      WHERE a.agent_type LIKE 'rta-%'
+        AND (a.disabled = false OR a.created_at >= now() - interval '1 day');
```
maxkondr: again, this metric will be shown only once

maxkondr (changes requested):

managed/services/telemetry/config.default.yml
```diff
@@ -991,69 +991,67 @@ telemetry:
...
+  # Real-Time Analytics (RTA) agents details
+  # Per-agent rows; agent_type is the row marker (consumer assumes 4 lines per agent).
+  # COALESCE keeps row alignment when version/collect_interval are NULL (empty values would be dropped by removeEmpty).
+  # 1_000_000_000 divisor converts the stored nanosecond CollectInterval to seconds. LIKE 'rta-%' is forward-compatible with future PG/MySQL RTA agents.
+  - id: RTAAgentsEnabledDetail
+    source: PMMDB_SELECT
+    query: >
+      a.agent_type AS agent_type,
+      COALESCE(s.version, 'unknown') AS db_version,
+      EXTRACT(EPOCH FROM a.updated_at - a.created_at)::int AS session_sec,
```
JiriCtvrtka, line 1042: Maybe I am wrong, but I think updated_at does not keep changing while the session runs. This will usually report 0 or the time until the last agent metadata update, not the active session age. If we want the current session duration, should this use now() - a.created_at instead?

maxkondr: > If we want the current session duration, should this use now() - a.created_at instead?

No, because `created_at` is filled once during agent creation. 
RTA session stop doesn't delete rta agent from DB, it only sets `disabled=true`.
All subsequent sessions start/stop update `updated_at` column. And yes, updating other agent fields (like labels, etc) bumps `updated_at` as well, that influence on session duration calculation.

managed/services/telemetry/config.default.yml
```diff
@@ -991,69 +991,67 @@ telemetry:
...
-      - metric_name: "agent_type"
-        column: "agent_type"
-      - metric_name: "duration_sec"
-        column: "duration_sec"
+      - metric_name: "rta_mongo_agents_enabled"
+        column: "agent_count"
 
-  - id: RTAAgentCollectInterval
+  # disabled=true means StopSession was called; mirrors the disabled=false logic of RTAMongoAgentsEnabled.
+  - id: RTAMongoAgentsDisabled
```
maxkondr, line 1021: this one metric can be combined together with the prev ones. As a result on storage side it will be possible to calculate:
- total number of RTA agents
- number of enabled RTA agents
- number of disabled RTA agents

In addition it is possible to make the query unified for all future RTA agents (`agent_type like 'rta-%' `)

managed/services/telemetry/config.default.yml
```diff
@@ -991,69 +991,67 @@ telemetry:
...
+    source: PMMDB_SELECT
+    query: >
+      a.agent_type AS agent_type,
+      COALESCE(s.version, 'unknown') AS db_version,
+      EXTRACT(EPOCH FROM a.updated_at - a.created_at)::int AS session_sec,
+      COALESCE(((a.rta_options->>'collect_interval')::bigint / 1000000000)::int, 0) AS collect_interval_sec
+      FROM agents a
+      LEFT JOIN services s ON a.service_id = s.service_id
+      WHERE a.agent_type LIKE 'rta-%'
+        AND a.disabled = false;
```
maxkondr, line 1047: why are only running agents counted here?

maxkondr (approved):

JiriCtvrtka (approved):

PR #5369, opened 2026-05-15
PMM-14858 Skip built-in ClickHouse scrape for external instances

4nte:
PMM-14858

[Feature build](https://github.com/Percona-Lab/pmm-submodules/pull/4356)

PMM kept scraping `127.0.0.1:9363` even when an external ClickHouse was configured.
Introduced new  `ClickHouseParams` model struct that validates and facilitates CH parameters. 

ClickHouse scraping is now skipped when `PMM_CLICKHOUSE_ADDR` points to an external host.


managed/services/victoriametrics/victoriametrics.go
```diff
@@ -60,26 +60,31 @@ type Service struct {
...
+	chParams *models.ClickHouseParams
 
 	l         *logrus.Entry
 	reloadCh  chan struct{}
 	haService haService
 }
 
 // NewVictoriaMetrics creates new VictoriaMetrics service.
-func NewVictoriaMetrics(scrapeConfigPath string, db *reform.DB, params *models.VictoriaMetricsParams, haService haService) (*Service, error) {
+func NewVictoriaMetrics(scrapeConfigPath string, db *reform.DB, params *models.VictoriaMetricsParams, chParams *models.ClickHouseParams, haService haService) (*Service, error) {
```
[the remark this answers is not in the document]
ademidoff: This needs `//nolint:lll`

managed/models/clickhouse_params.go
```diff
@@ -0,0 +1,74 @@
...
+
+// ClickHouseParams represents ClickHouse server params.
+type ClickHouseParams struct {
+	BuiltinDisabled bool
+	url             *url.URL
+}
+
+// ExternalClickHouse returns true if ClickHouse is configured externally.
+func (p *ClickHouseParams) ExternalClickHouse() bool {
+	return p.url.Hostname() != "127.0.0.1"
```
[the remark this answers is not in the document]
ademidoff: Yeah, why not, i.e. const localhost = "127.0.0.1"?

managed/cmd/pmm-managed/main.go
```diff
@@ -729,6 +729,7 @@ func main() { //nolint:gocognit,maintidx,cyclop
 	clickhouseAddrF := kingpin.Flag("clickhouse-addr", "Clickhouse database address").Default("127.0.0.1:9000").Envar("PMM_CLICKHOUSE_ADDR").String()
 	clickhouseUsernameF := kingpin.Flag("clickhouse-username", "Clickhouse database user").Default("default").Envar("PMM_CLICKHOUSE_USER").String()
 	clickhousePasswordF := kingpin.Flag("clickhouse-password", "Clickhouse database user password").Default("clickhouse").Envar("PMM_CLICKHOUSE_PASSWORD").String()
+	clickhouseBuiltinDisabledF := kingpin.Flag("clickhouse-disable-builtin", "Disable the built-in ClickHouse").Envar("PMM_DISABLE_BUILTIN_CLICKHOUSE").Bool()
```
maxkondr: I do not see any benefit from this additional var. The fact of set `PMM_CLICKHOUSE_ADDR` looks enough that external CH usage is requested.

managed/models/clickhouse_params.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// URL returns the ClickHouse DSN.
```
maxkondr:
```suggestion
// URL returns the ClickHouse URL.
```

managed/cmd/pmm-managed/main.go
4nte: I was considering following options:
1. Check if `PMM_CLICKHOUSE_ADDR` hostname is not `127.0.0.1`
2. Check if `PMM_DISABLE_BUILTIN_CLICKHOUSE` is set
3. Check both 1. and 2.

I agree `PMM_CLICKHOUSE_ADDR` is sufficient to determine this and would prefer it too. My concern is that `PMM_DISABLE_BUILTIN_CLICKHOUSE` is already explicitly documented as **the flag** that disables internal CH and [managed-init](https://github.com/percona/pmm/blob/PMM-14858-fix-external-clickhouse-scrape/managed/cmd/pmm-managed-init/main.go#L57) already relies on it. 
Not checking it here, while checking it in `managed-init` seemed inconsistent, unless we treat `PMM_DISABLE_BUILTIN_CLICKHOUSE` as an env var that is exclusive to `managed-init`.

Having all that said, do you still think we should just infer from `PMM_CLICKHOUSE_ADDR` and not lookup `PMM_DISABLE_BUILTIN_CLICKHOUSE`?

ademidoff: We agreed to deprecate all `PMM_DISABLE_BUILTIN_` env variables, since their use conflicts with `PMM_<TECH>_ADDR` variables, which should be the only source of truth. Therefore, this fix will rely on `PMM_CLICKHOUSE_ADDR`.

4nte: I've also discussed this with @ademidoff,  we all agree `PMM_CLICKHOUSE_ADDR` is technically sufficient to determine whether CH is internal or external.

I will omit `PMM_DISABLE_BUILTIN_CLICKHOUSE` and rely on `PMM_CLICKHOUSE_ADDR` as a single source of truth.

Since `PMM_DISABLE_BUILTIN_CLICKHOUSE` is obviously redundant and creates ambiguity, I will create a separate issue to deprecate `PMM_DISABLE_BUILTIN_CLICKHOUSE` in favor of inferring from `PMM_CLICKHOUSE_ADDR`

managed/services/victoriametrics/prometheus.go
```diff
@@ -266,13 +266,21 @@ func AddScrapeConfigs(l *logrus.Entry, cfg *config.Config, q *reform.Querier, //
 
 // AddInternalServicesToScrape adds internal services metrics to scrape targets.
 func addInternalServicesToScrape(s models.MetricsResolutions, svc *Service, pmmServerNodeName string) []*config.ScrapeConfig {
-	cfg := []*config.ScrapeConfig{
+	cfg := []*config.ScrapeConfig{}
+
+	if svc.params.ExternalVM() {
```
maxkondr: the logic has changed drastically.
Before: in case of external VM - scraping of CH was skipped only
Now: in case of external VM - scraping of all services is skipped.

Is in intentional?

managed/services/victoriametrics/victoriametrics.go
```diff
@@ -60,26 +62,37 @@ type Service struct {
...
+	db *reform.DB,
+	params *models.VictoriaMetricsParams,
+	chParams *models.ClickHouseParams,
+	haService haService,
+) (*Service, error) {
 	u, err := url.Parse(params.URL())
 	if err != nil {
 		return nil, err
 	}
+	if chParams == nil {
```
maxkondr: is it really required to pass the whole structure if the logic uses `isExternal` indication only?

managed/services/victoriametrics/prometheus.go
4nte: Yes, it was proposed as a correction by @ademidoff who is the original author of `if svc.params.ExternalVM()` check, with the idea being that if VM is configured externally, we should not produce scrape configs for any service, not just ClickHouse.

So the conclusion was that only skipping CH scrape config when VM is external was a mistake, and that **all** scrape configs should be skipped instead.

I just tried to understand user expectations and desired behavior when VM is external but current docs don't mention what should happen to Grafana / pmm-managed / qan-api2 / ClickHouse metrics.
The fact that only ClickHouse was excluded here makes me think that `if svc.params.ExternalVM()` was supposed to be `if svc.ExternalClickHouse()`, and that `svc.params.ExternalVM()` check was not supposed to be here to begin with.

maxkondr: I am just curious, why don't we want/need to scrape metrics from local pmm-server components in case VM is external? The same Q with CH? can't we scrape external components? In such a case we loose the observability of pmm-server itself

4nte: > I am just curious, why don't we want/need to scrape metrics from local pmm-server components in case VM is external?

You're right,  this prevents VMagent from scraping local pmm metrics, and VMagent is supposed to write them to VM (doesn't matter if VM is internal or external)

managed/services/victoriametrics/victoriametrics.go
4nte: It's technically not required for this, we could extract it and pass it as a boolean value to NewVictoriaMetrics.
I think architecturally it makes sense for VictoriaMetrics to have knowledge of ClickHouse parameters - specifically the address (even though it's not used now).

Today the address is hardcoded to `127.0.0.1:9363`, but I'd like to work towards having the address parameterized, so in the future when we decide to support scraping of external ClickHouse instances, we will be able to retrieve it from `ClickHouseParams` struct.

maxkondr (approved):

ademidoff (approved):

JiriCtvrtka (approved):

build/ansible/pmm-docker/post-build.yml
4nte, line 1: `victoriametrics-promscrape.yml` configuration file should be considered invalid when generated in image build  time since it doesn't honor user provided configuration in runtime (e.g `PMM_CLICKHOUSE_ADDR`), which affects the desired promscrape configuration for VM. 

The problem is that VM attempts to scrape before _managed_ generates / rewrites the appropriate `victoriametrics-promscrape.yml` configuration on container startup, and as a result, will use the misconfigured `victoriametrics-promscrape.yml` configuration file which always contains the builtin clickhouse scrape job. This results in a single failed scrape on container as reported by @peterSirotnak 

The fix is to clear the contents of the file in post-build step and let __managed__ write appropriate configuration in runtime.

maxkondr (approved):

managed/models/common_helpers.go
```diff
@@ -32,3 +32,13 @@ func NewInvalidArgumentError(format string, a ...any) *InvalidArgumentError {
 func (e *InvalidArgumentError) Error() string {
 	return "invalid argument: " + e.Details
 }
+
+// localhost is the IPv4 loopback address used by PMM Server's
+// co-located services.
+const localhost = "127.0.0.1"
```
ademidoff: Duplicate localhost constant: defined in both `managed/models/common_helpers.go` and `managed/services/victoriametrics/victoriametrics.go`. Acceptable since the models one is unexported and the packages are separate, but it's duplication — exporting one (e.g. models.LocalhostAddr) would avoid it.

managed/services/victoriametrics/victoriametrics.go
ademidoff: nit: marginally cleaner if we put this check before dereferencing `params.URL()`

ademidoff (approved):

PR #5390, opened 2026-05-20
PMM-7 Fix intrange linter warnings

maxkondr:
PMM-7


.golangci.yml
```diff
@@ -115,6 +115,7 @@ linters:
       - third_party$
       - builtin$
       - examples$
+      - ui/
```
ademidoff, line 118: Q: are there go files inside the `ui` directory?

ademidoff (approved):

.golangci.yml
maxkondr: in case developer downloads `node_modules` - some of them contain
```
❯ ./bin/golangci-lint run -c=./.golangci.yml --enable-only=intrange
ui/node_modules/flatted/golang/pkg/flatted/flatted.go:68:4: for loop can be changed to use an integer range (Go 1.22+)
Because the key is returned by a function or method, take care to consider side effects. (intrange)
                        for i := 0; i < rv.Len(); i++ {
                        ^
ui/node_modules/flatted/golang/pkg/flatted/flatted.go:100:4: for loop can be changed to use an integer range (Go 1.22+)
Because the key is returned by a function or method, take care to consider side effects. (intrange)
                        for i := 0; i < rv.NumField(); i++ {
                        ^
2 issues:
* intrange: 2
```

PR #5395, opened 2026-05-21
PMM-14594 Backup compatibility fix.

JiriCtvrtka:
PMM-14594

FB: https://github.com/Percona-Lab/pmm-submodules/pull/4368

- Fixed MySQL backup compatibility checks for MySQL 8.4 LTS by allowing Percona XtraBackup 8.4.x with MySQL 8.4.x.
- Kept existing MySQL 8.0.22+ behavior: XtraBackup 8.0.x must have the same or newer core version and stay below 8.1.
- Improved incompatible XtraBackup error messages so they explain which XtraBackup family/version should be used.
- Added unit coverage for MySQL 8.4 positive/negative compatibility cases and equivalence tests to ensure the refactor preserves existing behavior.
- Updated MySQL backup prerequisites documentation with the MySQL-to-XtraBackup compatibility matrix.


JiriCtvrtka:
@copilot review

managed/services/backup/compatibility_helpers.go
```diff
@@ -87,7 +88,7 @@ func init() {
 			mysqlMinVersion:      "8.0.21",
 			mysqlMaxVersion:      "8.0.22",
 			xtrabackupMinVersion: "8.0.14",
-			xtrabackupMaxVersion: "9.0",
+			xtrabackupMaxVersion: "8.1.0",
```
maxkondr: if later all those strings are converted into `version.Version` - isn't it better to declare them as `version.Version` already and avoid conversion on func calls?

JiriCtvrtka: Yes, it makes sense to change it this way. I only fixed the logic, but this refactor simplifies it nicely.

managed/services/backup/compatibility_helpers.go
```diff
@@ -75,68 +73,146 @@ func init() {
...
+func incompatibleXtrabackupError(message, xtrabackupVersionString, mysqlVersionString string) error {
+	return errors.Wrapf(
+		ErrIncompatibleXtrabackup,
+		message,
+		xtrabackupVersionString,
+		mysqlVersionString,
+	)
+}
+
+func mysqlAndXtrabackupCoreVersionsCompatible(mysqlVersion, xtrabackupVersion *version.Version) bool {
```
maxkondr, line 173: in `version/features.go` we have func
```
func (p *Parsed) IsFeatureSupported(f FeatureVersion) bool {
	return !p.Less(f)
}
```

Can we adopt this approach for checking compatibility? looks like the code will be simpleer

JiriCtvrtka: PTAL it should be adopted in similar way.

maxkondr (approved):

4nte (approved):

managed/services/backup/compatibility_helpers.go
```diff
@@ -75,68 +76,139 @@ func init() {
...
+	switch mysqlXtrabackupBandFor(mysqlVersion) {
+	case mysqlXtrabackupBand84:
+		return versionRange{min: mysql84Version, max: mysql85Version}.isSupported(xtrabackupVersion)
+	case mysqlXtrabackupBand80Aligned:
+		return versionRange{min: mysqlVersion, max: mysql81Version}.isSupported(xtrabackupVersion)
```
ademidoff, line 197: **Please confirm intent: MySQL Innovation releases 8.1/8.2/8.3 become un-backupable.**

`band80Aligned` covers `[8.0.22, 8.4.0)` and requires PXB in `[mysqlVersion, 8.1.0)`. For any MySQL >= 8.1.0 in this band the range is empty, so no PXB is ever compatible — `versionRange{min: 8.1.0, max: 8.1.0}.isSupported(...)` is always `false`.

The old code allowed PXB `[mysqlVersion, 9.0)`, so e.g. MySQL 8.1.0 + PXB 8.1.0 was compatible and now is not (the resulting error even tells an 8.1 user to "use Percona XtraBackup 8.0.x for MySQL 8.0.x"). Innovation releases are short-lived/EOL so this is plausibly fine, but it's a real behavior change not called out in the description, and the doc table also omits 8.1-8.3. Worth confirming + adding a test that pins 8.1/8.2/8.3 either way.

managed/services/backup/compatibility_helpers.go
```diff
@@ -75,68 +76,139 @@ func init() {
...
-		xtrabackupMinVersion := version.Must(version.NewVersion(s.xtrabackupMinVersion))
-		xtrabackupMaxVersion := version.Must(version.NewVersion(s.xtrabackupMaxVersion))
-
-		mysqlAndXtrabackupCompatibleVersions = append(mysqlAndXtrabackupCompatibleVersions, compatibility{
-			dbMinVersion:         mysqlMinVersion,
-			dbMaxVersion:         mysqlMaxVersion,
-			backupToolMinVersion: xtrabackupMinVersion,
-			backupToolMaxVersion: xtrabackupMaxVersion,
-		})
+func mysqlAndXtrabackupCompatible(mysqlVersionString, xtrabackupVersionString string) (bool, error) {
```
ademidoff, line 105: `mysqlAndXtrabackupCompatible` is now referenced only by tests — production uses `mysqlAndXtrabackupCompatibilityError`. It's a thin bool wrapper kept for the equivalence loop. Fine to keep, but consider a comment noting it's test-only so it doesn't read as a live production path.

managed/services/backup/compatibility_helpers.go
```diff
@@ -75,68 +76,139 @@ func init() {
...
-			if (mysqlVersion.GreaterThanOrEqual(cv.dbMinVersion) &&
-				mysqlVersion.LessThan(cv.dbMaxVersion)) &&
-				xtrabackupVersion.GreaterThanOrEqual(cv.backupToolMinVersion) &&
-				xtrabackupVersion.LessThan(cv.backupToolMaxVersion) {
-				return true, nil
+			if cv.dbVersions.isSupported(mysqlVersion) && cv.backupToolVersions.isSupported(xtrabackupVersion) {
+				return true
 			}
 		}
+	case mysqlXtrabackupBandUnsupported:
```
ademidoff, line 204: Nit: `case mysqlXtrabackupBandUnsupported: return false` duplicates the trailing `return false` below the switch. Harmless; the explicit case could be dropped.

documentation/docs/backup/mysql-backup/mysql_prerequisites.md
```diff
@@ -38,4 +38,14 @@ Before creating MySQL backups, make sure to:
...
\ No newline at end of file
+       The versions of each must be compatible with the installed version of MySQL.
+       Use the Percona XtraBackup version family that matches your MySQL version:
+
+       | MySQL version | Percona XtraBackup version |
+       |---------------|----------------------------|
+       | MySQL 5.5, 5.6, or 5.7 | Percona XtraBackup 2.4.x |
+       | MySQL 8.0.0 through 8.0.21 | Percona XtraBackup 8.0.x, using the minimum version required for your MySQL release |
+       | MySQL 8.0.22 and newer 8.0 releases | Percona XtraBackup 8.0.x with the same or newer core version |
+       | MySQL 8.4.x | Percona XtraBackup 8.4.x |
```
ademidoff: The matrix jumps from "8.0.22 and newer 8.0 releases" straight to "8.4.x", omitting 8.1-8.3. This is consistent with the code (see the `compatibility_helpers.go` comment about Innovation releases), but if 8.1-8.3 are intentionally unsupported it may be worth stating so explicitly to avoid user confusion.

managed/services/backup/compatibility_helpers.go
JiriCtvrtka: Should be fixed for missed ones.

managed/services/backup/compatibility_helpers.go
```diff
@@ -75,68 +85,185 @@ func init() {
...
+func mysqlAndXtrabackupCompatibilityError(mysqlVersionString, xtrabackupVersionString string) error {
+	mysqlVersion, xtrabackupVersion, err := mysqlAndXtrabackupCoreVersions(mysqlVersionString, xtrabackupVersionString)
+	if err != nil {
+		return err
+	}
+	if mysqlAndXtrabackupCoreVersionsCompatible(mysqlVersion, xtrabackupVersion) {
+		return nil
+	}
+
+	switch mysqlXtrabackupBandFor(mysqlVersion) {
```
ademidoff: Minor / non-blocking: the band is computed twice on the incompatible path. `mysqlAndXtrabackupCoreVersionsCompatible` (line 203) already calls `mysqlXtrabackupBandFor` internally, and this `switch` calls it again. The cost is negligible and the code reads fine as-is, so this is just a note — if you want to tidy it up, you could compute the band once and reuse it for both the compatibility check and the error `switch`.

ademidoff (approved):

JiriCtvrtka:
@copilot review and verify if https://github.com/percona/pmm/pull/5395#discussion_r3394174215 is addressed.

JiriCtvrtka:
@copilot  review and verify if https://github.com/percona/pmm/pull/5395#discussion_r3394174215 is addressed.

PR #5396, opened 2026-05-21
PMM-7 Fix modernize linter warnings

maxkondr:
PMM-7


maxkondr:
@copilot review

managed/cmd/pmm-managed/main.go
```diff
@@ -1214,14 +1214,12 @@ func main() { //nolint:gocognit,maintidx,cyclop
 		return nil
 	}))
 
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
+	wg.Go(func() {
 		err := haService.Run(ctx)
 		if err != nil {
 			l.Panicf("cannot start high availability service: %+v", err)
```
4nte:
```suggestion
			l.Fatalf("cannot start high availability service: %+v", err)
```

According to https://pkg.go.dev/sync#WaitGroup.Go
> The function f must not panic.

4nte (changes requested):

agent/utils/filereader/continuous_file_reader.go
```diff
@@ -65,25 +66,25 @@ func (r *ContinuousFileReader) NextLine() (string, error) {
 	r.m.Lock()
 	defer r.m.Unlock()
 
-	var line string
+	var line strings.Builder
 	for {
 		l, err := r.r.ReadString('\n')
 		r.l.Tracef("ReadLine: %q %v", l, err)
-		line += l
+		line.WriteString(l)
```
ademidoff, line 73: I believe concatenating strings using "+=" is more performant than using the `strings.Builder`. Moreover, it's a very simple case, so it looks like overkill.

ademidoff (approved):

JiriCtvrtka (approved):

agent/utils/filereader/continuous_file_reader.go
maxkondr: no, StringBuilder is better and always recommended for high-performance. the linter complained exactly for this reason.

4nte (approved):

PR #5397, opened 2026-05-21
PMM-14901: RTA details arrows respect filters

mattiasimonato:
PMM-14901

Link to the Feature Build: [SUBMODULES-4370](https://github.com/Percona-Lab/pmm-submodules/pull/4370)

Fixes RTA overview details pane navigation so prev/next arrows move only through filtered table rows, not the full query list from the API.

- `OverviewTable` syncs `navigableQueries` from Material React Table (after column/global filters, before pagination).
- `RealtimeOverview` uses that list for arrow navigation and first/last button state.
- Adds unit tests with a mocked table to ensure the next row is not taken from unfiltered API data.


ui/apps/pmm/src/pages/rta/overview/table/OverviewTable.tsx
```diff
@@ -11,54 +15,76 @@ import { filterElapsedTime } from './OverviewTable.utils';
...
-        'data-testid': `query-${row.original.queryId}-row`,
-      })}
-    />
-  </RealtimeTableWrapper>
-);
+        }}
+        state={{ columnFilters, globalFilter }}
+        onColumnFiltersChange={setColumnFilters}
+        onGlobalFilterChange={setGlobalFilter}
+        enableGlobalFilter={true}
```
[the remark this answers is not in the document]
fabio-silva: @mattiasimonato are we using global filtering for RTA? I think we only had column filtering, but please confirm

ui/apps/pmm/src/pages/rta/overview/RealtimeOverview.tsx
```diff
@@ -29,39 +29,36 @@ const RealtimeOverviewPage: FC = () => {
...
   const previousFetchingState = useRef<boolean>(fetching);
   const { data: sessions = [], isLoading } = useRealtimeSessions();
 
-  const handleQueryChange = (query: QueryData, index: number) => {
+  const selectedQueryIndex = selectedQuery
+    ? navigableQueries.findIndex(
+        (query) => query.queryId === selectedQuery.queryId
+      )
+    : -1;
+
```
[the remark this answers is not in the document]
fabio-silva: ☝️ @mattiasimonato I think users can't change filters when the pane is open, so I believe it's safe to ignore this comment

fabio-silva (changes requested):

ui/apps/pmm/src/pages/rta/overview/RealtimeOverview.tsx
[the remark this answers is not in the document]
mattiasimonato: no it doesn't, I just double checked it

ui/apps/pmm/src/pages/rta/overview/table/OverviewTable.tsx
```diff
@@ -11,54 +15,76 @@ import { filterElapsedTime } from './OverviewTable.utils';
...
+  const [globalFilter, setGlobalFilter] = useState('');
+
+  // Pre-pagination so navigation covers all filtered rows, not only the current page.
+  const getNavigableQueries = () =>
+    tableRef.current?.getPrePaginationRowModel().rows.map((row) => row.original) ??
+    queries;
+
+  useEffect(() => {
+    onNavigableQueriesChange(getNavigableQueries());
+  }, [queries, columnFilters, globalFilter, onNavigableQueriesChange]);
```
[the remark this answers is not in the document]
mattiasimonato: Yeah, I reproduced it. Fixed by controlling MRT `sorting` state, resyncing `navigableQueries` when sort changes, and refreshing the list on row click before opening the pane. Arrow navigation now follows the sorted table order.

The fix is [HERE](https://github.com/percona/pmm/pull/5397/changes/64567bbfd4a263c05470855da9dc5977a728bbb5)

ui/apps/pmm/src/pages/rta/overview/RealtimeOverview.tsx
```diff
@@ -29,39 +29,36 @@ const RealtimeOverviewPage: FC = () => {
...
-    const idx = (selectedQueryIndex || 0) - 1;
-    if (idx < 0) {
+  const handleAdjacentQuery = (offset: -1 | 1) => {
+    const nextIndex = selectedQueryIndex + offset;
+    if (nextIndex < 0 || nextIndex >= navigableQueries.length) {
       return;
     }
-    handleQueryChange(queries[idx], idx);
+    handleQuerySelected(navigableQueries[nextIndex]);
   };
```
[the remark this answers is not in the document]
mattiasimonato: While the details pane is open, the table is covered so users cannot change filters or sort. I still fixed the `-1` case for when `navigableQueries` updates without that row (e.g. refresh): next no longer jumps to the first item, and prev/next are disabled when the selection is not in the list.

fix is [HERE](https://github.com/percona/pmm/pull/5397/commits/f914530349d159fbfd701d71a006ba93dfabeff1)

ui/apps/pmm/src/pages/rta/overview/table/OverviewTable.tsx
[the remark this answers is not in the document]
mattiasimonato: @fabio-silva Thanks for flagging this. You are right, RTA should stay column-only.

Global filter got turned on by accident while I was wiring controlled filter state for the details pane navigation fix.

I reverted it to `enableGlobalFilter={false}` and removed the extra global filter state. Column filters and sorting still work, and details navigation still follows the visible rows.

The fix is [HERE](https://github.com/percona/pmm/pull/5397/commits/5aa136a8ded1add2c2267bd5e02515d3a3a74cc4)

ui/apps/pmm/src/pages/rta/overview/table/OverviewTable.utils.ts
```diff
@@ -1,6 +1,21 @@
-import { type MRT_Row } from 'material-react-table';
+import {
+  type MRT_ColumnFiltersState,
+  type MRT_Row,
+  type MRT_SortingState,
+} from 'material-react-table';
 import { QueryData } from 'types/rta.types';
```
fabio-silva, line 3: @mattiasimonato unused imports here ☝️

ui/apps/pmm/src/pages/rta/overview/table/OverviewTable.tsx
```diff
@@ -1,64 +1,133 @@
...
+    if (navigableQueryIdsKeyRef.current === nextKey) {
+      return;
+    }
+    navigableQueryIdsKeyRef.current = nextKey;
+    onNavigableQueriesChange(navigableQueries);
+  }, [getNavigableQueries, onNavigableQueriesChange]);
+
+  const handleColumnFiltersChange = useCallback(
+    (updater: MRT_ColumnFiltersState | ((old: MRT_ColumnFiltersState) => MRT_ColumnFiltersState)) => {
+      setColumnFilters((previous) => {
```
fabio-silva: @mattiasimonato is there a reason to use a custom callback instead of the state setter right away (i.e. `onColumnFiltersChange: setColumnFilters` and `onSortingChange: setSorting`), given we don't have any side-effect running inside?

By checking TanStack Table docs (the table lib that's used by material-react-table), I see they just go with the regular state setter: https://tanstack.com/table/v8/docs/guide/column-filtering#controlled-column-filter-state

fabio-silva (changes requested):

ui/apps/pmm/src/pages/rta/overview/table/OverviewTable.utils.ts
mattiasimonato: 👍

ui/apps/pmm/src/pages/rta/overview/table/OverviewTable.tsx
mattiasimonato: @fabio-silva I switched to `setColumnFilters` and `setSorting` directly like you suggested. 
Tests and build are green ✅ , thanks for pointing it out.

**Why I added this code**: the main goal was to fix prev/next in the details pane so it follows filtered/sorted rows in the table, not the full API list. The extra filter/sort wrappers were mostly to stop an infinite loop that hung the tests. Looking at it again, that part was a bit over-engineered, and plain setters are enough.

What is still there (`syncNavigableQueries`, controlled state) is for the arrow fix. The only tradeoffs I see are two lists to keep in mind and a bit more logic than passing setters only. Fine for arrows, but not the simplest setup.

fabio-silva (approved):

ui/apps/pmm/src/pages/rta/overview/table/OverviewTable.tsx
```diff
@@ -1,64 +1,108 @@
...
+    [queries]
+  );
+
+  const syncNavigableQueries = useCallback(() => {
+    const navigableQueries = getNavigableQueries();
+    const nextKey = getNavigableQueryIdsKey(navigableQueries);
+    if (navigableQueryIdsKeyRef.current === nextKey) {
+      return;
+    }
+    navigableQueryIdsKeyRef.current = nextKey;
```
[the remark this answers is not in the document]
fabio-silva: @mattiasimonato I'd probably go with option C). We can safely afford to not check query identity when the filters/sorting changes, it's not critical in terms or performance.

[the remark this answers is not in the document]
mattiasimonato: @fabio-silva Agreed on option C). I removed the id short-circuit and we always sync navigable queries now. Filter/sort updates are rare, so the extra sync is fine performance-wise, and we still use plain `setColumnFilters` / `setSorting` like you suggested earlier.

The fix is [HERE](https://github.com/percona/pmm/pull/5397/commits/a280c7a867b87ab7ef3b2b73d97799e0d418ac26)

fabio-silva (approved):

ui/apps/pmm/src/pages/rta/overview/RealtimeOverview.tsx
```diff
@@ -17,51 +17,54 @@ import Button from '@mui/material/Button';
...
-  const { data: queries = [], refetch } = useRealtimeQueries(
+  const { data: queries, refetch } = useRealtimeQueries(
     { serviceIds },
     {
       enabled: fetching,
       refetchInterval: refreshInterval,
     }
   );
-  const [selectedQueryIndex, setSelectedQueryIndex] = useState<number>();
+  const tableQueries = queries ?? EMPTY_QUERIES;
```
matejkubinec, line 34:
```suggestion
  const tableQueries: QueryData[] = queries ?? [];
```

Or is there other reason to declare `EMPTY_QUERIES`?

matejkubinec (commented):
@mattiasimonato looks good outside of the one comment, one thing, do you think it would make sense to provide this functionality on the percona-ui level? (I've used a similar pattern on the alerts page POC, and it might be good to upstream it if the pattern will be used more often)

@pmcf-percona is the next/prev table item in a details pane a pattern that would be used on more parts of the application?

ui/apps/pmm/src/pages/rta/overview/RealtimeOverview.tsx
pmcf-percona: Likely, yes. So now it's in RTA and planned for Alerts. I see it potentially coming to Inventory as well, so this panel may be restricted to always consuming lists from MRT at Peak Design level

mattiasimonato: @matejkubinec Yes, there’s a reason for `EMPTY_QUERIES`.

While queries is still undefined, `queries ?? []` creates a new empty array on every render. That was making `OverviewTable` re-sync navigable rows over and over and caused a render loop (fixed in f84b62cf0). The module-level constant keeps the same empty array reference, so that doesn’t happen

mattiasimonato: > @mattiasimonato looks good outside of the one comment, one thing, do you think it would make sense to provide this functionality on the percona-ui level? (I've used a similar pattern on the alerts page POC, and it might be good to upstream it if the pattern will be used more often)
> 
> @pmcf-percona is the next/prev table item in a details pane a pattern that would be used on more parts of the application?

@matejkubinec Yeah, I think it makes sense to upstream this if we’ll reuse it. I’d rather not block this PR on a percona-ui change though.

If you’re good with that, let’s add a follow-up ticket 👍

matejkubinec: > @matejkubinec Yes, there’s a reason for `EMPTY_QUERIES`.
> 
> While queries is still undefined, `queries ?? []` creates a new empty array on every render. That was making `OverviewTable` re-sync navigable rows over and over and caused a render loop (fixed in [f84b62c](https://github.com/percona/pmm/commit/f84b62cf09c9e83b043d18b7b480e91cc8df5ed8)). The module-level constant keeps the same empty array reference, so that doesn’t happen

Yeah, I thought it would be something like that.

matejkubinec: > > @mattiasimonato looks good outside of the one comment, one thing, do you think it would make sense to provide this functionality on the percona-ui level? (I've used a similar pattern on the alerts page POC, and it might be good to upstream it if the pattern will be used more often)
> > @pmcf-percona is the next/prev table item in a details pane a pattern that would be used on more parts of the application?
> 
> @matejkubinec Yeah, I think it makes sense to upstream this if we’ll reuse it. I’d rather not block this PR on a percona-ui change though.
> 
> If you’re good with that, let’s add a follow-up ticket 👍

Yes, let's create a follow up.

matejkubinec (approved):

PR #5400, opened 2026-05-21
PG-2338 Updated PPG latest versions to reflect Q2-2026 minor version …

Naeem-Akhter:
…release.


artemgavrilov:
As for me this check has no sense since it's bundled with PMM

Naeem-Akhter:
> As for me this check has no sense since it's bundled with PMM
@artemgavrilov 
I am not aware of whether this check is used in PMM. However, this is what I have been asked to do for the last 2 years or so to update this file for each Major or minor release. Earlier, it was done in the CHECKED repo, which has been archived. Since then, IIRC, @ademidoff asked me to update here moving forward. Maybe he can confirm whether this is needed moving forward. 

For background, here is the initial ticket where it started: https://perconadev.atlassian.net/browse/PMM-12010

ademidoff (approved):

ademidoff:
> As for me this check has no sense since it's bundled with PMM

Do you mean it makes no sense because once PMM is deployed, the version values will not be updated? 

What about users who regularly upgrade their PMM instances? They would still benefit from it, no?

artemgavrilov:
> > As for me this check has no sense since it's bundled with PMM
> 
> Do you mean it makes no sense because once PMM is deployed, the version values will not be updated?
> 
> What about users who regularly upgrade their PMM instances? They would still benefit from it, no?

Yep. I'm pretty sure that most of the users don't update PMM at the day of release. I assume for the majority of the user this check will just provide incorrect insights. If we still want to provide version checks than this should be done some other way. Now it's just bad user experience.

artemgavrilov:
@JiriCtvrtka @4nte PTAL, it should be merged or closed.

4nte (approved):

PR #5407, opened 2026-05-22
PMM-7 Dev env for MySQL backups.

JiriCtvrtka:
PMM-7
As side effect during working on https://github.com/percona/pmm/pull/5395 I created dev/mysql-backup-compose because testing MySQL physical backups requires mysqld, xtrabackup on the same host as pmm-agent. The compose setup builds a PMM Server image with MySQL and XtraBackup installed locally. This is reducing time to setup backups for dev significantly. Probably not clean and best solution, but I think it is better to have something than nothing. QA also does not have any setup and they doing everything manually every time.

Usage:
make env-up
make env-down


dev/mysql-backup-compose/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+USER root
+
+ARG MYSQL_SERVER_RPM_VERSION=8.4.6-6.1.el9
+ARG XTRABACKUP_RPM_VERSION=8.4.0-4.1.el9
+
+RUN dnf install -y epel-release https://repo.percona.com/yum/percona-release-latest.noarch.rpm \
+	&& percona-release setup ps84lts \
+	&& percona-release enable pxb-84-lts release \
+	&& percona-release enable tools release \
+	&& dnf install -y "percona-server-server-${MYSQL_SERVER_RPM_VERSION}" "percona-xtrabackup-84-${XTRABACKUP_RPM_VERSION}" qpress \
```
ademidoff, line 13: Installing PS and PXB to the PMM Server is a questionable design decision. It would be much cleaner to have a separate container for these two.

Even though I understand that this is being done for testing purposes, however this deployment doesn't seem to be close to any real-life scenario.

If we have to install the client as well to meet the deployment requirement, we can use the tarball from the feature build.

This will allow to reuse the same Docker compose file elsewhere.

JiriCtvrtka: So should I close it? I think it does not makes sense to waste time with changing how it works. 

About real life scenario. I think it is real life scenario. It does not matter that it is together with PMM server, but main is that we have PS, PXB and PMM client on same env.

PR #5408, opened 2026-05-22
PMM-7 Fix `staticcheck` and `noerrcheck` linter warnings

maxkondr:
PMM-7


maxkondr:
@copilot review

agent/agents/mongodb/profiler/internal/profiler_test.go
```diff
@@ -196,14 +207,22 @@ func testProfiler(t *testing.T, url string) {
...
+		return err
+	}
 	for _, dbname := range dbs {
 		if strings.HasPrefix(dbname, "test_") {
-			err = sess.Database(dbname).Drop(context.TODO())
-			require.NoError(t, err)
+			err = sess.Database(dbname).Drop(t.Context())
+			if err != nil {
+				t.Logf("failed to drop database %q: %v", dbname, err)
+				continue
```
ademidoff:
```suggestion
```

ademidoff (commented):
Mostly mechanical linter fixes that look correct. A few concerns inline that I'd suggest addressing (or confirming CI exercises) before merge.

api-tests/server/auth_test.go
```diff
@@ -126,7 +128,9 @@ func TestSwagger(t *testing.T) {
 
 				resp, err := http.DefaultClient.Do(req)
 				require.NoError(t, err)
-				defer resp.Body.Close() //nolint:errcheck
+				t.Cleanup(func() {
+					assert.Error(t, resp.Body.Close())
```
ademidoff: `assert.Error(t, resp.Body.Close())` looks wrong here: the body is never explicitly closed elsewhere in this test, so the cleanup is the first and only `Close()` call — which normally returns `nil`, failing this assertion. The original `defer resp.Body.Close() //nolint:errcheck` was just ignoring the close error; the equivalent linter-clean fix is `assert.NoError(...)`.

Same inverted assertion in this file at lines 162, 177, 392, 410, 440, 482, 507, 530, and also in:
- `api-tests/server/readyz_test.go:62`
- `api-tests/server/serialization_test.go:54`
- `api-tests/server/settings/settings_test.go:736, 772`

A subtler case: in `doRequest` (line 177) the cleanup is registered alongside callers that also register their own cleanup on the same `resp.Body` (e.g. line 440). Cleanups run LIFO, so one of the two `Close()` calls will succeed and the other may error — the current mix of `assert.Error`/`assert.NoError` doesn't reflect that ordering consistently.

Worth going through every `assert.Error(..., Close())` in this PR and confirming it's intentional; I suspect almost all should be `assert.NoError`.

agent/agents/supervisor/ports_registry_test.go
```diff
@@ -27,7 +28,10 @@ func TestRegistry(t *testing.T) {
 	r := newPortsRegistry(65000, 65002, []uint16{65000})
 	l1, err := net.Listen("tcp", "127.0.0.1:65001")
 	require.NoError(t, err)
-	defer l1.Close() //nolint:gosec,errcheck,nolintlint
+	t.Cleanup(func() {
+		err, _ = errors.AsType[*net.OpError](l1.Close())
```
ademidoff: This cleanup now performs a test assertion (verifying that double-close returns `*net.OpError`) rather than just cleaning up. That's beyond a linter fix and changes test semantics. It also depends on the test body having explicitly closed `l1`/`l2` first (line 56) — if a future edit removes those lines, the cleanup will fail confusingly.

If the goal is just to satisfy the linter, `_ = l1.Close()` is enough. If the double-close behavior is worth testing, consider an explicit subtest rather than putting the assertion in `t.Cleanup`.

Also note `err, _ = ...` shadows / mutates the outer-scoped `err` from line 29 — fine here but easy to misread.

agent/agents/mongodb/profiler/internal/collector/collector_test.go
```diff
@@ -64,21 +64,37 @@ func BenchmarkCollector(b *testing.B) {
...
-	cleanUpDBs(b, client) // Just in case there are old dbs with matching names
-	defer cleanUpDBs(b, client)
+	// Just in case there are old dbs with matching names
+	err = cleanUpDBs(b, client)
+	if err != nil {
+		b.Fatal(err)
+		return
+	}
+	defer func() {
+		err = cleanUpDBs(b, client)
```
ademidoff: This defer doesn't actually handle the error:
```go
defer func() {
    err = cleanUpDBs(b, client)
    if err != nil {
        return
    }
}()
```
The `if err != nil { return }` exits the deferred function but the error is silently dropped — strictly worse than the original `defer cleanUpDBs(b, client)` because it overwrites the outer `err`. Either log/Fatal the error, or use `_ = cleanUpDBs(b, client)` to be explicit about ignoring it.

Same pattern at line 138 in this file and line 155 in `agent/agents/mongodb/profiler/internal/profiler_test.go`.

agent/agents/mongodb/profiler/internal/collector/collector_test.go
```diff
@@ -64,21 +64,37 @@ func BenchmarkCollector(b *testing.B) {
 
 	client, err := createSession(url, "pmm-agent")
 	if err != nil {
+		b.Fatal(err)
 		return
```
ademidoff: The `return` after `b.Fatal(err)` is unreachable — `Fatal` calls `runtime.Goexit()`. Harmless, but `staticcheck` will likely flag this as dead code (SA4006-ish). Same at lines 75, 88, 158, 169.

agent/agents/mongodb/shared/fingerprinter/fingerprinter_test.go
```diff
@@ -156,15 +159,21 @@ func TestProfilerFingerprinter(t *testing.T) {
...
+		require.NoError(t, err)
 		profilerCollection := database.Collection("system.profile")
 		query := createQuery(dbName, time.Now().Add(-10*time.Minute))
 
 		cursor, err := createIterator(ctx, profilerCollection, query)
 		require.NoError(t, err)
 		// do not cancel cursor closing when ctx is canceled
-		defer cursor.Close(context.Background()) //nolint:errcheck
+		defer func() {
+			err = cursor.Close(t.Context())
```
ademidoff: The preceding comment (line 172) says *"do not cancel cursor closing when ctx is canceled"* — which was the whole point of the original `context.Background()`. Switching to `t.Context()` does exactly the opposite: it's cancelled when the test ends, which is when this defer runs. Either revert to `context.Background()` (and keep the comment honest) or delete the stale comment.

agent/agents/mongodb/profiler/internal/collector/collector_test.go
4nte: > strictly worse than the original defer cleanUpDBs(b, client) because it overwrites the outer err

Overwriting the outer `err` should not be a concern here since it's executed only after the function execution is finished.

 I agree we should either swallow the err if we don't care about it, or always report it to be safe. But whatever it is we should make it a rule/convention to keep the codebase consistent.

api-tests/server/auth_test.go
maxkondr: done

agent/agents/supervisor/ports_registry_test.go
maxkondr: done

agent/agents/mongodb/profiler/internal/collector/collector_test.go
maxkondr: done

agent/agents/mongodb/profiler/internal/collector/collector_test.go
maxkondr: done

agent/agents/mongodb/shared/fingerprinter/fingerprinter_test.go
maxkondr: adjusted

ademidoff (approved):

JiriCtvrtka (approved):

4nte (approved):

PR #5414, opened 2026-05-25
PMM-14931 Control built-in ClickHouse config via PMM_CLICKHOUSE_CONFIG env var

4nte:
PMM-14931 

[Feature Build](https://github.com/Percona-Lab/pmm-submodules/pull/4372)

Previously, switching the built-in ClickHouse required manually running `switch-config.sh` inside the container. 
`PMM_CLICKHOUSE_CONFIG` is a new env var that instructs supervisord to start ClickHouse with specified config.

If not set, ClickHouse will start with default config.

`switch-config.sh` has been deprecated in favor of configuring via env var. When ran, it will print out a deprecation message with link to docs.


documentation/docs/install-pmm/install-pmm-server/deployment-options/docker/env_var.md
```diff
@@ -25,6 +25,16 @@ Fine-tune data retention and collection intervals to balance monitoring detail w
...
 
+### Built-in ClickHouse configuration
+Select the configuration profile for the built-in ClickHouse instance 
+
+| Variable | Default | Description | Example |
+|----------|---------|-------------|----------|
+| `PMM_CLICKHOUSE_CONFIG` | `default` | Configuration profile for the built-in ClickHouse instance. Use `low-memory` for PMM Server environments with less than 16 GB RAM to avoid "memory limit exceeded" errors. | `low-memory` |
+
+!!! note
+    This applies only to the built-in ClickHouse instance. For details, see [ClickHouse memory issues](../../../../troubleshoot/qan_issues.md#clickhouse-memory-issues-in-low-memory-environments).
```
ademidoff: I would also mention the following:

"While this configuration attepmts to optimize the ClickHouse configuration for low-memory environments, we highly recommend you follow our configuration guide (provide a link) for the best performance and operational stability".

managed/cmd/pmm-managed-init/main.go
```diff
@@ -46,8 +52,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	err := models.ValidateSettings(envSettings)
-	if err != nil {
+	if err := models.ValidateSettings(envSettings); err != nil {
```
ademidoff: Let's not inline the "if err != nil", it facilitates refactoring :)

managed/services/clickhouse/config.go
```diff
@@ -0,0 +1,80 @@
...
+	availableConfigs, err := availableClickHouseConfigs(dir)
+	if err != nil {
+		return fmt.Errorf("unable to get available ClickHouse configs: %w", err)
+	}
+
+	path := filepath.Join(dir, config+"-config.xml")
+	if _, err := os.Stat(path); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf(
+				"invalid PMM_CLICKHOUSE_CONFIG=%q: %s not found; available configs: %v",
```
ademidoff:
```suggestion
				"invalid PMM_CLICKHOUSE_CONFIG=%s: %s not found; available configs: %v",
```
Quotes look extremely bad when escaped: `invalid PMM_CLICKHOUSE_CONFIG=\"pmm-non-existent\": /srv/config/non-existent not found; ...`

managed/cmd/pmm-managed-init/main.go
ademidoff: I added a little rule to `AGENTS.md` if you don't mind :)

maxkondr:
I have an alternative idea.
1. Create several configs for CH (default, low-memory, etc)
2. Keep switch-config.sh script that will do the following:
2.1 create symlink `/etc/clickhouse-server/config.xml` -> `/etc/clickhouse-server/config-default.xml`
2.2. run CH with `clickhouse --config-file=/etc/clickhouse-server/config.xml`

So that in general we may point:
```
/etc/clickhouse-server/config.xml to one of the available files:
 ->      /etc/clickhouse-server/config-default.xml  (by default)
         /etc/clickhouse-server/config-low-memory.xml
         /etc/clickhouse-server/config-high-memory.xml
```

So that no need to touch supervisord configuration (it always uses config.xml).
The only action required - switch symlink to the needed config and restart/reload CH.
The target config may be passed via ENV var as an option during container startup, but also allows to switch in already running container.

This approach allows switching configuration without passing ENV var with container restart.

managed/cmd/pmm-managed-init/main.go
4nte: Sure. To be fair, I inlined this myself, it's not the agent's fault 😄

4nte:
> The target config may be passed via ENV var as an option during container startup, but also allows to switch in already running container.
> 
> This approach allows switching configuration without passing ENV var with container restart.

@maxkondr I agree it's a nice solution to use symlinks the way it was previously implemented by @JiriCtvrtka. My issue is that we're now introducing a new way to control ClickHouse config through an env var, but still allow users to change the configuration in runtime meaning that `PMM_CLICKHOUSE_CONFIG` would not be single source of truth and not immutable.

For example it's possible for user to deploy PMM with `PMM_CLICKHOUSE_CONFIG=A`, then in runtime mutate the configuration with `./switch-config.sh B`, while `PMM_CLICKHOUSE_CONFIG` was set to `A`. Having two ways to achieve same thing, and one overshadowing the other.

On top of that, configuration applied through `switch-config.sh` could be wiped on container restart. This is the reason why we're introducing an env var for this in the first place.
For example, every pod restart in k8s (PMM upgrade, node drain, reschedule, etc.) would reset the change applied through `switch-config.sh`, and the configuration would silently fallback to "default" or whatever was set with `PMM_CLICKHOUSE_CONFIG`.

For these reasons I think `switch-config.sh` is better off deprecated since `PMM_CLICKHOUSE_CONFIG` achieves the same goal.

maxkondr (approved):

JiriCtvrtka:
@4nte could you check why test for config is failing? If this could be bug.

4nte:
> @4nte could you check why test for config is failing? If this could be bug.

@JiriCtvrtka  Thanks for noticing. Had to update the test to match the actual error message which was changed just recently (https://github.com/percona/pmm/pull/5414/commits/8a2479be99c0f0ceed2da1c376bf97245afb3cb3)

JiriCtvrtka (approved):

PR #5422, opened 2026-05-26
PMM-14915 Remove UI events.

JiriCtvrtka:
PMM-14915

FB: https://github.com/Percona-Lab/pmm-submodules/pull/4376


JiriCtvrtka:
@copilot review

ademidoff:
@JiriCtvrtka Let's re-generate `api/descriptor.bin`.

JiriCtvrtka:
> @JiriCtvrtka Let's re-generate `api/descriptor.bin`.

Yeah right.

api/uievents/v1/server.proto
```diff
@@ -2,9 +2,6 @@ syntax = "proto3";
 
 package uievents.v1;
 
-import "google/api/annotations.proto";
-import "protoc-gen-openapiv2/options/annotations.proto";
-
 message NotificationEvent {
```
ademidoff: 1. What about `NotificationEvent`, `FetchingEvent`, `DashboardUsageEvent`, `UserFlowEvent`, `StoreRequest`, `StoreResponse`? Should we drop them as well?

2. Another question: why would we keep the `api/uievents` specs?

3. Need to cleanup the Makefile once p.2 is done.

4. If `ExtensionType` is no longer used, let's drop it as well. This is dead code.

5. The method signature of `telemetry.NewService` needs to be updated since we no longer need to pass `extensions` to it (`nil` is being passed now).

ademidoff (changes requested):

api/uievents/v1/server.proto
JiriCtvrtka: Removed completely.

ademidoff (approved):

maxkondr (approved):

4nte (approved):

PR #5424, opened 2026-05-27
PMM-14618 Ability to disable collectors for RDS MySQL, PG technology.

JiriCtvrtka:
PMM-14618

FB: https://github.com/Percona-Lab/pmm-submodules/pull/4377


JiriCtvrtka:
@copilot review

ademidoff:
Two comments: 

1. collector names are not validated (out of scope). This needs to be mentioned both in the code and in the docs
2. If a caller sends `postgresql_disable_collectors` together with `engine = MYSQL`, the field is silently ignored (and vice versa).

@JiriCtvrtka Do you think it's acceptable or should we add minimal validation for this case?

api/management/v1/rds.proto
```diff
@@ -127,6 +127,10 @@ message AddRDSServiceParams {
   google.protobuf.Duration connection_timeout = 35 [(validate.rules).duration = {
     gte: {seconds: 0}
   }];
+  // List of collector names to disable in the MySQL database exporter.
+  repeated string mysql_disable_collectors = 36;
```
ademidoff, line 132: Pls leave a note: applies only when the corresponding engine is selected, otherwise no-op.

JiriCtvrtka:
> Two comments:
> 
> 1. collector names are not validated (out of scope). This needs to be mentioned both in the code and in the docs
> 2. If a caller sends `postgresql_disable_collectors` together with `engine = MYSQL`, the field is silently ignored (and vice versa).
> 
> @JiriCtvrtka Do you think it's acceptable or should we add minimal validation for this case?

@ademidoff 

1. It is not validated even on CLI. If we want to include it should be done for CLI and UI also together, so do we want to create ticket? Sure lets mention it in doc. I will add comment to code. 

2. Do you think it is problem? Since it has prefix for disable_collectors fields with technology? It is ignored which is same behaviour right now for example to wrong collector name, silently ignored.

4nte (approved):

api/management/v1/rds.proto
```diff
@@ -127,6 +127,12 @@ message AddRDSServiceParams {
   google.protobuf.Duration connection_timeout = 35 [(validate.rules).duration = {
     gte: {seconds: 0}
   }];
+  // List of collector names to disable in the MySQL database exporter.
+  // Applies only when engine is DISCOVER_RDS_ENGINE_MYSQL, otherwise it is ignored.
+  repeated string mysql_disable_collectors = 36;
```
maxkondr, line 132: why not to name var as `rds_mysql_disable_collectors` ?

api/management/v1/rds.proto
```diff
@@ -127,6 +127,12 @@ message AddRDSServiceParams {
...
+  // List of collector names to disable in the PostgreSQL database exporter.
+  // Applies only when engine is DISCOVER_RDS_ENGINE_POSTGRESQL, otherwise it is ignored.
+  repeated string postgresql_disable_collectors = 37;
```
maxkondr, line 135: why not to name var as `rds_postgresql_disable_collectors` ?

api/management/v1/rds.proto
JiriCtvrtka: It is inside the "rds" property, so I thought prefix would be redundant. Example:

```
"rds": {
      "region": "us-east-1",
      "az": "us-east-1b",
      "instance_id": "test-rds-postgresql",
      "node_model": "db.t3.micro",
      "address": "YOUR_RDS_POSTGRES_ENDPOINT",
      "port": 5432,
      "engine": "DISCOVER_RDS_ENGINE_POSTGRESQL",
      "pmm_agent_id": "pmm-server",
      "node_name": "test-rds-postgresql",
      "service_name": "test-rds-postgresql",
      "database": "postgres",
      "username": "DB_USER",
      "password": "DB_PASSWORD",
      "aws_access_key": "AWS_ACCESS_KEY",
      "aws_secret_key": "AWS_SECRET_KEY",
      "rds_exporter": true,
      "skip_connection_check": true,
      "postgresql_disable_collectors": [
        "stat_database",
        "stat_bgwriter"
      ]
    }
```

What do you think?

ademidoff: IMO it's reads OK since it's inside `rds.proto` already.

ademidoff:
> > Two comments:
> > 
> > 1. collector names are not validated (out of scope). This needs to be mentioned both in the code and in the docs
> > 2. If a caller sends `postgresql_disable_collectors` together with `engine = MYSQL`, the field is silently ignored (and vice versa).
> > 
> > @JiriCtvrtka Do you think it's acceptable or should we add minimal validation for this case?
> 
> @ademidoff
> 
> 1. It is not validated even on CLI. If we want to include it should be done for CLI and UI also together, so do we want to create ticket? Sure lets mention it in doc. I will add comment to code.
> 2. Do you think it is problem? Since it has prefix for disable_collectors fields with technology? It is ignored which is same behaviour right now for example to wrong collector name, silently ignored.

Reasonable, but if we want to be friendly, a lightweight warning log or a validation error would surface user mistakes earlier. This PR affects the API which is used by many, so letting them know of an unintentional mistake could be worth it.

ademidoff (approved):

JiriCtvrtka:
> > > Two comments:
> > > 
> > > 1. collector names are not validated (out of scope). This needs to be mentioned both in the code and in the docs
> > > 2. If a caller sends `postgresql_disable_collectors` together with `engine = MYSQL`, the field is silently ignored (and vice versa).
> > > 
> > > @JiriCtvrtka Do you think it's acceptable or should we add minimal validation for this case?
> > 
> > 
> > @ademidoff
> > 
> > 1. It is not validated even on CLI. If we want to include it should be done for CLI and UI also together, so do we want to create ticket? Sure lets mention it in doc. I will add comment to code.
> > 2. Do you think it is problem? Since it has prefix for disable_collectors fields with technology? It is ignored which is same behaviour right now for example to wrong collector name, silently ignored.
> 
> Reasonable, but if we want to be friendly, a lightweight warning log or a validation error would surface user mistakes earlier. This PR affects the API which is used by many, so letting them know of an unintentional mistake could be worth it.

Warn messages added.

ademidoff (approved):

maxkondr (approved):

PR #5426, opened 2026-05-27
PMM-14971:  Avoid duplicate navigation sidebar

mattiasimonato:
PMM-14971

Link to the Feature Build: SUBMODULES-4383

Fixes duplicate PMM sidebar and cases where sidebar/header stayed visible while Grafana was logged out or `localStorage` was cleared.

The PMM shell and Grafana iframe could get out of sync: the iframe sometimes loaded `/pmm-ui/...` (second full app + second sidebar), and the parent could still treat the user as logged in after Grafana navigated to login or after same-tab storage clear (no storage event).

- Keep sidebar drawer in the layout tree (`disablePortal`)
- Treat login as server session (`rotateToken`) plus client flag `pmm-ui.session.active`; react to `localStorage.clear` in the same tab
- Reset iframe when it hits `/pmm-ui`; remount on `src` change; load iframe only when logged in on a Grafana route
- On Grafana login URL, clear client session and invalidate `rotateToken`
- Hide sidebar and header when `isLoggedIn` is false


ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -63,8 +67,8 @@ export const GrafanaProvider: FC<PropsWithChildren> = ({ children }) => {
   const { colorMode, setFromGrafana } = useColorMode();
 
   useEffect(() => {
-    if (isGrafanaPage) setIsLoaded(true);
-  }, [isGrafanaPage]);
+    setIsLoaded(isGrafanaPage && isLoggedIn);
```
matejkubinec: Will this handle anonymous mode (a grafana iframe can be loaded when user is not logged in yet)

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
```diff
@@ -103,6 +107,10 @@ export const GrafanaProvider: FC<PropsWithChildren> = ({ children }) => {
           return;
         }
 
+        if (isGrafanaLoginPath(location.pathname)) {
```
matejkubinec, line 111: Will we receive message if the user has logged out (or was automatically logged out)?

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
mattiasimonato: You're right! With only `isLoggedIn` the iframe would never load in anonymous mode. I updated it so the iframe loads when you're on a Grafana page and either logged in or anonymous mode is enabled.

The fix is [HERE](https://github.com/percona/pmm/pull/5426/changes/24acdb39b0e3adb216bf2dc9539a8ef208fce8c2)

ui/apps/pmm/src/contexts/grafana/grafana.provider.tsx
mattiasimonato: When Grafana sends you to the login page after logout (manual or automatic), the code inside the Grafana iframe tells the PMM app about the new URL. We pick that up and sync the shell. `GrafanaPage` also checks the iframe on load, just in case that message is missed. There isn’t a separate “logged out” signal, we watch for the login page, same idea as Sign out.

matejkubinec (approved):

ui/apps/pmm/src/components/main/MainWithNav.tsx
```diff
@@ -10,9 +11,16 @@ import { SHOW_UPDATE_INFO_DELAY_MS } from 'lib/constants';
 import { isRenderingServer } from '@pmm/shared';
 import Header from './header/Header';
 
+const useMainNavVisible = () => {
```
ademidoff, line 14:
```suggestion
const useMainNavVisible: boolean = () => {
```

nit: I'm surprised how little TypeScript we use. Ok, it's inference anyway, but still could be better )

ademidoff (approved):

PR #5434, opened 2026-05-28
PMM-7 fix staticcheck linter warnings

maxkondr:
PMM-7

NOTES:
1. There are changes in establishing gRPC client connection. The deprecated `grpc.DialContext` + `grpc.WithBlock()` are replaced with `grpc.NewClient` + `grpc.WaitForReady(true)`.
2. OpenAPI client's transport doesn't support context anymore. The context is passed as a parameter to each API call instead.
3. Single-row queries in `qan-api2/models/metrics.go` are simplified to use `QueryRowxContext` instead of `QueryxContext`, removing unnecessary row iteration and adding explicit `sql.ErrNoRows` handling.
4. Test improvements: `t.Parallel()` moved from the `setup` helper to individual test functions; `t.Cleanup` used in place of `defer` for teardown; teardown signature simplified by capturing `t` in closure.


maxkondr:
@copilot review

managed/services/agents/channel/channel_test.go
```diff
@@ -181,7 +184,7 @@ func TestServerRequest(t *testing.T) {
 	}
 
 	stream, _, teardown := setup(t, connect, io.EOF) // EOF = server exits from handler
-	defer teardown(t)
+	t.Cleanup(teardown)
```
ademidoff: Non-blocking, but this is a real failure. In this file `setup` derives the stream context from `t.Context()` (line 79), so switching `defer teardown(t)` → `t.Cleanup(teardown)` breaks `TestServerRequest`: `t.Context()` is canceled *just before* `t.Cleanup` functions run, so by the time `teardown` calls `channel.Wait()` it sees `"context canceled"` instead of the expected `io.EOF`. It reproduces deterministically (6/6) when the package runs as a whole; passes in isolation.

Reverting just this test to `defer teardown()` makes the package pass 5/5. Either keep `defer teardown()` here (teardown is now `func()`), or derive the context from `context.Background()` like `agent/client/channel/channel_test.go` does — that's why the identical change is safe there.

Side note: the `codes.Canceled`-expecting tests in this file still pass, but now vacuously — the `t.Context()` cancellation produces exactly the error they assert.

qan-api2/models/metrics.go
```diff
@@ -1085,20 +1085,19 @@ func (m *Metrics) SchemaByQueryID(ctx context.Context, serviceID, queryID string
...
 
 	res := &qanv1.SchemaByQueryIDResponse{}
-	for rows.Next() {
-		err = rows.Scan(&res.Schema)
-		if err != nil {
-			return res, errors.Wrap(err, "failed to scan query")
+	err = row.Scan(&res.Schema)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, errors.New("query_id doesnt exists")
```
ademidoff: Non-blocking — please confirm this is intended. It flips `SchemaByQueryID`'s not-found contract: previously zero rows returned `(empty response, nil)`; now it returns an error. That surfaces as a gRPC error in `managed/.../actions.go` (the EXPLAIN action fails) for the narrow case where a query has explain-fingerprint data but no schema row — a fully-missing `query_id` already errors earlier in `ExplainFingerprintByQueryID`.

ademidoff (commented):
Two non-blocking notes from review — one real test failure, one behavior change to confirm. The rest (gRPC migration, `transport.Context` removal, `NginxError` simplification, test refactors) looks good.

qan-api2/models/metrics.go
ademidoff: @JiriCtvrtka I have the impression this may be problematic. Could you please check as well?

JiriCtvrtka: Like Alex pointed out, for no rows we should keep the current behavior, which is to return an empty response and no error. It is used in MySQL Explain, and even with a schema that is not found, we still want to run Explain.

managed/services/agents/channel/channel_test.go
maxkondr: done

qan-api2/models/metrics.go
maxkondr: fixed

maxkondr:
@copilot review

JiriCtvrtka (approved):

4nte (approved):

ademidoff (approved):

PR #5435, opened 2026-05-28
PMM-15039 Disable telemetry except for GA and RC builds

4nte:
PMM-15039

Feature Build: https://github.com/Percona-Lab/pmm-submodules/pull/4381

Disables telemetry by default with intent to prevent non-official builds (e.g forks) and dev environments from reporting telemetry.

Docker image now accepts `PMM_ENABLE_TELEMETRY` build arg which enables our build pipeline to explicitly ENABLE telemetry on GA and RC builds.

Check out related PR in our build pipeline: https://github.com/Percona-Lab/jenkins-pipelines/pull/4133


managed/services/telemetry/telemetry.go
```diff
@@ -159,12 +158,6 @@ func (s *Service) Run(ctx context.Context) {
 			return
 		}
 
-		p, err := version.Parse(s.pmmVersion)
-		// do not send telemetry if this is a feature build, match only clean release versions like "3.7.1"
```
4nte, line 163: This check is not necessary anymore and could be misleading future readers to think that this is how we control telemetry behavior.

Telemetry is now disabled by default everywhere and can be explicitly enabled via env var at startup or build step, like in our Jenkins pipeline for RC and GA builds.

build/docker/server/Dockerfile.el9
```diff
@@ -2,10 +2,12 @@ FROM oraclelinux:9-slim
 
 ARG VERSION
 ARG BUILD_DATE
+ARG PMM_ENABLE_TELEMETRY=0
 
 ENV LANG=en_US.utf8
 ENV GF_PLUGIN_DIR=/srv/grafana/plugins
 ENV PERCONA_TELEMETRY_DISABLE=1
+ENV PMM_ENABLE_TELEMETRY=${PMM_ENABLE_TELEMETRY}
```
ademidoff, line 10: 👍

managed/models/settings.go
```diff
@@ -26,7 +26,7 @@ import (
 const (
 	AdvisorsEnabledDefault             = true
 	AlertingEnabledDefault             = true
-	TelemetryEnabledDefault            = true
+	TelemetryEnabledDefault            = false
```
ademidoff, line 29: @4nte We don't want to do that actually :) Let's discuss.

managed/services/telemetry/telemetry.go
ademidoff: I don't think so. For example, should you still pass `-e PMM_ENABLE_TELEMETRY=1` to `docker run...` on a dev image, we would still get the telemetry data.

managed/services/telemetry/telemetry_test.go
```diff
@@ -186,68 +190,6 @@ func TestRunTelemetryService(t *testing.T) {
 	}
 }
 
-func TestRunSkipsNonReleaseVersion(t *testing.T) {
```
ademidoff, line 189: Let's keep this test, I still find it useful.

ademidoff (changes requested):

managed/services/telemetry/telemetry.go
4nte: Should we really ignore `PMM_ENABLE_TELEMETRY=1` in dev?

Since user needs to explicitly enable telemetry, I assume they must have a valid reason for providing `PMM_ENABLE_TELEMETRY=1` and would expect for telemetry to be really enabled, when set.


To be fair, there could still be a case where someone unintentionally enables telemetry in development, e.g their AI runs PMM with all env vars set and enables telemetry for some reason or they are being careless about it.
If the goal is to prevent telemetry pollution from misconfigured local dev environments I would propose that we also inject telemetry endpoint/address in our build step in tandem with current env var:


For example, in our build pipeline, official builds (RC/GA) could build with:
```
PMM_ENABLE_TELEMETRY=1
PMM_DEV_PERCONA_PLATFORM_ADDRESS="https://check.percona.com"

```

And also replace the hardcoded default address with a value that does not point to our prod telemetry
 ```go
defaultPlatformAddress = "https://devcheck.localhost"
```

This way we don't block users/devs from playing with telemetry in local dev, but we also prevent them from reporting to our prod if they just enable `PMM_ENABLE_TELEMETRY=1` by mistake.


What do you think?

ademidoff (approved):

JiriCtvrtka (approved):

maxkondr (approved):

PR #5439, opened 2026-05-31
PMM-7 Fix API tests

ademidoff:
PMM-7

We observe the following errors when running the API tests:

```
  00:08:22.000  === NAME  TestAgents/FilterList
  00:08:22.000      agents_test.go:185:
  00:08:22.000              Error Trace:    /go/pmm/api-tests/inventory/agents_test.go:185
  00:08:22.000              Error:          Received unexpected error:
  00:08:22.000                              [GET /v1/inventory/agents][404] ListAgents default {"code":5,"message":"Service with ID \"97a27633-7ab6-4295-9f10-664aeee5d761\" not
  found.","details":[]}
  00:08:22.000              Test:           TestAgents/FilterList/exit
```

and 

```
  00:09:23.669  === NAME  TestExternalService/Basic
  00:09:23.669      services_test.go:1222:
  00:09:23.669              Error Trace:    /go/pmm/api-tests/inventory/services_test.go:1222
  00:09:23.669              Error:          Should be empty, but was [0xc000998e60 0xc000998f00 0xc000998fa0]
  00:09:23.669              Test:           TestExternalService/Basic
```

Link to the Feature Build: SUBMODULES-4382 ✅

### DIFF
--- a/api-tests/inventory/agents_test.go
+++ b/api-tests/inventory/agents_test.go
@@ -174,7 +174,7 @@ func TestAgents(t *testing.T) {
 		assertPMMAgentNotExists(t, res, pmmAgentID)
 		assertNodeExporterNotExists(t, res, nodeExporterID)
 
-		// Filter by agent type, scoped to this test's pmm-agent to avoid the TOCTOU 404
+		// Filter by agent type, scoped to this test's pmm-agent to avoid the 404
 		// race an unscoped type filter hits (see the List subtest).
 		res, err = client.Default.AgentsService.ListAgents(
 			&agents.ListAgentsParams{
--- a/api-tests/inventory/agents_test.go
+++ b/api-tests/inventory/agents_test.go
@@ -174,11 +174,13 @@ func TestAgents(t *testing.T) {
 		assertPMMAgentNotExists(t, res, pmmAgentID)
 		assertNodeExporterNotExists(t, res, nodeExporterID)
 
-		// Filter by service ID.
+		// Filter by agent type, scoped to this test's pmm-agent to avoid the TOCTOU 404
+		// race an unscoped type filter hits (see the List subtest).
 		res, err = client.Default.AgentsService.ListAgents(
 			&agents.ListAgentsParams{
-				AgentType: new(types.AgentTypeMySQLdExporter),
-				Context:   pmmapitests.Context,
+				PMMAgentID: new(pmmAgentID),
+				AgentType:  new(types.AgentTypeMySQLdExporter),
+				Context:    pmmapitests.Context,
 			},
 		)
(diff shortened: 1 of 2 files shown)


api-tests/inventory/agents_test.go
```diff
@@ -174,11 +174,13 @@ func TestAgents(t *testing.T) {
...
 		assertNodeExporterNotExists(t, res, nodeExporterID)
 
-		// Filter by service ID.
+		// Filter by agent type, scoped to this test's pmm-agent to avoid the TOCTOU 404
+		// race an unscoped type filter hits (see the List subtest).
 		res, err = client.Default.AgentsService.ListAgents(
 			&agents.ListAgentsParams{
-				AgentType: new(types.AgentTypeMySQLdExporter),
-				Context:   pmmapitests.Context,
+				PMMAgentID: new(pmmAgentID),
```
maxkondr, line 181: filter by PMM Agent ID shall not influence on result in general - I mean the particular MySQLExporter shall be in response anyway. 
So what does it fix?

ademidoff: It's a fair question. Nevertheless, we are dealing with quite a tricky race condition here. Let me remind that several tests run in parallel.


The `client.Default.AgentsService.ListAgents` (agents_test.go:179) uses a filter by agent type only - `AgentType: MySQLdExporter`. On the server side, `AgentsService.List` returns every `mysqld_exporter` in the DB, not only the services directly related to the agent used in the scope of the test ( `managed/services/inventory/agents.go:97-109`):
```
  agents, err := models.FindAgents(tx.Querier, filters)   // 1 SELECT: all matching agents
  ...
  for i, a := range agents {
      res[i], err = toInventoryAgent(tx.Querier, a, as.r) // N more SELECTs, one per agent
      ...
  }
```

For each `mysqld_exporter`, `toInventoryAgent → ToAPIAgent` resolves that agent's service — `managed/services/converters.go:243-248`:
```
  if agent.ServiceID != nil {
      service, err := models.FindServiceByID(q, *agent.ServiceID)
      if err != nil {
          return nil, err   // codes.NotFound: "Service with ID %q not found."
      }
      ...
  }
```

That `FindServiceByID` is the source of error - `managed/models/service_helpers.go:201`.

Therefore, when a parallel test deletes its own service between the query and the conversion loop, the call fails with a `404 Service with ID … not found`.

Given the above, I consider this fix to be valid. The good news is there is nothing wrong with the `List` method, it's the test that needs a fix.

maxkondr (approved):

4nte (approved):

JiriCtvrtka (approved):

PR #5448, opened 2026-06-02
PMM-14671 Randomize QAN delivery timing within collection interval

JiriCtvrtka:
Ticket number: PMM-14671

Feature build: SUBMODULES-0

Adjusted scope based on review feedback: metric collection timing remains unchanged, and only QANCollectRequest delivery is randomized within the interval. This preserves existing QAN bucket timestamps/periods while distributing request delivery to reduce server-side request bursts.

If this PR adds, removes or alters one or more API endpoints, please review and update the relevant [API documentation](https://github.com/percona/pmm/tree/v3/docs/api) as well:

- [ ] API Docs updated

- Links to related work items (optional).


JiriCtvrtka:
@copilot review

JiriCtvrtka:
@copilot review

agent/agents/mongodb/mongolog/internal/mongolog.go
```diff
@@ -128,6 +129,8 @@ func (l *Mongolog) Start(ctx context.Context) error {
 	logsPathWithPrefix := path.Join(l.logFilePrefix, logsPath)
 	reader, err := filereader.NewContinuousFileReader(logsPathWithPrefix, l.logger)
 	if err != nil {
+		l.sender.Stop()
```
ademidoff: Is it is a bugfix?

ademidoff (commented):
I think we raised the right problem, but didn't validate the proposed solution.

It matters little when the collection takes place. "min:00" was quite a good approach to reason about and test. However, the sending part was causing a storm of requests and potentially even network congestion server-side. That's the real problem we want to solve.

Therefore, IMO we should change the goal from

`Metric collection and delivery should be evenly distributed or randomised within the collection interval` 

to

`Metric collection should stay the same but the delivery should be evenly distributed or randomised within the collection interval`.

That would be a much smaller change to review and test.

agent/agents/mongodb/mongolog/internal/mongolog.go
ademidoff: Btw, I'd ship the memory leak fixes asap, so they bring value in 3.8.1.

JiriCtvrtka: Yes I believe it should be fixed.

JiriCtvrtka:
> I think we raised the right problem, but didn't validate the proposed solution.
> 
> It matters little when the collection takes place. "min:00" was quite a good approach to reason about and test. However, the sending part was causing a storm of requests and potentially even network congestion server-side. That's the real problem we want to solve.
> 
> Therefore, IMO we should change the goal from
> 
> `Metric collection and delivery should be evenly distributed or randomised within the collection interval`
> 
> to
> 
> `Metric collection should stay the same but the delivery should be evenly distributed or randomised within the collection interval`.
> 
> That would be a much smaller change to review and test.

Fair point, I agree. I changed the approach so collection stays unchanged and randomization happens only before sending QANCollectRequest.

This keeps QAN bucket timestamps/periods the same while spreading delivery to PMM Server across the interval. The trade-off is that collection load on a single pmm-agent can still spike if many services collect at the same time.

JiriCtvrtka:
@copilot review

agent/agents/mongodb/mongolog/internal/mongolog.go
JiriCtvrtka: Leak fix PR: https://github.com/percona/pmm/pull/5458

JiriCtvrtka:
@copilot review

JiriCtvrtka:
@copilot review

JiriCtvrtka:
After discussion with @4nte, he will create a ticket related to server/CH changes.

4nte:
@JiriCtvrtka Do you think we should make the QAN delivery interval configurable via an env var?

The current implementation will benefit large deployments that suffer from spikes by essentially delaying the delivery but at cost of increased latency for QAN data.

What if we let users lower it (e.g. 60s to 20s) to trade some server load for fresher QAN data, do you think this is something users would care for?

JiriCtvrtka:
> @JiriCtvrtka Do you think we should make the QAN delivery interval configurable via an env var?
> 
> The current implementation will benefit large deployments that suffer from spikes by essentially delaying the delivery but at cost of increased latency for QAN data.
> 
> What if we let users lower it (e.g. 60s to 20s) to trade some server load for fresher QAN data, do you think this is something users would care for?

@4nte this is not possible at this moment. PGSM needs to have 1 minute otherwise it will desync. Not sure about other sources right now from top of my head.

agent/agents/supervisor/supervisor.go
```diff
@@ -688,6 +688,18 @@ func (s *Supervisor) startBuiltin(agentID string, builtinAgent *agentv1.SetState
 	go pprof.Do(ctx, pprof.Labels("agentID", agentID, "type", agentType), agent.Run)
 
 	go func() {
+		qanDeliveryRequests := make(chan *agentv1.QANCollectRequest, qanRequestsBufferSize)
+		qanDeliveryDone := make(chan struct{})
+		go func() {
```
maxkondr, line 693: this func `startBuiltin` is common for QAN and RTA agents. This particular go-routine is created unconditionally for all agent types 9even when it is not required (RTA case) .


one more point - waiting for each such nested go-routine finish is performed via done channel pattern, isn't it better to have one workGroup instead channel per each go-routine?

agent/agents/supervisor/supervisor.go
```diff
@@ -741,6 +756,54 @@ func (s *Supervisor) startBuiltin(agentID string, builtinAgent *agentv1.SetState
 	return nil
 }
 
+func (s *Supervisor) deliverQANRequests(ctx context.Context, l *logrus.Entry, agentID string, requests <-chan *agentv1.QANCollectRequest) {
```
maxkondr, line 759: there is an alternative via rateLimiter:
```
package main

import (
	"context"
	"log"
	"time"

	"golang.org/x/time/rate"
)

// Allow exactly 10 requests per second, with a maximum burst of 1.
// This enforces a strict, even flow to the backend.
var backendLimiter = rate.NewLimiter(rate.Every(100*time.Millisecond), 1)

func HandleRequest(ctx context.Context, payload string) error {
	// 1. We wait for permission to proceed. 
	// If the rate is exceeded, this Goroutine goes to sleep (is parked)
	// and wakes up exactly when it's allowed to hit the backend.
	if err := backendLimiter.Wait(ctx); err != nil {
		// This error ONLY triggers if the context expires or is canceled 
		// before a token becomes available.
		return err 
	}

	// 2. We are cleared to go. Hit the backing service.
	return callBackingService(payload)
}

func callBackingService(payload string) error {
	// Simulate backend work
	return nil
}
```

you may define a time distribution factor in `..rate.Every(100*time.Millisecond)...`. It looks less complex

agent/agents/supervisor/supervisor.go
```diff
@@ -688,6 +688,18 @@ func (s *Supervisor) startBuiltin(agentID string, builtinAgent *agentv1.SetState
 	go pprof.Do(ctx, pprof.Labels("agentID", agentID, "type", agentType), agent.Run)
 
 	go func() {
+		qanDeliveryRequests := make(chan *agentv1.QANCollectRequest, qanRequestsBufferSize)
```
maxkondr, line 691: `qanRequestsBufferSize` == only 100 items. It may become full very quickly if we postpone sending QAN requests to network on really loaded systems. And it results into the following:
1. this go-routine will stop reading from `agent.Changes()`
2. but this `agent.Changes()` is used for delivering agent's status changes as well.
3. agent's status delivery will be postponed till QAN requests are processed

It looks like there is a need in moving agent's status change channel into a separate high priority channel that is not blocked by the rest channels.

agent/agents/supervisor/supervisor.go
```diff
@@ -741,6 +756,54 @@ func (s *Supervisor) startBuiltin(agentID string, builtinAgent *agentv1.SetState
...
 
+func (s *Supervisor) deliverQANRequests(ctx context.Context, l *logrus.Entry, agentID string, requests <-chan *agentv1.QANCollectRequest) {
+	var qanDeliveryOffset time.Duration
+	qanDeliveryOffsetAssigned := false
+	var releaseQANDeliveryOffset func()
+
+	for request := range requests {
+		if !qanDeliveryOffsetAssigned {
+			var offset time.Duration
+			offset, releaseQANDeliveryOffset = agents.RandomMinuteOffset(agentID)
```
maxkondr, line 767: the problem with this approach - it is applied even when there is only 1 QAN agent because this logic is placed on Agent side that knows nothing about number of QAN agents registered on PMM server.

It still may create local spikes (yes, they will be not so critical in comparing to heavy-loaded systems) because it only shifts sending the whole collected/buffered QAN requests collection from 0-s second of each minute to N-s second of each minute.

PR #5462, opened 2026-06-07
PMM-7 Re-license  PMM "utils/"

ademidoff:
**Why**

PMM Client must remain fully open source under a permissive license. The client code (`agent/`, `admin/`) is already licensed under Apache 2.0, but it depends on several packages in `utils/`, which until now were licensed under AGPL-3.0:

- utils/sqlrows
- utils/logger
- utils/tlsconfig
- utils/sqlmetrics
- utils/pdeathsig
- utils/nodeinfo
- utils/enums
- utils/errors

Shipping an Apache-2.0 client that links AGPL-3.0 dependencies creates a licensing mismatch: the copyleft (AGPL) terms propagate into the otherwise permissively-licensed client, undermining the "fully open source / permissive" guarantee and creating friction for downstream consumers.

This PR re-licenses utils/ to Apache 2.0 so that all dependencies of PMM Client share a single, consistent permissive license. As Percona LLC is the sole copyright holder of this code, we are free to relicense it.

**Changes**

- Replaced the AGPL-3.0 header with the Apache 2.0 header in all 22 `utils/**/*.go` files (copyright line preserved).
- Added utils/LICENSE (Apache 2.0), matching the existing agent/LICENSE and admin/LICENSE.
- Updated `.licenserc.yaml`: moved `utils/**` into the Apache-2.0 header group and excluded it from the AGPL group, mirroring how `agent/` and `admin/` are configured.

The repository root LICENSE remains AGPL-3.0 — it governs PMM Server overall; this change only scopes the shared `utils/` packages used by the client.

**Verification**

`make check-license` passes with 0 invalid files across both the AGPL and Apache header groups.

### DIFF
--- a/utils/depstests/depstests.go
+++ b/utils/depstests/depstests.go
@@ -1,17 +1,16 @@
 // Copyright (C) 2023 Percona LLC
 //
-// This program is free software: you can redistribute it and/or modify
-// it under the terms of the GNU Affero General Public License as published by
-// the Free Software Foundation, either version 3 of the License, or
-// (at your option) any later version.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
 //
-// This program is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-// GNU Affero General Public License for more details.
+//	http://www.apache.org/licenses/LICENSE-2.0
 //
-// You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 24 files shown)


JiriCtvrtka (approved):

utils/depstests/depstests.go
```diff
@@ -1,17 +1,16 @@
 // Copyright (C) 2023 Percona LLC
```
maxkondr, line 1: if we change a license - shall year be changed?

maxkondr (approved):

utils/depstests/depstests.go
ademidoff: No, for two reasons:
1. Too much noise - we would have to be consistent and bump up the rest
2. The license year does not have to be this year, since we own the code since long

So, we retroactively changed the license of "utils/", which is good for community.

PR #5484, opened 2026-06-10
PMM-15143 Fix resource leaks in managed

ademidoff:
Ticket number: PMM-15143

Feature build: SUBMODULES-4401


managed/services/ha/services.go
```diff
@@ -74,13 +74,13 @@ func (s *services) StartAllServices(ctx context.Context) {
 	for id, service := range s.all {
 		if _, ok := s.running[id]; !ok {
 			s.running[id] = service
+			s.wg.Add(1)
```
ademidoff: This is the most important fix here )

JiriCtvrtka (approved):

managed/services/checks/checks.go
```diff
@@ -94,6 +94,11 @@ type Service struct {
 	startDelay      time.Duration
 	customCheckFile string // For testing
 
+	runCtxM sync.Mutex
+	// runCtx is the service lifecycle context recorded by Run. It bounds
+	// asynchronous work started via StartChecks so it is cancelled on shutdown.
+	runCtx context.Context //nolint:containedctx
```
maxkondr: this is anti-pattern. https://pkg.go.dev/context
```
Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.
```

managed/services/checks/checks.go
```diff
@@ -275,8 +286,11 @@ func (s *Service) StartChecks(checkNames []string) error {
 		return services.ErrAdvisorsDisabled
 	}
 
+	s.runCtxM.Lock()
+	ctx := s.runCtx
```
maxkondr: no need to pass upper-level context here, much simpler to implement via `on demand` channel that is read in `runChecksLoop` func

managed/services/checks/checks.go
```diff
@@ -1650,10 +1664,16 @@ func (s *Service) updateAdvisors(advisors []check.Advisor) {
 // UpdateIntervals updates advisor checks restart timer intervals.
 func (s *Service) UpdateIntervals(rare, standard, frequent time.Duration) {
 	s.tm.Lock()
+	defer s.tm.Unlock()
+	// Tickers are created by Run; if it has not started on this node (e.g. not
+	// the leader), there is nothing to reset - Run reads the new intervals from
+	// the persisted settings when it starts.
+	if s.rareTicker == nil || s.standardTicker == nil || s.frequentTicker == nil {
```
maxkondr, line 1667: shouldn't the update request be sent to the current leader node only?

managed/services/ha/services.go
maxkondr, line 90: 1. no need to create 2 loops, it is possible to spin up within the same loop
2. with using wg.Go() would avoid the issue that is fixed here now

maxkondr: looks like no need to create a copy, it is possible to stop within the same loop because `service.Stop()` just calls `context.Cancel()` - it is fast enough and doesn't block

maxkondr: this is conceptually wrong. `s.wg.Done()` must be called at the end of go-routine that is part of `sync.waitGroup` otherwise it leads to unpredictable results and go-routines may keep running even after `wg.Wait()` has returned.

maxkondr: removing from `s.running` shall be called when the service is really stopped but not when there is an intention to stop it.

managed/services/ha/services.go
```diff
@@ -122,7 +122,12 @@ func (s *services) Wait() {
 // removeService removes a service from the registry of running services.
 func (s *services) removeService(id string) {
```
maxkondr: the logic of this function shall be only in one place - at the ent of go-routine in line 90. This function is not needed at all because it creates a dangerous situation with `sync.WaitGroup` counter

managed/services/checks/checks.go
ademidoff: Good point, but routing the interval reset to the current leader needs cross-node coordination (the node handling the settings change is not necessarily the leader running the tickers), which is a pre-existing gap beyond the scope of this resource-leak PR.

The actual defect here was the nil-ticker panic when `UpdateIntervals` runs on a non-leader (where `Run` never created the tickers); the guard fixes that. On the leader, `Run` reads the intervals from the persisted settings when it starts, so the change is not lost. I'd prefer to track the leader-routing improvement separately.

managed/services/checks/checks.go
ademidoff: Addressed in a5dda11a - dropped the stored `runCtx` field and the `//nolint:containedctx`. `StartChecks` now hands the request to `runChecksLoop`, which runs it with the context it already owns.

managed/services/checks/checks.go
ademidoff: Addressed in a5dda11a - done exactly as suggested: `StartChecks` does a non-blocking send on an on-demand channel that `runChecksLoop` drains.

managed/services/ha/services.go
ademidoff: Addressed in 10dc37d6 - single loop now, and `wg.Go` owns the Add/Done so the double-Done this PR worked around can no longer occur.

ademidoff: Addressed in 10dc37d6 - no copy; `StopAllServices` iterates `s.running` and calls `Stop()` (a fast, non-blocking cancel) directly.

ademidoff: Addressed in 10dc37d6 - `Done` is now handled by `wg.Go` at the end of each service goroutine, not in `StopAllServices`.

ademidoff: Addressed in 10dc37d6 - the service is removed from `s.running` by its own goroutine, only after `Start` returns.

managed/services/ha/services.go
ademidoff: Addressed in 10dc37d6 - `removeService` is gone; the cleanup now lives only at the end of the service goroutine.

4nte (approved):

maxkondr (approved):

JiriCtvrtka (approved):

PR #5490, opened 2026-06-11
PMM-14576 Backup improvements.

JiriCtvrtka:
PMM-14576

FB: https://github.com/Percona-Lab/pmm-submodules/pull/4431

Fixes false failed to get backup status / failed to get restore status errors during MongoDB PBM backup and restore jobs when describe-backup / describe-restore temporarily fail while the operation is still running or metadata is not ready yet.

Refactors backup/restore completion polling into a shared describePoller that:
1. treats known PBM transient errors as retryable (no such file, file is empty, get backup meta: not found, etc.)
2. keeps waiting while pbm status reports the operation as running
3. falls back to pbm status snapshots (backup) or pbm list --restore (restore) when describe keeps failing after completion
4. preserves the full retry budget until the operation has finished


JiriCtvrtka:
@copilot review

JiriCtvrtka:
@copilot review

JiriCtvrtka:
@copilot review

JiriCtvrtka:
@copilot review

agent/runner/jobs/pbm_helpers.go
maxkondr, line 1: can it be implemented easier?
here is the native poller implementation (I would place it into `utils/`) and just call it with appropriate values:
```go
import (
	"context"
	"errors"
	"fmt"
	"time"
)

// ConditionFunc returns true when the condition is successfully met.
type ConditionFunc func(ctx context.Context) (done bool, err error)

// PollUntilContextTimeout native standard library equivalent
func PollUntilContextTimeout(ctx context.Context, interval time.Duration, condition ConditionFunc) error {
	// 1. Run an immediate initial check before starting the ticker.
	// If the condition is already true, we skip waiting.
	if done, err := condition(ctx); err != nil {
		return err
	} else if done {
		return nil
	}

	// 2. Initialize a ticker for subsequent intervals.
	ticker := time.NewTicker(interval)
	defer ticker.Stop()

	for {
		select {
		case <-ctx.Done():
			// The context timeout expired or was canceled by the caller
			return ctx.Err()

		case <-ticker.C:
			// Trigger the condition check on every tick
			done, err := condition(ctx)
			if err != nil {
				return err
			}
			if done {
				return nil
			}
		}
	}
}
```

JiriCtvrtka:
@copilot review

JiriCtvrtka:
@copilot review

JiriCtvrtka:
@copilot review

agent/runner/jobs/mongodb_restore_job.go
```diff
@@ -211,36 +212,38 @@ func (j *MongoDBRestoreJob) startRestore(ctx context.Context, backupName string)
...
-
-	ticker := time.NewTicker(statusCheckInterval)
-	defer ticker.Stop()
 	retryCount := 500
+	started := false
 
-	for {
-		select {
-		case <-ticker.C:
+	err := poll.UntilContextTimeout(ctx, statusCheckInterval, func(ctx context.Context) (bool, error) {
```
maxkondr:
```suggestion
	pollErr := poll.UntilContextTimeout(ctx, statusCheckInterval, func(ctx context.Context) (bool, error) {
```
just to avoid misleading

agent/runner/jobs/pbm_helpers.go
```diff
@@ -254,42 +265,303 @@ func getPBMStatus(ctx context.Context, dsn string) (*pbmStatus, error) {
...
-			case "done":
-				return nil
-			case "canceled":
-				return errors.New("backup was canceled")
-			case "error":
-				return errors.New(info.Error)
-			}
+func waitDescribe(ctx context.Context, cfg *describePoller) error {
+	return poll.UntilContextTimeout(ctx, cfg.interval(), func(ctx context.Context) (bool, error) {
+		err := ctx.Err()
```
maxkondr: move it to `pollDescribeOnce`

agent/runner/jobs/pbm_helpers.go
```diff
@@ -254,42 +265,303 @@ func getPBMStatus(ctx context.Context, dsn string) (*pbmStatus, error) {
...
+		if err != nil {
+			return false, err
+		}
+
+		done, err := pollDescribeOnce(ctx, cfg)
+		if err != nil {
+			return false, err
+		}
+		return done, nil
```
maxkondr:
```suggestion
		return pollDescribeOnce(ctx, cfg)
```

agent/runner/jobs/pbm_helpers.go
JiriCtvrtka: Removed ctx check as we discussed.

JiriCtvrtka:
@copilot review

maxkondr (approved):

4nte (approved):

agent/runner/jobs/pbm_helpers.go
```diff
@@ -213,23 +234,13 @@ func retrieveLogs(ctx context.Context, dsn string, event string) ([]pbmLogEntry,
...
-			status, err := getPBMStatus(ctx, dsn)
-			if err != nil {
-				return err
-			}
-			if status.Running.Type == "" {
-				return nil
-			}
-		case <-ctx.Done():
-			return ctx.Err()
+	return poll.UntilContextTimeout(ctx, statusCheckInterval, func(ctx context.Context) (bool, error) {
```
ademidoff, line 239: **The status is checked immediately, so this wait can finish too early after a force-resync.**

`poll.UntilContextTimeout` runs the check function one time right away (at t=0), before the first tick. The old `ticker` loop waited 5s (`statusCheckInterval`) before the first `getPBMStatus` call.

This function runs right after `pbmConfigure(forceResync=true)`, and `pbm config --force-resync` starts the resync in the background. So when we check immediately, the resync may not be registered yet in `status.Running.Type`. In that case the check sees `Running.Type == ""`, returns `nil`, and the job continues (to `findCurrentSnapshot` / restore) before the resync has finished. The old 5s delay was hiding this race.

Suggestion: skip the first immediate check here (the same way `startRestore` does with its `started` flag), or let the helper wait one interval before the first call.

agent/runner/jobs/pbm_helpers.go
```diff
@@ -254,44 +265,296 @@ func getPBMStatus(ctx context.Context, dsn string) (*pbmStatus, error) {
...
+func pollDescribeOnce(ctx context.Context, cfg *describePoller) (bool, error) {
+	info, describeErr := cfg.fetchDescribe(ctx)
+	if describeErr == nil {
+		cfg.retries = maxDescribeRetries
+		return checkDescribe(info, cfg.operation)
+	}
+
+	status, statusErr := cfg.getPBMStatus(ctx)
+	if statusErr != nil {
+		return false, fmt.Errorf("failed to get pbm status: %w", statusErr)
```
ademidoff: **A temporary `pbm status` error now stops the whole wait.**

When `describe` fails, the new code calls `getPBMStatus`, and if that call also fails it returns the error immediately and stops waiting. The old `waitForPBMBackup` never looked at `pbm status`; it simply retried a transient `describe` error up to 500 times.

The problem: during the normal startup window `describe-backup` returns the expected transient error ("no such file"), and at the same time a `pbm status` call can also fail for a short moment (for example a replica-set election or a command timeout). Now this single temporary failure ends the whole wait, and the job reports "failed to wait backup completion", even though the backup is still healthy.

Suggestion: treat a `pbm status` error here as retryable (keep polling) instead of returning it right away.

agent/runner/jobs/pbm_helpers.go
```diff
@@ -254,44 +265,296 @@ func getPBMStatus(ctx context.Context, dsn string) (*pbmStatus, error) {
...
+	i := slices.IndexFunc(list, func(r pbmListRestore) bool {
+		return r.Name == name
+	})
+	if i < 0 {
+		return nil
+	}
+	return &list[i]
+}
+
+func checkDescribe(info describeInfo, operation string) (bool, error) {
```
ademidoff, line 540: **Backups now fail immediately on `partlyDone`, which was not the case before.**

`checkDescribe` sends `partlyDone` to `groupDescribeErrs` for every operation, including backup. In the old `waitForPBMBackup` there was no `partlyDone` case, so a `partlyDone` backup did not fail: the loop kept polling until the status became `done` / `error` or the context timed out.

So this changes backup behavior: a backup reported as `partlyDone` now fails the job right away. This may be the intended fix (failing fast with node-level details is better than hanging until the timeout), but please confirm it is on purpose, because it is a new error path for backups.

agent/runner/jobs/pbm_helpers.go
```diff
@@ -254,44 +265,296 @@ func getPBMStatus(ctx context.Context, dsn string) (*pbmStatus, error) {
...
+}
+
+func (cfg *describePoller) opRunning(status *pbmStatus) bool {
+	if cfg.isRunning != nil {
+		return cfg.isRunning(status)
+	}
+	if cfg.operation != pbmCmdBackup && cfg.operation != pbmCmdRestore {
+		return false
+	}
+	return status.Running.Type == cfg.operation && status.Running.Name == cfg.name
```
ademidoff, line 507: **`opRunning` may not detect a running restore because of the name check.**

`opRunning` requires `status.Running.Name == cfg.name`. For restores, the name we pass (from `restoreInfo.Name` or the PITR list name) is not guaranteed to match PBM's `status.running.name`.

If they do not match while a restore is actually running, `opRunning` returns `false`, `trackFinished` sets `finishedAt` too early, and the code moves into the completion-grace / retry path. The `list --restore` fallback usually keeps the poll alive, but it means a running restore is then limited by the 5-minute `describeCompletionGrace` instead of the unlimited "still running" path.

Suggestion: please confirm that the restore name really equals `Running.Name` across PBM versions. If it does not, match on the type only, or on the operation id.

agent/runner/jobs/pbm_helpers.go
```diff
@@ -320,32 +583,45 @@ func findPITRRestoreName(ctx context.Context, dsn string, restoreInfo *pbmRestor
...
-					return "", errors.New("failed to start restore")
-				}
-				continue
-			} else {
-				return entry.Name, nil
-			}
-		case <-ctx.Done():
-			return "", ctx.Err()
+	err = poll.UntilContextTimeout(ctx, statusCheckInterval, func(ctx context.Context) (bool, error) {
+		err = ctx.Err()
```
ademidoff: **Redundant `ctx.Err()` check and reuse of the outer `err`.**

`poll.UntilContextTimeout` already checks `ctx.Err()` before the first call and on every tick, so the `err = ctx.Err()` line here is not needed.

Also, the closure writes to the function-level `err` variable (`err = ctx.Err()`, `err = execPBMCommand(...)`). This is left over from the old loop and is easy to misuse: if someone later reads `err` between the poll call and the `return`, they may read an old per-tick value.

Suggestion: remove the `ctx.Err()` check and use a local variable (for example `cmdErr`), the same way `startRestore` does.

agent/runner/jobs/pbm_helpers.go
JiriCtvrtka: partlyDone is a normal terminal state for PBM restore, but not a typical backup end state.
So failing fast on partlyDone is generally better (clearer, no hanging), but for backups it mostly protects an edge case rather than a common PBM flow.

What do you think?

JiriCtvrtka:
@copilot review

PR #5494, opened 2026-06-11
PMM-15129 fix depguard linter issues (p1)

maxkondr:
Ticket number: PMM-15129

https://github.com/Percona-Lab/pmm-submodules/pull/4407


sna0978117341:
Hi

ademidoff:
@maxkondr Can we address Copilot review comments?

maxkondr:
> @maxkondr Can we address Copilot review comments?

done

managed/services/minio/client.go
```diff
@@ -87,7 +88,7 @@ func (c *Client) RemoveRecursive(ctx context.Context, endpoint, accessKey, secre
 		}
 		for object := range mc.ListObjects(ctx, bucketName, options) {
 			if object.Err != nil {
-				return errors.WithStack(object.Err)
+				return fmt.Errorf("failed to list objects in bucket %q: %w", bucketName, object.Err)
```
ademidoff:
```suggestion
				return fmt.Errorf("failed to list objects in bucket %s: %w", bucketName, object.Err)
```

I suggest we don't use double quotes (`%q`), since they make it difficult to grep the logs (they show up as `\"something\"`).

managed/services/minio/client.go
```diff
@@ -151,7 +152,7 @@ func (c *Client) List(ctx context.Context, endpoint, accessKey, secretKey, bucke
 
 	for object := range mc.ListObjects(ctx, bucketName, options) {
 		if object.Err != nil {
-			return nil, errors.WithStack(object.Err)
+			return nil, fmt.Errorf("failed to list objects in bucket %q: %w", bucketName, object.Err)
```
ademidoff:
```suggestion
			return nil, fmt.Errorf("failed to list objects in bucket %s: %w", bucketName, object.Err)
```

managed/services/minio/client.go
maxkondr: we have a plenty of places with the following pattern:
```
fmt.Errorf("'%s' is not a supported data model for MongoDB backups", dataModel)
```

that produces a quoted error string (single quote) as well.

ademidoff: Yes, that why plain "%s" should be used.

maxkondr: > Yes, that why plain "%s" should be used.

this will produce `"<value>"` strings

ademidoff (approved):

JiriCtvrtka (approved):

PR #5499, opened 2026-06-15
PMM-14661 Fix QAN crashes, make inserts asynchronous

ademidoff:
Ticket number: PMM-14661

Feature build: SUBMODULES-4406

## Summary

Reduce QAN read-path scan cost and stabilize ingestion without changing the schema or result semantics. A larger structural change (a pre-aggregated rollup) seems necessary, but will be reconsidered separately.

## Changes

- **Single-scan filter dimensions** ([reporter.go](qan-api2/models/reporter.go)) — `queryDimension` uses `sumIf(<metric>, <active-dimension predicate>)` in one scan, replacing the filtered + "enumerate-with-0" `UNION ALL` (two scans → one per dimension), roughly halving the filters-panel scans.
- **Asynchronous inserts** ([default-users.xml](build/ansible/roles/clickhouse/files/default-users.xml), [low-memory-users.xml](build/ansible/roles/clickhouse/files/low-memory-users.xml)) — enable ClickHouse `async_insert` on the default profile so metric inserts are buffered server-side and flushed in batches, cutting per-insert overhead and write contention on the QAN tables. We rely on the default `wait_for_async_insert=1`, per ClickHouse's strong recommendation: the `INSERT` still blocks until the server flushes, so commit errors are surfaced synchronously (keeping the existing rollback/retry path in [data_ingestion.go](qan-api2/models/data_ingestion.go) meaningful) and the server can apply backpressure instead of being overrun. `wait_for_async_insert=0` would silence those errors and defeat backpressure, so it is deliberately not used.
- **Ingest buffer left at 100** — the in-process request queue (`requestsCap` in [data_ingestion.go](qan-api2/models/data_ingestion.go)) is intentionally unchanged. It is the backpressure/decoupling mechanism between the gRPC `Save` handler and the single DB-writer goroutine, not the batching layer, so `async_insert` does not make it redundant. With `wait_for_async_insert=1`, slower commits correctly chain into agent-side backpressure; enlarging the queue would only mask that and increase in-memory data loss on restart. Tune via ClickHouse-side settings (e.g. `async_insert_busy_timeout_ms`) if the `qan_api2_data_ingestion_requests_len` gauge rides near the cap under load.

## Considered and dropped

- **Parallel sparklines** — an earlier revision fetched per-row sparklines concurrently (a bounded `errgroup`) to cut report latency. Reverted: running several sparkline aggregations at once multiplies peak ClickHouse memory during a single report, which works against the OOM/crash fix this PR targets — most acutely on the low-memory profile. It also forced a larger shared connection pool (read load could starve the ingest writer). Sparklines remain serial (one query per row), and `maxOpenConns`/`maxIdleConns` stay at their baseline.

### DIFF
--- a/build/ansible/roles/clickhouse/files/default-users.xml
+++ b/build/ansible/roles/clickhouse/files/default-users.xml
@@ -6,6 +6,7 @@
         <!-- Default settings. -->
         <default>
             <async_insert>1</async_insert>
+            <wait_for_async_insert>1</wait_for_async_insert>
         </default>
 
         <!-- Profile that allows only read queries. -->
--- a/build/ansible/roles/clickhouse/files/default-users.xml
+++ b/build/ansible/roles/clickhouse/files/default-users.xml
@@ -5,6 +5,7 @@
     <profiles>
         <!-- Default settings. -->
         <default>
+            <async_insert>1</async_insert>
         </default>
 
         <!-- Profile that allows only read queries. -->
--- a/build/ansible/roles/clickhouse/files/low-memory-users.xml
+++ b/build/ansible/roles/clickhouse/files/low-memory-users.xml
@@ -6,6 +6,7 @@
         <!-- Default settings. -->
         <default>
             <async_insert>1</async_insert>
+            <wait_for_async_insert>1</wait_for_async_insert>
             <!-- PMM: Low-memory tuning: block size, download threads, parallel parsing/formatting -->
             <max_block_size>8192</max_block_size> <!-- PMM: Before 65409, now 8192 rows -->
             <max_download_threads>1</max_download_threads> <!-- PMM: Default no limit (0), now 1-->
--- a/build/ansible/roles/clickhouse/files/low-memory-users.xml
+++ b/build/ansible/roles/clickhouse/files/low-memory-users.xml
@@ -5,6 +5,7 @@
     <profiles>
         <!-- Default settings. -->
         <default>
+            <async_insert>1</async_insert>
             <!-- PMM: Low-memory tuning: block size, download threads, parallel parsing/formatting -->
             <max_block_size>8192</max_block_size> <!-- PMM: Before 65409, now 8192 rows -->
             <max_download_threads>1</max_download_threads> <!-- PMM: Default no limit (0), now 1-->
--- a/qan-api2/models/base.go
+++ b/qan-api2/models/base.go
@@ -25,12 +25,6 @@ import (
 
 const queryTimeout = 30 * time.Second
 
-// MaxParallelQueries bounds the number of ClickHouse queries a single request may
-// run concurrently. It is kept well below the connection pool size (see maxOpenConns
-// in main.go) so one request cannot monopolize the pool — which is shared with the
-// data ingestion writer — or flood ClickHouse with concurrent scans.
-const MaxParallelQueries = 4
-
 var sparklinePointAllFields = []string{
 	"point",
 	"timestamp",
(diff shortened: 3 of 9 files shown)


build/ansible/roles/clickhouse/files/default-users.xml
```diff
@@ -5,6 +5,7 @@
     <profiles>
         <!-- Default settings. -->
         <default>
+            <async_insert>1</async_insert>
```
JiriCtvrtka, line 8: I think we should set <wait_for_async_insert>1</wait_for_async_insert> explicitly. The PR relies on the default for rollback/retry semantics in ingestion, but better to pin it in config so a future CH default change can't silently break that.

build/ansible/roles/clickhouse/files/low-memory-users.xml
```diff
@@ -5,6 +5,7 @@
     <profiles>
         <!-- Default settings. -->
         <default>
+            <async_insert>1</async_insert>
```
JiriCtvrtka, line 8: async_insert buffers data in memory before flush, which may work against the low-memory tuning below (max_block_size, disabled parallel parsing).

qan-api2/models/base.go
```diff
@@ -25,6 +25,12 @@ import (
 
 const queryTimeout = 30 * time.Second
 
+// MaxParallelQueries bounds the number of ClickHouse queries a single request may
+// run concurrently. It is kept well below the connection pool size (see maxOpenConns
+// in main.go) so one request cannot monopolize the pool — which is shared with the
+// data ingestion writer — or flood ClickHouse with concurrent scans.
+const MaxParallelQueries = 4
```
JiriCtvrtka: With this is maxOpenConns enough? It is 10 and shared with the ingest writer, report Select(), filter queries, and up to 4 parallel sparklines per report. I believe in this case few concurrent QAN users could exhaust the pool and time out.

build/ansible/roles/clickhouse/files/low-memory-users.xml
4nte: Parallel parsing seems to be disabled for async inserts, so there is nothing we can do about it anyways: https://github.com/ClickHouse/ClickHouse/blob/f90551824bb90ade2d8a1d8edd7b0a3c0a459617/src/Interpreters/AsynchronousInsertQueue.cpp#L789-L790

Looking at the docs, my understanding is that `max_block_size` should have no effect on inserts or async buffering at all:
> The max_block_size setting indicates the recommended maximum number of rows to include in a single block when loading data from tables.

if we want to limit the buffer size used by async inserts, we should do it via `async_insert_max_data_size` which defaults to 100MiB.

4nte (approved):

build/ansible/roles/clickhouse/files/default-users.xml
ademidoff: Done.

qan-api2/models/base.go
ademidoff: Good catch — this is moot now. I've removed the sparkline parallelization (and `MaxParallelQueries`) in 1434e4387.

Running several sparkline aggregations concurrently multiplies peak ClickHouse memory per report, which works against the OOM/crash fix this PR targets — most acutely on the low-memory profile — and it coupled read concurrency to the shared pool (a burst of reports could starve the single ingest writer). Sparklines are serial again and `maxOpenConns`/`maxIdleConns` stay at baseline (10/5), so the pool-exhaustion concern no longer applies.

The report-latency win will be revisited separately, likely via a pre-aggregated rollup rather than more concurrency.

JiriCtvrtka (approved):

PR #5522, opened 2026-06-22
PMM-15129 Fix unparam linter issues

maxkondr:
Ticket number: PMM-15129

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/4422

- fixed logic in `managed/services/alerting/service.go`.  Before this PR, if the templates dir cannot be created (e.g. permissions/disk on /srv) the error was just logged and the service was still constructed and returned. Now, `pmm-managed` aborts at startup instead of starting in a degraded state.


managed/services/alerting/service.go
```diff
@@ -70,12 +70,13 @@ type Service struct {
...
 
 // NewService creates a new Service.
-func NewService(db *reform.DB, grafanaClient grafanaClient) (*Service, error) { //nolint:unparam
+func NewService(db *reform.DB, grafanaClient grafanaClient) (*Service, error) {
 	l := logrus.WithField("component", "management/alerting")
 
 	err := dir.CreateDataDir(userTemplatesDir, dirPerm)
 	if err != nil {
 		l.Error(err)
+		return nil, err
```
ademidoff, line 79: **Behavior change: a previously-tolerated error is now a fatal startup failure.**

Before this PR, a `dir.CreateDataDir` failure in `NewService` was logged (`l.Error(err)`) and the service was still constructed and returned. With `return nil, err` here, the sole caller in `managed/cmd/pmm-managed/main.go` does `l.Fatalf("Could not create alerting service: %s", err)`, so if the templates dir cannot be created (e.g. permissions/disk on `/srv`), `pmm-managed` now aborts at startup instead of starting in a degraded state.

This is reasonable hardening, but it is a real failure-mode change riding in a `unparam` linter-fix PR — just flagging to confirm it is intended.

qan-api2/services/receiver/receiver.go
```diff
@@ -42,12 +42,9 @@ func NewService(mbm *models.MetricsBucket) *Service {
 }
 
 // Collect implements rpc to store data collected from slowlog/perf schema etc.
-func (s *Service) Collect(ctx context.Context, req *qanpb.CollectRequest) (*qanpb.CollectResponse, error) {
+func (s *Service) Collect(ctx context.Context, req *qanpb.CollectRequest) (*qanpb.CollectResponse, error) { //nolint:unparam
```
ademidoff, line 45: All params are being used here, do we need `nolint:unparam`?

Makefile.include
```diff
@@ -48,7 +48,7 @@ api-test:             ## Run API tests on dev env.
 GOLANG_CI_LINT_RUN_OPTS ?=
 check:                ## Run required checks and linters
 	go tool buf lint -v api
-	LOG_LEVEL=error bin/golangci-lint run -c=.golangci.yml --new-from-rev=$(shell git merge-base v3 HEAD) --new $(GOLANG_CI_LINT_RUN_OPTS)
+	LOG_LEVEL=error bin/golangci-lint run -c=.golangci.yml --new-from-rev=$(shell git merge-base main HEAD) --new $(GOLANG_CI_LINT_RUN_OPTS)
```
ademidoff, line 51: 👍

ademidoff (approved):

managed/services/alerting/service.go
ademidoff: I suggest we highlight such changes to the reviewers going forward, since those could change how the app behaves. In this particular case it's a good one, but it could have been a breaking change in a different scenario.

agent/agents/mongodb/shared/sender/sender.go
```diff
@@ -89,7 +89,7 @@ func (s *Sender) Stop() {
 	s.wg.Wait()
 }
 
-func start(ctx context.Context, wg *sync.WaitGroup, reportChan <-chan *report.Report, w Writer, logger *logrus.Entry, doneChan <-chan struct{}) {
+func start(_ context.Context, wg *sync.WaitGroup, reportChan <-chan *report.Report, w Writer, logger *logrus.Entry, doneChan <-chan struct{}) {
```
JiriCtvrtka, line 92: Since it is used only in one place maybe would be better to remove context param completely.

JiriCtvrtka (approved):

qan-api2/services/receiver/receiver.go
maxkondr: it complains 
```
❯ ./bin/golangci-lint run -c .golangci.yml --enable-only=unparam
qan-api2/services/receiver/receiver.go:45:100: (*Service).Collect - result 1 (error) is always nil (unparam)
func (s *Service) Collect(ctx context.Context, req *qanpb.CollectRequest) (*qanpb.CollectResponse, error) {
```

ademidoff: OK then )

agent/agents/mongodb/shared/sender/sender.go
maxkondr: I think better to make it use context, but this out of scope

PR #5526, opened 2026-06-22
PMM-7 Introduce PR branch auto update

maxkondr:
Ticket number: PMM-7

This workflow automatically updates PR branch once its base branch is updated (new commit is merged).
To make it work it is required to set `auto-update-branch` label on PR.
There will be max 10 attempts to update PR branch.

### DIFF
--- a/.github/workflows/auto-update-base.yaml
+++ b/.github/workflows/auto-update-base.yaml
@@ -134,7 +134,7 @@ jobs:
             }
 
             if (errorEncountered) {
-              core.setFailed(`[ERROR] One or more PR base updates failed. Check logs for details.`);
+              core.setFailed('One or more PR branch updates failed. Check logs for details.');
             } else if (warningEncountered) {
               core.warning('Max attempts reached for one or more PRs; some mergeability states remained unknown. See workflow logs for details.');
             } else {
--- a/.github/workflows/auto-update-base.yaml
+++ b/.github/workflows/auto-update-base.yaml
@@ -1,8 +1,8 @@
 ---
-# This workflow automatically updates PR's base branch once it is updated
-# (new commit is merged).
+# This workflow automatically updates a PR branch with the latest changes from its base branch
+# (i.e. performs the equivalent of the GitHub UI "Update branch" action).
 # To make it work it is required to set "auto-update-base" label on PR.
-# There will be max 10 attempts to update base branch.
+# There will be max 10 attempts to update branches.
 name: Auto-Update PR Base
 
 on:
--- a/.github/workflows/auto-update-base.yaml
+++ b/.github/workflows/auto-update-base.yaml
@@ -11,6 +11,10 @@ on:
       - main
       - pmm-*
 
+permissions:
+  contents: write
+  pull-requests: write
+
 jobs:
   auto-update-base:
     runs-on: ubuntu-latest
--- a/.github/workflows/auto-update-base.yaml
+++ b/.github/workflows/auto-update-base.yaml
@@ -1,0 +1,118 @@
+---
+# This workflow automatically updates PR's base branch once it is updated
+# (new commit is merged).
+# To make it work it is required to set "auto-update-base" label on PR.
+# There will be max 1- attempts to update base branch.
+name: Auto-Update PR Base
+
+on:
+  push:
+    branches:
+      - main
+      - pmm-*
+
+jobs:
+  auto-update-base:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          github-token: ${{ secrets.ROBOT_TOKEN || secrets.GITHUB_TOKEN }}
+          script: |
+            let retriesNeeded = true;
+            let attemptCount = 0;
+            const maxAttempts = 10;  // Set a limit for the maximum number of attempts
+            let errorEncountered = false; // Track if any errors occur
+
+            while (retriesNeeded && attemptCount < maxAttempts) {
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-update-base.yaml
```diff
@@ -0,0 +1,137 @@
...
+
+                if (pendingPRs.length > 0 && attemptCount < maxAttempts) {
+                  console.log(`[INFO] Waiting 5 seconds before retrying ${pendingPRs.length} PR(s) (Attempt ${attemptCount}/${maxAttempts})...`);
+                  await new Promise(resolve => setTimeout(resolve, 5000));
+                }
+              }
+
+              if (pendingPRs.length > 0) {
+                console.error(`[ERROR] Max attempts reached. ${pendingPRs.length} PR(s) still have 'unknown' mergeable state.`);
+                errorEncountered = true;
```
ademidoff: Marking the whole run as failed here is arguably too noisy. A PR that's still in `unknown`/computing mergeability after 10 attempts isn't really an actionable failure of this job — it's a transient GitHub state, not something broken in the workflow or the PR. Setting `errorEncountered = true` makes the run go red and is indistinguishable from a genuine error (e.g. a non-409/422 API failure in the loop above).

Consider downgrading this case to a warning (e.g. `core.warning(...)` and leaving `errorEncountered` untouched), so a red run reliably signals a real problem and exhausted-retry timeouts don't generate false alarms.

---

ademidoff (commented):
Thanks for this — overall a clean, well-structured workflow. A couple of things worth a look before merging:

### Main thing to verify: `mergeable_state === "behind"` depends on branch protection

```js
if (pr.mergeable_state !== "behind") {
  // skip
}
```

GitHub only reports `behind` when the **"Require branches to be up to date before merging"** branch-protection rule is enabled on the base branch. If that setting is off on `main` / `pmm-*`, an out-of-date PR reports `clean`/`unstable`/`blocked` instead, and this workflow will skip it on every push — effectively becoming a silent no-op.

So the feature's effectiveness is coupled to that protection setting being enabled on each target branch. Could we confirm that's the case, and ideally note the dependency in the workflow header comment so future maintainers understand why nothing happens if the setting is later removed?

### Minor: exhausted-retry case marks the run as failed (noisy)

See the inline comment — a PR stuck in `unknown` mergeability after 10 attempts is a transient GitHub state, not an actionable failure, but it currently turns the whole run red and is indistinguishable from a genuine error.

### Other minor notes (non-blocking)

- `cancel-in-progress: true` + the up-to-~50s retry loop means rapid successive pushes to `main` can cancel an in-flight run mid-loop, potentially leaving some labeled PRs un-updated until the next push. Acceptable since `updateBranch` is a single fire-and-forget call with no partial-state risk — just noting it.
- `mergeable_state` is an unofficial/undocumented API field. It works and is widely used, but isn't formally guaranteed by GitHub.

The only thing I'd actually gate on is confirming the "require up-to-date branches" protection — otherwise the rest is good to go.

---

.github/workflows/auto-update-base.yaml
maxkondr: done

ademidoff: Thanks for the change — the warning approach is exactly right. One catch though: the new fix has a JavaScript syntax error that will break the whole workflow:

```js
} else if (warningEncountered) {
  core.warning(`[WARN] `;)
}
```

The `;` is inside the call's argument list, so `core.warning(`[WARN] `;)` is invalid JS. `actions/github-script` evaluates the `script:` body, so this throws a `SyntaxError` and the step fails on every run. It should be something like:

```js
} else if (warningEncountered) {
  core.warning(`Max attempts reached; some PR(s) still have 'unknown' mergeable state.`);
}
```

(Also worth giving the message real content rather than an empty `[WARN] ` so the annotation is useful.)

---

maxkondr:
> GitHub only reports `behind` when the **"Require branches to be up to date before merging"** branch-protection rule is enabled on the base branch. If that setting is off on `main` / `pmm-*`, an out-of-date PR reports `clean`/`unstable`/`blocked` instead, and this workflow will skip it on every push — effectively becoming a silent no-op.
> 

`main` branch has such rule, but for some reason `pmm-*` branches lack it

ademidoff:
Thanks for checking. Since `pmm-*` branches lack the "Require branches to be up to date before merging" rule, `mergeable_state` will never be `"behind"` for PRs targeting them — so as written this workflow will silently skip every `pmm-*` PR and only actually do anything for `main`.

Two ways to close that gap:

1. **Enable the same protection rule on the `pmm-*` branches** so behavior is consistent with `main`. Simplest, and keeps the workflow logic unchanged.
2. **Don't depend on the `"behind"` state at all** — drop the `mergeable_state !== "behind"` gate and just call `updateBranch` for every non-conflicting PR, treating the `422 "already up to date"` response (already handled by the 409/422 catch) as a no-op. This makes the workflow self-sufficient regardless of per-branch protection settings.

I'd lean toward (2) since it removes the hidden coupling to branch-protection config, but either works. Not a blocker for merging as long as the intended scope is just `main` for now.

---

maxkondr:
> Thanks for checking. Since `pmm-*` branches lack the "Require branches to be up to date before merging" rule, `mergeable_state` will never be `"behind"` for PRs targeting them — so as written this workflow will silently skip every `pmm-*` PR and only actually do anything for `main`.
> 
> Two ways to close that gap:
> 
> 1. **Enable the same protection rule on the `pmm-*` branches** so behavior is consistent with `main`. Simplest, and keeps the workflow logic unchanged.

done this way for `main` and `pmm-*` branches

4nte (approved):

JiriCtvrtka (approved):

PR #5259, opened 2026-04-16
PMM-14911 Alerts page

matejkubinec:
PMM-14911

BE: https://github.com/percona/pmm/pull/5527

SUBMODULES-4444

PR #5527, opened 2026-06-23
PMM-14911 Alerts page - BE

JiriCtvrtka:
Ticket number: PMM-14911

FE: http://github.com/percona/pmm/pull/5259
FB: https://github.com/Percona-Lab/pmm-submodules/pull/4302


JiriCtvrtka:
@copilot review

4nte (commented):
https://github.com/percona/pmm/pull/5527/changes/5b01d7f94ee9cb0a109efc5657cddd27e1dc4849
These (previously single expression alert templates) are chosen as good candidates to be converted to multiple-expression because they are more readable in the multiple-expression format.

Just for the record, they're technically not required to be a multiple-expression.

api/alerting/v1/alerting.proto
```diff
@@ -88,6 +106,12 @@ message Template {
   google.protobuf.Timestamp created_at = 10;
   // YAML template file content. Empty for built-in and SaaS templates.
   string yaml = 11;
+  // PromQL query steps for multi-query templates. Empty for single-expression templates.
+  repeated TemplateQuery queries = 12;
+  // Server-side expression steps for multi-query templates. Empty for single-expression templates.
+  repeated TemplateExpression expressions = 13;
+  // Reference ID of the step used as the alert condition (e.g. "C"). Empty for single-expression templates.
+  string condition = 14;
```
4nte, line 114: Template message now exposes fields for a multiple-expression variant. 
They are empty for a single expression.

api-tests/alerting/multiexpr_filter_test.go
```diff
@@ -0,0 +1,179 @@
...
+		rt = basicAuthRoundTripper{user: u.Username(), pass: pass, base: rt}
+	}
+	promHTTP, err := api.NewClient(api.Config{
+		Address:      pmmapitests.BaseURL.ResolveReference(&url.URL{Path: "/prometheus"}).String(),
+		RoundTripper: rt,
+	})
+	require.NoError(t, err)
+	promClient := promapi.NewAPI(promHTTP)
+
+	result, _, err := promClient.Query(context.Background(), exprB, time.Now())
```
ademidoff:
```suggestion
	result, _, err := promClient.Query(t.Context(), exprB, time.Now())
```

managed/services/alerting/rule_builder.go
```diff
@@ -0,0 +1,204 @@
...
+	}, nil
+}
+
+func newMathExpressionData(refID, expression string) (services.Data, error) {
+	model, err := json.Marshal(mathExpressionModel{
+		Type:       expressionTypeMath,
+		Expression: expression,
+		RefID:      refID,
+		Datasource: map[string]string{
+			"type": grafanaExprDatasourceUID,
```
ademidoff, line 173: This reads a little bit oddly as a type. Maybe leave a comment?

managed/services/alerting/rule_builder.go
```diff
@@ -0,0 +1,204 @@
...
+	filledExpr, err := fillExprWithParams(expr, params)
+	if err != nil {
+		return "", err
+	}
+
+	for _, filter := range filters {
+		switch filter.Type {
+		case alertingv1.FilterType_FILTER_TYPE_MATCH:
+			// Preserve series that don't carry the label (e.g. constant/threshold queries)
+			filledExpr = fmt.Sprintf(`label_match(%s, "%s", "(%s)|")`, filledExpr, filter.Label, filter.Regexp)
```
ademidoff, line 130: The trailing empty alternative `"(%s)|"` doesn't only rescue constant/threshold queries — it also lets through any series where `node_name` (or whatever label) is **absent or empty**. So a `MATCH` filter on a real metric query now admits label-less series that were previously excluded. In the shipped templates this is masked because their `by(node_name,...)` guarantees the label, but a user template filtering a metric that sometimes lacks the label would get unexpected alert instances.

This reads as an unintended side effect of the constant-query fix rather than a deliberate loosening of `MATCH`. Can we scope the fix so it doesn't change `MATCH` semantics for ordinary metric queries (e.g. only apply the empty-alternative to constant/threshold legs), or at minimum document the widening here? Note `MISMATCH` was correctly left untouched (an absent label already survives `label_mismatch`).

managed/data/alerting-templates/mongodb_pbm_agent_down.yml
```diff
@@ -3,7 +3,7 @@ templates:
   - name: pmm_mongodb_pbm_agent_down
     version: 1
     summary: PBM agent down
-    expr: 'avg by (node_name) (mongodb_pbm_agent_status{role=~"P|S|H",self="1"}) > 0'
+    expr: 'avg by (node_name,service_name) (mongodb_pbm_agent_status{role=~"P|S|H",self="1"}) > bool 0'
```
ademidoff, line 6: Switching `> 0` to `> bool 0` changes the query from "return only breaching series (raw value)" to "return every series as 0/1". The set of firing series should be equivalent (non-zero ⇒ fire), but the value surfaced in `{{ $value }}` / state history changes for this alert. This looks incidental rather than intended — was it a deliberate, QA-verified change? Same question applies to the `> bool [[ .threshold ]]` edits in `managed/data/alerting-templates/mongodb_pbm_backup_duration.yml` and `managed/data/alerting-templates/mongodb_pbm_backup_size.yml`.

managed/data/alerting-templates/mongodb_pbm_backup_duration.yml
```diff
@@ -4,8 +4,8 @@ templates:
     version: 1
     summary: PBM backup duration
     expr: |-
-      (max by (name,node_name) (mongodb_pbm_backup_duration_seconds{status="done"}))
-      > [[ .threshold ]]
+      (max by (name,node_name,service_name) (mongodb_pbm_backup_duration_seconds{status="done"}))
+      > bool [[ .threshold ]]
```
ademidoff, line 8: Same `> bool` behavior-change question as in `managed/data/alerting-templates/mongodb_pbm_agent_down.yml:6` — the value reported for this alert changes to 0/1. Please confirm this is intended.

managed/data/alerting-templates/mongodb_pbm_backup_size.yml
```diff
@@ -4,8 +4,8 @@ templates:
     version: 1
     summary: PBM backup size
     expr: |-
-      (max by (name,node_name) (mongodb_pbm_backup_size_bytes{status="done"})/1073741824)
-      > [[ .threshold ]]
+      (max by (name,node_name,service_name) (mongodb_pbm_backup_size_bytes{status="done"})/1073741824)
+      > bool [[ .threshold ]]
```
ademidoff, line 8: Same `> bool` behavior-change question as in `managed/data/alerting-templates/mongodb_pbm_agent_down.yml:6` — the value reported for this alert changes to 0/1. Please confirm this is intended.

managed/services/alerting/service.go
```diff
@@ -552,9 +566,49 @@ func convertTemplate(l *logrus.Entry, template models.Template) (*alerting.Templ
...
 
+	if err := populateMultiExpressionFields(t, template.Yaml); err != nil {
+		return nil, err
+	}
+
 	return t, nil
 }
 
+// populateMultiExpressionFields fills the multi-query fields (queries/expressions/condition)
+func populateMultiExpressionFields(t *alerting.Template, yamlContent string) error {
```
ademidoff: `populateMultiExpressionFields` re-parses the template YAML (`alert.Parse`) on every `ListTemplates` call, once per template, since `convertTemplate` runs per template. It's fine at current scale but is avoidable work — the queries/expressions/condition could be parsed once at `CollectTemplates` time and cached on `models.Template` instead of on every list request.

managed/services/alerting/service.go
```diff
@@ -644,20 +698,14 @@ func (s *Service) CreateRule(ctx context.Context, req *alerting.CreateRuleReques
 		forDuration = req.For.AsDuration()
 	}
 
-	expr, err := fillExprWithParams(sourceTemplate.Expr, paramsValues.AsStringMap())
+	alertTemplate, err := parseAlertTemplate(sourceTemplate.Yaml)
 	if err != nil {
-		return nil, fmt.Errorf("failed to fill rule expression with parameters: %w", err)
+		return nil, status.Errorf(codes.InvalidArgument, "Invalid template %s: %v.", req.TemplateName, err)
```
ademidoff: If a **stored** template's YAML fails to parse here, that's server-side data corruption, not a bad client request — `codes.Internal` would be more accurate than `codes.InvalidArgument`. (The request-time validation already happened at create/update.)

api-tests/alerting/multiexpr_filter_test.go
ademidoff: Prefer `t.Context()` over `context.Background()` here per the project test guideline.

ademidoff (commented):
Code review notes for the multi-query alerting BE work. Overall solid and well-tested — the integration test that evaluates the built query against VictoriaMetrics is especially valuable. A few behavior-change and quality items are left as inline comments; the two I'd most want confirmed before merge are the `MATCH` filter widening and the `> bool` template changes, since both alter existing alerting behavior.

managed/services/alerting/rule_builder.go
JiriCtvrtka, line 130: I think this changes the existing behavior for single-expression alerts, because filters now also match series without the specified label.

managed/data/alerting-templates/mongodb_pbm_agent_down.yml
4nte: It was intentional, the goal is to ensure that both firing and not-firing states produce a data instance.
bare `>` check would produce results only for the firing / unhealthy state, while healthy would produce a label-less `NoData` which means that it wouldn't show up in the new alerts page.

`> bool` changes this behavior so that the healthy state also produces an instance that will be present in the new alerts page.
  
Firing set is unchanged since the condition is essentially the same.
However, state history does record `0 / 1` instead of the raw value now, but also there is no user facing annotation change. We should discuss if this is acceptable.

managed/services/alerting/service.go
4nte: `CollectTemplates` now caches the parsed `alert.Template` per template, so `ListTemplates` no longer re-parses each template's YAML on every request

managed/data/alerting-templates/mongodb_pbm_backup_size.yml
4nte: It's intentional: https://github.com/percona/pmm/pull/5527#discussion_r3593472485

managed/data/alerting-templates/mongodb_pbm_backup_duration.yml
4nte: It's intentional: https://github.com/percona/pmm/pull/5527#discussion_r3593472485

4nte:
@ademidoff @JiriCtvrtka  We will come back to the filtering functionality to make it work fully for both single expressions and multiple-expressions without compromising on either. Current change (`"(%s)|"`)  does slightly change the behavior for single-expression but it's in favor of fixing a bug that would affect multiple-expressions templates for a much more common scenarios (when one of the expressions is a constant e.g threshold value), which would render the template completely unusable in combination with the MATCH filter without this fix.

As discussed, this is an interim fix. We'll revisit filtering so it works fully for both single and multi-expression templates without compromising either.

ademidoff (approved):

4nte (approved):

JiriCtvrtka:
LGTM. I cannot approved PR I created so Ante approved instead.

PR #5531, opened 2026-06-23
PMM-7 Improve GH Actions

maxkondr:
PMM-7

This PR was originally a large GitHub Actions refactor. That work has since been superseded by later PMM-7 changes already on `main`, so the branch has been reset to `main` and repurposed to carry two small, targeted CI changes. Everything except `.github/workflows/agent.yml` is back to `main`'s content.

## 1. Trigger the pmm-agent workflow with a `paths` allowlist

`.github/workflows/agent.yml` switches from `paths-ignore` to `paths`.

`paths-ignore` is a denylist: it has to be extended every time a new top-level directory appears, and until someone notices, the full pmm-agent matrix (11 jobs, each spinning up MySQL/MongoDB/PostgreSQL containers) runs for changes that cannot possibly affect pmm-agent. The allowlist instead names what pmm-agent actually builds and tests against:

```yaml
paths:
  - ".github/workflows/agent.yml"
  - ".github/actions/setup-go/**"
  - "agent/**"
  - "api/**"
  - "utils/**"
  - "version/**"
  - "go.*"
  - "!**/*.md"
```

`api/`, `utils/` and `version/` are the only other repo packages imported from `agent/`; Markdown-only changes are excluded.

## 2. Remove `GOPRIVATE` from every workflow

`GOPRIVATE` marks module path prefixes as private, which makes the go command skip the public module proxy (`GONOPROXY`) and the checksum database (`GONOSUMDB`) — neither of which can see a private repository. It was introduced in 61c670cf6 as `GOPRIVATE: github.com/percona-platform`, paired with a `Configure git for private modules` step authenticating as `percona-platform-robot`, because `github.com/percona-platform/*` (saas, dbaas-api) were private repos at the time.

Neither precondition holds any more:

- **Nothing it covers is private.** Every `percona` / `Percona-Lab` module in `go.sum` is a public repo and is served by `proxy.golang.org`: `exporter_shared`, `go-mysql`, `percona-toolkit`, `platform`, `promconfig`, `Percona-Lab/spec`, `Percona-Lab/kingpin`. The last genuinely private one, `percona/saas`, was replaced by the public `percona/platform` in 17b30d8f8.
- **The other half of the mechanism is gone.** The `git config … insteadOf` step with `ROBOT_TOKEN` was dropped in cc3ead8a7. No Go workflow sets up credentials any more, so if a private module did reappear, `GOPRIVATE` alone would not help — the fetch would fail on auth.

And it is not free: because `GONOPROXY` inherits it, CI fetches those modules over git from github.com instead of the CDN-backed proxy — slower, subject to GitHub rate limits, uncached — and `GONOSUMDB` means a newly added Percona dependency would land in `go.sum` without checksum-database corroboration. Existing `go.sum` entries are still verified, so this is a small hardening win rather than a fixed hole.

Removed from `main.yml`, `admin.yml`, `agent.yml`, `qan-api2.yml`, `vmproxy.yml`, `percona-intelligence.yml` (top-level `env:` blocks, each of which held nothing else) and from the `docker run` invocation in `managed.yml`. The same `ENV GOPRIVATE` line is dropped from `.devcontainer/Dockerfile` and `api-tests/Dockerfile`, which set it for the same obsolete reason; `GOPRIVATE` no longer appears anywhere in the repo.

## Verification

CI is the check: every workflow still parses, `agent.yml` must trigger on this PR (it touches `.github/workflows/agent.yml`), and the Go jobs must still resolve modules with `GOPRIVATE` unset.


JiriCtvrtka (approved):

maxkondr:
agreed with @ademidoff to postpone this PR till 3.9.0 release

fabio-silva (approved):

ademidoff (approved):

4nte (approved):

PR #5541, opened 2026-06-25
PMM-14519 Omit skip-verify flag when TLS is disabled

4nte:
PMM-14519

[Feature build](https://github.com/Percona-Lab/pmm-submodules/pull/4432)

When a MySQL (or MariaDB) service is added with `tls=false` and `tls_skip_verify=true`, it would forcefully attempt a TLS connection and fail if server is not configured to support it.

Fix is to ignore `tls_skip_verify` when TLS is disabled.

This was a regression caused by a change in the mysqld_exporter upstream, fixing it in pmm-managed is probably the better option than patching it in our mysqld_exporter to avoid future merge conflicts.


JiriCtvrtka:
Does it still support partial certificates?

maxkondr (approved):

4nte:
@JiriCtvrtka 
> Does it still support partial certificates? 


Yes, standard TLS behavior is unaffected by this change.
This fix addresses the issue where TLS is supposed to be disabled completely but `tls_skip_verify` is also provided, which should be a no-op.

ademidoff (approved):

PR #5547, opened 2026-06-25
PMM-15155 fix missing datasource on upgrade

fabio-silva:
Ticket number: PMM-15155


ademidoff (commented):
I think it would work, but it's not the best solution IMO.

It seems more like a band-aid and it does not solve the root cause - a race between `managed` starting Grafana and ansible applying new dashboards while Grafana is running.

I propose a different approach, which triggers only one restart and avoids the race (see https://github.com/percona/pmm/pull/5552).

ademidoff (changes requested):

ademidoff:
@fabio-silva I'm closing this PR in favor of #5552.

PR #5552, opened 2026-06-26
PMM-15155 Fix missing datasource on upgrade

ademidoff:
Ticket number: PMM-15155

SUBMODULES-4436

## Problem

On init/upgrade, `pmm-init` (ansible) replaced the bundled Grafana plugins under `/srv/grafana/plugins` while `pmm-managed` independently started Grafana via supervisord. Grafana discovers plugins only at startup, so a Grafana started mid-sync registered an incomplete plugin set — e.g. the ClickHouse datasource went missing after an upgrade.

## Approach

This resolves the **race condition between the two components** structurally instead of applying a band-aid (an extra restart that merely recovers from the race after the fact): the plugin sync moves into the container entrypoint, where it runs **before `exec supervisord`**. At that point no process that could launch Grafana exists yet, so Grafana can never observe a partially synced plugin directory. Grafana starts exactly once, via pmm-managed, on the final plugin set — no stop/start/restart choreography anywhere.

## Changes

- **`build/docker/server/entrypoint.sh`** — new sync step, replacing the fresh-volume-only copy (fresh install and upgrade are now the same code path):
  - runs only when the bundled plugin set changed, keyed on a marker file (`/srv/grafana/plugins/.pmm-synced-version`) compared against the image's `/usr/share/percona-dashboards/VERSION`; plain restarts skip the sync entirely
  - the marker is owned exclusively by the entrypoint and written only after a successful copy, so an interrupted sync is retried on the next start (`set -o errexit`); wiping the plugins dir also self-heals, since the marker goes with it
  - delete-then-copy per bundled plugin dir (same semantics as the removed ansible role): files dropped between plugin versions don't linger, user-installed plugins are untouched
- **`build/ansible/roles/dashboards`** — role removed. Only the `PERCONA_DASHBOARDS_VERSION` write survives, moved into the initialization role: it remains a pure init/upgrade-pipeline marker with a single writer (ansible), no longer coordinating anything between components.
- **`managed/services/supervisord`** — no net change: Grafana keeps unconditional `autostart = true` and is started by pmm-managed as before.

Also includes a small `pmm-server` cleanup fix in `build/ansible/pmm-docker/post-build.yml` (exclude static files).

## Behavior matrix

| Scenario | Result |
| --- | --- |
| Fresh volume | marker absent → sync, marker written |
| Upgrade (bundled version changed) | marker differs → sync before anything starts |
| Plain container restart | marker matches → skip |
| Sync interrupted (crash/kill) | marker not written → re-sync on next start |
| Plugins dir wiped manually | marker gone too → self-heals on next start |


build/ansible/roles/dashboards/tasks/main.yml
```diff
@@ -31,3 +36,12 @@
     dest: /srv/grafana/PERCONA_DASHBOARDS_VERSION
     mode: 0664
     remote_src: yes
+
+- name: Restart Grafana after plugin sync
+  supervisorctl:
+    name: grafana
+    state: restarted
+  failed_when: false
```
JiriCtvrtka: I think that failed_when: false on Grafana restart can hide startup failures. With autostart=false during init/upgrade, a failed restart may leave Grafana down and the playbook still green.

maxkondr:
As soon as `supervisord` doesn't natively support start services dependencies, we still may emulate it using wrapper scripts. 

So that the graph is like the following:
1. Run init server - initialise everything that is required (`srv/`, `grafana plugins`, `etc`)
2. Grafana service is started via wrapper like:
```
#!/bin/bash

# Wait until the init step is finished
while ! <check for init step finish>; do   
  echo "Waiting for init step to finish..."
  sleep 1
done

# CRITICAL: Use 'exec' so the Grafana replaces the bash script in memory.
# This ensures Supervisord tracks the correct PID and can stop/restart it.
exec /usr/sbin/grafana server <CLI params>
```

3. Similar approach with pmm-managed if shall be running only after init step.

Has this approach been considered earlier?

Alternative robust solution (but requires more effort) - using Nomad instead of Supervisord. 
https://developer.hashicorp.com/nomad/docs/concepts/scheduling/schedulers#service
https://developer.hashicorp.com/nomad/docs/job-declare/task-dependencies

build/ansible/pmm-docker/post-build.yml
```diff
@@ -53,7 +53,7 @@
         - nomad-server.ini
 
     - name: Clean pmm-server dir
-      shell: find /usr/share/pmm-server -mindepth 1 -maxdepth 1 -print0 | xargs -0 rm -rf --
+      shell: find /usr/share/pmm-server -mindepth 1 -maxdepth 1 -not -name static -print0 | xargs -0 rm -rf --
```
ademidoff, line 56: This is a bugfix that prevents `/usr/share/pmm-server/static` directory from being removed in the post-build phase. The directory is being used to render the maintenance page when the user hits the browser refresh button during upgrade.

build/ansible/roles/dashboards/tasks/main.yml
ademidoff: Thanks! I ended up by removing the role :)

maxkondr (approved):

4nte (approved):

PR #5559, opened 2026-06-28
PMM-15181 Migrate to grafana-openapi-client-go

ademidoff:
Ticket number: PMM-15181

Feature build: SUBMODULES-4439

## What

Replace the archived `github.com/grafana/grafana-api-golang-client` with the maintained `github.com/grafana/grafana-openapi-client-go` (go-swagger generated client).

## Changes

- `managed/services/grafana/client.go`: build the client via `client.NewHTTPClientWithConfig`, preserving auth-header forwarding through `TransportConfig.HTTPHeaders`. Folder create/get/delete and the alerting `grafanaClient` interface now use `models.Folder`; delete uses `folders.DeleteFolderParams`.
- `GetDatasourceUIDByID`: the new client has no by-numeric-ID lookup, so it lists datasources and matches on `ID` (single caller, the Metrics datasource `id=1`). Returns a clearer "not found" error when absent.
- `api-tests`: `GetGrafanaClient` returns `*client.GrafanaHTTPAPI` built on the existing go-swagger `Transport` helper; folder and user operations moved to `Folders.*` / `AdminUsers.*`.
- Regenerated the alerting `grafanaClient` mock for the `models.Folder` return type.
- `go mod tidy`: dropped the old dependency, added the new one as direct.

## Testing

- `go build ./...` and `go vet` on the changed packages are clean; all three api-test binaries compile; gofumpt reports no changes.
- DB-backed alerting tests run in CI (no local Postgres).


managed/services/grafana/client.go
```diff
@@ -784,26 +786,31 @@ func (c *Client) GetDatasourceUIDByID(ctx context.Context, id int64) (string, er
...
 	if err != nil {
 		return "", err
 	}
-	return ds.UID, nil
+	for _, ds := range resp.Payload {
+		if ds.ID == id {
+			return ds.UID, nil
+		}
+	}
+	return "", fmt.Errorf("datasource with ID %d not found", id)
```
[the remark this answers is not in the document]
ademidoff: The snippet here is from the first commit (the `GetDataSources()` list-and-filter), which a later commit replaced with a direct `Datasources.GetDataSourceByName("Metrics")` lookup, so the loop and its plain `fmt.Errorf("...not found")` no longer exist.

The underlying point still applied, so I've wrapped the lookup error with context to match the sibling folder methods (commit `fc21a519d`):

```go
resp, err := grafanaClient.Datasources.GetDataSourceByName(name)
if err != nil {
    return "", fmt.Errorf("failed to get datasource %q: %w", name, err)
}
```

The go-swagger error already carries the HTTP status, so a missing datasource surfaces clearly.

On returning `codes.NotFound`: I left that out intentionally. This grafana client layer doesn't emit gRPC status codes for its other methods (status mapping lives in the service layer), and a missing internal "Metrics" datasource isn't really a client-facing not-found — the caller is creating a rule, not fetching a datasource — so `NotFound` would be misleading (closer to an internal/failed-precondition condition).

api-tests/alerting/alerting_test.go
```diff
@@ -46,10 +47,12 @@ func TestRulesAPI(t *testing.T) {
...
 	// Create grafana folder for test alert rules
 	gClient := pmmapitests.GetGrafanaClient(t)
-	folder, err := gClient.NewFolder(pmmapitests.TestString(t, "test-folder"))
+	createdFolder, err := gClient.Folders.CreateFolder(&models.CreateFolderCommand{Title: pmmapitests.TestString(t, "test-folder")})
 	require.NoError(t, err)
+	folder := createdFolder.Payload
 	t.Cleanup(func() {
-		_ = gClient.DeleteFolder(folder.UID, gapi.ForceDeleteFolderRules())
+		force := true
+		_, _ = gClient.Folders.DeleteFolder(folders.NewDeleteFolderParams().WithFolderUID(folder.UID).WithForceDeleteRules(&force))
```
maxkondr:
```suggestion
		_, _ = gClient.Folders.DeleteFolder(folders.NewDeleteFolderParams().WithFolderUID(folder.UID).WithForceDeleteRules(new(true)))
```

maxkondr (approved):

4nte (approved):

api-tests/alerting/alerting_test.go
ademidoff: Done.

PR #5572, opened 2026-06-30
PMM-7 Setup format check on UI

matejkubinec:
Setups format check on UI and fix current format issues.


fabio-silva (approved):

.github/workflows/ui.yml
```diff
@@ -48,6 +48,9 @@ jobs:
           cache: yarn
           cache-dependency-path: ui/yarn.lock
 
+      - name: Check formatting
+        run: make format-check
```
maxkondr, line 42: this command may produce some changes in source code.
but we need to fail GH Action in case such changes are produced (code is not formatted properly in commits).


```suggestion
        run: |
          make format-check
          git status
          git diff --exit-code
```

mattiasimonato (approved):

.github/workflows/ui.yml
matejkubinec: @maxkondr It shouldn't make changes in the source code as it runs `prettier --check` under the hood (only reports if code needs to be formatted)

PR #5578, opened 2026-07-01
PMM-7 Adjust GH Action for updating Dependabot's PRs

maxkondr:
Ticket number: PMM-7

This pull request updates the Dependabot GitHub Actions workflow to improve security and automation. The main changes involve switching authentication to use a GitHub App token, tightening permissions, and automatically labeling Dependabot pull requests.

**Authentication and Permissions:**
* The workflow now generates a GitHub App token using the `actions/create-github-app-token` action, replacing the use of the default `GITHUB_TOKEN` for all subsequent steps. This enhances security and allows for more granular permissions.
* Permissions are now set to an empty object (`permissions: {}`) at the workflow level, relying on job-level permissions or the generated token for required access.

**Dependabot Automation:**
* All steps that previously used `GITHUB_TOKEN` now use the generated GitHub App token, including fetching Dependabot metadata and enabling auto-merge for version update PRs.
* Added a new step to automatically add the `auto-update-branch` label to Dependabot pull requests using the `actions/github-script` action and the generated token.

**Workflow Trigger:**
* The workflow trigger for `pull_request` events is now limited to the `opened` type, reducing unnecessary workflow runs.


4nte (approved):

.github/workflows/dependabot.yml
```diff
@@ -14,15 +17,33 @@ jobs:
...
+      - name: Add auto-update-branch label
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          github-token: ${{ steps.generate-token.outputs.token }}
+          script: |
+            await github.rest.issues.addLabels({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              labels: ['auto-update-branch']
```
ademidoff, line 48: I don't really understand why this label needs to be set on every PR.

It's wasteful, because it will trigger a storm of actions on every single PR. Let's be careful as this is a paid feature. Even if it wasn't I still believe such an spike in resource consumption does not justify this change.

Moreover, I end up waiting when I click on the "Update" button, since GH is not able to run my actions before it finishes dealing with the storm.

Screenshot 2026-07-06 at 14 12 14

ademidoff (changes requested):

.github/workflows/dependabot.yml
maxkondr: the goal was: make PR ready for merge as fast as possible. The less actions developers need to do to merge the commit is better.
Current flow:
1. open all (usually there are 3-5)  PRs from Dependabot. 
1.1 it already has "auto-merge" feature enabled.
3. review it - approve
4. in case PR is not up-2-date:
4.1 click "Update" button on each PR
4.2 wait for all checks passed

Once 1 Dependabot PR is merged p4.1 - 4.2 have to be repeated for all remaining PRs and someone from developers has to do this.

maxkondr: Regarding actions storm:
1. GH allocates some limit to each org. If some PRs "eat" this budget - rest will be waiting in queue and I observed it on regular basis.
2. Dependabot PRs have to be merged ASAP without long waiting in queue for review and merge. the faster they are merged -> more slots will be available for the rest of PRs.
3. We already have cleanup from worthless runs in https://github.com/percona/pmm/pull/5565 that will provide more free slots to PRs.
4. We have even more improvements (avoid running not-needed actions) in - https://github.com/percona/pmm/pull/5531 -> as result more available slots.

ademidoff: Oh, sorry, I guess I misunderstood the goal :)

IMO Dependabot is a well-engineered piece of software, and we can hardly make it better. It already auto-updates itself whenever its PRs get behind the merge base.

It does however stop updating its own PRs when they get stale, which effectively adresses the "storm" issue. Otherwise, I personally don't see a problem.

ademidoff: Also, there should never be a rush when merging dependabot-authored PRs as they are prone to occasionally breaking the main branch. We have seen this many times already.

I believe, only the GH action bumps are relatively innocent, as they are not part of our code base and they have little to no impact on our tests.

maxkondr: > ... It already auto-updates itself whenever its PRs get behind the merge base.

haven't seen it in action though. Here is example:
https://github.com/percona/pmm/pull/5620 has been merged
but https://github.com/percona/pmm/pull/5622 is still behind base branch and there is no attempt to update it even after N hours.

Screenshot 2026-07-09 at 3 24 52 PM

JiriCtvrtka (approved):

.github/workflows/dependabot.yml
ademidoff: Dependabot has an algorithm where it gives up on stale PRs. This change would reverse that. For example, isn't it wasteful to keep the PR #5327, which was open on May 6 and which was abandoned, updated? I think it is.

Therefore, let's not patch Dependabot. Consider contributing to Dependabot instead: if they accept the PR, we will get it for granted, no change.

Image

PR #5581, opened 2026-07-01
PMM-14678 Route vmagent remote-write through PMM server in HA

4nte:
PMM-14678

[Feature Build](https://github.com/Percona-Lab/pmm-submodules/pull/4451)

In HA, `PMM_VM_URL` points at an in-cluster-only endpoint (`vmauth`) that is unreachable for external pmm agents.

To fix this, when HA is enabled, we set vmagent's remote-write to the PMM server address.
vmagent requests then get routed through PMM server `/victoariametrics/` proxy using server credentials.


managed/services/agents/vmagent.go
```diff
@@ -96,6 +99,12 @@ func vmAgentConfig(scrapeCfg string, params victoriaMetricsParams) *agentv1.SetS
 		}
 	}
 
+	// When routing metric writes through the PMM server, authenticate with the PMM server credentials.
+	if !useExternalVM {
+		delete(systemEnvs, "VMAGENT_remoteWrite_basicAuth_username")
+		delete(systemEnvs, "VMAGENT_remoteWrite_basicAuth_password")
```
ademidoff: The strip only removes injected `VMAGENT_remoteWrite_basicAuth_*`, not an injected `VMAGENT_remoteWrite_url`, so a deployment that passes the VM endpoint via that env var bypasses the fix. 

Failure scenario: an HA deployment injects VMAGENT_remoteWrite_url=http://vmauth:8427/api/v1/write alongside the basic-auth creds. After this change `useExternalVM` is false, so the auth keys are deleted and server credentials are added via `addEnvIfNotSet` — but the injected `VMAGENT_remoteWrite_url` survives the `addEnvIfNotSet` in line 116 and still points at the unreachable in-cluster `vmauth`, now presented with PMM server credentials → writes go to the wrong endpoint and get rejected.

ademidoff: Also, the `delete(systemEnvs, ...)` block also changes behavior for the non-HA internal VM path: previously a deployment-injected `VMAGENT_remoteWrite_basicAuth_*` would win; now server credentials always override it. Let's confirm the deployment injection won't catch a bug.

maxkondr:
The current implementation (this PR) points all VMAgent (internal inside the same K8S cluster and external outside K8S) to the NGINX's external address, and all traffic from VMAgents will pass through NGINX. 
Such approach fixes the issue and makes it work. BUT it contradicts to one of HA mode purpose - cope with higher load on the system then single node PMM server can "swallow". So this is step back to single-node direction.

The good approach would be: provide different VM addresses (K8S internal or external) to different VMAgents (internal within the same K8S cluster or outside the K8S) like it is depicted on the diagram below.

Screenshot 2026-07-02 at 3 05 43 PM

managed/services/agents/vmagent.go
```diff
@@ -53,35 +53,38 @@ func extractCredentialsFromURL(urlStr string) (string, string) {
 }
 
 // vmAgentConfig returns desired configuration of vmagent process.
-func vmAgentConfig(scrapeCfg string, params victoriaMetricsParams) *agentv1.SetStateRequest_AgentProcess {
+func vmAgentConfig(scrapeCfg string, params victoriaMetricsParams, haEnabled bool) *agentv1.SetStateRequest_AgentProcess {
```
maxkondr: I would recommend splitting this func into several ones:
- prepare config for non-HA internal VM case
- prepare config for non-HA external VM case 
- prepare config for HA case 
it will be easier for understanding.

managed/models/victoriametrics_params.go
```diff
@@ -121,16 +121,3 @@ func (vmp *VictoriaMetricsParams) URLFor(path string) (*url.URL, error) {
...
-	if vmp.url.User != nil {
-		username := vmp.url.User.Username()
-		password, _ := vmp.url.User.Password()
-		return []string{
-			"-remoteWrite.basicAuth.username=" + username,
-			"-remoteWrite.basicAuth.password=" + password,
-		}
-	}
-	return []string{}
-}
```
4nte, line 136: This is now dead code, credentials are passed to vmagent exclusively through environment variables as opposed to cmd args.

Previously, vmagent could receive credentials through both args and env at the same time, which is hard to reason about. There is now only one channel for vmagent to receive credentials - through environment variables.

managed/services/agents/deps.go
```diff
@@ -69,7 +69,6 @@ type victoriaMetricsParams interface {
 	ExternalVM() bool
 	URLFor(path string) (*url.URL, error)
 	URL() string
-	VMAgentArgs() []string
```
4nte, line 72: Dead code since vmagent now receives credentials through env vars.

managed/services/agents/vmagent.go
4nte: I created two functions that configure for a specific configuration mode: `vmAgentConfigServerProxy` and `vmAgentConfigExternalVM`.

Both function calls are annotated with deployment scenarios (per your suggestion) that will trigger each branch, so it's easier to understand what path will be taken depending on the deployment scenario.

managed/services/agents/vmagent.go
```diff
@@ -53,34 +53,72 @@ func extractCredentialsFromURL(urlStr string) (string, string) {
...
+	interfaceToBind := envvars.GetInterfaceToBind()
 	// Only keep the specified exceptions as command line arguments
-	args := append([]string{
+	args := []string{
 		"-envflag.enable=true",
 		"-envflag.prefix=VMAGENT_",
 		"-remoteWrite.tmpDataPath={{.tmp_dir}}/vmagent-temp-dir",
 		"-promscrape.config={{.TextFiles.vmagentscrapecfg}}",
 		"-httpListenAddr=" + interfaceToBind + ":{{.listen_port}}",
-	}, params.VMAgentArgs()...)
```
4nte, line 81: This simplifies configuration, credentials now flow through environment variables only as opposed to both environment and cmd args where one would overshadow another and could have conflicting values.

It's also a security improvement since credentials no longer appear in the process command line, where, for example,`ps` would previously expose them.

managed/services/agents/vmagent.go
```diff
@@ -96,6 +134,15 @@ func vmAgentConfig(scrapeCfg string, params victoriaMetricsParams) *agentv1.SetS
...
 	}
 
+	// An injected remote-write URL means the target is no longer the server proxy, so injected
+	// credentials belong with it and must not be dropped.
+	if settings.dropInjectedAuth {
+		if _, urlInjected := systemEnvs["VMAGENT_remoteWrite_url"]; !urlInjected {
+			delete(systemEnvs, "VMAGENT_remoteWrite_basicAuth_username")
+			delete(systemEnvs, "VMAGENT_remoteWrite_basicAuth_password")
+		}
+	}
```
4nte, line 160: We drop credentials when `dropInjectedAuth` is true (for example, in HA deployment) and when `VMAGENT_remoteWrite_url` env var is **NOT** set.

in HA deployment scenario, `VMAGENT_remoteWrite_basicAuth_username/password` is injected by default with VM credentials, but the vmagent needs server credentials instead.
Fix is to delete injected `VMAGENT_remoteWrite_basicAuth_username/password` and set correct credentials [here](https://github.com/percona/pmm/pull/5581/changes#diff-2d22b4ddd67626775c4f9f26b9a02431d5d3bb1a5c13428822520f894a996b09R160-R163).

If user sets `VMAGENT_remoteWrite_url`, we keep injected `VMAGENT_remoteWrite_basicAuth_username/password` assuming that if user decided to override the address via env var, they are expected to be responsible for providing appropriate credentials too (`VMAGENT_remoteWrite_basicAuth_username/password`)

managed/services/agents/vmagent.go
4nte: We don't inject `VMAGENT_remoteWrite_url` by default, if present, we should treat it as an user override.

An injected `VMAGENT_remoteWrite_url` is now treated as a deliberate user override, and the credential drop is skipped whenever a `VMAGENT_remoteWrite_url` is set. 
This assumes that users who set `VMAGENT_remoteWrite_url` are expected to also set appropriate credentials (`VMAGENT_remoteWrite_basicAuth_*`) in tandem.


> Also, the delete(systemEnvs, ...) block also changes behavior for the non-HA internal VM path: previously a deployment-injected VMAGENT_remoteWrite_basicAuth_* would win; now server credentials always override it. Let's confirm the deployment injection won't catch a bug.

Credential drop is now gated on haEnabled only, so the non-HA internal VM path keeps deployment-injected VMAGENT_remoteWrite_basicAuth_*.

managed/services/agents/vmagent.go
```diff
@@ -53,34 +53,72 @@ func extractCredentialsFromURL(urlStr string) (string, string) {
...
+	// Scenarios:
+	// a) standalone deployment with internal VM
+	// b) HA deployment (external VM by default)
+	return vmAgentConfigServerProxy(scrapeCfg, haEnabled)
+}
 
-	if params.ExternalVM() {
-		serverURL = params.URL()
+// vmAgentConfigServerProxy configures the vmagent so that writes go through the PMM server's VM proxy,
```
maxkondr, line 76: I thought we use VMProxy for read requests only (it was introduced as an additional filter in LBAC).
But write are send to VM directly, isn't it?

managed/services/agents/vmagent.go
```diff
@@ -53,34 +53,72 @@ func extractCredentialsFromURL(urlStr string) (string, string) {
...
-		maxScrapeSize = space
+	var auth *basicAuth
+	if vmUsername, vmPassword := extractCredentialsFromURL(vmURL); vmUsername != "" {
+		auth = &basicAuth{username: vmUsername, password: vmPassword}
 	}
 
-	interfaceToBind := envvars.GetInterfaceToBind()
+	return buildVMAgentProcess(scrapeCfg, vmAgentSettings{
+		remoteWriteURL: vmURL + "api/v1/write",
+		auth:           auth,
```
maxkondr, line 102: do we really need to extract auth credentials into a separate param if they are still passed in `vmUrl`?

managed/services/agents/vmagent.go
4nte: I was not aware that VMProxy is reserved for read requests only, if that's what you're claiming.

The problem is that an external pmm-agent is unable to reach VM directly, because VM is not exposed and external agents cannot reach in-cluster VM. `vmproxy` can proxy both read and write requests, and this fix leverages it.

There is room for optimization in terms of reducing number of hops taken for a single request from vmagent to VM but it increases the scope significantly, as we discussed already, this should be a bigger improvement that affects arhictecture.

FWIW, standalone deployments do support direct writes to external VM's, but for all other deployment modes, writes go through vmproxy. In the future we can hopefully make this possible for all deployment modes.

managed/services/agents/vmagent.go
4nte: It's redundant, credentials are now stripped from the remote-write URL and vmagent receives credentials through environment only.

maxkondr (approved):

managed/models/victoriametrics_params.go
ademidoff: This sounds like a breaking change and is probably not necessary. We apply the same principle all across the board: if both are present, the environment variable must win.

managed/services/agents/vmagent.go
```diff
@@ -31,56 +31,95 @@ var (
...
-	if space := os.Getenv(maxScrapeSizeEnv); space != "" {
-		maxScrapeSize = space
+	var auth *basicAuth
+	if vmUsername != "" || vmPassword != "" {
+		auth = &basicAuth{username: vmUsername, password: vmPassword}
 	}
 
-	interfaceToBind := envvars.GetInterfaceToBind()
+	return buildVMAgentProcess(scrapeCfg, vmAgentSettings{
+		remoteWriteURL: vmURL + "api/v1/write",
```
ademidoff: This URL concatenation leans on the guaranteed trailing slash. I see it matches prior behavior, but `url.JoinPath/URLFor` would be more robust against the potential absence of the trailing `/`.

ademidoff:
Let's make the linters happy :)

Screenshot 2026-07-09 at 09 51 58

ademidoff (approved):

managed/services/agents/vmagent.go
4nte: Agreed, done!

managed/services/agents/vmagent.go
```diff
@@ -31,56 +31,107 @@ var (
...
+
+	return parsedURL.String(), username, password
+}
 
-	return username, password
+// vmAgentDeployment describes the deployment context that selects the
+// vmagent remote-write path.
+type vmAgentDeployment struct {
+	haEnabled     bool // PMM Server runs in HA (clustered) mode
+	isServerAgent bool // this is PMM Server's own built-in agent
```
ademidoff:
```suggestion
    // PMM Server runs in HA (clustered) mode
	haEnabled     bool
	// this is PMM Server's own built-in agent
	isServerAgent bool
```
Inline comments fight with potential `//nolint`, pls always put them above :)

managed/services/agents/vmagent.go
```diff
@@ -31,56 +31,107 @@ var (
...
 
-	if params.ExternalVM() {
-		serverURL = params.URL()
+// vmAgentConfigServerProxy configures the vmagent so that writes go through the PMM server's VM proxy,
+// authenticated by default with each client's own PMM server credentials.
+func vmAgentConfigServerProxy(scrapeCfg string, dropInjectedAuth bool) *agentv1.SetStateRequest_AgentProcess {
+	return buildVMAgentProcess(scrapeCfg, vmAgentSettings{
+		remoteWriteURL:   "{{.server_url}}/victoriametrics/api/v1/write",
+		dropInjectedAuth: dropInjectedAuth,
+		auth:             &basicAuth{username: "{{.server_username}}", password: "{{.server_password}}"}, //nolint:gosec // template placeholder, not a secret
```
ademidoff: Inline comments - it doesn't obey https://github.com/percona/pmm/blob/main/AGENTS.md?plain=1#L176

managed/services/agents/vmagent.go
```diff
@@ -31,56 +31,107 @@ var (
...
 
+// basicAuth holds default remote-write basic-auth credentials.
+type basicAuth struct {
+	username string
+	password string
+}
+
+// vmAgentSettings captures settings that differ between deployment scenarios.
+type vmAgentSettings struct {
+	remoteWriteURL   string     // default for VMAGENT_remoteWrite_url, can be overridden by an injected value
```
ademidoff: Inline comments

managed/services/agents/vmagent.go
```diff
@@ -31,56 +31,107 @@ var (
...
+	dropInjectedAuth bool       // dropInjectedAuth discards injected VMAGENT_remoteWrite_basicAuth_*.
```
ademidoff:
```suggestion
    // discards injected VMAGENT_remoteWrite_basicAuth_*
	dropInjectedAuth bool
```

managed/services/agents/vmagent.go
```diff
@@ -31,56 +31,107 @@ var (
...
 // vmAgentConfig returns desired configuration of vmagent process.
-func vmAgentConfig(scrapeCfg string, params victoriaMetricsParams) *agentv1.SetStateRequest_AgentProcess {
-	serverURL := "{{.server_url}}/victoriametrics/"
-	var vmUsername, vmPassword string
+func vmAgentConfig(scrapeCfg string, params victoriaMetricsParams, deployment vmAgentDeployment) *agentv1.SetStateRequest_AgentProcess {
+	// Push straight to VictoriaMetrics when the vmagent can reach it directly:
+	//   a) PMM Server's own agent in HA is in-cluster and reaches the VM the same way it did
+	//     before the write-proxy existed. Only remote clients need the server write-proxy, so
+	//     in HA the built-in agent is not routed through it.
+	//   b) standalone deployment with an externally configured VM.
```
ademidoff: Hard to follow, and way more that the code it explains.

A simpler variant of it:
```
* server agent in HA → direct
* external VM standalone → direct
* everything else → proxy
```

ademidoff (commented):
### Security

One behavior to be aware of, documented by "HA honors an operator-injected remote-write URL": injecting only a URL, with no credentials, still sends {{.server_username}}/{{.server_password}} to that arbitrary endpoint. It's the operator's own configuration, but it is a credential-forwarding path.

### Performance

All HA client writes now traverse nginx → vmproxy → vmauth on the node the agent is connected to, versus a direct push before. 

As a result, `nginx client_max_body_size 10m` now bounds remote-write payload size, and vmproxy is an untuned `httputil.ReverseProxy` with no explicit transport timeouts that becomes the write path for every agent on the node.

### Design
Tying `dropInjectedAuth` to haEnabled encodes "the HA chart injects vmauth credentials" as "HA drops injected credentials". The asymmetry is real and tested (standalone internal-VM honors injected basic-auth, HA does not), but it means an HA operator has no way to set basic-auth for the proxy path except by also injecting a URL. Consider whether the chart should stop injecting `VMAGENT_remoteWrite_basicAuth_*` for the client path instead — or at minimum, `vmAgentConfigServerProxy(scrapeCfg, deployment.haEnabled)` should take a named/self-describing argument, since `haEnabled → dropInjectedAuth` is invisible at the call site.

### Tests
Gaps and nits:
  - No test for HA + isServerAgent + internal VM URL (above).
  - No test at the state.go layer that isServerAgent is wired from models.PMMServerAgentID — the flag's whole value depends on that comparison.
  - `TestVMAgentStandaloneServerAgent` / `TestVMAgentExternalVMBasicAuthOverridePreserved` are top-level funcs while peers are subtests of `TestMaxScrapeSize` — organization is drifting (partly pre-existing: `TestMaxScrapeSize` already holds unrelated cases).

managed/services/agents/vmagent.go
```diff
@@ -104,23 +167,19 @@ func vmAgentConfig(scrapeCfg string, params victoriaMetricsParams) *agentv1.SetS
...
 	addEnvIfNotSet("VMAGENT_remoteWrite_maxDiskUsagePerURL", "1073741824") // 1GB disk queue size
 	addEnvIfNotSet("VMAGENT_loggerLevel", "INFO")
 
-	// Set authentication based on VM type
-	if params.ExternalVM() && vmUsername != "" {
-		// Use credentials from external VM URL
-		addEnvIfNotSet("VMAGENT_remoteWrite_basicAuth_username", vmUsername)
-		if vmPassword != "" {
-			addEnvIfNotSet("VMAGENT_remoteWrite_basicAuth_password", vmPassword)
+	if settings.auth != nil {
```
JiriCtvrtka, line 176: Severity: Blocker
Problem: default basic-auth is applied even when the remote-write URL was injected
Desc: The credential drop at line 155 is correctly guarded on whether the URL was injected:


if settings.dropInjectedAuth {
    if _, urlInjected := systemEnvs["VMAGENT_remoteWrite_url"]; !urlInjected {
        delete(systemEnvs, "VMAGENT_remoteWrite_basicAuth_username")
        delete(systemEnvs, "VMAGENT_remoteWrite_basicAuth_password")
    }
}
but the default auth at line 176 is not:


if settings.auth != nil {
    addEnvIfNotSet("VMAGENT_remoteWrite_basicAuth_username", settings.auth.username)
    ...
}
The case where both the URL and basic-auth are injected is handled. The case where only the URL is injected is not: addEnvIfNotSet then fires and vmagent authenticates to the injected non-PMM endpoint (e.g. vmauth) with {{.server_username}} / {{.server_password}} — Grafana credentials → 401 and silent metric loss. Previously that URL got the PMM_VM_URL credentials.

Hoisting the lookup fixes both sites:
```
_, urlInjected := systemEnvs["VMAGENT_remoteWrite_url"]

if settings.dropInjectedAuth && !urlInjected {
    delete(systemEnvs, "VMAGENT_remoteWrite_basicAuth_username")
    delete(systemEnvs, "VMAGENT_remoteWrite_basicAuth_password")
}
...
if settings.auth != nil && !urlInjected {
    ...
}
```
Worth a test case for HA + injected URL + no injected auth — the existing suite doesn't cover it.

PR #5600, opened 2026-07-06
Clarify QAN data sources and prepared-statement visibility for MySQL

michaelcoburn:
## What

Adds a short, expectation-setting note so readers recognize why PMM Query Analytics (QAN) may show incomplete data for MySQL prepared-statement workloads.

## Why

Server-side prepared statements (used by most application drivers, e.g. JDBC) are exposed by upstream MySQL to Performance Schema with placeholder parameters (`?`) rather than literal values. As a result, QAN query examples appear as placeholders and EXPLAIN is unavailable for them. This is upstream MySQL behavior, **not** a PMM limitation. The note helps users self-diagnose instead of assuming data is missing.

Because this page lets users choose the query source, the note also points out that the file-based slow query log records literal values, so it gives fuller examples and EXPLAIN than Performance Schema.

## Scope

Documentation only — a single admonition in the "Choose and configure a source" section of `documentation/docs/install-pmm/install-pmm-client/connect-database/mysql/mysql.md`. Part of a coordinated set of PRs across pmm, psmysql-docs, pxc-docs, k8sps-docs, and k8spxc-docs.

DCO: signed off.


catalinaadam:
Thanks for the contribution! I've incorporated your changes in #5712 with a bit of restructuring. Closing this one in favor of that PR.

PR #5618, opened 2026-07-08
PMM-15130 Validate connection on agent params change

4nte:
PMM-15130

[Feature Build](https://github.com/Percona-Lab/pmm-submodules/pull/4464)

When a change touches connection parameters (username, password, TLS options, etc.), pmm-managed now verifies the new settings by having the pmm-agent connect to the service (with changed params). If the check fails, the change is rolled back and the agent keeps its previous working config.

User has an option to opt-out via `--skip-connection-check` to bypass validation.


managed/services/inventory/agents.go
```diff
@@ -220,7 +220,7 @@ func (as *AgentsService) ChangeNodeExporter(ctx context.Context, agentID string,
 		MetricsResolutions: convertMetricsResolutions(p.MetricsResolutions),
 	}
 
-	agent, err := as.executeAgentChange(ctx, agentID, params)
+	agent, err := as.executeAgentChange(ctx, agentID, params, true)
```
ademidoff: We need a note here, something like: `// Node exporter does not support connection check`

managed/services/inventory/agents.go
```diff
@@ -1579,7 +1579,7 @@ func (as *AgentsService) ChangeRDSExporter(ctx context.Context, agentID string,
...
```
ademidoff: Need a note here: `// RDS connects via cloud APIs; connection check unsupported`

managed/services/inventory/agents.go
```diff
@@ -1756,7 +1756,7 @@ func (as *AgentsService) ChangeAzureDatabaseExporter(
...
```
ademidoff: Same here

managed/services/inventory/agents.go
```diff
@@ -1779,7 +1779,7 @@ func (as *AgentsService) ChangeNomadAgent(ctx context.Context, agentID string, p
 		Enabled: params.Enable,
 	}
 
-	agent, err := as.executeAgentChange(ctx, agentID, changeParams)
+	agent, err := as.executeAgentChange(ctx, agentID, changeParams, true)
```
ademidoff: Same here

api/inventory/v1/agents.proto
```diff
@@ -1348,6 +1348,8 @@ message ChangeProxySQLExporterParams {
   google.protobuf.Duration connection_timeout = 13 [(validate.rules).duration = {
     gte: {seconds: 0}
   }];
+  // Skip connection check.
+  optional bool skip_connection_check = 14;
```
maxkondr, line 1352: no need to set `optional`, in proto v3 all params are optional by default

managed/services/inventory/agents.go
```diff
@@ -1991,16 +1991,28 @@ func convertMetricsResolutions(mrs *common.MetricsResolutions) *models.ChangeMet
 }
 
 // Helper function to execute agent change and build response.
-func (as *AgentsService) executeAgentChange(ctx context.Context, agentID string, params *models.ChangeAgentParams) (inventoryv1.Agent, error) { //nolint:ireturn
+func (as *AgentsService) executeAgentChange(ctx context.Context, agentID string, params *models.ChangeAgentParams, skipConnectionCheck bool) (inventoryv1.Agent, error) { //nolint:ireturn,lll
```
maxkondr: isn't it better to move `skip_connection_check` flag into `models.ChangeAgentParams` because this flag is logical tied to these params?

ademidoff (approved):

ademidoff:
I think we need to document well this feature for those who's agent 1) lost the password or 2) is in the maintenance window or just down for some reason.

Clearly, the agent won't be able to verify the password, so documenting something like "please pass the skipConnectionCheck = true " if such case would be great.

managed/services/inventory/agents.go
4nte: All fields in `models.ChangeAgentParams` describe agent configuration, and every field is persisted to the database, while `skipConnectionCheck` is a transient option specific to this request/action that should not be persisted.

For that reason I think they should remain separate to keep agent state separate from transient options that are not.

api/inventory/v1/agents.proto
4nte: True, it's just following an existing convention from other messages, for example:https://github.com/percona/pmm/blob/da99944e59acdbebdf4542895e22e08bc9a0dfd0/api/inventory/v1/agents.proto#L1259

maxkondr: this is not a convention but leftover from proto v2

managed/services/inventory/agents.go
maxkondr: no, they are not persisted automatically, Look at `managed/models/agent_helpers.go ChangeAgent()` they are used to update structure fetch from DB

4nte: `skipConnectionCheck` has been moved to `ChangeAgentParams` struct.
 https://github.com/percona/pmm/pull/5618/commits/ff5e63b995bc01c69c2ce2b6b883d45d714f3bb5

managed/services/inventory/agents.go
```diff
@@ -241,72 +241,43 @@ func (as *AgentsService) ChangeNodeExporter(ctx context.Context, agentID string,
...
-		return nil
-	})
-	if e != nil {
-		return nil, e
+	agent, err := as.executeAgentAdd(ctx, models.MySQLdExporterType, params, true)
+	if err != nil {
+		return nil, err
 	}
 
+	mysqldExporter := agent.(*inventoryv1.MySQLdExporter) //nolint:forcetypeassert
```
maxkondr: is it possible to correctly handle `forcetypeassert` linter warning in order to no create tech debt for us ?

4nte: Done, I also cleaned up existing supressions that were already here so we don't have any remaining supressions for `forcetypeassert`

maxkondr (approved):

JiriCtvrtka (approved):

PR #5625, opened 2026-07-09
fix: always show alerting for anon users

fabio-silva:
Ticket number: PMM-15067

Anonymous view:
Screenshot 2026-07-29 at 12 41 02


ui/apps/pmm/src/contexts/navigation/navigation.provider.tsx
```diff
@@ -69,18 +71,22 @@ export const NavigationProvider: FC<PropsWithChildren> = ({ children }) => {
 
     items.push(NAV_QAN);
 
-    if (user && settings) {
-      if (settings.frontend.exploreEnabled && user.isEditor) {
+    if (user) {
+      if (settings?.frontend.exploreEnabled && user.isEditor) {
```
matejkubinec, line 75: I think settings should be accessible even when users are not logged in (only readonly)

ui/apps/pmm/src/contexts/navigation/navigation.provider.tsx
```diff
@@ -69,18 +71,22 @@ export const NavigationProvider: FC<PropsWithChildren> = ({ children }) => {
...
         items.push(
           addExplore('grafana-metricsdrilldown-app' in settings.frontend.apps)
         );
       }
 
-      if (settings.frontend.unifiedAlertingEnabled) {
-        items.push(addAlerting(settings?.alertingEnabled, user));
-      }
+      items.push(
+        addAlerting(
```
matejkubinec, line 82: If alerting is turned off on grafana side, does it make sense to add the item at all?

ui/apps/pmm/src/contexts/navigation/navigation.provider.tsx
fabio-silva: @matejkubinec not sure I understand. You mean alerting settings or do you refer to `settings` prop?

ui/apps/pmm/src/contexts/navigation/navigation.utils.test.ts
```diff
@@ -0,0 +1,122 @@
+import { OrgRole, User } from 'types/user.types';
+import { addAlerting } from './navigation.utils';
+
+const baseUser = (overrides: Partial<User> = {}): User =>
```
matejkubinec: I think you could replace this with test users in `ui/apps/pmm/src/utils/testStubs.ts`, maybe with adding an anonymous user.

ui/apps/pmm/src/contexts/settings/settings.provider.tsx
```diff
@@ -19,7 +19,31 @@ export const SettingsProvider: FC<PropsWithChildren> = ({ children }) => {
   const frontendSettings = useFrontendSettings({
     refetchOnMount: false,
   });
+
   const combinedSettings = useMemo<CombinedSettings | null>(() => {
+    if (user?.isAnonymous && frontendSettings.data) {
+      return {
```
matejkubinec, line 25: I think it would be good to have an util function `createAnonymousUser` - for example in `ui/apps/pmm/src/contexts/user/user.utils.ts`

mattiasimonato (approved):

PR #5641, opened 2026-07-14
PMM-15129 Fix `gosec` Go linter issues (p1)

maxkondr:
Ticket number: PMM-15129

This pull request introduces several improvements focused on resource management, context handling, and defensive programming in the MongoDB, MySQL, and PostgreSQL agents. The main themes are improved error handling for resource cleanup, enhanced support for context cancellation in long-running goroutines, and more robust handling of time and histogram data to prevent panics and ensure correctness.

**Resource Management and Error Handling:**

* Replaced unchecked `Close()` and `Stop()` calls with error-checked versions in multiple agents (MongoDB, MySQL, PostgreSQL). Errors are now logged if resource cleanup fails, improving observability and reliability. (`admin/commands/summary.go`, `agent/agents/mongodb/mongolog/mongodb.go`, `agent/agents/mongodb/profiler/mongodb.go`, `agent/agents/mongodb/profiler/internal/profiler.go`, `agent/agents/mysql/perfschema/perfschema.go`, `agent/agents/postgres/pgstatmonitor/pgstatmonitor.go`, `agent/agents/postgres/pgstatstatements/pgstatstatements.go`) [[1]](diffhunk://#diff-94c9acae2ec502cab403aaf3d6cea9645ee1b17bbf711aca77b05ebc56c20c59L192-R195) [[2]](diffhunk://#diff-0b1347ff020d3bdf2716693bd884a22ddef7f597a6f750cfc6a72283db81572dL77-R80) [[3]](diffhunk://#diff-d2c489b9cfee850fa89b81301a6dd9ab2f53828ab98d3cb6774b338eed000915L75-R78) [[4]](diffhunk://#diff-12ddf04b5a0368011e538ccd48f162bd4abbebf767c52249cc91f8c81c699feaL148-R151) [[5]](diffhunk://#diff-38f6f59ca3bdd05ce32c696f1e4d7dd0ef8635b1b8c0a087d3df0e42f971d440L196-R199) [[6]](diffhunk://#diff-ae12be77883993fc7d2b5ebc501e2fb0b0078ea1d4645c2f9327095ae0661d30L290-R293) [[7]](diffhunk://#diff-e7e74230423ceb11899f9ed2ca5ca572cc331e8214a61727c2f71b88fd610bceL158-R161)

**Context Handling and Graceful Shutdown:**

* Collector and parser goroutines now properly respond to context cancellation, ensuring that background operations terminate cleanly when requested. New tests verify that context cancellation reliably shuts down these components. (`agent/agents/mongodb/profiler/internal/collector/collector.go`, `agent/agents/mongodb/profiler/internal/collector/collector_test.go`, `agent/agents/mongodb/profiler/internal/parser/parser.go`, `agent/agents/mongodb/profiler/internal/parser/parser_test.go`) [[1]](diffhunk://#diff-3001453c9916a590a73bbd6e2ffc0ff7fc5bd2f10c491a68e84d237a9d56f40dL66-R66) [[2]](diffhunk://#diff-3001453c9916a590a73bbd6e2ffc0ff7fc5bd2f10c491a68e84d237a9d56f40dL89) [[3]](diffhunk://#diff-3001453c9916a590a73bbd6e2ffc0ff7fc5bd2f10c491a68e84d237a9d56f40dL178-R184) [[4]](diffhunk://#diff-3001453c9916a590a73bbd6e2ffc0ff7fc5bd2f10c491a68e84d237a9d56f40dL207-R206) [[5]](diffhunk://#diff-4552f42c85dd3521d26e69e263d41b971058176f37937dea101b1310c2a71ed1R181-R219) [[6]](diffhunk://#diff-d47b34d9fe9fe792d2023ee03b2ba24c7d8f8044c59f950f4e305b172524d078L57-R57) [[7]](diffhunk://#diff-d47b34d9fe9fe792d2023ee03b2ba24c7d8f8044c59f950f4e305b172524d078L73) [[8]](diffhunk://#diff-d47b34d9fe9fe792d2023ee03b2ba24c7d8f8044c59f950f4e305b172524d078R128-R129) [[9]](diffhunk://#diff-d47b34d9fe9fe792d2023ee03b2ba24c7d8f8044c59f950f4e305b172524d078R153-R154) [[10]](diffhunk://#diff-5aa0e248e18e73b5daa9d75801b77f4499b3ec0033e3f39d6e812b3b2280fb38R84-R115)

**Defensive Programming for Time and Histogram Data:**

* Added defensive casts and comments to clarify and suppress linter warnings for Unix timestamps and durations stored as `uint32`, ensuring correctness up to the year 2106. (`agent/agents/mongodb/shared/aggregator/aggregator.go`, `agent/agents/mysql/perfschema/perfschema.go`, `agent/agents/mysql/slowlog/slowlog.go`, `agent/agents/postgres/pgstatmonitor/pgstatmonitor.go`)
* In PostgreSQL pgStatStatements, `periodStart` is now clamped to the valid `uint32` range to prevent overflows. (`agent/agents/postgres/pgstatstatements/pgstatstatements.go`) [[1]](diffhunk://#diff-8ea496b37353a34d26e564c82f078c08e7527698acaaa44103ec6f0a686fbd95L284-R285) [[2]](diffhunk://#diff-38f6f59ca3bdd05ce32c696f1e4d7dd0ef8635b1b8c0a087d3df0e42f971d440L348-R351) [[3]](diffhunk://#diff-620cacdc024d46674cb14c53c30204951f52e34c1016a7d189ff9ab99a0a7418L457-R457) [[4]](diffhunk://#diff-ae12be77883993fc7d2b5ebc501e2fb0b0078ea1d4645c2f9327095ae0661d30L612-R615) [[5]](diffhunk://#diff-e7e74230423ceb11899f9ed2ca5ca572cc331e8214a61727c2f71b88fd610bceL315-R324)

**Histogram Parsing Robustness:**

* Refactored histogram parsing to use `ParseUint` instead of `ParseInt`, preventing negative values and ensuring only valid, non-negative frequencies are processed. Added logic to cap at zero if counters reset and to ignore excess buckets, avoiding panics. Comprehensive tests verify correct behavior for normal, overflow, counter reset, and invalid data cases. (`agent/agents/postgres/pgstatmonitor/pgstatmonitor.go`, `agent/agents/postgres/pgstatmonitor/pgstatmonitor_test.go`) [[1]](diffhunk://#diff-ae12be77883993fc7d2b5ebc501e2fb0b0078ea1d4645c2f9327095ae0661d30L728-R758) [[2]](diffhunk://#diff-55635f56d85ff6db15c97d0d4baf5a450931b4724b269ea35afd65fb9621cdeeR28) [[3]](diffhunk://#diff-55635f56d85ff6db15c97d0d4baf5a450931b4724b269ea35afd65fb9621cdeeR598-R670)

**Linting and Code Quality:**

* Updated `.golangci.yml` to disable the `gosec` linter for tests, acknowledging that security checks are less critical in test code and would otherwise require excessive suppression.

These changes together improve the reliability, maintainability, and correctness of the agents, especially in production environments where resource leaks or unhandled errors could cause subtle issues.


agent/agents/mongodb/profiler/internal/collector/collector.go
```diff
@@ -63,7 +63,7 @@ type Collector struct {
 }
 
 // Start starts but doesn't wait until it exits.
-func (c *Collector) Start(context.Context) (<-chan proto.SystemProfile, error) {
+func (c *Collector) Start(ctx context.Context) (<-chan proto.SystemProfile, error) {
```
ademidoff, line 66: The `ctx` threaded in here (and into `parser.Start`) is `context.Background()` in production: `Profiler.Start` hardcodes `ctx := context.Background()` at `agent/agents/mongodb/profiler/internal/profiler.go:114` and passes it down through `start` -> `Monitors.MonitorAll` -> `Monitor.Start` -> here. So the new `case <-ctx.Done()` branches (parser) and the `context.TODO()`->`ctx` swaps (collector) never fire in production; shutdown still relies entirely on `doneChan`/`Stop()`, and the new context-cancel tests exercise a path prod does not use.

Not a regression (behavior is unchanged, `Background()` == `TODO()`), but the "graceful shutdown via context" goal is not reached end-to-end. Suggest either wiring the real ctx down from `Profiler.Start` now, or adding a code comment explaining that the ctx is intentionally not cancelable yet (it is `context.Background()`), so a future reader does not assume context cancellation already stops profiling.

agent/agents/mongodb/profiler/internal/parser/parser_test.go
```diff
@@ -81,6 +81,38 @@ func TestParserStartStop(t *testing.T) {
 	parser1.Stop()
 }
 
+func TestParserContextCancel(t *testing.T) {
+	docsChan := make(chan pm.SystemProfile)
+	a := aggregator.New(time.Now(), "test-id", logrus.WithField("component", "aggregator"), truncate.GetMongoDBDefaultMaxQueryLength())
+
+	ctx, cancel := context.WithCancel(context.Background())
```
ademidoff: Per the project guideline (prefer `t.Context()` in tests), use `context.WithCancel(t.Context())` here to match `TestCollectorContextCancel`, which already does.

agent/agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -725,7 +728,11 @@ func (m *PGStatMonitorQAN) makeBuckets(current, cache map[time.Time]map[string]*
 func parseHistogramFromRespCalls(respCalls pq.StringArray, prevRespCalls pq.StringArray, vPGSM pgStatMonitorVersion) ([]*agentv1.HistogramItem, error) {
 	histogram := getHistogramRangesArray(vPGSM)
 	for k, v := range respCalls {
-		val, err := strconv.ParseInt(v, 10, 32)
+		if k >= len(histogram) {
```
ademidoff, line 735: Good defensive bound (this prevents a real index-out-of-range panic). Consider a `Debugf` when truncating, so silently dropping excess buckets is observable if a future pg_stat_monitor version returns more buckets than the static range array.

agent/agents/postgres/pgstatstatements/pgstatstatements.go
```diff
@@ -312,7 +315,13 @@ func (m *PGStatStatementsQAN) getNewBuckets(ctx context.Context, periodStart tim
 	}
 
 	buckets := m.makeBuckets(current, prev)
-	startS := uint32(periodStart.Unix())
+	unix := periodStart.Unix()
```
ademidoff: This runtime clamp addresses the same `uint32(time)` gosec concern that `aggregator.go`, `perfschema.go`, `slowlog.go`, and `pgstatmonitor.go` handle with a `//nolint:gosec // Unix timestamp fits in uint32 until 2106` comment. Worth unifying on one approach across the PR; the comment-only route is fine since these are all `time.Now()`-derived.

agent/agents/mongodb/mongolog/mongodb.go
```diff
@@ -74,7 +74,10 @@ func (m *MongoDB) Run(ctx context.Context) {
 	var log Mongolog
 
 	defer func() {
-		log.Stop() //nolint:errcheck
+		err := log.Stop()
```
4nte: `Profiler.Stop()`  and `Mongolog.Stop()`  return `nil` on every path. So:

```go
err := prof.Stop()
if err != nil {  // <----- dead code
    m.l.Errorf("Can't stop profiler, reason: %v", err)
}
```

The one error `Stop()` can actually produce, `p.client.Disconnect(ctx)`, is logged and dropped at `profiler.go:151-154`. The `//nolint:errcheck` was removed but the silence moved one frame down, which is the opposite of what the PR sets out to do. This is now more reachable than before, because the PR added a 5s bound to Disconnect.

Fix: `return fmt.Errorf("failed to disconnect from MongoDB: %w", err)` from `Profiler.Stop()`. If the intent really is best-effort, drop the error return from `Stop()` and delete the caller-side checks rather than leaving them as decoration.

agent/agents/mongodb/profiler/internal/parser/parser_test.go
```diff
@@ -81,6 +81,38 @@ func TestParserStartStop(t *testing.T) {
...
+		parser.Stop()
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		// Success: the parser stopped correctly via context cancellation
+	case <-time.After(1 * time.Second):
+		t.Fatal("Parser did not stop after context cancellation within timeout")
+	}
```
4nte, line 113: This test passes even if you delete both new case `<-ctx.Done()`: return arms from `start()`, so it doesn't actually cover them.

The culprit is `Stop()`. It closes `doneChan`, and both selects in `start()` have always had a `case <-doneChan: return`. So the parser shuts down here whether or not it can see the cancelled context, and the assertion passes either way.

The fix is to not call `Stop()` until after the check, so cancellation is the only thing that can end the goroutine. Waiting on the WaitGroup directly does that. I ran the version below 10x under -race: passes on this branch, fails all 10 with the `ctx.Done()` arms removed.

```go
   // Cancel alone must terminate the goroutine. Don't call Stop() first: it closes
   // doneChan, which would shut the parser down on its own and hide a missing
   // ctx.Done() case.
   cancel()

   exited := make(chan struct{})
   go func() {
           parser.wg.Wait()
           close(exited)
   }()

   select {
   case <-exited:
   case <-time.After(1 * time.Second):
           t.Fatal("parser goroutine did not exit on context cancellation")
   }

   parser.Stop()
```

agent/agents/mongodb/mongolog/mongodb.go
ademidoff: Right on both counts — fixed in 7740384.

`Profiler.Stop()` now returns the disconnect error (`failed to disconnect from MongoDB: %w`) instead of logging it, with `p.running = false` still set before the return so the state transition completes on the error path too. The caller-side check in `agent/agents/mongodb/profiler/mongodb.go` is now live.

`Mongolog.Stop()` turned out to be the other case you described: it has no fallible step at all — the monitor/aggregator/sender stops are void, and the client is disconnected back in `Start()` — so I dropped the error return from it and from the `Mongolog` interface rather than keep a check that can never fire. Nothing else implements or mocks that interface.

agent/agents/mongodb/profiler/internal/parser/parser_test.go
ademidoff: Confirmed, and your repro is exactly right — fixed in 7740384 with your version: wait on the WaitGroup after `cancel()`, call `Stop()` afterwards. Passes 10x under `-race` here, and fails 10/10 with both `ctx.Done()` arms removed from `agent/agents/mongodb/profiler/internal/parser/parser.go`.

The same flaw was in `TestCollectorContextCancel` (`agent/agents/mongodb/profiler/internal/collector/collector_test.go`), which is easy to miss because it needs a live MongoDB — same fix applied there. Verified against a throwaway `mongo:8 --profile=2`: all three collector tests pass, and `TestCollectorContextCancel` fails on its 5s timeout once the `ctx.Done()` arms are stripped from `agent/agents/mongodb/profiler/internal/collector/collector.go`.

agent/agents/mongodb/profiler/internal/collector/collector_test.go
```diff
@@ -178,6 +179,74 @@ func TestCollector(t *testing.T) {
 	assert.Len(t, profiles, maxDocs*maxLoops)
 }
 
+func TestCollectorContextCancel(t *testing.T) {
+	url := "mongodb://root:root-password@127.0.0.1:27017"
```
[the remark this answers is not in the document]
ademidoff: Remember: this is allowed in `_test.go` files.

agent/cmd/pmm-agent-entrypoint/main.go
```diff
@@ -113,7 +113,7 @@ func runPmmAgent(ctx context.Context, commandLineArgs []string, restartPolicy re
 
 func commandPmmAgent(args []string) *exec.Cmd {
 	const pmmAgentCommandName = "pmm-agent"
-	command := exec.Command(pmmAgentCommandName, args...)
+	command := exec.Command(pmmAgentCommandName, args...) //nolint:gosec,noctx
```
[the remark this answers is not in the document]
ademidoff: This PR was getting big, we have a follow up PR, where this will be addressed.

agent/agents/mongodb/profiler/internal/collector/collector.go
```diff
@@ -204,7 +205,7 @@ func connectAndCollect(ctx context.Context, collection *mongo.Collection, dbName
 	}()
 
 	for {
-		for cursor.TryNext(context.TODO()) {
+		for cursor.TryNext(ctx) {
```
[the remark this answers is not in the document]
ademidoff: Fixed in b1c5e26 — that was the one select in `connectAndCollect` still ignoring the context, so thanks for catching it.

One correction on severity: this was not a hang. `Collector.Stop()` closes `doneChan` *before* `wg.Wait()`, so a send blocked on a full `docsChan` was always released at shutdown. What it did mean is that the collector kept waiting for `Stop()` even though the parser had already exited on the same canceled context and would never drain the channel again. Now cancellation is self-sufficient, consistent with the other selects in that function.

I could not build a test that reliably covers the new arm: reaching it needs the goroutine to be holding a decoded document while all 100 buffer slots are taken, and every setup I tried (pre-created collection, filled buffer, cancel inside the cursor window) exited through `cursor.TryNext(ctx)` / `cursor.Err()` instead — verified with a goroutine dump. I dropped the attempt rather than commit a test that passes with the bug present.

agent/agents/mongodb/profiler/internal/collector/collector_test.go
```diff
@@ -178,6 +179,74 @@ func TestCollector(t *testing.T) {
...
+	client, err := createSession(url, "pmm-agent-test")
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
ademidoff: Done in b1c5e26 — all four clients now register a bounded `Disconnect` through a small `disconnectOnCleanup(tb, client)` helper, using the file's existing `mgoTimeoutSessionSync` (5s). Registered immediately after client creation so that, with LIFO cleanup order, it runs after the `cleanUpDBs` cleanups that still need the client.

agent/runner/actions/pt_mysql_summary_action.go
```diff
@@ -97,7 +97,10 @@ func (a *ptMySQLSummaryAction) Run(ctx context.Context) ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to write to temporary file: %w", err)
 	}
-	tmpFile.Close() //nolint:errcheck
+	err = tmpFile.Close()
+	if err != nil {
+		return nil, fmt.Errorf("failed to close temporary file: %w", err)
+	}
```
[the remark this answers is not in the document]
ademidoff: Correct — fixed in b1c5e26 with a deferred `tmpFile.Close()` right after the remove defer, so it runs before the remove and covers the `fmt.Fprint` error path; the double close on the happy path is ignored, as you suggested. Same class of leak as the one already fixed in `agent/runner/jobs/pbm_helpers.go` in this PR.

agent/agents/postgres/pgstatmonitor/pgstatmonitor.go
```diff
@@ -609,7 +616,7 @@ func (m *PGStatMonitorQAN) makeBuckets(current, cache map[time.Time]map[string]*
 					NumQueries:          count,
 					ClientHost:          currentPSM.ClientIP,
 					AgentType:           inventoryv1.AgentType_AGENT_TYPE_QAN_POSTGRESQL_PGSTATMONITOR_AGENT,
-					PeriodStartUnixSecs: uint32(currentPSM.BucketStartTime.Unix()),
+					PeriodStartUnixSecs: uint32(currentPSM.BucketStartTime.Unix()), //nolint:gosec
```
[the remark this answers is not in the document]
ademidoff: Skipping this one. This PR deliberately settles on a single approach for Unix-timestamp casts — `//nolint:gosec` at every site — and an earlier review round had me remove the one runtime clamp that existed (in `agent/agents/postgres/pgstatstatements/pgstatstatements.go`) for exactly that consistency. Re-introducing a range check here would put the PR back to two competing styles.

On the risk itself: `bucket_start_time` comes from the monitored server, so it would take a system clock outside 1970..2106 to wrap, and the consequence would be a wrong `period_start` on a QAN bucket, not memory unsafety or data loss.

qan-api2/db_test.go
```diff
@@ -69,7 +69,7 @@ func cleanupDB(t *testing.T, dbName string) {
 	t.Helper()
 
 	cmdStr := fmt.Sprintf(`docker exec pmm-clickhouse-test clickhouse client --password=clickhouse --query='DROP DATABASE IF EXISTS %s;'`, dbName)
-	out, err := exec.CommandContext(context.Background(), "/bin/sh", "-c", cmdStr).Output() //nolint:gosec
+	out, err := exec.CommandContext(context.Background(), "/bin/sh", "-c", cmdStr).Output()
```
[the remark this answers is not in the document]
ademidoff: The stated defect is not present: `cleanupDB` at line 72 already uses `context.Background()`, not `t.Context()` — the `t.Context()` call is at line 62, in `setupDB`, where it is correct because it runs during the test rather than in a cleanup callback. Your general point about `t.Context()` being canceled before `t.Cleanup` runs is right, it just does not apply to this line.

Adding a timeout so a hung `docker exec` cannot stall the suite is a fair suggestion, but it is pre-existing test code that this PR only touched to delete a now-dead `//nolint:gosec` (gosec is excluded for `_test.go` files as of this PR), so I am leaving it out of scope. The OpenGrep shell-invocation note is likewise pre-existing and covered by that same policy.

4nte (approved):

ademidoff (approved):

PR #5649, opened 2026-07-15
PMM-12279 Enable perf_schema.memory_events collector and add memory panels

theTibi:
Ticket number: PMM-12279

Feature build: SUBMODULES-0

## What / why

PMM's bundled `mysqld_exporter` already ships a `perf_schema.memory_events` collector over `performance_schema.memory_summary_global_by_event_name`, exposing memory usage per `event_name`:

- `mysql_perf_schema_memory_events_used_bytes` (= `CURRENT_NUMBER_OF_BYTES_USED`)
- `mysql_perf_schema_memory_events_alloc_bytes_total`
- `mysql_perf_schema_memory_events_free_bytes_total`

The collector was never enabled by PMM and nothing rendered it. This PR turns it on and graphs it, giving memory observability for PS/PXC (especially relevant in k8s).

## Changes

- **Enable the collector** in both synced generators (one lightweight global-summary query, LR resolution):
  - `managed/services/agents/mysql.go` — add `--collect.perf_schema.memory_events` to the LR args.
  - `managed/services/victoriametrics/scrape_configs.go` — add `perf_schema.memory_events` to the LR scrape options, keeping the two lists in sync.
- **Golden tests** updated: `mysql_test.go`, `scrape_configs_test.go`, `victoriametrics_test.go`.
- **Dashboard** — new "Performance Schema Memory" section in *MySQL Performance Schema Details*: current bytes used (topk by event), allocation rate, and free rate, plus a "Memory Event" filter variable. Panels degrade to "no data" when memory instruments are not enabled server-side (`performance-schema-instrument='memory/%=ON'`).

## Testing

Validated end-to-end against two monitored instances, **Percona Server 8.0.46** and **MySQL 9.1.0**:

- Both `mysqld_exporter` processes carry `--collect.perf_schema.memory_events`.
- All three metrics present in VictoriaMetrics (~215–220 event series each); values match the source query (top consumer `innodb/buf_buf_pool` = 130.9 MB on both).
- `up=1` on all scrape jobs; no scrape errors and no parse/"unsupported" warnings on MySQL 9.1.
- All three new dashboard panels render populated in the PMM UI for both versions.
- `TestMySQLdExporterConfig*` and `TestScrapeConfig` pass; both packages build.

No regression on 8.0; works on the newer 9.x series.

- [ ] API Docs updated (n/a — no API endpoint changes)

- https://perconadev.atlassian.net/browse/PMM-12279

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TNZLMVAW9gHXcccq8uYBip

---
_Generated by [Claude Code](https://claude.ai/code/session_01TNZLMVAW9gHXcccq8uYBip)_


managed/services/agents/mysql.go
```diff
@@ -52,6 +52,7 @@ func mysqldExporterConfig(
 		"--collect.info_schema.clientstats",
 		"--collect.info_schema.userstats",
 		"--collect.perf_schema.eventsstatements",
+		"--collect.perf_schema.memory_events",
```
JiriCtvrtka, line 55: Since collector is enabled globally, I think this collector fails on MariaDB versions before 10.5.2 because the required table does not exist. If so could we add message and include it in tests?

I just added message about compatibility on different MySQL dashboard and it looks like this: 
Screenshot 2026-07-16 at 10 53 12

Because it is hard to tell if "NO DATA" are because of error or I am using unsupported version.

EDIT: My mistake, apparently we no longer officially support MariaDB.

dashboards/dashboards/MySQL/MySQL_Performance_Schema_Details.json
```diff
@@ -3547,6 +3547,329 @@
...
+            "pointradius": 5,
+            "points": false,
+            "renderer": "flot",
+            "seriesOverrides": [],
+            "spaceLength": 10,
+            "stack": false,
+            "steppedLine": false,
+            "targets": [
+                {
+                    "expr": "topk(10, rate(mysql_perf_schema_memory_events_free_bytes_total{service_name=~\"$service_name\", event_name=~\"$mem_event_name\"}[$interval]))",
```
JiriCtvrtka: What if interval is set to 30s? This collector runs at low resolution, but the queries use rate(...[$interval]). When $interval is shorter than the scrape interval, rate() does not have enough samples and could returns no data.

dashboards/dashboards/MySQL/MySQL_Performance_Schema_Details.json
```diff
@@ -3547,6 +3547,329 @@
...
+            "pointradius": 5,
+            "points": false,
+            "renderer": "flot",
+            "seriesOverrides": [],
+            "spaceLength": 10,
+            "stack": false,
+            "steppedLine": false,
+            "targets": [
+                {
+                    "expr": "topk(10, rate(mysql_perf_schema_memory_events_alloc_bytes_total{service_name=~\"$service_name\", event_name=~\"$mem_event_name\"}[$interval]))",
```
JiriCtvrtka: I think it could be same case as https://github.com/percona/pmm/pull/5649#discussion_r3595010729

dashboards/dashboards/MySQL/MySQL_Performance_Schema_Details.json
```diff
@@ -3919,6 +4242,38 @@
...
+                    "selected": true,
+                    "text": [
+                        "All"
+                    ],
+                    "value": [
+                        "$__all"
+                    ]
+                },
+                "datasource": "Metrics",
+                "definition": "label_values(mysql_perf_schema_memory_events_used_bytes, event_name)",
```
mattiasimonato: Should this be scoped to the service?

```
label_values(mysql_perf_schema_memory_events_used_bytes{service_name=~"$service_name"}, event_name)
```

Right now it lists event names from every MySQL instance in the install, and the list doesn't change when you switch `$service_name`. The other variables here (`node_name`, `version`, `username`) all scope. With `refresh: 2` it also re-runs unscoped on every time range change.

Also, was it on purpose to make it visible (`hide: 0`)? It adds a "Memory Event" dropdown at the top of the whole dashboard, but it only drives the 3 new panels at the bottom.

dashboards/dashboards/MySQL/MySQL_Performance_Schema_Details.json
```diff
@@ -3547,6 +3547,329 @@
...
+        {
+            "collapsed": false,
+            "datasource": null,
+            "gridPos": {
+                "h": 1,
+                "w": 24,
+                "x": 0,
+                "y": 200
+            },
+            "id": 1100,
```
mattiasimonato: Can we move this row above `MySQL Summary` (id 293) and `Node Summary` (id 291)?

Those two are the standard last rows in all the MySQL dashboards (InnoDB Details, Instance Summary, Replication Summary all end with them), so the new section reads a bit lost below them.

Small thing while you're there: `y: 200/201/210` looks arbitrary. Grafana repacks it so nothing breaks, but the previous row is at `y: 81`.

mattiasimonato (commented):
Nice one, the exporter side is clean and I like that both synced lists got updated. Left a few inline questions. Just one more things from me:

1. Could we update `documentation/docs/reference/dashboards/dashboard-mysql-performance-schema-details.md`? It lists the sections of this dashboard and the new one is missing. More importantly the ticket title is basically "show data if `performance-schema-instrument='memory/%=ON'`", and right now that instruction only lives in a panel tooltip. Feels like it should be in the docs.

managed/services/agents/mysql.go
mattiasimonato, line 55: Was always-on the agreed call here, or is it still open?

The triage note on the ticket listed it as an open question (tablestats-group toggle vs. always-on) and I don't see it answered anywhere.

Not saying it's wrong, I lean the same way, the query is one cheap global summary and nothing like the per-table cost the tablestats toggle exists for. But by your own numbers it's ~215 events x 3 metrics, so roughly 650 extra series per MySQL service, on by default, for everyone after upgrade. On a few hundred services that adds up. Can we get someone to sign off on that explicitly rather than it slipping in? Or did that already happen offline?

(I checked the version-gate side and it's fine, no `IsFeatureSupported` needed. `ScrapePerfMemoryEvents` is in the scrapers map from `pmm-2.42.0` through `main`, so the flag exists on any client we care about. Just noting it so nobody else has to go look.)

JiriCtvrtka, line 55: Enabling perf_schema.memory_events for every MySQL service unconditionally is a significant cardinality change. Measured on an idle mysql:8.0.46: 212 of 502 instruments report non-null, × 3 metrics = ~636 new series per instance, and a busy server pushes that toward ~1500. These are scraped at the LR interval and stored unfiltered.

The only per-service gate in this file is the tablestats group's table-count limit, so this would be the largest collector turned on for the whole fleet by default. Could we either put it behind a service-level opt-in / agent flag, or drop the instrument set to the ones that actually get looked at?

theTibi: @Tusamarco can you advise here? should this be collected by default all the time?

Tusamarco: My Point is that if we do not have performance impact in collecting the events at MySQL level this should be ON by default. If this cause a problem for you in VM then this is a different issue.
In either case I would agree in having the option to disable as we have for tablestat, but if we do that then I think we should have it also for all other major consumers, to be consistent and modular.
However the memory usage information is one crucial aspect of the service health and as said I would like to have it ON by default. 
@theTibi

theTibi:
Thanks all — a follow-up commit is coming with everything below.

**Decision on always-on:** settled by @Tusamarco above — the collector stays enabled by default. No `IsFeatureSupported` gate, no new toggle in this PR. The Go side is therefore unchanged from what you already reviewed. Per-collector opt-out already exists today (details in the thread on `mysql.go:55`).

**Review items addressed**

| Item | Change |
|---|---|
| Variable not scoped to service (@mattiasimonato) | `label_values(mysql_perf_schema_memory_events_used_bytes{service_name=~"$service_name"}, event_name)`, `refresh: 2 → 1` |
| Section sits below `MySQL Summary` / `Node Summary` (@mattiasimonato) | Row moved above both; `gridPos` now contiguous (80/81/89, then 97/98) |
| `rate(...[$interval])` fallback (@JiriCtvrtka, CodeRabbit) | `>0` added to the `irate` branch too, so both branches match the other 11 rate panels in this file |
| Docs missing the new section + the `memory/%=ON` prerequisite (@mattiasimonato) | `documentation/docs/reference/dashboards/dashboard-mysql-performance-schema-details.md` — three list entries plus a "Server configuration required" note with the config snippet |

**Also fixed while in there, not raised by anyone**

- Panel titles used em dashes (`Memory Usage — Current Bytes Used (Top 15 by Event)`). There isn't a single em-dash panel title anywhere else in `dashboards/dashboards/`, so they now follow house style: `Performance Schema Memory Usage (Current Bytes)`, `... Memory Allocation Rate (Bytes)`, `... Memory Free Rate (Bytes)`.
- The three queries skipped the `avg by (service_name,event_name)` wrapper that every other perf-schema panel here uses. Added, for consistency and to collapse duplicate label sets.
- Panel 1101 had `yaxes[].min: 0`. `CURRENT_NUMBER_OF_BYTES_USED` can legitimately go negative for instruments whose memory was allocated before instrumentation was enabled, and `min: 0` clipped that out of view silently. Removed.
- `gridPos.h` on panel 1101 was 9; every other graph in this dashboard is 8.
- Panel descriptions now name the actual source columns (`SUM_NUMBER_OF_BYTES_ALLOC` / `SUM_NUMBER_OF_BYTES_FREE`) and repeat the server prerequisite on all three panels rather than only the first.

**Verification of this round**

- `cleanup-dash.py --check-only` clean; JSON round-trips byte-identically to the canonical `sort_keys/indent=4` form, so the diff contains only the intended changes (21 changed leaves — the rest of the churn is the row relocation).
- All 15 targets in the dashboard parse against VictoriaMetrics 2.24.0 with PMM's flags.
- Replayed the three panels against 215 instruments × 2 services at the 60s LR cadence: 15/10/10 series as intended, `event_name` and `service_name` survive the `avg by()`, and results are identical at `$interval` of 1m and 18s.
- `TestMySQLdExporterConfig*` and `TestScrapeConfig` pass.

One correction to the record while verifying the rate fallback — see the thread on the rate queries. The "no data when `$interval` < scrape interval" failure mode does not reproduce on VictoriaMetrics; MetricsQL rollups look back past the window. The fallback is kept for consistency with the rest of the file, but it isn't load-bearing.

**Not doing, deliberately**

- Keeping the `Memory Event` variable visible (`hide: 0`) — reasoning in @mattiasimonato's thread.
- Not adding `memory/%=ON` to the `connect-database/mysql/mysql.md` table. That table is scoped to what QAN needs; putting it there would read as a QAN requirement. Happy to add it if you'd rather have it in both places.

dashboards/dashboards/MySQL/MySQL_Performance_Schema_Details.json
theTibi: Fixed — and worth recording what I found while verifying it, because the premise turns out not to hold on VictoriaMetrics.

Both panels now use the same shape as every other rate panel in this file:

```
topk(10, avg by (service_name,event_name) (rate(...[$interval]))>0) or topk(10, avg by (service_name,event_name) (irate(...[5m]))>0)
```

The earlier fix in 1b2fce8 had `>0` on the `rate` branch only; both branches have it now, matching panels 47–49, 52 and 77–80.

Your concern was actually sharper than you framed it — but only under Prometheus semantics. `$interval` defaults to `$__auto_interval_interval` with `auto_count: 200, auto_min: 1s`, so on any range under ~3.3h the auto step is already below the 60s LR cadence. It was never just the manual `1s`/`5s` options.

Except it doesn't reproduce. Measured against the VM we ship (v2.24.0, PMM's own flags, no staleness overrides), 215 instruments × 2 services at a 60s cadence:

```
rate([ 1s]) -> 430 series, of which >0: 305
rate([ 5s]) -> 430 series, of which >0: 305
rate([18s]) -> 430 series, of which >0: 305
rate([30s]) -> 430 series, of which >0: 305
rate([ 1m]) -> 430 series, of which >0: 305
rate([ 5m]) -> 430 series, of which >0: 305
```

MetricsQL rollups look back past the window to the previous sample, so `rate()` doesn't go empty on a sub-scrape-interval window the way Prometheus' does. The `or irate(...[5m])` pattern throughout this dashboard is a Prometheus-era carry-over — harmless, and I've kept it so these panels don't look different from their neighbours, but it isn't load-bearing on VM.

dashboards/dashboards/MySQL/MySQL_Performance_Schema_Details.json
theTibi: Scoped, thanks:

```
label_values(mysql_perf_schema_memory_events_used_bytes{service_name=~"$service_name"}, event_name)
```

Also dropped `refresh` from 2 to 1, so it re-runs on dashboard load rather than on every time-range change — matching `username`, `schema` and `database`, which are the other high-cardinality scoped variables here. Good catch on that part.

On `hide: 0` — that one is deliberate and I'd like to keep it. With ~215 instruments and `topk(15)`, the top consumers are nearly always the same handful of InnoDB buffers; the filter is the only way to see past them. Setting `hide: 2` would pin the variable at `.*` permanently and make the three panels un-drillable, which costs more than the tidier top bar buys.

You're right that it's the only visible variable driving a subset of panels. If that bothers you more than losing the drill-down, say so and I'll flip it — but my preference is to keep it, and it's now called out on the dashboard's docs page so the behaviour isn't a surprise.

dashboards/dashboards/MySQL/MySQL_Performance_Schema_Details.json
theTibi: Done. `Performance Schema Memory` now sits after panel 80 and before `MySQL Summary` (293) and `Node Summary` (291), and the coordinates are contiguous rather than arbitrary:

| Panel | Before | After |
|---|---|---|
| row 1100 | `y: 200` | `y: 80` |
| 1101 | `y: 201, h: 9` | `y: 81, h: 8` |
| 1102 / 1103 | `y: 210` | `y: 89` |
| 293 `MySQL Summary` | `y: 80` | `y: 97` |
| 291 `Node Summary` | `y: 81` | `y: 98` |

`h: 9 → 8` on 1101 while I was in there — every other graph in this dashboard is 8.

Fair warning that the relocation makes the JSON diff look much bigger than the change is: reordering the `panels` array moves ~600 lines. The semantic delta is 21 changed leaves, and `cleanup-dash.py --check-only` is clean.

managed/services/agents/mysql.go
theTibi: Answered by @Tusamarco above: on by default. Memory usage is core service-health data, the server-side cost is one global summary query, and series volume on the VM side is a separate concern rather than a reason not to collect. So no `IsFeatureSupported` gate and no new toggle in this PR — the Go side stays as you reviewed it.

Thanks for pre-checking the version gate. Agreed, and the client floor is 3.0.0 regardless (`NodeExporterNewTLSConfig`; the `2.36.0` constant next to it is vestigial), so every supported client has the scraper.

On the opt-out @Tusamarco wants: it already exists and needs no new code. `mysql.go:111` runs

```go
args = collectors.FilterOutCollectors("--collect.", args, exporter.ExporterOptions.DisabledCollectors)
```

mirrored on each resolution list in `scrape_configs.go`, so `--disable-collectors=perf_schema.memory_events` on `pmm-admin add mysql` already turns this off per service and keeps the exporter args and the scrape config in agreement. Making that consistently discoverable across all the major consumers, as he suggests, is a reasonable separate ticket rather than something to bolt onto this one.

theTibi: Your measurement matches mine and the concern is a fair one to raise, but three of the specifics don't hold — and @Tusamarco has called the decision above: on by default.

**1. Not the largest default-on collector, by a wide margin.** `--collect.perf_schema.eventsstatements` is already unconditional in the same LR list (`mysql.go:54`), and `eventsstatements.limit` defaults to 250 digests × 15 metric descriptors — up to ~3750 series per instance. That's roughly 6× memory_events' ~636, on by default, today.

**2. A per-service gate already exists.** Beyond the tablestats group, `mysql.go:111` runs `collectors.FilterOutCollectors("--collect.", args, exporter.ExporterOptions.DisabledCollectors)`, mirrored in `scrape_configs.go`, so `pmm-admin add mysql --disable-collectors=perf_schema.memory_events` already disables this per service with no new API surface. Inverting it to opt-in would need a new inventory field, a pmm-admin flag and UI, for a collector that on your own numbers is a sixth of eventsstatements.

**3. Reducing the instrument set isn't available today.** The collector's only tunable is `--collect.perf_schema.memory_events.remove_prefix`. There's no instrument filter, and `scrape_configs.go` has no `metric_relabel_configs` plumbing to drop them at ingest, so this would be an upstream `mysqld_exporter` change — its own ticket, not a review fix.

One point in your favour worth writing down: the ~1500 figure isn't load-driven, it's just the instrument count. The collector query is

```sql
... FROM performance_schema.memory_summary_global_by_event_name WHERE COUNT_ALLOC > 0
```

so the ceiling is 502 × 3 = 1506 on 8.0.46 no matter how busy the server gets. It grows as instruments are touched for the first time and then plateaus — worth knowing for capacity planning, but it's a bounded step change rather than something that scales with workload.

theTibi:
@coderabbitai review

theTibi:
Pushed as fd44b398d — everything described in my comment above is now on the branch. The PR is 7 files, `+389/-2`.

The dashboard diff looks larger than the change is: moving the `Performance Schema Memory` row above `MySQL Summary` reorders the `panels` array, so ~600 lines shift position. Against `main` it nets out to `+355/-2` on that file, the `-2` being the two `gridPos.y` values on rows 293 and 291.

mattiasimonato (approved):

PR #5653, opened 2026-07-16
PMM-15216 Migrate the SEP UI into PMM

fabio-silva:
Ticket number: PMM-15216

Feature build: [https://github.com/Percona-Lab/pmm-submodules/pull/4474](https://github.com/Percona-Lab/pmm-submodules/pull/4474)

> **Stacked PR.** Base is #5728 (PMM-15288, the pnpm + oxlint/oxfmt + library upgrade). Review that one first; the diff shown here is only the UI migration on top of it. This branch was rebuilt from `main` to split the two, so the commit history changed — the previous tip was `9a507d0c`.

## What

Brings SEP's frontend into `ui/` and mounts the migrated plugins as native PMM routes, so SEP surfaces render inside the PMM shell instead of an iframe.

### Packages ported from SEP's frontend workspace

| PMM path | From |
|---|---|
| `ui/packages/sep/api` | typed API client, generated OpenAPI surfaces, hooks |
| `ui/packages/sep/framework` | schema-driven form/list/task components |
| `ui/packages/sep/shared` | shared primitives |
| `ui/packages/plugins/atw` | Collect Diagnostic Data (ATW) |

SEP's "app" vocabulary is renamed to "plugin" throughout the port, since "app" already means a workspace app in `ui/`: `SchemaDrivenApp` → `SchemaDrivenPlugin`, `useAppSchema` → `usePluginSchema`, `useAppTasks` → `usePluginTasks`, `app-schema.ts` → `plugin-schema.ts`. Ported files also carry PMM's AGPL header.

### Wiring in `apps/pmm`

- `router.tsx` mounts the plugins under their own routes; `SepPage` gives them the standard PMM `Page` chrome (padding, width, auth gate, footer).
- Navigation gates the SEP entries behind admin + the inventory settings flag.
- `main.tsx` calls `initSepAuth`, pointing SEP's axios client at PMM's session. Auth is still the interim Option D: the dev proxy injects `SEP_INTERNAL_TOKEN` server-side, so no token reaches the browser.
- `vite.config.ts` proxies SEP's paths (`/api`, `/sep_app`, `/stream-logs`, `/execution-events`, `/files`) to `SEP_BACKEND_URL`, and lets `PMM_SERVER_URL` override the PMM target.
- A `SyntaxHighlighter` component backs the schema renderer's script/JSON fields.
- `Page` gains `maxWidth` so SEP pages can opt into the full-width container from `@percona/percona-ui`; `Settings.tsx` moves to it in place of the removed `fullWidth` flag.

## SEP catch-up (second commit)

SEP moved on after the initial port. Ported commit by commit rather than by copying files, so the app→plugin rename and license headers survive:

- **SEP-1629 / SEP-1684 / SEP-1689 / SEP-1696 / SEP-1668** — refresh the generated OpenAPI surface from SEP head. Two schema components are now namespaced: `ConnectivityWarning` → `framework__ConnectivityWarning`, `TaskExecuteWrite` → `framework__TaskExecuteWrite`; their consumers move with them.
- **SEP-1663** — honor `HostRef` / `HostField.allow_custom`. `HostField` passes it to `HostSelector`, which renders `FreeSoloSelect` instead of the closed `AutoCompleteInput` and commits a scalar id/string (including from cascade auto-select). `FreeSoloSelect` now resolves a stored string against option **ids**, not just labels, so string host ids like `"nomad-1"` display as their option.
- **SEP-1653** — hide the task-history **Download files** button unless the files API returns a non-empty listing. `has_logs` was the wrong signal: logs exist even when the output dir holds only the hidden `.sep-run-result.json` marker, leaving a dead download action. Probes are cached for 30s so the history table's poll loop doesn't re-hit the files API every tick.
- **SEP-1692** — add `postSession` / `postSessionExchange` to `@sep/api`. The exchange endpoint trades PMM's session cookie for a short-lived SEP bearer, which is what replaces the interim `SEP_INTERNAL_TOKEN` wiring (that token's service principal hardcodes `is_admin = False` and so 403s every admin-gated surface).

### Deliberately not ported

- **SEP-1663's multi-host half** (`FreeSoloMultiSelect`, `MultiHostField`) — PMM's snapshot has no multi-host selector to extend. Porting it means first porting SEP's multi-host feature, which is its own change.
- **Flipping `bootstrap.ts` to the token exchange (Option B)** — only the `@sep/api` client surface lands here. Switching over needs a SEP backend carrying `POST /api/oauth/session/exchange` and live verification, so it gets its own ticket.

## Review comments addressed

- **@ademidoff** — *"I don't really understand what this PR has to do with the entrypoint and `postgres-migration`. Please remove that code, it's a separate PR."* Removed. `build/ansible/roles/postgres/*`, `build/docker/server/entrypoint.sh`, `docker-compose.yml`, `.env.example` and `managed/utils/envvars` are no longer touched by either branch; those changes belong to #5700.
- **@ademidoff** (*pnpm and yarn*) and **@maxkondr** (*pnpm version in the Dockerfile*) — both addressed in the base PR, #5728, where the Dockerfiles live.

Also gitignores `*.tsbuildinfo`; one had been committed by accident.

## Verification

From `ui/`, on this branch (i.e. including the base PR):

| Check | Result |
|---|---|
| `pnpm install` | clean |
| `make format-check` (oxfmt) | 769 files, all formatted |
| `make lint` (oxlint) | 0 errors across all 6 packages |
| `make test` (vitest) | 116 files passed, 1084 tests passed, 1/13 skipped |
| `make build` | pmm + pmm-compat build |
| `turbo run check-types` | clean |

Not yet verified against a live PMM Server with a SEP backend attached — that needs a feature build, which is tracked below.

## Related

- Base: #5728 (PMM-15288, toolchain + library upgrade)
- PostgreSQL/entrypoint changes: #5700

- [ ] API Docs updated


build/docker/rpmbuild/Dockerfile.el8
```diff
@@ -29,6 +29,9 @@ RUN dnf update -y && \
     dnf install -y --enablerepo=ol8_codeready_builder glibc-static && \
     dnf clean all && rm -rf /var/cache/dnf
 
+# UI build uses pnpm (packageManager in ui/package.json); nodesource nodejs ships npm/yarn only.
+RUN npm install -g pnpm@11.1.3
```
maxkondr: Is it possible to pass ppm version during docker build (read from some file in repo)? It will simplify the maintenance in future (less files are required to be changed during version bump).

build/docker/rpmbuild/Dockerfile.el9
```diff
@@ -29,6 +29,9 @@ RUN dnf update -y && \
     dnf install -y --enablerepo=ol9_codeready_builder glibc-static && \
     dnf clean all && rm -rf /var/cache/dnf
 
+# UI build uses pnpm (packageManager in ui/package.json); nodesource nodejs ships npm/yarn only.
+RUN npm install -g pnpm@11.1.3
```
maxkondr: the same as above

yyyyyyyan:

build/docker/rpmbuild/Dockerfile.el8
ademidoff: Do we really need both `pnpm` and yarn?

build/docker/server/entrypoint.sh
```diff
@@ -8,6 +8,7 @@ declare CURRENT_GID CURRENT_UID CURRENT_USER
 is_enabled() { [ "$1" = "1" ] || [ "$1" = "true" ]; }
 declare POSTGRES_DATA_DIR="/srv/postgres14"
 declare POSTGRES_PASSWORD_FILE="/srv/.postgres_password"
+declare POSTGRES_BIN_DIR="/usr/pgsql-14/bin"
```
ademidoff: @fabio-silva I don't really understand what this PR has to do with the entrypoint and `postgres-migration`.

Please remove that code, it's a separate PR that needs to be tested stand-alone and merged separately, possibly even prior to this one.

https://github.com/percona/pmm/pull/5700

ademidoff (changes requested):

build/docker/rpmbuild/Dockerfile.el8
nachodd: Good call — done, and the hardcoded version is gone entirely rather than being passed in as a build arg.

All three rpmbuild Dockerfiles now just do:

```dockerfile
RUN corepack enable
```

corepack ships with nodesource's nodejs 22 and resolves pnpm from the `packageManager` field in `ui/package.json` at build time, so a pnpm bump only touches that one field. `ui/Makefile`'s `setup` target does `corepack enable pnpm` for the same reason, and `.devcontainer/setup.sh` no longer pins a version either.

This moved to the split-out toolchain PR: #5728.

build/docker/rpmbuild/Dockerfile.el9
nachodd: Same fix as on el8 — see the reply above. Applied to el8, el9 and hetzner-el9 in #5728.

build/docker/rpmbuild/Dockerfile.el8
nachodd: Yes — they cover different projects, and both are built from this image:

- `ui/` (the PMM UI workspace) is pnpm.
- `dashboards/pmm-app` (the QAN Grafana plugin) is still yarn 1 — `packageManager: yarn@1.22.21`, and `dashboards/Makefile` runs `yarn install --frozen-lockfile`.

Migrating pmm-app off yarn is out of scope here, so the image needs both until that happens.

I've made the Dockerfile say so, and dropped the hardcoded pnpm version at the same time (per @maxkondr's comment above):

```dockerfile
# The UI workspace (ui/) builds with pnpm; the QAN Grafana plugin (dashboards/pmm-app)
# is still on yarn 1, so both are needed. corepack ships with nodesource's nodejs 22
# and resolves pnpm from the `packageManager` field in ui/package.json, so the version
# is pinned in exactly one place and needs no change here on a bump.
RUN corepack enable
```

This moved to the split-out toolchain PR: #5728.

build/docker/server/entrypoint.sh
nachodd: Agreed, and removed. Neither branch touches that code any more — `build/docker/server/entrypoint.sh`, `build/ansible/roles/postgres/*` (including the new `postgres-sep` script), `docker-compose.yml`, `.env.example` and `managed/utils/envvars` are all gone from the diff. It's yours to land stand-alone in #5700.

While removing it we also split the rest of the PR in two, since the toolchain change and the UI migration were reviewing as one 320-file diff:

- **#5728 (PMM-15288)** — pnpm + oxlint/oxfmt + the library upgrades (React 19, RR7, Vite 8, Vitest 4, TS 6). Targets `main`. This is where the Dockerfile/devcontainer/CI changes now live.
- **#5653 (this PR, PMM-15216)** — the SEP UI migration only, now stacked on top of #5728.

That meant rebuilding this branch from `main`, so its history changed; the previous tip was `9a507d0c`.

ademidoff: Great, thanks!

@Nailya Let's give priority to testing #5700, we need it merged asap.

nachodd:
@coderabbitai Thanks — went through all 30 major findings. 24 applied in bb8c522d6, 6 declined with reasons below.

**Applied — correctness / stability**

| Finding | What changed |
|---|---|
| `client.ts` `_onRefreshed` | Handler call wrapped in `try/catch` so a throwing handler can't reject the shared refresh promise. Confirmed: the call sat inside the async executor, contradicting its own comment. |
| `SchemaFormRenderer` section violations | Gate moved into `handleSubmitEvent` (`preventDefault` + return) and dropped from `handleFormSubmit`, so RHF never sets `isSubmitSuccessful` on a blocked submit. Added a regression test that fails on the old code. |
| `PluginTaskEditPage.normalizeChoiceDefaults` | Now reads/writes via `getAtPath`/`setAtPath`, covering dotted one-of names. No `structuredClone` needed: `setAtPath` shallow-clones the intermediates it walks (see its docstring), so `{...form}` already leaves the input untouched — test asserts that. |
| `PluginListPage` / `PluginDetailPage` `list_view` | `PluginListPage` renders `Not found` when the optional `list_view` is absent (guard placed after all hooks); `OverviewTab` falls back to an empty column set. `PluginDetailPage:917` needed no change — that value is only dereferenced inside the `multi` branch, where `PluginEntitySchema.list_view` is required. |
| `SchemaListView.formatCellValue` | Loose null check; `date`/`relative`/`chip` no longer render `Invalid Date` / `NaNd ago` / `undefined`. Test added. |
| `HostSelector` error lookup | `get(errors, name)` from react-hook-form. |
| `extractId` | Decimal-integer regex + `Number.isSafeInteger`, so `'   '`, `'1.5'`, `'0x10'` return `null`. New test file. |
| `validationMapper` | Numeric `validate` rule (`Number.isFinite`, `Number.isInteger` for integers) and `Number()` in `coerceFormValues`. |
| `useTaskLogs` stepless lines | `typeof step === 'string'`; `''` already routes to the `STEPLESS_STEP_KEY` "General" bucket in `LogStepTabs`. |
| `useExecutionEvents` retries | Consecutive-failure counter (cap 5, reset on successful open) that sets `sseError`, clears `sseLoading`, and re-throws to stop the loop. |
| `StandaloneHostSelector` | `disabled` no longer includes `isError`, so `onOpen` can retry; the existing test was inverted to cover the new behavior. |
| `ScheduledTasksPanel` `kwargs` | Preserved when the response carries it, `'{}'` only as fallback. |
| `useCascadingField` | `''` sentinel in both the clear path and `ready`. |

**Applied — contract / conventions**

- `SepPage` now gates on the nav's own predicate (`user.isPMMAdmin`) via `Page`'s `roles`, so direct navigation is denied too. Note `isPMMAdmin` is `isGrafanaAdmin || orgRole === Admin`, which `roles` alone can't express — hence the computed value rather than a bare `[OrgRole.Admin]`.
- `renderEditForm` slot in `SchemaDrivenPlugin` now receives `capabilities`, `submitError`, `fieldErrors`, matching `PluginTaskEditPage`.
- Shared `SEP_ATW_PATH` / `SEP_MYSQL_BACKUPS_PATH` in `lib/constants.ts`, consumed by both `router.tsx` (paths + `routeBase`) and `navigation.utils.tsx`.
- Renamed to `PMM_DEV_SEP_BACKEND_URL` / `PMM_DEV_SEP_INTERNAL_TOKEN` (`vite.config.ts`, `turbo.json`, comments).
- `SchemaListView` table surface is now mode-aware instead of pinning `common.white` — PMM does ship dark mode (`hooks/theme.ts`), so this was a real bug. `pmmThemeOptions` lives in `@percona/percona-ui` and is out of scope here.
- `FileField` picker button has an `aria-label`.
- `useResolvedServiceField` exposes `isError` / `error`.
- Deleted the orphaned `framework/test/setup.ts` (unreferenced; `vitest.config.ts` uses `tests/setup.ts`).

**Declined**

1. **Production SEP routing / auth (`bootstrap.ts`), and `/files/*` for `useTaskHistoryFiles`** — real gap, but intentional and documented: this is interim Option D (dev proxy + internal token). Production routing plus the `postSessionExchange()` provider (Option B, SEP-1692) is tracked separately; inventing it here isn't a review fix.
2. **`CollectPane` dotted-name flattening** — dotted names denote nested backend models in SEP: `setAtPath`/`coerceFormValues` produce nested objects for every other plugin's submit, so passing them through is the consistent behavior. Flattening to literal `"source.path"` keys would need a backend-contract confirmation first.
3. **`ScriptPreviewField` → TanStack Query** — deferring. It's a live debounced preview with abort handling; converting it also means adding a query client to the field-level form tests. Worth doing, but not as part of this migration PR.
4. **`freeSoloValue.normalizeChange` committing the raw string** — trimming the committed value is deliberate (it prevents whitespace-padded custom ids). Inner spaces already survive, and MUI does not reset the visible input while the committed value is unchanged, so a trailing space can still be typed.
5. **Test co-location for `@sep/api` and `atw`** — valid per the guideline, but moving those suites means relocating their shared `tests/` helpers (`msw-server.ts`, `queryWrapper.tsx`) and touching vitest config; a mechanical follow-up rather than added diff on this PR. `@sep/framework` is already co-located.
6. **`related_apps` + `entities` on one schema** — no shipped schema declares both, so this would add an untested path. Happy to take it if the schema contract is meant to allow the combination.

`pnpm format`, `pnpm check-types`, `pnpm lint` (0 errors) and `pnpm test` (1092 passing) are all green.

nachodd:
Sorry — the commit was local at the time. Pushed now as `1c5963e44` (rebased onto `db970422c`, the SEP catch-up sync that landed in the meantime; no overlap with these hunks — that commit only added the `remote_choice` case to `fieldDefault`).

Head is now `1c5963e44`, and it contains all of it: `_onRefreshed` wrapped in `try/catch`, the `hasSectionViolations` gate moved from `handleFormSubmit` into `handleSubmitEvent`, `normalizeChoiceDefaults` on `getAtPath`/`setAtPath`, and `extractId` on a decimal-integer test plus `Number.isSafeInteger`. Gates after the rebase: check-types clean, 1092 tests passing, lint 0 errors.

@coderabbitai review

nachodd:
Second pass addressed in `1e1a6ab3e`. Four applied, two declined.

**Applied**

- **`extractId` numeric branch** — correct, and the asymmetry was mine to begin with: I tightened the string branch and left `1.5` passing through the numeric one. Now `Number.isSafeInteger`, with cases for `1.5`, `MAX_SAFE_INTEGER + 1`, and `{ id: 1.5 }` (the recursive path).
- **`kwargs` wire format** — right that a string-only check is a guess. Both shapes are now accepted: a JSON string passes through, a decoded object is re-serialised with `JSON.stringify`, and `'{}'` is left only for a genuinely missing or blank value. On the broader point: a partial-update route for the toggle would delete the whole class of problem, but that is a backend change — the client can only stop making it worse.
- **Log the throwing refresh handler** — agreed, the silent catch traded one failure mode for an invisible one. Logged at error level via the file's existing `console` pattern, with neither the token nor the expiry in the message.
- **Spy restore in teardown** — fixed with `onTestFinished`.

**Declined**

- **Removing `ctrl.abort()` from `useExecutionEvents`' terminal path.** The premise doesn't hold for `@microsoft/fetch-event-source@2.0.1`. From `lib/cjs/fetch.js`:

  ```js
  inputSignal?.addEventListener('abort', () => {
    dispose();      // clears retryTimer and aborts curRequestController
    resolve();
  });
  ```

  `curRequestController` is the library's own per-attempt controller; the signal we pass is `inputSignal`, and aborting it runs `dispose()` — clearing the retry timer and aborting the in-flight request. So the call is effective, not misleading. It is *redundant* alongside `throw err`, which is why I'm keeping it: `finish` and `sep-error` in this same hook both stop the stream by aborting that controller, and having the third terminal path do something different would be the confusing version. Thanks for checking the return contract against the library docs, though — that part matched what I'd concluded.
- **Binding the ATW `pagesByOffset` fixture to `ATW_PAGE_SIZE`.** Sound in isolation, but that file is a mirror of SEP's `packages/apps/atw` test, kept aligned so the periodic sync between the two trees stays a clean no-op. A cosmetic edit on one side only makes the next sync diff noisier for a constant that has never changed. If it moves, it should move in SEP first.

**Two extra fixes** in the same commit, same defect class as the reviewed ones, found while porting this to SEP:

- `HostSelector` had the identical wedge to `StandaloneHostSelector` — both its free-solo and standard branches passed `disabled={disabled || isError}` while `onOpen` held the only `refetch()`. The review's diff hunk didn't reach those lines. Its existing "disables the input on error" test was inverted to cover retry-on-open.
- `SchemaListView.formatCellValue` now renders an absent `date` / `relative` value as an empty cell rather than an em dash, adopting SEP's variant of the guard I ported: an em dash for a task that never ran reads as a placeholder for a time that exists. SEP was ahead of the port here.


Gates: check-types clean, oxlint 0 errors, 1105 tests pass, oxfmt clean.

yyyyyyyan (approved):

PR #5670, opened 2026-07-20
PMM-15228 Improve NGINX and pmm-managed performance

maxkondr:
Ticket number: PMM-15228

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/4484

This pull request makes:
- significant improvements to the NGINX configuration for PMM, focusing on performance optimization, authentication and caching enhancements, and better static asset handling. It also updates code generation and linting configuration files to support new interfaces and generic types.
- adds detailed Prometheus metrics to the `AuthServer` component in order to monitor authentication request flows, cache usage, and request durations. It also ensures that metrics are accurately labeled, especially in error scenarios, and improves code clarity around request path handling.

**Prometheus Metrics Integration:**

* Added a new `authMetrics` struct to `AuthServer` that defines and registers Prometheus counters and histograms for tracking authentication requests, cache hits/misses, Grafana backend requests, cache size, and operation durations. (`managed/services/grafana/auth_server.go`)
* Implemented the `prom.Collector` interface for `AuthServer`, allowing it to be registered with Prometheus and report its metrics. (`managed/services/grafana/auth_server.go`)
* Registered the `AuthServer` as a Prometheus collector in the main application entrypoint. (`managed/cmd/pmm-managed/main.go`)

**Metrics Collection in Auth Flow:**

* Updated the authentication flow to increment the appropriate counters and observe durations for total request time, Grafana backend requests, database lookups, cache hits, and cache misses. (`managed/services/grafana/auth_server.go`)
* Ensured that metrics for failed requests (e.g., bad requests) use cleaned and normalized route labels for consistency and security. (`managed/services/grafana/auth_server.go`)

**Testing and Path Handling Improvements:**

* Added tests to verify that metrics for bad requests use the cleaned route or fall back to the raw route if cleaning fails. (`managed/services/grafana/auth_server_test.go`)
* Centralized and clarified the use of cleaned request paths by ensuring path normalization occurs early and is consistently used throughout the authentication logic. (`managed/services/grafana/auth_server.go`)

These changes provide better observability into authentication operations and improve the maintainability and correctness of metrics reporting.

**NGINX Configuration Improvements:**

* **Authentication and Caching Enhancements:**
- Get rid of an additional auth sub-request in case the initial auth request fails.
- Refactored the authentication flow to use separate locations for cached and non-cached authentication requests (`/auth_request_cached` and `/auth_request_no_cache`), extracting custom headers for richer error responses, and constructing JSON error payloads for 401 responses. Introduced caching for authentication responses to reduce backend load. 
  - Added and documented two dedicated proxy cache paths (`STATIC` for static assets and `AUTH_CACHE` for authentication responses), optimizing cache usage and resource allocation.

* **Performance and Connection Handling:**
  - Increased `keepalive_requests` for upstreams to handle higher loads and reduce connection churn, and added new upstream blocks for `victoriametrics` and `vmalert` with custom connection settings for observability data. Enabled `tcp_nodelay` for low latency.
  - Disabled body size limits and increased buffer sizes for metric endpoints to support large payloads.

* **Static Asset and UI Optimization:**
  - Improved static file serving for PMM UI and Grafana assets with optimized caching headers, file descriptor caching, and sendfile/tcp optimizations. Added dedicated locations for asset caching and bypassed authentication for static assets.

* **Route and Proxy Improvements:**
  - Updated proxy locations to use `^~` for more precise matching, added missing HTTP/1.1 and connection headers, and improved proxy buffering settings for metrics and alerts endpoints. 
  - point `write` requests from vm-agents to VictoriaMetrics directly. Previously the scheme was the following:
  `vm-agent -> NGINX -> vm-proxy -> VictoriaMetrics`. Now it will be `vm-agent -> NGINX -> VictoriaMetrics`. vm-proxy is extra in this scenario, it is involved in *query* metrics from VictoriaMetrics but not in *write* metrics.

* **`PMM-Managed` Improvements:**
- Use an external library to properly setup GOMEMLIMIT
- Replace caches implementations with optimised for high performance in (Auth Server, Agents State)
- Split DB connections pool into 2:
-- one pool for internal stuff (handles internal system logic like init config for VM, Check/Jobs run)
-- one for handling gRPC/REST API requests and interactions with PMM Agents. So that it doesn't interfere with internal stuff.
- Pools sizes are calculated based on available CPUs on start.
- RateLimiters are introduced in the following places:
-- PMM Agent connection handler (agent stream connect and RTA stream connect).
-- PMM Agent State update handler. 
- Get rid of useless DB queries (select common data for all agents, use in-memory data instead of querying DB).
- Optimize objects creations to avoid heap allocation and pressure on GC.


4nte (approved):

managed/services/grafana/auth_server.go
```diff
@@ -168,49 +173,193 @@ type authError struct {
...
 	u       authUser
 	created time.Time
 }
 
+// authRequestKey defines the labels for the requests_total metric.
+type authRequestKey struct {
+	method string
+	route  string
+	code   int
+}
```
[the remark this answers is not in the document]
ademidoff: This is a major concern of this PR and it's pretty valid. We can't afford blowing up metrics cardinality, especially on self-introspection tickets.

ademidoff (commented):
Another general PromQL gotcha is the use of `rate(sum(...))`, while the best practice is `sum(rate(...))`. Let's update the formulas.

managed/services/grafana/auth_server.go
```diff
@@ -266,7 +438,16 @@ func (s *AuthServer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 		}
-		s.returnError(rw, httpStatusForAuthError(authErr.code), m, l)
+
+		status := httpStatusForAuthError(authErr.code)
+		// Use cleaned path for metrics to avoid high cardinality due to URL-encoded characters.
+		// TODO: move cleaning to the beginning of ServeHTTP and use it for both logging and metrics.
+		cleanedPath, cleanErr := cleanPath(req.URL.Path)
+		if cleanErr != nil {
+			cleanedPath = req.URL.Path
+		}
```
ademidoff: `req.URL.Path` here is the full original URI *including the query string* — `extractOriginalRequest` assigns `X-Original-Uri` verbatim. `cleanPath` normally strips the query, but this fallback runs exactly when `cleanPath` fails (which is also why `authenticate` returned an error for the same path), so the raw query string ends up in the `route` label. Query strings often carry unique timestamps/tokens, so each becomes a distinct series — and this path is reachable by unauthenticated clients (any invalid-escape URI).

The bad-request branch above already trims at `?` before cleaning; this fallback should do the same (or better — see the comment below about hoisting the cleaning to the top of `ServeHTTP`, which fixes this for free).

managed/services/grafana/auth_server.go
```diff
@@ -285,9 +466,24 @@ func (s *AuthServer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 		return
 	}
+
+	// Use cleaned path for metrics to avoid high cardinality due to URL-encoded characters.
+	// TODO: move cleaning to the beginning of ServeHTTP and use it for both logging and metrics.
+	cleanedPath, cleanErr := cleanPath(req.URL.Path)
+	if cleanErr != nil {
+		cleanedPath = req.URL.Path
+	}
+	s.incAuthRequests(req.Method, cleanedPath, http.StatusOK)
```
ademidoff: This clean-with-fallback block now appears three times in `ServeHTTP` (plus a variant in the bad-request branch), each with a TODO saying it should move to the top. Since all of this code is new in this PR, I'd do it now rather than leave the TODO: clean once right after `extractOriginalRequest` succeeds and reuse the result for all `incAuthRequests` calls. It could also be passed down to `authenticate`, removing the second `cleanPath` call per request.

Note also that the fallbacks here and in the LBAC-error branch are effectively dead code: these branches are only reachable after `authenticate` already ran `cleanPath` successfully on the same unchanged `req.URL.Path`, so `cleanErr` can't be non-nil here.

managed/services/grafana/auth_server.go
```diff
@@ -168,49 +173,193 @@ type authError struct {
...
+	s.mDurations.Collect(ch)
+}
+
+func (s *AuthServer) incAuthRequests(method, route string, code int) {
+	key := authRequestKey{method: method, route: route, code: code}
+	v, _ := s.mAuthRequests.LoadOrStore(key, new(atomic.Uint64))
+	if counter, ok := v.(*atomic.Uint64); ok {
+		counter.Add(1)
 	}
 }
```
ademidoff, line 315: The `sync.Map` + `atomic.Uint64` + `Range`/type-assertion machinery (including the `fmt.Sprint(key)` conversion in `Collect`) reimplements what `prom.CounterVec` already provides. Other collectors in the codebase follow the vec-based pattern — e.g. `managed/services/agents/registry.go` holds `prom.Counter`/`prom.CounterVec`/`prom.GaugeFunc` fields and delegates `Describe`/`Collect` to them.

Switching `mAuthRequests`/`mGrafanaAuthRequests` to `CounterVec` and the cache hit/miss counters to `prom.Counter` would remove ~70 lines, including the defensive type assertions in `Collect` that can never fire; only `cache_size` would still need a const metric (or a `GaugeFunc`). `WithLabelValues` on the hot path is comparable in cost to a `sync.Map` lookup, so nothing is lost performance-wise.

managed/services/grafana/auth_server.go
maxkondr: wanted to add metrics without the logic changes, but OK - will modify logic

managed/services/grafana/auth_server.go
maxkondr: wanted to add metrics without the logic changes, but OK - will modify logic

managed/services/grafana/auth_server.go
maxkondr: done

managed/services/grafana/auth_server.go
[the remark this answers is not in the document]
maxkondr: the "route" label is cleaned from query parameters, so yes - it increases metrics number but not drastically. Most of the metrics will have really small numbers and are updated infrequently

maxkondr:
> Another general PromQL gotcha is the use of `rate(sum(...))`, while the best practice is `sum(rate(...))`. Let's update the formulas.

updated

JiriCtvrtka (approved):

mattiasimonato (approved):

fabio-silva (approved):

Nailya (approved):

managed/services/realtimeanalytics/service.go
```diff
@@ -521,15 +526,27 @@ func (s *Service) Collect(stream grpc.ClientStreamingServer[rtav1.CollectRequest
...
+			return nil, status.Error(codes.ResourceExhausted, "is rejected by ratelimit, please retry later.")
+		}
+		defer s.rateLimiter.Release()
+
+		// Validate that the pmm-agent exists
+		agent, err := models.FindAgentByID(s.db.WithContext(streamCtx), agentMD.ID)
+		if err != nil {
+			l.Warnf("Disconnecting client: agent validation failed: %v", err)
+			return nil, status.Error(codes.InvalidArgument, "Invalid Agent ID: "+agentMD.ID)
+		}
```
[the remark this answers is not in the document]
maxkondr: agent must be in DB, otherwise this is invalid agentMD.ID

managed/cmd/pmm-managed/main.go
```diff
@@ -933,11 +1012,11 @@ func main() { //nolint:gocognit,maintidx,cyclop
...
 
 	pbmPITRService := backup.NewPBMPITRService()
-	backupRemovalService := backup.NewRemovalService(db, pbmPITRService)
-	backupRetentionService := backup.NewRetentionService(db, backupRemovalService)
+	backupRemovalService := backup.NewRemovalService(internalDB, pbmPITRService)
+	backupRetentionService := backup.NewRetentionService(internalDB, backupRemovalService)
 	prom.MustRegister(agentsRegistry)
 
-	inventoryMetrics := inventory.NewInventoryMetrics(db, agentsRegistry)
+	inventoryMetrics := inventory.NewInventoryMetrics(internalDB, agentsRegistry)
```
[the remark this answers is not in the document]
maxkondr: this is intentionally for now

build/ansible/roles/nginx/files/conf.d/pmm.conf
```diff
@@ -127,6 +206,44 @@
...
+      # If the client request method is listed in this directive then the response will be cached.
+      # “GET” and “HEAD” methods are always added to the list,
+      # though it is recommended to specify them explicitly.
+      proxy_cache_methods GET HEAD POST;
+      # Cache entry key = bearer/basic auth header|original request method|original request URL path.
+      # Same pair reuses one cached auth result.
+      # Scope restriction: /auth_request_cached is only used by /victoriametrics/api/v1/write
+      # (credential-only auth). Do not enable it for cookie-authenticated routes unless
+      # $http_cookie is added to this key to match authentication inputs.
+      proxy_cache_key "$http_authorization|$request_method|$request_uri";
```
ademidoff, line 222: **Auth bypass risk — verified.**

This key is `$http_authorization|$request_method|$request_uri` only. Any cookie-authenticated request to `/victoriametrics/api/v1/write` caches a 200 under the *empty*-Authorization key, and that entry is then served to clients presenting no credentials at all.

Concretely: a logged-in browser (session cookie, no `Authorization` header) hits this path. The subrequest returns 200 and nginx caches it as `"|POST|/victoriametrics/api/v1/write"`. For the next 5 minutes (`proxy_cache_valid 200 5m`) anyone POSTing there unauthenticated hits the same key and writes arbitrary metrics into VictoriaMetrics. The inverse also holds: one cached 401 blocks legitimate writers for 30s.

The comment just above documents exactly this hazard, but nothing enforces the scope restriction. Adding `$http_cookie` to the key would close it.

managed/services/grafana/auth_server.go
```diff
@@ -247,127 +344,81 @@ func (s *AuthServer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
-// wrongly pass method-specific rules. Authentication and internal errors stay 401.
-func httpStatusForAuthError(code codes.Code) int {
-	if code == codes.PermissionDenied {
-		return http.StatusForbidden
+	if authRes.vmProxyFilters != "" {
+		// Add HTTP headers to response based on filled fields in authResult.
+		rw.Header().Set(lbacHeaderName, authRes.vmProxyFilters)
 	}
-	return authenticationErrorCode
+	s.incAuthRequests(req.Method, req.URL.Path, http.StatusOK)
```
ademidoff, line 385: **Unbounded metric cardinality — verified.**

`extractOriginalRequest(req)` at line 345 rewrites `req.URL.Path` to the *original client URI*, so the `route` label here is per-URL, not per-route.

Every distinct URL creates a `CounterVec` child that is never freed: `/v1/inventory/services/<uuid>`, `/graph/d/<dashboard-uid>/<slug>`, `/graph/api/datasources/proxy/<id>/...`. A browsing user or an agent fleet produces thousands of unique paths within minutes — unbounded memory growth in pmm-managed plus ballooning `/debug/metrics` scrapes.

The matched rule prefix already computed by `resolveRule` is the bounded value to use here (and at lines 350, 371, 375).

managed/services/grafana/helpers.go
```diff
@@ -0,0 +1,324 @@
...
+		}
+		prefix = nextPrefix(prefix)
+	}
+}
+
+// isLocalAgentConnection reports whether the request is a local PMM agent
+// connection for endpoints that are allowed from localhost.
+// This func expects that req.Method and req.URL.Path are already replaced
+// with original request values - extractOriginalRequest(req) has been called beforehand.
+func isLocalAgentConnection(req *http.Request) bool {
```
ademidoff, line 192: **Possible auth widening — not independently verified, please confirm.**

This now compares the cleaned, query-stripped `req.URL.Path` instead of the raw `X-Original-Uri` header, which widens the set of requests that receive an unauthenticated `admin` authUser.

Behind nginx `req.RemoteAddr` is always loopback, so the address check is unconditionally true — this path comparison is the only real gate. On `main` it was an exact match on the raw URI; now `extractOriginalRequest` runs `cleanPath` first, so `/agent.v1.AgentService/Connect?x=1` also matches. The loopback test was also widened from `== "127.0.0.1"` to any loopback address.

Caveat on my own analysis: nginx normalizes URIs before setting `X-Original-Uri`, so the `..`-traversal variant of this probably does *not* reach here. The query-string widening looks real regardless. Worth confirming the exact-match behaviour was not load-bearing.

managed/cmd/pmm-managed/main.go
```diff
@@ -870,45 +911,91 @@ func main() { //nolint:gocognit,maintidx,cyclop
...
+	sqlInternalDB, err := models.OpenDB(setupInternalDBParams)
 	if err != nil {
 		l.Panicf("Failed to connect to database: %+v", err)
 	}
-	defer sqlDB.Close() //nolint:errcheck
+	defer func() {
+		_ = sqlInternalDB.Close()
+	}()
+
+	migrateDB(ctx, sqlInternalDB, setupInternalDBParams)
```
ademidoff, line 945: **HA regression — verified against `main`.**

`migrateDB` moved to here (945), but the `if *haEnabled` override of `models.AgentConfigFilePath` is at line 981 — *after* this call. On `main` the order was the reverse (override at 892, `migrateDB` at 896).

So with `--ha-enable`, `migrateDB` → `models.SetupDB` → the fixture code opens `models.AgentConfigFilePath` while it is still the default `/usr/local/percona/pmm/config/pmm-agent.yaml`. In HA the real file is `/srv/pmm-agent/config/pmm-agent.yaml`, so the open fails or the wrong agent ID gets seeded.

Moving the `haEnabled` block above `migrateDB` restores the original ordering.

managed/services/agents/registry.go
```diff
@@ -244,9 +248,13 @@ func (r *Registry) register(stream agentv1.AgentService_ConnectServer) (*pmmAgen
 	})
 	if err != nil {
 		l.Warnf("Failed to authenticate connected pmm-agent %+v.", agentMD)
-		return nil, err
+		return zero, err
 	}
 	l.Infof("Connected pmm-agent: %+v.", agentMD)
+	pmmAgentVersion, err := version.Parse(agentMD.Version)
```
ademidoff, line 254: **Backward-compatibility concern — not independently verified.**

`register()` now hard-fails the whole agent connection when `version.Parse(agentMD.Version)` errors. Previously the version was read from the DB inside `sendSetStateRequest`, so a parse failure only broke state updates.

An agent whose `Version` string `version.Parse` can't handle (empty, custom branch suffix, PMM 2.x variant) now gets a plain non-gRPC-status error returned before `SendServerConnectMetadata`; `Handler.Run` marks `disconnectReason="auth"` and drops the stream. That agent can never connect at all rather than merely failing state pushes — a break against agents already deployed in the field.

Could this degrade gracefully instead (log + treat as unknown version)?

build/ansible/roles/nginx/files/conf.d/pmm.conf
```diff
@@ -18,28 +18,76 @@
...
+    server 127.0.0.1:9090;
+    # Allocates a pool of idle TCP connections to the backend.
+    # It's bigger than for the rest of upstreams because it has more incoming traffic to handle.
+    keepalive 64;
+    # Prevent connection churn during high-throughput metric shipping.
+    # The NGINX default (100 or 1000) is insufficient for observability data streams.
+    # Increasing this prevents frequent socket teardowns.
+    keepalive_requests 100000;
+    # Ensure connections survive standard 30s-60s scrape intervals
+    keepalive_timeout 75s;
```
ademidoff, line 42: **Timeout inversion — verified against the running server.**

`keepalive_timeout 75s` here is *longer* than VictoriaMetrics' own `-http.idleConnTimeout`, which defaults to **1m** and is not overridden in `managed/services/supervisord/supervisord.go`. So nginx keeps idle connections pooled for up to 75s that VM closed at 60s.

nginx reaps most of those via its close handler, but there is a real race where a request is dispatched between VM's FIN and the reap. For `location = /victoriametrics/api/v1/write` that request is a POST, and `proxy_next_upstream` defaults to `error timeout` *without* `non_idempotent` — so nginx will not retry, and vmagent sees a sporadic 502.

Suggest setting this below the backend's idle timeout (~30-50s), or raising VM's `-http.idleConnTimeout`.

Same issue on the `vmalert` upstream at line 51: `keepalive_timeout 60s` sits exactly on VM's idle boundary, and 60s is also the nginx upstream default, so that line is a no-op.

build/ansible/roles/nginx/files/conf.d/pmm.conf
```diff
@@ -267,54 +495,63 @@
...
+      tcp_nodelay on;
     }
 
-    # This location stores static content for general pmm-server purposes.
-    # Ex.: local-rss.xml - contains Percona's news when no internet connection.
-    location /pmm-static {
-      auth_request off;
-      alias /usr/share/pmm-server/static;
-    }
+    ############  UNPROTECTED ENDPOINTS BLOCK  ##############
```
ademidoff, line 539: **Dropped locations — verified against `main`.**

The `location /pmm-static` and `location = /percona-blog/feed` blocks exist on `main` (lines 302-318) and are gone in this rewrite, with no `location /` catch-all to absorb them — only `location = /` exists.

`dashboards/dashboards/PMM Health/Environments_Overview.json` sets `"feedUrl": "/percona-blog/feed"`, so the Percona news panel on the Home dashboard now 404s (and requires auth). Its offline fallback `error_page 500 502 503 504 /pmm-static/local-rss.xml` is gone too, which silently reverts the PMM-6153 fix.

Intentional removal, or lost in the restructure?

managed/services/agents/handler.go
```diff
@@ -69,8 +76,15 @@ func (h *Handler) Run(stream agentv1.AgentService_ConnectServer) error { //nolin
 
 	ctx := stream.Context()
 	l := logger.Get(ctx)
+
+	if !h.rateLimiter.TryAcquire() {
+		disconnectReason = "RESOURCE_EXHAUSTED"
+		return status.Error(codes.ResourceExhausted, "is rejected by ratelimit, please retry later.")
+	}
 	agent, err := h.r.register(stream)
+	h.rateLimiter.Release()
```
ademidoff, line 85: **Concurrency concern — not independently verified.**

The limiter slot is held across the whole of `h.r.register(stream)` and `Release()` is not deferred.

Two consequences. First, `register()` can reach the duplicate-ID branch and call `r.ping(...)` → `channel.SendAndWaitResponse`, which blocks on a channel receive with no timeout until the *old* TCP connection dies — minutes on a half-open socket. That slot comes from `pmmAgentsConnectionsLimiter`, shared with `StateUpdater` state updates and RTA Collect, so a handful of stuck duplicate reconnects can starve state updates.

Second, any panic between `TryAcquire` (line 80) and this line permanently burns a slot, since `Release` caps at `maxSlots` and never recovers it. A `defer` would fix the second half regardless.

managed/services/agents/state.go
```diff
@@ -78,161 +99,249 @@ func (u *StateUpdater) RequestStateUpdate(ctx context.Context, pmmAgentID string
...
 	start := time.Now()
 	defer func() {
 		if dur := time.Since(start); dur > time.Second {
 			l.Warnf("sendSetStateRequest took %s.", dur)
 		}
 	}()
-	pmmAgent, err := models.FindAgentByID(u.db.Querier, agent.id)
+
+	// Use singleflight to avoid fetching settings for each pmm-agent separately.
+	fetchedSettings, err, _ := u.dbGroup.Do("settings", func() (any, error) { //nolint:contextcheck
```
ademidoff, line 216: **Concurrency concern — not independently verified.**

This singleflight call (and the `node/<id>` one at line 266) runs its DB query on `context.Background()` with its own 5s timeout, so `sendSetStateRequest` can block well past the 5s `stateChangeTimeout` its caller set — while holding a rate-limiter slot.

`runStateChangeHandler` creates `nCtx` with `stateChangeTimeout` and acquires a slot. A waiter can then block up to 5s on the leader's Background-context settings query, then another 5s on the node query; a waiter whose own `nCtx` already expired still cannot abandon `singleflight.Do`. Under DB pressure each state update pins a shared slot for ~10s instead of 5s, draining a limiter sized at 70% of `apiDbMaxOpenConns` and shared with agent registration and RTA Collect.

utils/cache/common.go
```diff
@@ -0,0 +1,53 @@
...
+type item[V any] struct {
+	value   V
+	expires time.Time
+	rev     uint64
+}
+
+// shard contains the actual map and a lock, padded to prevent false sharing.
+type shard[V any] struct {
+	mu    sync.RWMutex
+	items map[uint64]item[V]
```
ademidoff, line 48: **Hardening — not independently verified, and low probability.**

The shard map is keyed on `uint64` (a 64-bit maphash) and the original key is never stored or compared, so a hash collision silently returns another entry's value.

`Registry.agentsCache` (`cache.Cache[pmmAgentInfo]`) does no key verification at all: on a collision, `LoadOrStore` during `register` returns the *other* agent's `pmmAgentInfo`, `r.ping` pings the wrong channel, and a healthy pong causes the new agent to be rejected with `AlreadyExists`. `AuthServer` works around this by re-checking `authorization`/`cookie` in `cachedAuthUser`, which suggests the hazard was known — the same guard is just missing in `Cache[V]`.

Storing the key alongside the value removes the class outright. Given 64-bit collisions are astronomically unlikely, treat this as hardening rather than a live bug — but the asymmetry with `cachedAuthUser` is worth resolving one way or the other.

managed/services/victoriametrics/victoriametrics.go
```diff
@@ -474,21 +474,20 @@ func scrapeConfigForVMAlert(interval time.Duration, pmmServerNodeName string) *c
...
-		settings, err := models.GetSettings(tx)
-		if err != nil {
-			return err
-		} // In HA mode, skip ExternalExporter agents if this node is not the leader
-		skipExternalExporter := !svc.haService.IsLeader()
-		return AddScrapeConfigs(svc.l, &cfg, tx.Querier, new(settings.MetricsResolutions), new(pmmAgentID), true, skipExternalExporter)
-	})
-	if e != nil {
-		return nil, e
+	settings, err := models.GetSettings(q)
```
ademidoff, line 482: **Consistency concern — not independently verified.**

`BuildScrapeConfigForVMAgent` appears to have lost its `InTransactionContext` wrapper, so `models.GetSettings(q)` here and `AddScrapeConfigs(..., q, ...)` below are now two independent reads on a plain `Querier` rather than one consistent snapshot.

If metrics resolutions change, or an agent is added/removed between the two reads, the generated vmagent scrape config mixes the old resolution with the new agent set and is pushed to the agent as though consistent.

Was dropping the transaction deliberate (contention?) or incidental?

managed/services/grafana/auth_server.go
```diff
@@ -425,223 +482,231 @@ func (s *AuthServer) getLBACFilters(ctx context.Context, userID int) ([]string,
...
+	if minRole == none && !needLbacFilter {
+		l.WithField("path", req.URL.Path).Debug("Minimum required role is 'none', granting access without authentication.")
+		return authResult{}, nil
 	}
 
-	if origURI == "" {
-		return errors.New("empty X-Original-Uri")
+	user, err := s.authenticateUser(req, l) //nolint:contextcheck
+	if err != nil {
+		l.WithError(err).Error("Failed to authenticate user.")
```
ademidoff, line 504: **Log noise — not independently verified.**

Ordinary authentication failures look like they went from a single `l.Warnf` on `main` to three `Error`-level lines per request: `getGrafanaAuthUser`'s "Failed to retrieve user info from Grafana." (~line 693), `getAuthUser`'s "Grafana user lookup failed." (~line 638), and this one.

A browser hitting the app before login, or a fleet with one stale token, will flood pmm-managed logs at ERROR and pollute any error-rate alerting keyed on log severity. These are expected conditions — WARN at a single site seems right.

managed/services/grafana/auth_server.go
```diff
@@ -247,127 +344,81 @@ func (s *AuthServer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+			s.incAuthRequests(req.Method, req.URL.Path, status)
+			writeResponseErrorStatus(rw, status, int(authErr.code), authErr.message, authErr.message)
+			return
 		}
-		l.Errorf("Failed to add VMProxy filters: %s", errF)
-
-		s.returnError(rw, authenticationErrorCode, m, l)
+		s.incAuthRequests(req.Method, req.URL.Path, http.StatusInternalServerError)
+		writeResponseErrorStatus(rw, http.StatusInternalServerError, http.StatusInternalServerError,
+			statusCodeToString(http.StatusInternalServerError), statusCodeToString(http.StatusInternalServerError))
```
ademidoff, line 377: **Minor — not independently verified.**

`statusCodeToString(http.StatusInternalServerError)` passed as both the error and message args means the `X-Auth-Error` and `X-Auth-Message` headers carry the literal string `"500"` rather than anything descriptive.

Also note the status is 500 rather than 401/403, and this file's own comment says nginx's `auth_request` cannot handle 500 — so nginx emits its default 500 page and these headers are discarded anyway. If that's the intent, the header values are dead; if not, the status is wrong.

managed/services/grafana/auth_server.go
```diff
@@ -425,223 +482,231 @@ func (s *AuthServer) getLBACFilters(ctx context.Context, userID int) ([]string,
...
+			return cached.user, nil
 		}
 	}
+	s.incCacheMiss()
+
+	// Cold-path: Cache miss / Stale data.
+	// Use single-flight to avoid calling Grafana for the same user's authHeaders.
+	// It appears when after restart the same vm-agent starts to send buffered metrics
+	// to server in parallel and all such requests have to be authenticated.
+	for attempt := range maxAuthCollisionRetries {
```
ademidoff, line 599: **Possible dead code — not independently verified.**

This retry loop and the singleflight-bypass fallback after it look unreachable. `getAuthCacheKey` (`managed/services/grafana/helpers.go:302`) concatenates the raw `Authorization` + `\x00` + `Cookie` values, and `singleflight.Group` keys on that exact string — not on a hash. Two different header pairs therefore can never share a group key, so the `cached.authorization == authorization && cached.cookie == cookie` check always passes on the first iteration.

That leaves ~40 lines plus a second full Grafana call that deliberately bypasses stampede protection permanently unreachable. The only real collision risk is in the TTL cache's 64-bit key, which the `cachedAuthUser` field comparison already handles.

If I've misread the key construction, please say — otherwise this could be deleted.

utils/cache/cache.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// calculateKeyHash builds a deterministic cache key directly from passed key.
+func (c *Cache[V]) calculateKeyHash(key string) uint64 {
+	return maphash.String(c.seed, key)
+}
+
+// getShard returns the shard for a precomputed key hash.
+func (c *Cache[V]) getShard(keyHash uint64) *shard[V] {
+	shardKey := maphash.Comparable(c.seed, keyHash)
```
ademidoff, line 50: **Cleanup — not independently verified.**

Two things here.

`maphash.Comparable(c.seed, keyHash)` runs a second hash over a value that is already a well-distributed 64-bit maphash. `keyHash & shardMask` gives the same distribution for free, on what the package doc calls the zero-allocation hot path.

More broadly, `utils/cache/cache.go` and `utils/cache/cache_ttl.go` are near-identical copies — `getShard`, `Load`, `Store`, `Delete`, `LoadAndDelete`, `Size` and the shard-init loop are duplicated, now including the new `rev` bookkeeping, plus two duplicated test and bench suites. Any fix to size accounting or shard math has to be applied twice. The same duplication shows up in the byte-identical `Limiter` interfaces in `managed/services/agents/deps.go` and `managed/services/realtimeanalytics/deps.go`.

Generics or a shared embedded base would collapse most of it.

build/ansible/roles/nginx/files/conf.d/pmm.conf
```diff
@@ -18,28 +18,76 @@
...
 
+  upstream victoriametrics {
+    server 127.0.0.1:9090;
+    # Allocates a pool of idle TCP connections to the backend.
+    # It's bigger than for the rest of upstreams because it has more incoming traffic to handle.
+    keepalive 64;
+    # Prevent connection churn during high-throughput metric shipping.
+    # The NGINX default (100 or 1000) is insufficient for observability data streams.
+    # Increasing this prevents frequent socket teardowns.
+    keepalive_requests 100000;
```
ademidoff, line 40: **Inert setting — verified.**

`keepalive_requests 100000` on this upstream can never be reached. VictoriaMetrics runs with the default `-http.connTimeout` (2m, not overridden in `managed/services/supervisord/supervisord.go`), which closes every incoming connection after 120s. An HTTP/1.1 upstream connection is serialized, so hitting 100000 requests inside that window would need ~833 req/s on a *single* connection — about 1.2ms per remote_write round trip including body upload.

Also, the comment above says the nginx default is "100 or 1000". On the shipped nginx 1.26.3 it is unambiguously **1000**; 100 applied only before 1.19.10. So the real change here is 100x over a default that already wasn't binding.

The 10000 values on the other upstreams are cheap and harmless (`keepalive_requests` is just a counter — only `keepalive N` sizes the pool), and genuinely useful on `managed-json`, where the per-request `auth_request` subrequest makes the cap reachable. It's specifically this one that's dead.

ademidoff (changes requested):
Review of the performance work. 16 comments below.

**Verification status is stated on each comment.** Six are verified against this PR head (`52d55b96`) — I checked them against `main` and/or the running server. The other ten are analysis I have *not* independently confirmed; please treat those as questions rather than defects, and push back where I'm wrong.

The items I'd consider blocking are the first four: the `AUTH_CACHE` cache key, the unbounded `route` label, the `isLocalAgentConnection` widening, and the HA `migrateDB` ordering.

One earlier finding is already fixed here and is not included: the `$auth_code_guarded` block was moved out of the `server` block into `location @auth_failed`, which correctly puts it after the auth subrequest completes.

Context for the nginx upstream comments: the shipped nginx is 1.26.3, where `keepalive_requests` defaults to **1000** (the 100 default applied only before 1.19.10), and VictoriaMetrics runs with default `-http.idleConnTimeout` (1m) and `-http.connTimeout` (2m), neither overridden in `managed/services/supervisord/supervisord.go`.

PR #5679, opened 2026-07-21
PMM-13776 Create a pmm-ui package to cache its artifacts

ademidoff:
PMM-13776

SUBMODULES-4490

## Problem

The PMM UI was built inside `build/packages/rpm/server/SPECS/pmm-managed.spec` and had no package name of its own. Its S3 build-cache key therefore tracked the pmm repo `HEAD` commit, so the UI was rebuilt on **every push** even when nothing under `ui/` changed.

## Solution

Split the UI into its own `pmm-ui` RPM package and make its cache key track only UI changes, using native git.

- **New `build/packages/rpm/server/SPECS/pmm-ui.spec`** — runs `make -C ui release` (turbo builds both apps) and installs to the same paths the runtime already expects:
  - `ui/apps/pmm/dist` → `/usr/share/pmm-ui` (served by nginx)
  - `ui/apps/pmm-compat/dist` → `/usr/share/percona-dashboards/panels/pmm-compat-app` (Grafana panel plugin)
- **`build/packages/rpm/server/SPECS/pmm-managed.spec`** — dropped the UI build/install/files; bumped `%define release` 20→21 with a changelog entry.
- **`build/scripts/build-server-rpm`** — `prepare_specs`/`build` take an optional 3rd arg (a git path). When set, the package's `%global commit` (which feeds both the S3 cache key and the source archive) is derived from `git log -1 --format=%H -- <path>` instead of `HEAD`, with a `HEAD` fallback. So the key stays constant across pushes that don't touch the path — the package is rebuilt only when the branch actually differs from its base in that path.
- **`build/scripts/build-server-rpm-all`** — registers `build-server-rpm pmm-ui pmm ui`.
- **`build/ansible/roles/pmm-images/tasks/main.yml`** — installs `pmm-ui`.

The dual ownership of `/usr/share/percona-dashboards/...` between `pmm-ui` and `percona-dashboards` mirrors the arrangement `pmm-managed` previously had, so there's no new RPM conflict.

## Testing

Verified locally: `bash -n` on the changed scripts, `shellcheck` (no new warnings), and the `git log -1 --format=%H -- ui` path-scoping on the repo. Full RPM build / `rpmspec` parsing is delegated to CI (needs the pmm-submodules workspace layout, the rpmbuild image, and S3).


fabio-silva (approved):

mattiasimonato (approved):

build/packages/rpm/server/SPECS/pmm-managed.spec
```diff
@@ -6,7 +6,7 @@
 %global commit          8f3d007617941033867aea6a134c48b39142427f
 %global shortcommit     %(c=%{commit}; echo ${c:0:7})
 %define build_timestamp %(date -u +"%y%m%d%H%M")
-%define release         20
+%define release         21
 %define rpm_release     %{release}.%{build_timestamp}.%{shortcommit}%{?dist}
```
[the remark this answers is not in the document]
ademidoff: UI is not part of managed, so it's not really an issue.

These RPM packages are only necessary for the feature builds, they are not uploaded to any repository.

build/packages/rpm/server/SPECS/pmm-ui.spec
```diff
@@ -0,0 +1,50 @@
...
+
+Name:		pmm-ui
+Version:	%{version}
+Release:	%{rpm_release}
+Summary:	Percona Monitoring and Management web UI
+
+License:	Apache-2.0
+URL:	    https://%{provider}
+Source0:	https://%{provider}/archive/%{commit}/%{repo}-%{shortcommit}.tar.gz
+
```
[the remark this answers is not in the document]
ademidoff: To build all assets, both BE and FE, we use the same `rpmbuild` container, which has NodeJS provisioned.

But OK, it's harmfull :)

ui/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
JiriCtvrtka, line 1: The UI still declares AGPLv3 in ui/package.json, while this package uses Apache-2.0. Is this license change ok?

ademidoff: Yes, this is what the license should be for an open-source project, so let's update package.json to keep them in sync :)

ademidoff: Done!

build/scripts/build-server-rpm-all
```diff
@@ -5,8 +5,11 @@ set -o xtrace
 . "$(dirname $0)"/vars
 
 # 1st-party
-"${bin_dir}/build-server-rpm" percona-dashboards pmm
+# The 3rd arg scopes the cache key to the given paths (the package sources plus
+# its spec) so the package is rebuilt only when those actually change.
+"${bin_dir}/build-server-rpm" percona-dashboards pmm "dashboards build/packages/rpm/server/SPECS/percona-dashboards.spec"
 "${bin_dir}/build-server-rpm" pmm-managed pmm
+"${bin_dir}/build-server-rpm" pmm-ui pmm "ui build/packages/rpm/server/SPECS/pmm-ui.spec"
```
JiriCtvrtka, line 12: Possible edge case: Do changes to the shared build scripts or build image invalidate this cache elsewhere? Otherwise, the cached RPM could become stale when the build process changes without changes under ui/.

build/packages/rpm/server/SPECS/pmm-ui.spec
```diff
@@ -0,0 +1,52 @@
...
+Name:		pmm-ui
+Version:	%{version}
+Release:	%{rpm_release}
+Summary:	Percona Monitoring and Management web UI
+
+License:	Apache-2.0
+URL:	    https://%{provider}
+
+BuildRequires:	nodejs
```
JiriCtvrtka, line 19: ui/package.json requires Node.js >=22 and Yarn 1.22.22. Should we require nodejs >= 22 and provision Yarn explicitly instead of relying on the build image?

JiriCtvrtka (approved):

build/scripts/build-server-rpm-all
ademidoff: If you are talking about S3 cache, then yes - the server RPMs are purged on a regular basis. Otherwise, there is no other cache, no repo that hosts them.

build/packages/rpm/server/SPECS/pmm-ui.spec
ademidoff: No, then it becomes a burden to maintain. I believe the way it's implemented here is the most flexible.

4nte (approved):

PR #5691, opened 2026-07-23
PMM-13860: finalize HA Overview dashboard

fabio-silva:
Ticket number: PMM-13860

Screenshot 2026-07-28 at 15 38 39
Screenshot 2026-07-28 at 15 38 51
Screenshot 2026-07-28 at 15 39 01
Screenshot 2026-07-28 at 15 39 11
Screenshot 2026-07-28 at 15 39 17
Screenshot 2026-07-28 at 15 39 28
Screenshot 2026-07-28 at 15 39 34


documentation/docs/install-pmm/install-HA-clustered.md
mattiasimonato: @fabio-silva  quick question on the docs vs UI here

From what I can tell:
- this PR removes the sidebar leader item
- the Helm `release` variable looks hidden (`hide: 2`)

But the install docs still mention the sidebar leader name, and the dashboard reference still talks about using the Helm release filter.

Was that intentional (e.g. leader now only on the Overview dashboard), or should the docs/screenshot still match the old sidebar UX?

dashboards/dashboards/PMM Health/PMM_HA_Health_Overview.json
mattiasimonato, line 1: It looks like Alive/Expected is encoded as `alive * 1000 + expected`, then mapped back with a large value-mappings list (repeated across the health panels). That seems to be what makes the dashboard JSON quite large (~2MB), and it may also cap the display around `40/40`.

Was that the intended approach for showing `3/3`-style counts in Grafana?

If not, a simpler option could be:
1. **Two series in the same Stat** (`Alive`, `Expected`) with `textMode: value_and_name`
2. Or a **small overview table**: Component | Alive | Expected | Health

mattiasimonato (changes requested):

documentation/docs/install-pmm/install-HA-clustered.md
fabio-silva: @mattiasimonato thanks, I've removed the reference to Helm release and leader from the docs

dashboards/dashboards/PMM Health/PMM_HA_Health_Overview.json
fabio-silva: Thanks for the tip! I've improved the dashboard

mattiasimonato (approved):

ui/apps/pmm/src/contexts/navigation/navigation.constants.ts
```diff
@@ -790,13 +790,14 @@ export const NAV_HIGH_AVAILABILITY: NavItem = {
   id: 'high-availability',
   icon: 'cluster',
   text: 'PMM HA',
-  url: `${PMM_NEW_NAV_GRAFANA_PATH}/high-availability`,
+  url: `${PMM_NEW_NAV_GRAFANA_PATH}/d/pmm-ha-health-overview/pmm-ha-health-overview`,
 };
 
-export const NAV_HIGH_AVAILABILITY_LEADER: NavItem = {
```
matejkubinec, line 828: What's the reason for removing the leader information? In the initial design (not sure if still applies) - https://www.figma.com/design/pWEsw8pSWBFDiXIfCFaFkd/PMM-High-Availability-Groundwork?node-id=100-1716&m=dev - the dashboard link and leader info coexist.

fabio-silva: @matejkubinec yes, it was in the initial design.
However, we discussed this subject and decided to remove it for now, given the dashboard is now immediately accessible from the sidebar and tells users the leader right away.
The element with the leader didn't seem a good fit on the sidebar, so for now we will make this way.

matejkubinec (approved):

api/ha/v1beta1/ha.proto
```diff
@@ -36,6 +36,9 @@ message StatusRequest {}
 message StatusResponse {
   // Status of HA mode: "Enabled" or "Disabled".
   string status = 1;
+  // Kubernetes namespace this instance is running in. Empty if unknown
+  // (e.g. non-Kubernetes installs).
```
maxkondr: is HA supposed to be running in non-Kubernetes env?

managed/services/ha/haservice.go
```diff
@@ -208,6 +213,23 @@ func setupRaftStorage(nodeID string, l *logrus.Entry) (*raftboltdb.BoltStore, *r
...
 
+// DetectNamespace returns the Kubernetes namespace this process is running in,
+// or an empty string if it can't be determined (e.g. non-Kubernetes installs,
+// or a service account without the token automounted). Missing the file is an
+// expected, non-fatal condition, not an error worth surfacing to the user.
+func DetectNamespace(l *logrus.Entry) string {
+	return detectNamespaceAt(serviceAccountNamespaceFile, l)
+}
+
+func detectNamespaceAt(path string, l *logrus.Entry) string {
```
maxkondr: why extra func is needed here?

fabio-silva: I mistakenly pushed this, but reverted in the meanwhile

api/ha/v1beta1/ha.proto
fabio-silva: yeah, in the scope of this proto, I guess that comment doesn't make sense

mattiasimonato (approved):

PR #5694, opened 2026-07-24
PMM-15241: Restore update popup snooze using UI-owned duration after settings field removal

mattiasimonato:
Ticket number: PMM-15241

Feature build: SUBMODULES-4496

Fixes`UpdateModal` reopens after hard refresh because `useSnooze` treated an active snooze as expired.

The bug was introduced as a side effect of  [PMM-14969](https://perconadev.atlassian.net/browse/PMM-14969) / [#5394](https://github.com/percona/pmm/pull/5394) (UI upgrade removal, Jul 2026): `update_snooze_duration` was removed from the Settings API (proto fields reserved) and from managed settings models, but the UI still computed snooze validity via `settings.updateSnoozeDuration` → `parseDuration(...)`. With the field gone, duration resolved to `0ms`. Per-user persistence (`snoozedPmmVersion`, `snoozedAt`) still worked, so the popup closed in-session and reappeared after reload once user/version data finished loading.

### Solution:
stop reading `settings` for snooze duration. Use per-user snooze fields plus a UI-owned 7-day window. Gate (`DEFAULT_UPDATE_SNOOZE_DURATION_MS`) in `isUpdateSnoozeActive`.

### Cleanup:
delete dead `updateSnoozeDuration` from settings types and test stubs; remove `useSettings` / `parseDuration` from `useSnooze`.

### QA: localStorage snooze override

For QA only. Default snooze is still **7 days**. To test expiry without waiting:

```js
localStorage.setItem('pmm-ui.dev.updateSnoozeDurationMs', '10000'); // 10s
// Remind me later → wait >10s → refresh → popup should show again
localStorage.removeItem('pmm-ui.dev.updateSnoozeDurationMs'); // reset
```

[PMM-14969]: https://perconadev.atlassian.net/browse/PMM-14969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update snoozing behavior across missing data, version changes, expiration, and release timing.
  * Update prompts now remain suppressed for a default of seven days after being snoozed.

* **Configuration**
  * Added support for configuring the snooze duration through a local browser setting, with validation and a safe default.


ui/apps/pmm/src/hooks/updates.utils.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        userInfo: {
+          snoozedPmmVersion: '3.1.0',
+          snoozedAt: new Date(NOW - 60_000).toISOString(),
+        },
+        now: NOW,
+      })
+    ).toBe(false);
+  });
+
+  it('returns true when the latest release is younger than one hour', () => {
```
fabio-silva, line 107: Not sure I understand this test. The description says 'younger than one hour' but the release is (1 week + 60secs) old. Should we be testing a release that's actually one hour old (i.e. NOW - 60_000) or is the description misleading?

ui/apps/pmm/src/hooks/updates.utils.ts
```diff
@@ -0,0 +1,38 @@
...
+}): boolean => {
+  if (!latest || !userInfo) {
+    return true;
+  }
+
+  if (
+    latest.timestamp &&
+    now - new Date(latest.timestamp).getTime() < SHOW_UPDATE_MODAL_AFTER_MS
+  ) {
+    return true;
```
fabio-silva, line 42: I think this is not correct.
Even though there's a test that asserts "returns false when the user has never snoozed", that test suite's release version has a null timestamp (which it shouldn't I believe), that's why it passes.
According to this code, if we release a version today, and assuming it has a timestamp (which it should), then the user won't see anything for a week.

fabio-silva (changes requested):

ui/apps/pmm/src/hooks/updates.utils.ts
```diff
@@ -0,0 +1,38 @@
+import {
+  DEFAULT_UPDATE_SNOOZE_DURATION_MS,
+  SHOW_UPDATE_MODAL_AFTER_MS,
+} from 'lib/constants';
+import { LatestInfo } from 'types/updates.types';
+import { UserInfo } from 'types/user.types';
+
+type SnoozeUserInfo = Pick<UserInfo, 'snoozedAt' | 'snoozedPmmVersion'>;
+
+export const isUpdateSnoozeActive = ({
```
fabio-silva: we probably should use three different args, not an object arg with three props

fabio-silva (changes requested):

ui/apps/pmm/src/hooks/updates.utils.test.ts
mattiasimonato: Good catch. The description was right, but the math looked confusing: it used `SHOW_UPDATE_MODAL_AFTER_MS` (1 hour, not 7 days), so the release was actually ~59 minutes old. I switched it to `NOW - 60_000` to make that obvious.

**Context:** that case covers the existing 1-hour quiet period after a new release (`SHOW_UPDATE_MODAL_AFTER_MS`). Even if the user never snoozed, we hide the popup until the release is at least 1 hour old. This test locks that in; the 7-day window is only for after clicking on **Remind me later**.

ui/apps/pmm/src/hooks/updates.utils.ts
mattiasimonato: You’re right about the fixture, default `timestamp` was `null`, so that path wasn’t really covered. Fixed: fixtures now use a realistic older timestamp.

On the duration, same as I wrote down on the previous comment: that early return uses `SHOW_UPDATE_MODAL_AFTER_MS` (1 hour), not the 7-day snooze. That’s the existing “don’t show the popup for the first hour after a release” behavior. The 7-day window only kicks in after Remind me later.

ui/apps/pmm/src/hooks/updates.utils.ts
mattiasimonato: updated to `isUpdateSnoozeActive(latest, userInfo, now?)` 👍

ui/apps/pmm/src/hooks/updates.utils.ts
fabio-silva: ah, you're right, I confused this one with `DEFAULT_UPDATE_SNOOZE_DURATION_MS`. Thanks!

fabio-silva (approved):

ui/apps/pmm/src/hooks/updates.utils.ts
mattiasimonato: no worries! all good 😄

matejkubinec (approved):

matejkubinec:
@mattiasimonato One thing that just came to mind, I think it would be good to have local storage option for an override, so the QA folks can set a lower time for the snooze duration.

mattiasimonato:
> @mattiasimonato One thing that just came to mind, I think it would be good to have local storage option for an override, so the QA folks can set a lower time for the snooze duration.

@matejkubinec Good idea. For this ticket, refresh persistence is the main risk and is covered by tests + manual check. A `localStorage` override for a shorter snooze TTL would help QA verify expiry without waiting 7 days. Happy to add that as a small follow-up (or in this PR if you prefer before merge).

matejkubinec (approved):

PR #5707, opened 2026-07-28
PMM-15186 Fix for invalid TLS when using change.

JiriCtvrtka:
PMM-15186

FB: https://github.com/Percona-Lab/pmm-submodules/pull/4575

**Problem**
The PR addresses four interconnected issues with pmm-admin and pmm-agent communicating over HTTPS:

Global transport mutation — Both CLIs reconfigured http.DefaultTransport in place, causing TLS settings to leak between unrelated HTTP clients.
Opaque error messages — Certificate verification failures and authentication errors surfaced as raw cryptographic errors without mentioning the --server-insecure-tls flag.
Regressions from initial fixes — The first attempt left pmm-agent still mutating the global transport, narrowed authentication hints too far, corrupted server messages, and leaked hints into JSON output.
Credential leakage — Invalid PMM Server URLs could echo plaintext passwords in logs, and nginx-error handling had no buffer bounds.
Solution Summary

The fix introduces isolated transport handling through a shared utils/apitransport package that clones transports before modification rather than mutating the global one. apitransport.SetAuth also centralizes the service_token/API-key-vs-basic-auth selection that pmm-admin and pmm-agent each reimplemented separately.

New utils/servererror utilities provide TLS certificate detection with helpful hints (WrapTLSError), distinguish authentication failures from permission errors (AuthHint), bound nginx response parsing to 64 KiB (NginxConsumer), assemble the shared consumer map both CLIs install on their transports (Consumers), and combine the TLS/nginx hint-wrapping sequence both CLIs needed (Explain) — replacing what had been several separate, drifting implementations of the same logic.

The api-tests integration-test harness, which still mutated the global transport and read nginx error bodies unboundedly, was also migrated onto apitransport/servererror, closing the one call site the original fix missed.

Error formatting preserves internal gRPC codes for diagnostic purposes while keeping JSON output unchanged. Credential redaction now covers edge cases where URLs lack // separators or contain slashes within passwords: a URL that parses with a real authority goes through url.URL.Redacted, opaque/authority-less/unparseable ones fall back to a userinfo regex, and anything still carrying an "@" that neither path can make sense of is replaced with a (redacted) placeholder rather than echoed back.

**Testing**
Comprehensive unit tests span transport isolation, URL validation, flag precedence, credential redaction (including passwords containing a slash), TLS mismatches, and nginx error handling. All builds, linting (including reviewdog/golangci-lint feedback), and race-condition tests pass.


JiriCtvrtka:
@copilot review

admin/commands/servererror_test.go
```diff
@@ -0,0 +1,128 @@
...
+		resp := agents.NewChangeAgentDefault(401)
+		resp.Payload = &agents.ChangeAgentDefaultBody{ //nolint:exhaustruct
+			Code:    13, // codes.Internal
+			Message: "Internal server error.",
+		}
+
+		e := GetError(resp)
+		assert.Equal(t, int32(13), e.GRPCCode)
+		assert.Equal(t, "Internal server error. Please check PMM Server logs.", ServerErrorMessage(e))
+	})
```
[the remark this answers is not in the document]
JiriCtvrtka: Not quite valid as written. Those constants are declared inside TestServerErrorMessage's function body, so they're not in scope for TestGetErrorFromGeneratedResponse, a separate test function.

agent/commands/setup.go
```diff
@@ -132,6 +133,36 @@ func checkStatus(configFilepath string, l *logrus.Entry) (string, bool) {
...
+	e, ok := errors.AsType[*mservice.RegisterNodeDefault](err)
+	if ok {
+		msg = e.Payload.Message
+		if e.Code() == http.StatusConflict {
+			msg += " If you want override node, use --force option"
+		}
+		// The HTTP status alone cannot tell an authentication failure apart from an
+		// internal one, since PMM Server maps several gRPC codes onto HTTP 401.
+		if hint := servererror.AuthHint(e.Code(), e.Payload.Code); hint != "" {
+			msg += "\n" + hint
```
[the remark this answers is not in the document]
JiriCtvrtka: Not really valid. The generated go-swagger client (register_node_responses.go) always does o.Payload = new(RegisterNodeDefaultBody) before returning this error type, so in the real runtime path Payload can never be nil when you get a *RegisterNodeDefault.

agent/commands/setup.go
```diff
@@ -132,6 +133,36 @@ func checkStatus(configFilepath string, l *logrus.Entry) (string, bool) {
...
+		}
+	}
+
+	if _, ok := err.(nginxError); ok { //nolint:errorlint
+		msg += ".\nPlease check pmm-managed logs."
```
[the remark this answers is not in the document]
JiriCtvrtka: Fixed.

utils/servererror/servererror.go
```diff
@@ -0,0 +1,166 @@
...
+// nginx page surfaces as an error of its own instead of as a JSON decoding failure.
+//
+// The body is trimmed: nginx pages end with a newline, which would otherwise leave a blank line
+// between the page and whatever the caller prints after it.
+func NginxConsumer() runtime.ConsumerFunc {
+	return func(reader io.Reader, _ any) error {
+		b, _ := io.ReadAll(reader)
+
+		return NginxError(strings.TrimSpace(string(b)))
+	}
```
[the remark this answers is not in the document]
JiriCtvrtka: check now

4nte (approved):

ademidoff (commented):
Code review of this branch (head `86dbd76b`): built, vetted, ran the touched packages' tests and `bin/golangci-lint` — all green, remaining lint hits are pre-existing. 12 comments below, most severe first.

**Correctness**
- `admin/commands/base/setup.go:70` — `credentialPattern` mangles host/port/path of the URL it redacts (reproduced)
- `utils/apitransport/apitransport.go:82` — cloned transport advertises h2 over ALPN with no h2 handler installed (reproduced)
- `utils/servererror/servererror.go:96` — an HTML 401/403 from a fronting proxy gets the "pmm-managed is down" hint
- `utils/servererror/servererror.go:82` — an expired certificate is reported as "self-signed or not valid for host"
- `admin/commands/base.go:137` — `GetError` can still panic on a missing/nil `Payload`

**Convention / tests / cleanup**
- `%q` in log and error messages (3 sites), `printResponse`'s `os.Exit` inside a test, `Consumers(nil)` + mutation in api-tests, an undrained response body, a single-use wrapper, and a duplicated test helper.

admin/commands/base/setup.go
```diff
@@ -31,24 +29,87 @@ import (
...
+// a string. It is a defensive final pass applied on top of structured parsing: url.URL.Redacted
+// only recognises userinfo in a properly structured "scheme://user:pass@host" URL. A URL typed
+// without "//" - a missing scheme being exactly the kind of mistake this code has to diagnose -
+// parses as opaque instead, with "user:password" sitting in Opaque, in cleartext, which
+// Redacted does not look at; a URL that fails to parse at all is not touched by it either.
+//
+// The password half deliberately allows "/": a scheme prefix such as "https:" would otherwise
+// also satisfy "name:" and swallow it into the match, but the password can legitimately contain
+// a slash, and excluding it let such a password slip through unredacted.
+var credentialPattern = regexp.MustCompile(`([^/@:\s]+):(?:[^/@\s][^@\s]*)?@`)
```
ademidoff: **`credentialPattern` corrupts the host, port and path of the URL it is redacting**

The regex runs unconditionally on `RedactDSN`'s output and treats *any* later `@` as a userinfo terminator, so it mangles the very URL this fatal log exists to help the user fix. Probed against the code on this branch:

```
"https://pmm-server:8443/v1/x@y"              -> "https://pmm-server:xxxxx@y"                  (port and path destroyed)
"https://admin:secret@pmm-server:8443/a@b"    -> "https://admin:xxxxx@pmm-server:xxxxx@b"      (second, bogus substitution
                                                                                                after RedactDSN already
                                                                                                redacted correctly)
```

A user who mistypes `--server-url` with an `@` anywhere after the authority is shown a mangled string and can't tell what they actually passed.

Suggestion: branch instead of layering — use `u.Redacted()` when `url.Parse` succeeds and `u.User != nil`, and fall back to the regex only for the opaque/unparseable cases it was written for.

utils/apitransport/apitransport.go
```diff
@@ -0,0 +1,89 @@
...
+
+// clone returns a copy of rt's HTTP transport which speaks HTTP/1.1 only. Working on a copy is
+// what keeps PMM's settings off http.DefaultTransport.
+func clone(rt *httptransport.Runtime) *http.Transport {
+	base, ok := rt.Transport.(*http.Transport)
+	if !ok {
+		panic(fmt.Sprintf("cannot configure a %T: an *http.Transport is required", rt.Transport))
+	}
+
+	httpTransport := base.Clone()
```
ademidoff, line 82: **`clone()` leaves the h2 ALPN advertisement in place while emptying `TLSNextProto`**

`http.Transport.Clone()` first runs `onceSetNextProtoDefaults` on `http.DefaultTransport`, which installs both `TLSNextProto{h2}` **and** a `TLSClientConfig{NextProtos: ["h2", "http/1.1"]}` on the global — so this package, written to keep PMM's settings off the global, still triggers that mutation — and the clone copies the config. Line 86 then empties `TLSNextProto`, but nothing clears `NextProtos`.

Probed on this branch: after `ConfigureLocal(rt)`, `rt.Transport.(*http.Transport).TLSClientConfig.NextProtos == [h2 http/1.1]` with an empty `TLSNextProto`. `Configure()` only overwrites `TLSClientConfig` for `scheme == "https"`, so this survives for `ConfigureLocal` and for `Configure(scheme != "https")`.

If such a transport ever does a TLS handshake (an `http://` server URL that redirects to `https`), the server negotiates h2, `TLSNextProto["h2"]` is absent, and net/http falls through to HTTP/1.1 framing on an h2 connection.

Suggestion: set `TLSClientConfig = nil` (or clear `NextProtos`) on the non-https path.

utils/servererror/servererror.go
```diff
@@ -0,0 +1,220 @@
...
+	return fmt.Errorf("%w.\n%s\nRe-run the command with %s to skip PMM Server TLS certificate validation, "+
+		"or configure PMM Server with a certificate valid for that host", err, reason, InsecureTLSFlag)
+}
+
+// Explain decorates err - a failure seen while contacting PMM Server that never reached its
+// API, so neither AuthHint nor a commands.ErrorResponse applies - with whichever hint fits: a
+// NginxError, when nginx answered instead of the API, gets NginxHint; anything else goes
+// through WrapTLSError, which is a no-op unless it was a certificate verification failure.
+// Host is the address the certificate was checked against.
+func Explain(err error, host string, insecureTLS bool) error {
```
ademidoff, line 96: **`NginxError` carries no HTTP status, so a 401/403 HTML page gets the "pmm-managed is down" hint**

`Explain` attaches `NginxHint` ("check pmm-managed logs, and that PMM Server is running") to *every* non-JSON response — the opposite diagnosis to the one this PR exists to give.

PMM's own nginx answers auth failures in JSON, so the shipped path is fine, but the PR's own fixture is an HTML 401: `admin/cli/cli_test.go:60` feeds `servererror.NginxError("<html><body><h1>401 Authorization Required</h1></body></html>")` through `printResponse` and asserts the pmm-managed-is-down hint. Any reverse proxy or LB in front of PMM Server that returns a plain 401/403 page — a common deployment — sends the user off to check pmm-managed logs when the real problem is their credentials or role.

Suggestion: carry the response status into `NginxError` so `Explain` can defer to `AuthHint` for 401/403.

utils/servererror/servererror.go
```diff
@@ -0,0 +1,220 @@
...
+// failures seen while certificate validation is already disabled, are returned unchanged.
+func WrapTLSError(err error, host string, insecureTLS bool) error {
+	if insecureTLS || !IsTLSCertificateError(err) {
+		return err
+	}
+
+	reason := "PMM Server TLS certificate could not be verified: it is either self-signed or not valid for the requested host."
+	if host != "" {
+		reason = fmt.Sprintf(
+			"PMM Server TLS certificate could not be verified: it is either self-signed or not valid for host %q.",
```
ademidoff: **An expired certificate is reported as "self-signed or not valid for host"**

`IsTLSCertificateError` also matches `x509.CertificateInvalidError` (expired, bad name constraints, too many intermediates) and `x509.UnknownAuthorityError`, but `WrapTLSError` emits one hardcoded sentence for all four.

Concretely: PMM Server presents a correctly-issued certificate for the right host that expired yesterday. It matches via `x509.CertificateInvalidError{Reason: x509.Expired}` — the package's own test in `utils/servererror/servererror_test.go` constructs exactly this — and the user is told the certificate is "self-signed or not valid for host pmm-server" and advised to re-run with `--server-insecure-tls`, hiding an expiry that should be renewed, not skipped.

Suggestion: derive the reason from the matched error rather than emitting one sentence for four distinct failures.

admin/commands/base.go
```diff
@@ -123,16 +123,30 @@ type ErrorResponse interface {
...
+	// the JSON output to preserve the documented shape of `pmm-admin --json` errors.
+	// PMM Server maps several gRPC codes onto HTTP 401, so Code alone cannot tell an
+	// authentication failure apart from an internal error - see ServerErrorMessage.
+	GRPCCode int32 `json:"-"`
 }
 
 // GetError converts an ErrorResponse to an Error.
 func GetError(err ErrorResponse) Error {
 	v := reflect.ValueOf(err)
 	p := v.Elem().FieldByName("Payload")
```
ademidoff, line 137: **`GetError` still dereferences a possibly-invalid `reflect.Value`, and evaluates `p.Elem()` twice**

Line 142 adds a `c.IsValid() && c.CanInt()` guard for `Code`, but `p` itself is unguarded: if an `ErrorResponse` implementation has no `Payload` field, or a nil `Payload` pointer, `p.Elem()` is the zero `Value` and `FieldByName` panics on line 138 — a panic inside error handling, i.e. exactly where pmm-admin is already in trouble.

Suggestion: hoist `pv := p.Elem()` once and guard `pv.IsValid()` before reading either field.

admin/commands/base/setup.go
```diff
@@ -67,62 +128,43 @@ func SetupClients(globalFlags *flags.GlobalFlags) {
...
-		switch globalFlags.ServerURL.Scheme {
-		case "http", "https":
-			// nothing
-		default:
-			logrus.Fatalf("Invalid PMM Server URL %q: scheme (https:// or http://) is missing.", globalFlags.ServerURL)
+		err = applyAgentServerParams(globalFlags, status)
+		if err != nil {
+			// status.ServerURL and err may both carry a password - reported by pmm-agent
+			// in the first case, embedded by url.Parse's own error text in the second.
+			logrus.Fatalf("Invalid PMM Server URL %q reported by local pmm-agent: %s.\n"+
```
ademidoff: **`%q` in log/error messages goes against the repo convention**

`AGENTS.md` (Logging, and again under Error Handling): "Don't interpolate strings with `%q` in log messages; use `%s`, or `'%s'` when the value can contain spaces."

Added by this PR:
- `admin/commands/base/setup.go:135` — `logrus.Fatalf("Invalid PMM Server URL %q reported by local pmm-agent: %s.\n"...)`
- `admin/commands/base/setup.go:145` — `logrus.Fatalf("Invalid PMM Server URL %q: %s."...)`
- `utils/servererror/servererror.go:82` — `"...not valid for host %q."`

The last one is asserted on in three tests (`not valid for host "pmm-server-second"`), so changing it means updating those assertions too.

admin/commands/base/setup.go
```diff
@@ -31,24 +29,87 @@ import (
...
+	}
 
-func (e nginxError) Error() string {
-	return "response from nginx: " + string(e)
+	return nil
 }
 
-func (e nginxError) GoString() string {
-	return fmt.Sprintf("nginxError(%q)", string(e))
+// credentialPattern matches a URL userinfo component - "user:password@" - occurring anywhere in
```
ademidoff: **Altitude: the whole redaction layer exists only to echo the user's own URL back at them**

`credentialPattern` + `sanitizeURLError` (a regex, a helper, and a new `dsnutils` dependency) are here so the fatal log can print the URL the user just typed. The diagnostic value of `Invalid PMM Server URL "<the string you just typed>"` is near zero, while it drags in three defensive layers — `RedactDSN`, a URL-shaped regex, and `*url.Error` unwrapping — each of which can fail open (see the `credentialPattern` comment above).

Suggestion: `logrus.Fatalf("Invalid PMM Server URL: %s. ...", sanitizeURLError(err))`, or at most printing `u.Scheme + "://" + u.Host`, is the smaller and strictly safer change to the underlying mechanism, and removes both helpers.

admin/cli/cli_test.go
```diff
@@ -0,0 +1,239 @@
...
+	t.Run("without --server-insecure-tls", func(t *testing.T) {
+		opts := &flags.GlobalFlags{ServerURL: cloneURL(t, serverURL)} //nolint:exhaustruct
+		base.SetupClients(opts)
+		redirectToTestServer(t, srv)
+
+		cmd := &inventory.ChangeAgentPostgresExporterCommand{AgentID: agentID} //nolint:exhaustruct
+		_, cmdErr := cmd.RunCmd()
+		require.Error(t, cmdErr)
+
+		wrapped := printResponse(opts, nil, cmdErr)
```
ademidoff: **Driving `printResponse` from a test means a fixture change can `os.Exit(1)` the whole test binary**

`printResponse` calls `os.Exit(1)` for any `commands.ErrorResponse`. It's only safe here because `cmdErr` is a transport error. If the httptest certificate or `redirectToTestServer` ever changes so the request reaches the handler and the handler returns an error status, `cmdErr` becomes a `commands.ErrorResponse`, `printResponse` exits, and `go test ./admin/cli/...` fails with no assertion output and no indication which test exited.

Suggestion: assert on `servererror.Explain` / `explainTransportError` directly instead of on the `os.Exit`-carrying `printResponse`.

api-tests/init.go
```diff
@@ -71,49 +68,18 @@ var (
...
-		b, _ := io.ReadAll(reader)
-		return new(NginxError(b))
-	})
-	transport.Consumers = map[string]runtime.Consumer{
-		runtime.JSONMime:    runtime.JSONConsumer(),
-		runtime.HTMLMime:    errorConsumer,
-		runtime.TextMime:    errorConsumer,
-		runtime.DefaultMime: errorConsumer,
-	}
+	transport.Consumers = servererror.Consumers(nil)
```
ademidoff, line 79: **`Consumers(nil)` then mutated, bypassing the `extra` parameter added for exactly this**

Here it's `servererror.Consumers(nil)`, and `api-tests/init.go:162` then mutates the returned map with `transport.Consumers["application/zip"] = runtime.ByteStreamConsumer()`, while `admin/commands/base/setup.go:157` passes the same entry through the new `extra` argument — two idioms for one API across three call sites.

Worse: the mutation only reaches the transport built in `init()`, so the Grafana client built in `api-tests/helpers.go:465` silently has no zip consumer.

utils/servererror/servererror.go
```diff
@@ -0,0 +1,220 @@
...
+// Both CLIs install it for the content types the PMM Server API never answers with, so that an
+// nginx page surfaces as an error of its own instead of as a JSON decoding failure.
+//
+// The body is trimmed: nginx pages end with a newline, which would otherwise leave a blank line
+// between the page and whatever the caller prints after it. A body over maxNginxBodySize is
+// truncated rather than rejected outright, since even a partial nginx page - unlike a read
+// failure - is still evidence of what happened and worth reporting.
+func NginxConsumer() runtime.ConsumerFunc {
+	return func(reader io.Reader, _ any) error {
+		b, err := io.ReadAll(io.LimitReader(reader, maxNginxBodySize+1))
```
ademidoff, line 177: **`NginxConsumer` never drains the remainder of the body, so the connection can't be reused**

For a response larger than the cap the `LimitReader` is abandoned mid-body; net/http sees an unread body on `Close` and tears the connection down instead of returning it to the keep-alive pool.

Harmless for a one-shot CLI, but `Consumers()` is also installed on the api-tests transport (`api-tests/init.go:79`), where every oversized non-JSON response then costs a fresh TCP+TLS handshake.

Suggestion: `io.Copy(io.Discard, reader)` after the limited read restores reuse.

admin/cli/cli.go
```diff
@@ -137,7 +138,25 @@ func printResponse(opts *flags.GlobalFlags, res commands.Result, err error) erro
...
+	if opts.JSON {
+		return err
+	}
+
+	return explainTransportError(opts, err)
+}
+
+// explainTransportError appends a hint to the failures which never reach PMM Server's API, and
+// returns every other error unchanged.
+func explainTransportError(opts *flags.GlobalFlags, err error) error {
```
ademidoff, line 154: **`explainTransportError` is a single-use wrapper around a nil check plus one call**

Called from exactly one place — `printResponse` at `admin/cli/cli.go:148` — and it adds a second name for the concept documented three lines above it. Inlining

```go
host := ""
if opts.ServerURL != nil {
    host = opts.ServerURL.Hostname()
}
return servererror.Explain(err, host, opts.SkipTLSCertificateCheck)
```

costs the same three lines and removes an indirection the reader has to follow.

admin/cli/cli_test.go
```diff
@@ -0,0 +1,239 @@
...
+	inventoryClient "github.com/percona/pmm/api/inventory/v1/json/client"
+	managementClient "github.com/percona/pmm/api/management/v1/json/client"
+	serverClient "github.com/percona/pmm/api/server/v1/json/client"
+	"github.com/percona/pmm/utils/servererror"
+)
+
+// restoreClients puts the package-level PMM Server API clients back the way they were once the
+// test is done. SetupClients reconfigures them for the whole test binary, and this test points
+// them at a server of its own, so without this a later test would dial a closed port.
+func restoreClients(t *testing.T) {
```
ademidoff: **`restoreClients` is duplicated byte-for-byte in two new test files**

This helper and the one at `admin/commands/base/setup_test.go:37` are identical, doc comment included. Adding a fourth generated client later means remembering to edit both; forgetting one leaves a test binary whose later tests dial a closed port — the exact failure the helper was written to prevent.

admin/commands/base/setup.go
JiriCtvrtka: Severity: medium

Confirmed, reproduced both strings exactly. Fixed: redactedServerURL now branches instead of layering — url.Parse + Redacted() when the URL has a real authority (u.Opaque == ""), and the regex only for the opaque and unparseable shapes it was written for. The regex is also anchored now, so it can't consume an @ from a path. Added two regression cases covering both of your examples, and dropped the dsnutils dependency since Redacted() is called directly.

utils/apitransport/apitransport.go
JiriCtvrtka: Severity: low (not a regression — main had the same hazard)

Confirmed: NextProtos == [h2 http/1.1] with an empty TLSNextProto after ConfigureLocal. Fixed by setting TLSClientConfig = nil in clone(); Configure replaces it wholesale on the https path. Worth noting this wasn't introduced here — the pre-PR code also only touched TLSClientConfig for https, so the same transport shipped before. Tightened TestConfigureLocal/TestConfigureHTTPSchemeSkipsTLS from if c != nil to assert.Nil, which is why they didn't catch it, and added a dedicated ALPN regression test.

admin/commands/base/setup.go
JiriCtvrtka: Severity: low (stated repo rule)

Fixed all three sites to '%s' and updated the three assertions in admin/cli, utils/servererror and agent/commands. Left the %q in NginxError.GoString — that one is a Go-syntax representation, which is what %q is for.

utils/servererror/servererror.go
JiriCtvrtka: Severity: low

Fixed: io.Copy(io.Discard, reader) on the truncated path only, so the connection goes back to the keep-alive pool.

utils/servererror/servererror.go
JiriCtvrtka: Severity: low — follow-up

Correct, and PMM's own nginx answering auth in JSON is exactly why the shipped path is fine. Plumbing the status into NginxError means changing the consumer signature, which go-openapi doesn't hand the response to — that's a larger change than this PR should carry. Filing a follow-up.

utils/servererror/servererror.go
JiriCtvrtka: Severity: low-medium — follow-up

Right, and the test constructing Reason: x509.Expired makes it concrete. Deriving the sentence from the matched error is the correct fix but it's a new message per branch plus its tests. Follow-up, tracked with #3.

admin/commands/base.go
JiriCtvrtka: Severity: trivial — out of scope

The panic is real, but p := v.Elem().FieldByName("Payload") and p.Elem().FieldByName("Message") are pre-existing and unchanged by this PR — the diff only adds the Code lookup, which is guarded. Hardening unchanged error-handling code is a separate change; happy to do it in its own PR.

admin/commands/base/setup.go
JiriCtvrtka: Severity: n/a — declined

Keeping it. A scheme-less admin:hunter2@pmm-server:8443 parses as opaque and Redacted() misses the password entirely, so the fallback isn't redundant — TestRedactedServerURL covers three such shapes. The #1 fix does remove one of the three layers, though: dsnutils.RedactDSN is gone and the regex no longer runs on top of parsed output.

api-tests/init.go
JiriCtvrtka: Severity: low — out of scope

The inconsistency is fair. But the transport.Consumers["application/zip"] mutation at init.go:162 and the Grafana client at helpers.go:465 are both untouched by this PR — the diff only swaps how the map is built. Routing them through extra is an api-tests cleanup I'd rather not fold into a pmm-admin fix.

admin/cli/cli.go
JiriCtvrtka: Severity: trivial · addressed, not as suggested

You were right that it was a wrapper around a nil check plus one call — but #8 asks for exactly this function as a test target, so inlining it would have left the os.Exit problem with nowhere to go. Instead I moved the --json carve-out down into it, so it now owns the complete transport-error policy rather than being an indirection over one call. printResponse ends with a single return explainTransportError(opts, err), and the function is no longer a second name for the same concept.

admin/cli/cli_test.go
JiriCtvrtka: Severity: low · fixed

Fixed as you suggested. All seven call sites now go through explainTransportError, which after the #11 change covers everything printResponse does with a transport error — including the --json passthrough, so no coverage was lost moving off it. Renamed TestPrintResponseTLSError/TestPrintResponseNginxError to TestExplainTransportErrorTLS/TestExplainTransportErrorNginx to match. printResponse no longer appears in the test file at all, so the fixture change you describe can't exit the binary.

admin/cli/cli_test.go
JiriCtvrtka: Severity: trivial · fixed

Fixed. Both copies are gone; the helper now lives once in a new admin/pkg/clienttest package as RestoreDefaults(t), called from all four sites. A shared package rather than a local helper because the two copies are in different packages — that's what made them diverge-able in the first place. Dropping the copies also freed several now-unused generated-client imports from both test files.

admin/commands/base/setup.go
```diff
@@ -31,24 +29,98 @@ import (
...
+var credentialPattern = regexp.MustCompile(`^((?:[a-zA-Z][a-zA-Z0-9+.-]*://)?[^/@:\s]+):(?:[^/@\s][^@\s]*)?@`)
+
+// redactedServerURL returns raw with any password it carries replaced by a placeholder, so a
+// PMM Server URL can be logged without leaking its credentials - however malformed the URL
+// turns out to be.
+func redactedServerURL(raw string) string {
+	// A URL with a proper authority: Redacted covers its userinfo exactly, and one without
+	// userinfo has nothing to redact. Either way credentialPattern must not run on top, or
+	// an "@" later in the path would be mistaken for a userinfo terminator.
+	u, err := url.Parse(raw)
```
ademidoff, line 94: **This branch leaks credentials the previous revision redacted**

`Redacted()` only redacts what landed in `u.User`. A URL can parse with `u.Opaque == ""` and still carry its credentials somewhere else — a one-slash typo puts them in `Path` with `User == nil`, so `Redacted()` returns them verbatim, and the anchored fallback on line 89 never runs.

Walked the real `--server-url` path on this head: kong parses the flag, `normalizeServerURL` returns "host is missing", and line 156 logs

```
Invalid PMM Server URL 'https:/admin:hunter2@pmm-server:8443': host is missing.
```

Shapes that regressed between `86dbd76b` and `6de2e064` — all four redacted correctly before, all four cleartext now:

| input | at `86dbd76b` | at `6de2e064` |
|---|---|---|
| `https:/admin:hunter2@pmm-server:8443` | `https:/admin:xxxxx@…` | **`…admin:hunter2@…`** |
| `https:///admin:hunter2@pmm-server:8443` | `https:///admin:xxxxx@…` | **`…admin:hunter2@…`** |
| `" https://admin:hunter2@pmm-server:8443"` (leading space/tab) | `…admin:xxxxx@…` | **`…admin:hunter2@…`** |
| `https://ad min:hunter2@pmm-server:8443` (space in userinfo) | `…admin:xxxxx@…` | **`…admin:hunter2@…`** |

The first is a plausible typo, and a missing slash is the same family of mistake as the missing scheme this function documents as its reason to exist.

Gating the branch on `u.User != nil` isn't enough on its own: I ran the fallback in isolation and it turns `https://pmm-server:8443/v1/x@y` into `https://pmm-server:xxxxx@y`, i.e. the bug this commit just fixed. `a:b/c@d` is genuinely ambiguous between `user:pass@host` and `host:port/path@…`, so no single pattern decides both.

Suggestion: make the fallback fail closed. If parsing yields no `User` and the raw string still contains an `@`, print a placeholder rather than the string — the message keeps its diagnostic value (the user typed the URL) and stops being a redaction problem.

utils/servererror/servererror.go
```diff
@@ -0,0 +1,226 @@
...
+
+		truncated := len(b) > maxNginxBodySize
+		if truncated {
+			b = b[:maxNginxBodySize]
+
+			// net/http only returns a connection to the keep-alive pool once its body
+			// has been read to the end. Abandoning an oversized body mid-read would cost
+			// a fresh handshake for every such response - irrelevant to a one-shot CLI,
+			// but these consumers are installed on the api-tests transport too.
+			_, _ = io.Copy(io.Discard, reader)
```
ademidoff: **The drain is unbounded — my suggestion, and I should have said bounded**

`io.Copy(io.Discard, reader)` has no limit, and it sits directly under `maxNginxBodySize`, whose comment promises that "a hostile server answering with an unbounded body cannot be used to exhaust memory". Memory is safe, but the read isn't bounded in bytes or in time, and there's no deadline anywhere on this path.

Probed on this head with a reader that never ends: **9.3 GiB consumed in 3 s and still going**. Against a slow-drip or endless body, pmm-admin hangs inside the consumer instead of printing the nginx page.

Worth weighing against what the drain buys: net/http only reuses a connection whose body was read to the *end*, so a partial drain buys nothing, and bodies over 64 KiB on this path are pathological to begin with. Either bound it (`io.CopyN(io.Discard, reader, 1<<20)` — reuse for anything plausibly real, one bounded read for the rest) or drop it and accept a fresh handshake for oversized responses. Connection reuse in api-tests was never worth an unbounded read; that's on me for suggesting the unqualified form.

ademidoff (commented):
Second pass over the fixes, against head `6de2e064`. I re-checked each of the twelve replies rather than taking them as given: built, ran the touched packages' tests and `bin/golangci-lint` (green; the two remaining lint hits are pre-existing on `main`), and re-probed the behaviour each fix claims.

**Verified fixed:** the redaction mangling (both reported strings now come out clean, with regression cases), the ALPN advertisement (probed: no `NextProtos` on `ConfigureLocal`, `Configure(http)` or `Configure(https)`), the `%q` sites (three assertion files updated, `%q` correctly kept in `GoString`), `printResponse` gone from `admin/cli/cli_test.go` entirely, `explainTransportError` now owning the `--json` carve-out rather than being an indirection, and both `restoreClients` copies replaced by `admin/pkg/clienttest.RestoreDefaults`.

**Deferred/declined:** agreed on all five. The `GetError` reflection and the api-tests consumer map are pre-existing and untouched here — I checked — so follow-ups are the right call.

Two new comments below, both in the fixes themselves: the redaction fix leaks credentials in four shapes the previous revision redacted, and the body drain I asked for is unbounded.

One nit with nowhere to anchor: `admin/commands/base/setup.go:129` carries an unused `//nolint:contextcheck`. It's pre-existing on `main`, but this PR rewrites that file, and the repo rule is to clear pre-existing lint in files you touch.

JiriCtvrtka:
> Second pass over the fixes, against head `6de2e064`. I re-checked each of the twelve replies rather than taking them as given: built, ran the touched packages' tests and `bin/golangci-lint` (green; the two remaining lint hits are pre-existing on `main`), and re-probed the behaviour each fix claims.
> 
> **Verified fixed:** the redaction mangling (both reported strings now come out clean, with regression cases), the ALPN advertisement (probed: no `NextProtos` on `ConfigureLocal`, `Configure(http)` or `Configure(https)`), the `%q` sites (three assertion files updated, `%q` correctly kept in `GoString`), `printResponse` gone from `admin/cli/cli_test.go` entirely, `explainTransportError` now owning the `--json` carve-out rather than being an indirection, and both `restoreClients` copies replaced by `admin/pkg/clienttest.RestoreDefaults`.
> 
> **Deferred/declined:** agreed on all five. The `GetError` reflection and the api-tests consumer map are pre-existing and untouched here — I checked — so follow-ups are the right call.
> 
> Two new comments below, both in the fixes themselves: the redaction fix leaks credentials in four shapes the previous revision redacted, and the body drain I asked for is unbounded.
> 
> One nit with nowhere to anchor: `admin/commands/base/setup.go:129` carries an unused `//nolint:contextcheck`. It's pre-existing on `main`, but this PR rewrites that file, and the repo rule is to clear pre-existing lint in files you touch.

Thanks @ademidoff — both the leak and the drain were real, fixed in f782b619.

1. Credential leak. Reproduced all four shapes. Three changes in redactedServerURL: Added u.Host != "" to the Redacted() guard. https:/user:pass@host and https:///user:pass@host parse without error and with Opaque == "", so they were taking that path — but with an empty authority the credentials sit in Path, which Redacted never looks at, and it returned them verbatim. credentialPattern's scheme prefix went from :// to :/*, so the fallback recognises userinfo after any number of slashes. Still anchored, so it does not go back to mangling an @ in a path. A fail-closed arm, which is what covers the whitespace cases: the pattern excludes \s deliberately, and those URLs are exactly the ones url.Parse already refused. If neither can say where the credentials end and an @ remains, the URL prints as (redacted) rather than in full. Both call sites print sanitizeURLError(err) next to it, so the user still gets the reason. Five regression cases added to TestRedactedServerURL; I confirmed all five fail against the previous implementation. There is also a NotContains(got, "ter2") assertion so the whitespace-in-password case cannot pass vacuously.
Worth recording why this regressed: the unanchored pattern in the earlier revision did redact these, but it also rewrote https://pmm-server:8443/path/with@sign into https://pmm-server:xxxxx@sign — it mangled the very URL the user is being asked to fix. The anchoring fixed that and opened this. The current version keeps both properties, and both are now covered by tests.

2. Unbounded drain. Correct, and the comment above it made it worse by treating a bounded read as sufficient. Now io.CopyN(io.Discard, reader, maxDrainSize) with maxDrainSize = 1 MiB — keep-alive reuse is preserved for anything resembling a real nginx page, and a hostile body costs a handshake instead of the process. New TestNginxConsumer subtest feeds an endless reader and asserts the consumer returns; it hangs forever on the old code.

3. //nolint:contextcheck. Leaving this one. It is on main at admin/commands/base/setup.go:57 and is one of eight such directives across the repo, so removing it here is a change to a pre-existing convention rather than to this PR — and since .golangci.yml runs default: all with no nolintlint settings, dropping it is as likely to break the lint gate as to clean anything. Happy to do a sweep of all eight separately if you want them gone.

ademidoff (approved):
Approving — re-verified both findings from the last pass against `f782b619`.

Redaction: probed 26 URL shapes, no leaks. The six that leaked at `6de2e064` now either redact (`https:/admin:…`, `https:///admin:…`) or fail closed to `(redacted)`, and the shapes the earlier fix got right stay right. Drain: an endless body now returns after exactly 64 KiB + 1 + 1 MiB instead of running unbounded, and a finite oversized body still drains to EOF so reuse is kept. Build, the touched packages' tests (also under `-race`) and `golangci-lint` are clean, with nothing attributable to this PR.

One thing to fix before merge, in the description rather than the code: it still says the redaction

> reuses utils/dsnutils.RedactDSN for the well-formed case instead of reimplementing it

which stopped being true when `redactedServerURL` was rewritten to branch on `url.Parse` — the `dsnutils` import is gone from `admin/commands/base/setup.go`. Since this squash-merges, that sentence is what ends up in the commit message.

PR #5720, opened 2026-07-30
PMM-15028 Flatten dashboards/, move pmm-app into the ui/ monorepo

fabio-silva:
Ticket number: PMM-15028

## Summary

Developer-experience restructuring requested in PMM-15028:

- **Flatten `dashboards/`**: `dashboards/dashboards/*` is moved up to `dashboards/*`, so the top-level `dashboards/` folder holds only JSON dashboard definitions (plus `misc/` tooling and `LICENSE`/`README.md`, kept for the RPM spec's `%license`/`%doc` lines).
- **Move `pmm-app` into the `ui/` monorepo**: `dashboards/pmm-app` → `ui/apps/pmm-app`, now a Yarn workspace + Turborepo member alongside `pmm` and `pmm-compat`. The plugin id (`pmm-app`) and its dist layout are unchanged, so no downstream Grafana paths change.
- **Symlink preserved**: `src/dashboards` is recreated at its new location, retargeted to the flattened top-level `dashboards/` folder. `plugin.json`'s 68 dashboard `includes` paths resolve unchanged — verified against the new symlink target.
- **No manual reload needed**: added a chokidar-based watcher (`ui/apps/pmm-app/scripts/watch-dashboards.mjs`) that mirrors dashboard JSON edits into the provisioned dashboards directory during local dev; wired into `pmm-app`'s `dev` script (`concurrently` alongside the existing webpack watcher) and into `.devcontainer/Makefile`'s `run-ui`/`build-ui` targets. QAN JS/TS changes already auto-reload via the existing webpack + livereload setup, just relocated.
- Fixed `pmm-app`'s `lint`/`test` scripts to be non-mutating/one-shot (previously `lint` ran `eslint --fix` and `test` ran `jest --watch`), matching `pmm-compat`'s conventions — required for these to behave correctly once folded into the shared workspace pipeline.
- Updated `AGENTS.md` guides (root, `dashboards/`, `ui/`, `ui/apps/pmm-app/`) and `CONTRIBUTING.md` docs to match the new layout.

## CI and packaging

- **`.github/workflows/dashboards.yml`**: dropped the `build`/`tests` jobs (now redundant — `pmm-app` lint/test/build is covered by `ui.yml` via Turborepo); fixed the changed-dashboard-files path pattern for the flattened tree; repointed the Slack notification jobs at the `check` job.
- **`.github/workflows/ui.yml`**: stopped ignoring `dashboards/**` changes (a renamed/deleted dashboard JSON can break `plugin.json` includes) and moved the "verify plugin.json dashboard paths exist" check here, fixed so it checks paths directly against the repo root instead of double-prefixing with `dashboards/`.
- **`build/packages/rpm/server/SPECS/percona-dashboards.spec`**: `%build`/`%install` now build and package `ui/apps/pmm-app` via `turbo run build --filter=pmm-app` instead of the now-removed `dashboards/Makefile`. `%license`/`%doc` are untouched (still `dashboards/LICENSE`/`dashboards/README.md`), which is exactly why those two files were kept at the flattened `dashboards/` root instead of moving with the rest of the plugin.
- **`.github/CODEOWNERS`**: added an explicit `/ui/apps/pmm-app/` entry so QAN review routing isn't silently absorbed by the generic `/ui/` rule.

[PMM-15028]: https://perconadev.atlassian.net/browse/PMM-15028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


build/docker/rpmbuild/Dockerfile.el9
```diff
@@ -8,12 +8,10 @@ FROM oraclelinux:9-slim
 RUN microdnf install -y dnf
 # enable nodesource repo for nodejs
 RUN curl -sL https://rpm.nodesource.com/setup_22.x | bash -
-RUN curl -sL https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo
 
 RUN dnf update -y && \
     dnf install -y --setopt=skip_missing_names_on_install=False \
     nodejs \
-    yarn && \
     dnf install -y gcc gcc-c++ \
```
mattiasimonato, line 15: Question on the yarn removal here: wasn't the `&& \` that terminated the first `dnf install` sitting on the `yarn` line? If so, dropping that line looks like it merges the two commands, and dnf would end up receiving `nodejs dnf install gcc ...` as a single package list.

If that reading is right, `install` isn't a package name, and `skip_missing_names_on_install=False` would turn that into a hard failure. Would something like this be what you intended?

```suggestion
    nodejs && \
    dnf install -y gcc gcc-c++ \
```

I haven't rebuilt the image to confirm, so happy to be wrong. Just flagging it since the rebuild has to happen for the yarn to pnpm switch to take effect. 🙂

build/docker/rpmbuild/Dockerfile.el8
```diff
@@ -8,12 +8,10 @@ FROM oraclelinux:8-slim
 RUN microdnf install -y dnf
 # enable nodesource repo for nodejs
 RUN curl -sL https://rpm.nodesource.com/setup_22.x | bash -
-RUN curl -sL https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo
 
 RUN dnf update -y && \
     dnf install -y --setopt=skip_missing_names_on_install=False \
     nodejs \
-    yarn && \
     dnf install -y gcc gcc-c++ \
```
mattiasimonato, line 15: Same question as in `Dockerfile.el9`, if that one holds:

```suggestion
    nodejs && \
    dnf install -y gcc gcc-c++ \
```

PR #5730, opened 2026-08-04
PMM-14956 HA alert templates

4nte:
Ticket number: PMM-14956

Feature build: [SUBMODULES-4506](https://github.com/Percona-Lab/pmm-submodules/pull/4506)

## What this adds

PMM Server gets working self-monitoring alert rules on first boot, and a shipped
expression fix reaches existing servers on upgrade. Two parts:

**1. Nine alert templates.** Five covering an HA cluster, four covering PMM's own
components.

| Template | Fires when | Severity | Bundle |
| --- | --- | --- | --- |
| `pmm_ha_no_leader` | No node holds the Raft leader lease | critical | ha |
| `pmm_ha_split_brain` | More than one node claims leadership | critical | ha |
| `pmm_ha_leader_flapping` | Raft term changes more than N times in 10m | warning | ha |
| `pmm_ha_node_unreachable` | Fewer nodes report than `PMM_HA_PEERS` declares | warning | ha |
| `pmm_ha_quorum_at_risk` | Live voters at or below the smallest quorum majority | critical | ha |
| `pmm_victoriametrics_down` | The metrics database stops being scraped | critical | components |
| `pmm_clickhouse_down` | The query-analytics database stops being scraped | critical | components |
| `pmm_grafana_down` | Grafana stops being scraped | critical | components |
| `pmm_qan_api2_down` | qan-api2 stops being scraped | warning | components |

**2. Rules provisioned from those templates automatically**, so nobody has to click
through the UI. pmm-managed renders a Grafana alerting provisioning file into
`/srv/grafana/provisioning/alerting/` and Grafana ingests it at startup. The rules
are read-only in the UI (`provenance: file`) and land in two folders: **PMM High
Availability** and **PMM Server**.

> [!IMPORTANT]
> **Upgrading turns on four PMM Server component alerts for every install**, standalone
> included. Users who already have a contact point configured will start receiving
> them. This is intended, not a side effect - a PMM Server with no self-monitoring is
> the problem being fixed - but it needs a release note. Set
> `PMM_ENABLE_COMPONENT_ALERTS=false` to opt out.

## Configuration

Two environment variables, both defaulting to true, read once at start:

| Variable | Effect |
| --- | --- |
| `PMM_ENABLE_HA_ALERTS` | Provision the five HA rules. No effect unless HA is on |
| `PMM_ENABLE_COMPONENT_ALERTS` | Provision the four PMM component rules |

Both are additionally gated on **Percona Alerting** being enabled, since the rules are
built from Percona alert templates.

## Two deliberate trades, so they are not read as oversights

**Configuration is environment-only, because the hard part is changing it at runtime.**
A runtime toggle would have to solve three problems that an environment variable simply
does not have:

- **Applying it means restarting Grafana.** Grafana reads provisioning only at startup,
  and pmm-managed holds no Grafana credentials to ask for a reload - that permission
  belongs to Grafana Admin, which a service account can never be. On HA the leader's
  Grafana is the one serving the request, so flipping the switch would kill the user's
  own session.
- **The change reaches one node; every node has to act on it.** Each node writes its own
  file and each Grafana reads its own copy, so a follower left behind would silently
  revert the cluster the next time its Grafana restarted. Nothing distributes settings
  between nodes, so closing that gap means polling, which converges in minutes and
  leaves nodes disagreeing meanwhile - the existing Percona Alerting gate behaves exactly
  this way.
- **Every one of those paths needs its own state.** A restart that does not come back
  needs rollback; a failed apply needs a retry record; a leader elected mid-change needs
  to resume work the last one dropped. Three more failure modes to get right, for a
  switch nobody flips twice.

An environment variable removes all three. It can only change by recreating the
container, which restarts Grafana anyway, so the file is read as part of a restart that
was already happening; and every node takes the value from the same place, so there is
nothing to synchronise. What is lost is turning a bundle off without a restart, which is
an install-time decision in practice - and a Grafana silence already quietens a rule
immediately.

**The metric reports what PMM did, not what Grafana holds.** `pmm_alerting_provisioning_info`
carries `state="written"`, not `applied`, and a follower deferring to the leader is
**not** counted as an error. PMM cannot know that Grafana ingested a file, so it does
not claim to; proving it would mean querying Grafana's schema at runtime to establish
something whose failure is already loud, because a Grafana that cannot load its
provisioning does not start.

## Design points worth a reviewer's attention

- **Every node renders and writes; only the leader restarts Grafana.** Alert rules live
  in the Grafana database all nodes share, so one ingestion serves the cluster. A node
  that stopped writing would silently revert the cluster the next time its own Grafana
  restarted, which is why writing is not leader-gated.
- **The datasource UID is never guessed.** Grafana only began deriving it from the
  datasource name in 8.3.4, and PMM shipped older Grafanas until 2.28.0, so servers
  first installed before then store a random one. A wrong UID leaves rules reporting
  healthy while querying nothing, so resolution fails closed: a missing row or absent
  table means a first boot and derives, anything else refuses to render.
- **Rule UIDs are frozen and append-only.** Removing one leaves the rule behind forever
  on every server that ever created it; retiring means moving the UID to
  `retiredRuleUIDs`.
- **A UID is only PMM's while PMM provisions it.** UIDs owned by someone else are
  omitted from both the group and the deletions - claiming one back kills Grafana on a
  provenance mismatch, and deleting one silently destroys a user's rule.
- **A bad provisioning file stops Grafana starting**, which takes the whole UI and API
  down with it. Hence validate-by-reparsing, atomic write, and last-known-good rollback.

## Template expression details

- **`last_over_time(...[2m])` on every query, not a bare selector.** The metrics
  database's own staleness window is longer than the alert needs. Without the bounded
  lookback a departed leader's final sample keeps counting, which delays
  `pmm_ha_no_leader` and can make an ordinary failover look like a split brain. The 2m
  window is deliberately shorter than every `for:`.
- **`max by(node_id)` before aggregating.** Collapses duplicate series for one node, so
  extra scrape labels cannot make a single node count twice and raise a split brain on
  its own.
- **`< bool` / `> bool` rather than a bare comparison.** A bare comparison filters the
  series away and Grafana evaluates the empty result as a label-less NoData instance,
  which breaks alert identity and grouping. The `bool` form keeps one instance carrying
  its labels.
- **`pmm_ha_quorum_at_risk` suppresses itself below three nodes** via
  `and on() (... > 2)`. On one- and two-node clusters the condition is permanently true,
  so it would be noise. `pmm_ha_node_unreachable` covers those sizes.
- **`pmm_ha_node_unreachable` uses `unless` plus an `absent()`-guarded fallback.** A node
  that is down emits nothing, so it cannot be named from a live series. The first term
  names nodes seen in the last 6h but not the last 2m; the fallback catches a node that
  never reported or has been gone longer than 6h, and reports it without a name. The
  guard stops both terms firing for the same node.
- **A node that is down stops emitting `pmm_ha_up` rather than reporting 0.** That is why
  the counting alerts compare against `pmm_ha_expected_nodes` instead of summing a
  health gauge.
- **Nothing is emitted at all when HA is disabled**, which is what keeps the HA templates
  silent on standalone PMM.

## Verification

Unit tests and `-race` green for every touched package; CI-equivalent lint clean; each
commit builds individually.

End to end on a dedicated 3-node kind cluster and on standalone containers:

| Case | Result |
| --- | --- |
| Fresh install, standalone | 4 component rules, `health: ok`, no HA folder |
| Fresh install, HA | 9 rules across both folders, one set for the cluster |
| Both bundles off, either topology | 0 rules, all 9 UIDs listed for deletion |
| Upgrade with a changed template | Live rule carries the new text, no duplicates |
| Steady state | 30-minute soak on both topologies, no rule left Normal, no restart churn |
| Component alert fires | qan-api2 stopped: pending at t+69s, firing at t+360s, recovered |
| Percona Alerting off | Rules removed cluster-wide in 188s (HA), 203s (standalone) |
| Divergent datasource UID | Rendered file follows the database; rules healthy |
| Grafana database unreachable | Fails closed, no file written, previous kept; recovery with **zero** Grafana restarts |
| Corrupt provisioning file | Grafana goes FATAL, PMM repairs the file and starts it unattended |
| A UID squatted by a user rule | Grafana still starts, the other rules provision, the conflict is logged and counted |
| Retiring a rule | UID moves to deletions, rule removed, the rest untouched |

## Documentation

Docs are in percona/pmm#5975, against `doc-3.10.0`, because documentation is released
ahead of the code. It covers the templates and that the rules are provisioned
automatically, both bundles and how to turn them off, the Helm values and the
rolling-upgrade caveat, and the two new environment variables. The two merge together.


theTibi:
@coderabbitai full review

managed/cmd/pmm-managed/main.go
```diff
@@ -1240,6 +1237,31 @@ func main() { //nolint:gocognit,maintidx,cyclop
...
+	seen := make(map[string]struct{})
+
+	for node := range strings.SplitSeq(peers, ",") {
+		node = strings.TrimSpace(node)
+		if node == "" {
+			continue
+		}
+		if _, ok := seen[node]; ok {
+			logrus.Warnf("Ignoring duplicate entry %q in PMM_HA_PEERS.", node)
+			continue
```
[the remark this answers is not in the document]
4nte: Fixed in c66f5e8d0. `parseHAPeers` now takes the `main` logrus entry and logs the peer as a field: `Ignoring duplicate entry in PMM_HA_PEERS. component=main peer=n1`. Verified by running the binary with a duplicate peer list.

documentation/docs/install-pmm/install-HA-clustered.md
```diff
@@ -838,6 +851,12 @@ View detailed role and health information for all PMM nodes in one place.
     - **Follower** status: which nodes are on standby
     - **Health** status: whether each node is responding
 
+### Get alerted about cluster problems
+
+Checking the High Availability page tells you the state of the cluster right now, but it does not notify you when that state changes. PMM ships alert templates that cover the failure modes of an HA cluster: no active leader, split-brain, a flapping leader, an unreachable node, and a quorum at risk.
+
+Like all alert templates, they do not create alert rules on their own: create a rule from each template and point it at a contact point. For the template list, the setup steps and the coverage limitations, see [PMM High Availability templates](../alert/templates_list.md#pmm_ha_alerts).
```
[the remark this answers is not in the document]
4nte: Fixed in c66f5e8d0. The paragraph now says PMM creates and maintains the HA rules, that the only setup is a contact point, and how to turn the rules off with `PMM_ENABLE_HA_ALERTS=false`. Both links resolve in the rendered site.

managed/services/alerting/provisioning.go
```diff
@@ -0,0 +1,662 @@
...
+	previous, changed, err := p.write(content)
+	if err != nil {
+		p.l.Errorf("Failed to write the alert rule provisioning file: %s.", err)
+		p.metrics.recordError(stageWrite)
+		return
+	}
+
+	p.metrics.setRendered(hash, bundles)
+	p.metrics.setWritten(hash)
+	p.retryBackoff = 0
```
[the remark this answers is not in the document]
4nte: Confirmed and fixed in c66f5e8d0 as proposed: the reset moved into the three terminal branches (no-op, successful apply, leader deferral), so consecutive failed applies grow the delay. Covered by `TestProvisionerApplyRetryBacksOff`, which fails on the previous code. Also observed live on a 3-node kind cluster with a Grafana that never became ready: retries followed the failures after 2, 4 and 8 seconds.

managed/services/alerting/provisioning.go
```diff
@@ -0,0 +1,662 @@
...
+	case err == nil:
+		p.applyPending = false
+		p.metrics.setApplyPending(false)
+
+	case errors.Is(err, errDeferredToLeader):
+		// Not a failure and not this node's work to retry: the leader applies, or Grafana reads the
+		// file itself as it starts. Counting this would make every healthy follower look broken.
+		p.applyPending = false
+		p.metrics.setApplyPending(false)
+		p.l.Debugf("Alert rules written; %s.", err)
```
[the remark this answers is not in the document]
4nte: Confirmed, and it was worse than described: in HA `IsLeader` is false on every node during `setup()` because the HA service starts later, so the startup apply deferred on every node, every boot. The rules still usually landed because pmm-managed wrote the file a few seconds before Grafana reached its provisioning phase, which is why the E2E rows passed. The gap opened whenever that write was delayed, for example by the datasource retry.

Fixed in c66f5e8d0, but not by keeping `applyPending` set on deferral: that would report every healthy follower as pending for good and restart Grafana on whichever node becomes leader next, right after a failover. Instead the boot records a debt (`startupApplyOwed`) that the startup reconcile sets and an unchanged file or a successful apply clears. While it is owed, `apply` skips the leader gate and restarts the local Grafana; a starting node serves no traffic, so the restart costs nothing. The debt survives a render that fails at boot, so the retry that finishes the recovery still restarts without waiting for a leader. The tick stays leader-gated, because the leader sees the same shared-state change itself.

Verified on a 3-node kind cluster: a pod booting with changed content restarts its Grafana about 0.2 s after the write with no leader elected, and the rules land in the shared database; unchanged content produces no write and no restart.

managed/utils/dir/dir.go
```diff
@@ -41,6 +41,55 @@ func CreateDataDir(path string, perm os.FileMode) error {
...
+		tmp.Close() //nolint:errcheck,gosec
+		return fmt.Errorf("cannot chmod '%s': %w", path, err)
+	}
+
+	err = tmp.Close()
+	if err != nil {
+		return fmt.Errorf("cannot close '%s': %w", path, err)
+	}
+
+	err = os.Rename(tmpName, path)
```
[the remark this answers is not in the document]
4nte: Fixed in c66f5e8d0. `WriteFileAtomic` now calls `Sync` on the temporary file before closing and renaming it. strace of a real call shows `openat` of the temp file, `fsync`, then `renameat`. The directory is deliberately not synced: a rename lost to a crash leaves the previous, valid file, which is safe here.

managed/services/alerting/provisioning.go
```diff
@@ -0,0 +1,695 @@
...
+	p.l.Infof("Restarting Grafana to apply alert rule changes.")
+
+	err := p.supervisord.RestartSupervisedService(ctx, grafanaProgramName)
+	if err != nil {
+		return fmt.Errorf("failed to restart Grafana: %w", err)
+	}
+
+	err = p.waitForGrafana(ctx)
+	if err != nil {
+		p.rollback(previous)
```
JiriCtvrtka, line 717: Severity: Blocker
Problem: restart loop after a failed apply
Desc: restartGrafana calls p.rollback(previous) when Grafana doesn't come back, then returns an error. The default arm at provisioning.go:427 sets applyPending = true and arms the backoff (datasourceRetryMax = 1 min, provisioning.go:87). The next reconcile renders the same content, write() compares it against the rolled-back file, sees changed = true, and calls apply() again → another restart, another readyTimeout wait, another rollback.

Nothing in the state machine remembers that this content was already rejected, so it repeats at ≤1-minute intervals indefinitely, each cycle taking Grafana (and the whole UI) down for up to the ready timeout. The rollback is the thing that makes it a loop rather than a single failure — without it, the unchanged-file early return at provisioning.go:403 would stop the retries.

Fix: record the hash of the content that failed to come up (p.rejectedHash = hash next to the rollback) and skip apply() in reconcile while hash == p.rejectedHash, clearing it whenever the rendered content changes. The file stays rolled back, the error metric stays set, and Grafana is restarted once per bad revision instead of forever.

managed/services/alerting/provisioning.go
```diff
@@ -0,0 +1,695 @@
...
+// with which to ask for a reload, so restarting Grafana is the only lever there is. It is rarely
+// needed: the file changes only when the configuration or the shipped templates change, and both of
+// those arrive by recreating the container, which restarts Grafana anyway. What is
+// left is the boot where the rendered content changed, and the case where the datasource resolved
+// late.
+func (p *Provisioner) apply(ctx context.Context, trigger provisioningTrigger, previous []byte) error {
+	running := p.supervisord.ProgramState(ctx, grafanaProgramName)
+
+	switch {
+	case running == nil:
```
JiriCtvrtka: Severity: Blocker
Problem: unknown program state silently discharges the apply debt
Desc: apply() returns nil for running == nil. ProgramState (managed/services/supervisord/supervisord.go:179) returns nil not only for EXITED/UNKNOWN but for any supervisorctl status invocation whose output it can't parse into two fields — i.e. a transient command failure. On that nil, the err == nil arm at provisioning.go:410 clears applyPending, startupApplyOwed and retryBackoff, and the metric reports state="written".

A Grafana that is actually running then keeps serving the old rules, and no retry is armed: every later tick renders identical content and returns early at provisioning.go:403 with !changed && !applyPending. The changed rules never reach Grafana until some unrelated restart, and the metrics say everything succeeded.

Fix: return a sentinel (errGrafanaStateUnknown) instead of nil, and give it its own arm in the reconcile switch that keeps applyPending/startupApplyOwed set and calls armRetryLocked() — without recordError(stageApply), so the legitimate first-boot case doesn't look broken.

managed/services/alerting/provisioning.go
4nte: Confirmed, the rollback turns one failure into a loop. Fixed along the lines you suggested, with two changes. Each rendered revision gets 2 tries rather than 1, because a single readiness timeout is often not about the content (a slow host, the Grafana DB briefly unreachable). And only a "Grafana did not come back" failure counts against the revision; a supervisorctl command that fails to run does not. The check runs before the write, so the file Grafana last accepted stays on disk and a later unrelated restart can't bring the bad one back. If Grafana was left FATAL, it gets started on that file. Any change to the rendered content clears the count. Tests: `TestProvisionerStopsOfferingARevisionGrafanaRejects`, `...OffersNewContentAfterGivingUpOnARevision`, `...StartsAGrafanaLeftDownByARevisionItGaveUpOn`, `...ChargesOnlyGrafanaFailuresToTheRevision`, `...RollsBackAStartGrafanaDoesNotSurvive`.

managed/services/alerting/provisioning.go
4nte: Confirmed, a status command that fails to run looks the same as "not configured yet" and cleared the owed apply. I went with a different fix than a sentinel plus retry. On first boot that retry would find Grafana running on its next pass and, with `startupApplyOwed` still set, restart a Grafana that had already read the new file. Instead, when the state is unknown, `apply()` asks Grafana itself. If it answers, it is serving rules it read earlier, so it's treated as running and goes through the normal restart path (leader check included). If it doesn't answer, it hasn't started yet and will read the new file when it does. Test: `TestProvisionerAppliesToAGrafanaThatServesWhileItsStateIsUnknown`.

The one case still not covered is a supervisorctl failure that coincides with a running Grafana failing its health check at that same moment; that needs two independent failures at once, and the next config change or restart applies the rules.

managed/services/alerting/provisioning.go
```diff
@@ -0,0 +1,809 @@
...
+		if p.rejectedApplies >= maxApplyAttemptsPerRevision {
+			// Out of attempts. Stop the fast retry as well: everything it would reach now returns
+			// at the gate above, and only a change to the rendered content can make this worth
+			// trying again. The slow tick keeps checking for that, and for a Grafana left down.
+			p.retryBackoff = 0
+			p.l.Errorf("Giving up on applying these alert rules: Grafana did not come back %d times, "+
+				"and the file it last accepted has been restored. The rules Grafana holds are the "+
+				"previous ones, and PMM will try again once the rendered rules change: %s.", p.rejectedApplies, err)
+			return
+		}
```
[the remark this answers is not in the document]
4nte: Confirmed: giving up cancels the fast retry, so a Grafana left FATAL by the last attempt waited a whole tick although the rolled-back file was already on disk. The give-up branch now calls `recoverGrafanaLocked` before returning, and `TestProvisionerStartsAGrafanaLeftDownByARevisionItGaveUpOn` now requires the start to happen in that same reconcile rather than on a later tick (it fails without the fix).

managed/services/alerting/provisioning_render.go
```diff
@@ -0,0 +1,377 @@
...
+	if err != nil {
+		return provisioningRule{}, fmt.Errorf("failed to fill template labels: %w", err)
+	}
+
+	labels["percona_alerting"] = "1"
+	labels["severity"] = common.Severity(template.Severity).String()
+	labels["template_name"] = template.Name
+	labels[provisionedLabel] = "1"
+
+	for name, value := range labels {
```
JiriCtvrtka, line 222: Severity: Blocker
Desc: $ guard covers labels only; Grafana os.ExpandEnvs annotations too, so every shipped {{ $labels.x }} annotation becomes {{ .x }} and ha_node_unreachable's {{ if $labels.node_id }} collapses to {{ if .node_id }}. Emit $$labels or extend the guard to annotations.

managed/services/alerting/provisioning.go
```diff
@@ -0,0 +1,814 @@
...
+		// previous one, and no other node can see that: the leader renders the same content from
+		// the same shared state and finds its own file unchanged, so it applies nothing. Nor is
+		// anyone leader yet while this node starts. A starting node serves no traffic, so
+		// restarting its Grafana costs nothing, and it is the only way the new rules reach the
+		// cluster before some Grafana happens to restart.
+		if !p.startupApplyOwed && !p.leader.IsLeader() {
+			return fmt.Errorf("%w on %s", errDeferredToLeader, trigger)
+		}
+
+		return p.restartGrafana(ctx, previous)
```
JiriCtvrtka, line 701: Severity: High
Desc: apply() restarts Grafana from supervisord process state without an IsReady check; on first boot this can restart Grafana mid-start, and the 2s retry loop can exhaust maxApplyAttemptsPerRevision leaving no rules file. PLAUSIBLE — needs cold start >2min.

managed/services/alerting/provisioning.go
```diff
@@ -0,0 +1,814 @@
...
+}
+
+// restartGrafana restarts Grafana and waits for it to answer again, restoring the previous file if
+// it does not.
+func (p *Provisioner) restartGrafana(ctx context.Context, previous []byte) error {
+	p.l.Infof("Restarting Grafana to apply alert rule changes.")
+
+	err := p.supervisord.RestartSupervisedService(ctx, grafanaProgramName)
+	if err != nil {
+		return fmt.Errorf("failed to restart Grafana: %w", err)
```
ademidoff, line 712: Severity: High
Problem: a failed `supervisorctl restart` isn't always "a command that would not run"

Grafana runs with `startsecs = 1`, and supervisor 4.2.4 (the pinned version) makes `restart`/`start` exit non-zero when the process dies before it reaches RUNNING. Its output is `ERROR (abnormal termination)` or `spawn error`. Content Grafana can't start from produces exactly that. The corrupt-file row in the description ends in FATAL, and a process only reaches FATAL by dying before `startsecs` on every retry.

On this path:
- the new file stays on disk, because there's no `rollback`
- the revision isn't charged, because the error isn't `errGrafanaNotBack`
- `applyPending` stays set.

As a result, every retry restarts Grafana on the same file, and at boot so does every `setup()` retry (see the `ProvisionAtStartup` comment). Once supervisord marks it FATAL, the retry moves to the start path below, which then loops.

Suggestion: roll back here as well and wrap the error as `errGrafanaNotBack`. Alternatively, after a failed command, read `ProgramState` and treat anything other than running as Grafana not coming back, so that only a genuinely missing `supervisorctl` goes uncharged.

---

managed/services/alerting/provisioning.go
```diff
@@ -0,0 +1,814 @@
...
+		//
+		// Deliberately not leader-gated. The gate exists to elect a single actor for a disruptive
+		// action across nodes sharing one database; a dead Grafana serves nobody, so there is no
+		// blast radius to contain, and at startup no node is leader yet.
+		p.l.Warnf("Grafana is down and supervisord will not restart it; starting it to apply the alert rules.")
+		err := p.startGrafana(ctx)
+		if err != nil {
+			// Same reasoning as the restart below: a Grafana that will not start on this file has
+			// to be left the one it last accepted, or it can never come up again.
+			p.rollback(previous)
```
ademidoff, line 678: Severity: High
Problem: a start failure is rolled back but not charged, and Grafana is never started on the restored file

This has the same root cause as the restart path, but here the loop has no exit:
1. `StartSupervisedService` fails because Grafana died within `startsecs`.
2. The file is rolled back, but nothing starts Grafana on the restored file.
3. `rejectedHash` was never set, so the next retry renders the same content and writes it again, since it differs from the rolled-back file.
4. The retry finds Grafana FATAL, starts it on the bad file, and fails again.

The budget never runs out, so `recoverGrafanaLocked` is never reached and the UI stays down indefinitely. `TestProvisionerChargesOnlyGrafanaFailuresToTheRevision` currently pins this behaviour. The mock returns an error from `StartSupervisedService` and asserts `rejectedApplies == 0`.

Charging this failure would let the gate plus `recoverGrafanaLocked` close the loop. So would starting Grafana on the restored file right after the rollback.

---

managed/cmd/pmm-managed/main.go
```diff
@@ -581,6 +582,15 @@ func setup(ctx context.Context, deps *setupDeps) bool {
...
 	}
 
+	// Runs after UpdateSettingsFromEnv above, which has already written grafana.ini and started
+	// Grafana, so the rules are usually applied by restarting it rather than by being in place
+	// first. Moving this earlier would avoid that restart but would read the settings before the
+	// environment has been applied to them, so the Percona Alerting gate could see a stale value.
+	// Never fails setup: this is retried every couple of seconds until it succeeds, and a permanent
+	// failure here would mean Grafana was never started at all.
+	deps.l.Infof("Provisioning built-in alert rules...")
+	deps.provisioner.ProvisionAtStartup(ctx)
```
ademidoff, line 592: Severity: Medium
Problem: the first `setup()` call now blocks on a Grafana restart

The first `setup()` call is synchronous. It runs before the gRPC and HTTP servers, the auth server and `haService.Run` start. On every boot where the rendered content changed, which means every fresh install and every upgrade that changes a template, it now waits here for the Grafana restart and the readiness polling. When Grafana is slow, that's up to `grafanaReadyTimeout` (2m). Until now `setup()` only did non-blocking readiness checks and handed off to the retry goroutine on failure.

Could the startup apply move into the provisioner's own goroutine, for example as the first reconcile in `Run`, so that `setup()` only renders and writes the file?

---

managed/services/alerting/provisioning.go
```diff
@@ -0,0 +1,814 @@
...
+}
+
+// ProvisionAtStartup writes the rules while PMM starts.
+//
+// Grafana is normally already running by this point, on a fresh container as well as an existing
+// one: setup() writes grafana.ini and starts it from UpdateSettingsFromEnv, before this runs. So
+// this may restart Grafana - once, on the boot where the content changed. That restart does not
+// wait for a leader, and the debt survives a render that fails here: the retry that finishes the
+// recovery settles it. See startupApplyOwed.
+func (p *Provisioner) ProvisionAtStartup(ctx context.Context) {
```
ademidoff, line 360: Severity: Medium
Problem: the `setup()` retry loop calls this repeatedly, outside the provisioner's own retry logic

The `setup()` retry goroutine calls this every 2s for as long as `setup()` keeps failing, which is common on first boot while VictoriaMetrics and VMAlert come up. That has two side effects:

- **It bypasses the backoff and re-arms the boot exemption.** Each call ignores `retryBackoff` and sets `startupApplyOwed` again. With `applyPending` set, for example after the restart failure above, Grafana is restarted every 2s whatever the backoff says. The leader-gate exemption also lasts as long as `setup()` keeps failing, not just for the boot.
- **`Run` doesn't see a backoff armed here.** `Run` has already computed `retry` as nil and only re-reads it after its `select` wakes. If one of these calls arms the backoff and that same `setup()` call then succeeds, the retry waits for the next tick, 4 to 6 minutes away.

Setting `startupApplyOwed` only on the first call, plus a wake channel that `armRetryLocked` signals to `Run`, would address both.

---

managed/services/alerting/provisioning.go
```diff
@@ -0,0 +1,814 @@
...
+
+	err = p.apply(ctx, trigger, previous)
+	switch {
+	case err == nil:
+		p.applyPending = false
+		p.metrics.setApplyPending(false)
+		p.retryBackoff = 0
+		p.startupApplyOwed = false
+
+	case errors.Is(err, errDeferredToLeader):
```
ademidoff, line 453: Severity: Medium
Problem: a deferral is lost if leadership moves before the leader applies

Clearing `applyPending` here assumes the node that is leader when a follower defers is the node that will apply. If leadership moves in between, no node applies:

1. B, a follower, ticks, writes and defers.
2. Leadership transfers to B.
3. A, now a follower, ticks, writes and defers.
4. B's next tick finds its file unchanged with `applyPending == false` and returns early.

The shared database keeps the old rules until some Grafana happens to restart. For example, rules keep firing after Percona Alerting is turned off.

One option is to remember that this node deferred a hash its own Grafana hasn't read, and treat it as owed if the node becomes leader.

---

managed/services/alerting/provisioning.go
```diff
@@ -0,0 +1,814 @@
...
+			if lastErr == nil {
+				return nil
+			}
+		}
+	}
+}
+
+// rollback puts back the content Grafana last started from, so that a server left in this state
+// still has a Grafana that starts.
+func (p *Provisioner) rollback(previous []byte) {
```
ademidoff, line 790: Severity: Low
Problem: the rollback target is whatever was on disk, not a known-good file

`previous` is just the file that was on disk before the write. Nothing establishes that Grafana ever accepted it, although the description calls it a last-known-good rollback.

Suppose Grafana is FATAL on the on-disk file, for example one written by an older release that a newly shipped Grafana rejects, and the new content then times out. The rollback restores the file that broke Grafana, and `recoverGrafanaLocked` starts Grafana on it.

Recording the hash of the content Grafana last came back healthy on would make the rollback target explicit.

---

managed/services/alerting/provisioning_metrics.go
```diff
@@ -0,0 +1,230 @@
...
+	for bundle, enabled := range m.bundles {
+		var state string
+		switch {
+		case !enabled:
+			state = stateDisabled
+		case m.failed:
+			state = stateError
+		case m.applyPending:
+			state = statePending
+		case m.renderedHash != "" && m.renderedHash == m.writtenHash:
```
ademidoff, line 205: Nit: `renderedHash` and `writtenHash` are only ever set one after the other, after a successful write (provisioning.go:433-434). Once set they are always equal, so this case can't tell anything apart, and the `default: statePending` below is reachable only before the first render. A single hash set after the write would do the same job.

---

managed/services/alerting/provisioning_datasource.go
```diff
@@ -0,0 +1,384 @@
...
+		User:     url.UserPassword(bundledGrafanaUser, bundledGrafanaPassword),
+		Host:     fallback.Addr,
+		Path:     bundledGrafanaDB,
+		RawQuery: fallback.SSLParams,
+	}
+
+	return dsn.String(), nil
+}
+
+func envOrDefault(name, fallback string) string {
```
ademidoff, line 147: Nit: this duplicates `envvars.GetEnv(key, fallback)` in `managed/utils/envvars`, which supervisord.go already uses for the `PMM_CLICKHOUSE_*` defaults.

---

managed/services/alerting/provisioning_test.go
```diff
@@ -0,0 +1,955 @@
...
+
+	f := newProvisionerFixture(t, true)
+	f.expectSettings(1, true)
+
+	// A nil status is "state unknown", and a Grafana that does not answer either has not read any
+	// file yet, so it is left alone: the file is written and whoever starts Grafana next reads it.
+	f.supervisord.On("ProgramState", mock.Anything, grafanaProgramName).Return(nil)
+	f.grafana.On("IsReady", mock.Anything).Return(errors.New("connection refused"))
+
+	f.provisioner.reconcile(context.Background(), triggerStartup)
```
ademidoff, line 149: Nit (AGENTS.md, Testing): the new tests call `context.Background()` 59 times across the added test files. Prefer `t.Context()`.

---

managed/services/alerting/provisioning_test.go
```diff
@@ -0,0 +1,955 @@
...
+	f.supervisord.AssertNotCalled(t, "RestartSupervisedService", mock.Anything, grafanaProgramName)
+}
+
+// TestProvisionerLeavesRulesItDoesNotOwnAlone is the guard against the two harms a squatted UID
+// causes: a rule under provenance "api" makes Grafana refuse to start, and a rule a user made in
+// the interface would be silently overwritten or deleted.
+func TestProvisionerLeavesRulesItDoesNotOwnAlone(t *testing.T) {
+	t.Parallel()
+
+	f := newProvisionerFixture(t, false) // standalone: the components bundle only
```
ademidoff, line 560: Nit (AGENTS.md, Code Style): the trailing comments here and at lines 566 and 586 should go on their own lines.

---

ademidoff (commented):
Reviewed at 59d9130.

The two high-severity comments share one root cause: the apply budget assumes a failed `supervisorctl restart`/`start` means the command never ran. With `startsecs = 1`, the same non-zero exit also means Grafana died on the new file. So the one failure the rollback and budget exist for goes uncharged. On the start path it also loops with Grafana down.

Not repeated here: the first-boot restart during Grafana startup and the `$` expansion in annotations are already open as threads from @JiriCtvrtka.

The branch also conflicts with `main` again, in `managed/services/ha/ha_metrics.go` and `managed/services/ha/ha_metrics_test.go`.

---

PR #5738, opened 2026-08-06
PMM-15227 Remove stale HA replicas from Inventory

4nte:
PMM-15227

https://github.com/Percona-Lab/pmm-submodules/pull/4513

## Problem

Every PMM Server replica in an HA deployment registers its own inventory Node at first boot,
and nothing ever removed it. After scaling the deployment down, the departed replica kept
appearing under **Inventory > Nodes** with status **Unknown**.

## Scope

HA deployments only, and inventory rows only. No API or schema change; metrics already
written to VictoriaMetrics are left alone.

## Solution

At startup, remove PMM Server Nodes whose name is absent from the configured `PMM_HA_PEERS`.
The peer list is the signal because the chart regenerates it from the replica count and
recreates every pod when it changes, whereas a member missing from the memberlist cluster may
simply be restarting.

Cleanup is skipped when the peer list cannot be trusted (an entry that carries no node name,
or a list that omits this replica), and a departing replica that still monitors Services is
kept with a warning instead of having those exporters cascade-deleted.


managed/models/node_helpers.go
```diff
@@ -334,3 +342,105 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
...
+				"Re-add them from a running replica and remove the node from Inventory.", node.NodeName, node.NodeID, monitored)
+			continue
+		}
+
+		err = removeNode(q, node.NodeID, RemoveCascade, true)
+		switch {
+		case err == nil:
+			logrus.Infof("Removed stale HA node %q (%s), it is not a part of the cluster anymore.", node.NodeName, node.NodeID)
+		case errors.Is(err, reform.ErrNoRows), status.Code(err) == codes.NotFound:
+			logrus.Infof("Stale HA node %q (%s) was already removed by another replica.", node.NodeName, node.NodeID)
```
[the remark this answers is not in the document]
4nte: @coderabbitai Done, `RemoveStaleHANodes` now logs through a `*logrus.Entry`: a base entry with
`component=ha` and `ha_node_id`, a per-node entry adding `node_id` and `node_name`,
and per-event `peer`, `ha_peers` and `service_ids` fields.
Control flow is unchanged.

managed/models/node_helpers_test.go
```diff
@@ -267,3 +267,158 @@ func TestNodeHelpers(t *testing.T) {
 		require.Len(t, nodes, 2) // PMM Server + HA PMM Server node
 	})
 }
+
+func TestRemoveStaleHANodes(t *testing.T) {
+	sqlDB := testdb.Open(t, models.SetupFixtures, nil)
+	t.Cleanup(func() {
+		require.NoError(t, sqlDB.Close())
+	})
```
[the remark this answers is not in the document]
4nte: @coderabbitai Keeping `testdb.Open` here.

The guideline reserves it for tests that genuinely need fixtures or migrations, and
this one does on both counts:

1. It asserts that the `SetupFixtures`-created PMM Server Node survives the cleanup,
   so the fixture is part of what is under test.
2. What is verified is which rows survive a cascade, not which SQL is emitted.
   `RemoveStaleHANodes` calls `removeNode(..., RemoveCascade, true)`, which fans out
   into `RemoveAgent` and `RemoveService`, each running its own multi-query cascade
   over tables with real foreign keys into `nodes`. The assertions are post-state
   `FindNodeByID` / `FindAgentByID` lookups; under go-sqlmock those return whatever
   the mock was scripted to return, so they would restate the script and verify no
   behaviour.

managed/models/node_helpers.go
```diff
@@ -334,3 +342,105 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
...
+		if err != nil {
+			return err
+		}
+		if len(monitored) != 0 {
+			logrus.Warnf("Keeping stale HA node %q (%s): it still monitors services %v, which would be removed with it. "+
+				"Re-add them from a running replica and remove the node from Inventory.", node.NodeName, node.NodeID, monitored)
+			continue
+		}
+
+		err = removeNode(q, node.NodeID, RemoveCascade, true)
```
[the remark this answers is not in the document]
4nte: @coderabbitai Not taking the skip, but the finding did surface a wrong comment, which
is now fixed.

`node.NodeID == PMMServerNodeID` cannot match in any supported deployment:

- A fresh HA cluster never runs `setupPMMServerAgents` (`SetupDB` branches on `HANodeID`),
  so no Node with ID `pmm-server` exists. HA replica Nodes get UUID IDs.
- `PMMServerNodeID` is still `"pmm-server"` when the cleanup runs; it is reassigned to the
  replica's own Node afterwards.
- A `pmm-server` Node can only reach an HA database by reusing a standalone deployment's
  PostgreSQL, which is not a supported or documented migration path, and only if that
  deployment was created fresh on a version carrying `is_pmm_server_node` (added
  `DEFAULT false` with no backfill).

managed/models/node_helpers.go
```diff
@@ -334,3 +342,105 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
...
+	host, _, _ := strings.Cut(strings.TrimSpace(peer), ":")
+	if net.ParseIP(host) != nil {
+		return "", false
+	}
+	// "/" is memberlist's "name/address" form, "[" an IPv6 literal; neither starts with a node name.
+	label, _, _ := strings.Cut(host, ".")
+	if label == "" || strings.ContainsAny(label, "/[") {
+		return "", false
+	}
+	return label, true
```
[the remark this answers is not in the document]
4nte: @coderabbitai
Fixed. The parser now tests the whole trimmed entry before cutting at `":"`. Rather than
`net.ParseIP` alone, it rejects any entry with more than one colon: that covers
every IPv6 form, including `2001:db8::7:9761` and zone IDs like `fe80::1%eth0`,
which `net.ParseIP` does not parse and which would otherwise have yielded the
names "2001" and "fe80". `net.ParseIP(host)` stays for bare IPv4, and the `/`/`[`
label check for memberlist's "name/address" form.

Checked the change against 13 peer forms: only those three IPv6 entries change
behaviour, every valid hostname form is untouched. Added the unbracketed IPv6
case to `DoesNothingWhenPeersCantBeTrusted`, plus the bracketed one to pin down
behaviour that was previously only implicit.

documentation/docs/install-pmm/install-HA-clustered.md
```diff
@@ -829,6 +829,7 @@ When you scale PMM HA up or down, **all PMM pods will be recreated**. This happe
   - HAProxy continues routing to available pods during rollout
   - No data loss (distributed storage)
   - Rolling update strategy minimizes downtime
+  - The Nodes of removed replicas disappear from **Inventory > Nodes** once the remaining pods restart
```
ademidoff:
```suggestion
  - The Nodes of removed replicas get removed from **Inventory > Nodes** once the remaining pods restart
```

`disappear` sounded more like an unwanted consequence rather than a necessary cleanup action.

managed/models/node_helpers.go
```diff
@@ -334,3 +342,119 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
...
+
+	nodes, err := FindNodes(q, NodeFilters{})
+	if err != nil {
+		return fmt.Errorf("failed to list Nodes for stale HA node cleanup: %w", err)
+	}
+
+	for _, node := range nodes {
+		// Set by HA replicas, and by the PMM Server Node of a non-HA deployment; every other
+		// Node is one the user monitors.
+		if !node.IsPMMServerNode {
```
ademidoff: **The legacy `pmm-server` Node is classified as stale.**

The comment right above acknowledges that a non-HA deployment's PMM Server Node also carries the flag, but nothing skips it. On a DB that was first set up non-HA, `setupPMMServerAgents` (`managed/models/database.go:1618`) created a Node with `NodeID = NodeName = "pmm-server"` and `IsPMMServerNode: true`. That name is never in `expected`, so the loop treats it as a scaled-down replica. Two outcomes:

* **Usually** its pmm-agent still owns the `pmm-server-postgresql` exporter, so `len(monitored) != 0` and every replica logs `Keeping stale HA node ... Re-add them from a running replica and remove the node from Inventory` on every single boot. That advice can't be followed: both public delete paths (`managed/services/inventory/nodes.go:340`, `managed/services/management/node.go:181`) go through `models.RemoveNode` -> `removeNode(..., allowPMMServerNode=false)` -> `PermissionDenied: PMM Server node can't be removed.`
* **If that service is gone**, `monitored` is empty and `removeNode(q, "pmm-server", RemoveCascade, true)` hits the `id == PMMServerNodeID` guard on line 269, returns `PermissionDenied`, falls into the `default:` branch on line 405 and aborts the migration transaction. `managed/cmd/pmm-managed/main.go:637` then retries for 5 minutes and calls `l.Fatalf("Could not migrate DB: timeout")` — the replica crash-loops.

Reproduced locally: with the fixture's `pmm-server-postgresql` service deleted, `RemoveStaleHANodes` returns `failed to remove stale HA node "pmm-server": rpc error: code = PermissionDenied desc = PMM Server node can't be removed.`

Suggested fix: `if node.NodeID == PMMServerNodeID { continue }` at the top of the loop, and tolerate `PermissionDenied` in the switch below.

managed/models/database.go
```diff
@@ -1528,6 +1528,12 @@ func setupPMMServerHAAgents(q *reform.Querier, params SetupDBParams) error {
 	// create PMM Server Node and associated Agents in HA mode
 	logrus.Infof("Setting up PMM Server agents in HA mode, Node ID: %s", params.HANodeID)
 
+	// Before the "agent already exists" early return, so restarted replicas still clean up.
+	err := RemoveStaleHANodes(q, params.HANodeID, params.HAPeers)
```
ademidoff: **Altitude: a cosmetic cleanup can stop the server from booting.**

`RemoveStaleHANodes` runs inside the schema-migration transaction and returns hard errors, so anything it trips over aborts `db.InTransaction` in `migrateDB`, and `managed/cmd/pmm-managed/main.go:640` eventually calls `l.Fatalf("Could not migrate DB: timeout")`.

Candidates that reach that path today: `PermissionDenied` on the pre-HA `pmm-server` Node (see the comment on `RemoveStaleHANodes`), a `NotFound` out of the `haNodeMonitoredServices` pre-check while another replica commits, a row-lock/serialization failure when three replicas restart at once.

Tidying stale rows out of Inventory should never keep a PMM Server replica from starting. Suggest logging the error and continuing instead of returning it — nothing up the stack can recover from it.

managed/models/node_helpers.go
```diff
@@ -334,3 +342,119 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
...
+	if label == "" || strings.ContainsAny(label, "/[") {
+		return "", false
+	}
+	return label, true
+}
+
+// haNodeMonitoredServices returns the IDs of Services whose exporters run under a replica's pmm-agent.
+// Remote instances bind theirs to the replica that added them (see management.RDSService), so removing
+// that replica's Node takes them with it.
+func haNodeMonitoredServices(q *reform.Querier, nodeID string) ([]string, error) {
```
ademidoff, line 481: **The safety check misses Services attached to the Node itself.**

This only walks agents whose `pmm_agent_id` is one of the replica's pmm-agents, so a Service bound to the stale Node via `services.node_id` is invisible here. That shape exists in the codebase — `managed/services/checks/checks.go:1377` special-cases `service.NodeID == models.PMMServerNodeID`, and under HA `PMMServerNodeID` is the local replica's Node.

If such a Service's exporter runs under a *different* replica's pmm-agent, `monitored` comes back empty, the guard passes, and `removeNode(..., RemoveCascade, true)` reaches `managed/models/node_helpers.go:318` and calls `RemoveService(..., RemoveCascade)` on it. The Service and its agents are deleted silently — no warning, which is exactly what the guard exists to prevent.

Worth also counting `SELECT ... FROM services WHERE node_id = $1` (and agents bound via `node_id`) before deciding a Node is safe to reap.

managed/models/node_helpers.go
```diff
@@ -334,3 +342,119 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
...
+		}
+		if _, ok := expected[node.NodeName]; ok {
+			continue
+		}
+
+		nodeL := l.WithFields(logrus.Fields{"node_id": node.NodeID, "node_name": node.NodeName})
+
+		monitored, err := haNodeMonitoredServices(q, node.NodeID)
+		if err != nil {
+			return err
```
ademidoff: **The pre-check doesn't get the race tolerance `removeNode` was given.**

Four lines down, the `switch` deliberately tolerates `reform.ErrNoRows` / `codes.NotFound` because another replica may be reaping the same Node concurrently. This `return err` bypasses that.

`haNodeMonitoredServices` -> `FindAgents` performs a `FindAgentByID(filters.PMMAgentID)` existence probe (`managed/models/agent_helpers.go:241`). If replica A commits the deletion of `ha-node-2`'s pmm-agent while replica B sits between its `FindNodes` snapshot and this call, B gets `codes.NotFound` here and aborts the whole migration.

Same `errors.Is(err, reform.ErrNoRows) || status.Code(err) == codes.NotFound` -> `continue` treatment would close it.

managed/models/node_helpers.go
```diff
@@ -252,13 +254,19 @@ func CreateNode(q *reform.Querier, nodeType NodeType, params *CreateNodeParams)
...
+// removeNode removes a single Node. The allowPMMServerNode flag lifts the ban on Nodes flagged as PMM
+// Server Nodes; only the HA cleanup sets it, to reap replicas that are no longer part of the cluster.
+func removeNode(q *reform.Querier, id string, mode RemoveMode, allowPMMServerNode bool) error { //nolint:gocognit
 	n, err := FindNodeByID(q, id)
 	if err != nil {
 		return err
 	}
 
-	if n.IsPMMServerNode || id == PMMServerNodeID {
+	if id == PMMServerNodeID || (!allowPMMServerNode && n.IsPMMServerNode) {
```
ademidoff: **With `allowPMMServerNode = true`, the only remaining protection is a mutable package variable.**

`PMMServerNodeID` starts as `"pmm-server"` (`managed/models/node_helpers.go` / `node_model.go:44`) but `setupPMMServerHAAgents` reassigns it to this replica's UUID at `managed/models/database.go:1561` and `:1610` — *after* the cleanup call at `:1532`.

`migrateDB` retries `SetupDB` in a loop (`managed/cmd/pmm-managed/main.go:637`), so on a second attempt in the same process (attempt 1 reached `:1610`, then the COMMIT failed and rolled back) `RemoveStaleHANodes` runs with `PMMServerNodeID` pointing at a Node that no longer exists. The pre-HA `pmm-server` Node is then unguarded and, with the flag lifted, gets deleted outright.

`RemoveAgent`'s `id == PMMServerAgentID` guard (`managed/models/agent_helpers.go:1450`) has the same order dependence. A guard that this very code path mutates is a fragile place to hang "can't delete the server's own node" on — better keyed off the caller's known NodeID.

managed/models/node_helpers.go
```diff
@@ -334,3 +342,119 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
...
+	if len(haPeers) == 0 {
+		return nil
+	}
+
+	l := logrus.WithFields(logrus.Fields{"component": "ha", "ha_node_id": haNodeID})
+
+	expected := make(map[string]struct{}, len(haPeers))
+	for _, peer := range haPeers {
+		name, ok := haPeerNodeName(peer)
+		if !ok {
```
ademidoff, line 379: **One empty entry in `PMM_HA_PEERS` silently disables the feature.**

`managed/cmd/pmm-managed/main.go:805` splits the env var on `,`, so `"pmm-ha-0.pmm-ha:9761,pmm-ha-1.pmm-ha:9761,"` — a trailing comma, easy to get out of a Helm `join` over a list with a blank element — yields a third, empty element. `haPeerNodeName("")` returns `ok=false` on the `label == ""` check, and the whole cleanup returns nil with only a WARN buried in the startup log.

An empty entry carries no information about the cluster, so it isn't the "partial list" hazard this branch guards against. Suggest skipping blank entries (`if peer == "" { continue }`) rather than treating them as untrusted.

managed/models/node_helpers.go
```diff
@@ -334,3 +342,119 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
 	}
 	return nil
 }
+
+// RemoveStaleHANodes removes the PMM Server Nodes of HA replicas that are no longer configured peers,
+// e.g. after a scale-down. Peers are the source of truth because they are regenerated from the replica
+// count and restart every replica, while a missing memberlist member may just be restarting.
+func RemoveStaleHANodes(q *reform.Querier, haNodeID string, haPeers []string) error {
+	if len(haPeers) == 0 {
```
ademidoff: **Does the headline scenario — scaling down to one replica — actually reach the cleanup?**

If the chart emits an empty `PMM_HA_PEERS` for a single-replica cluster (nothing to gossip with), this early return fires. If it emits a list that doesn't include the surviving pod, the `expected[haNodeID]` guard below fires. Either way the two departed replicas' Nodes stay in Inventory forever, which is the case the PR is meant to fix.

Nothing here pins down what the chart produces for `replicas: 1`, and there's no test for it. Worth confirming against the chart before merge — and if the list can legitimately be empty, that state needs its own handling.

managed/models/node_helpers_test.go
```diff
@@ -267,3 +267,160 @@ func TestNodeHelpers(t *testing.T) {
...
+
+		// the removal cascades to the agents of the stale node
+		for _, agentID := range []string{"ha-agent-2", "ha-node-exporter-2"} {
+			_, err := models.FindAgentByID(q, agentID)
+			tests.AssertGRPCErrorCode(t, codes.NotFound, err)
+		}
+
+		// neither monitored nodes nor the pre-HA pmm-server Node are touched
+		assertNodeExists(t, q, "monitored-node")
+		assertNodeExists(t, q, models.PMMServerNodeID)
```
ademidoff: **This assertion passes by accident, and the interesting path is untested.**

The comment above says the pre-HA `pmm-server` Node isn't touched, but it survives only because `SetupFixtures` attaches `pmm-server-postgresql` to it, so the run takes the `len(monitored) != 0` branch. The test log shows it:

```
Keeping stale HA node "pmm-server" (pmm-server): it still monitors services [5bf3df6d-... 5bf3df6d-...]
```

Nothing here exercises protection for PMM Server Nodes. Drop that one fixture service and `RemoveStaleHANodes` returns `PermissionDenied`, failing this subtest at the `require.NoError` on line 343.

Worth an explicit case for a PMM-Server-flagged Node with no monitored services. Also `assertNodeExists` uses `assert.NoError` (line 335) — `require` would stop the subtest at the real failure instead of cascading.

managed/models/node_helpers.go
```diff
@@ -334,3 +342,119 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
...
+// that replica's Node takes them with it.
+func haNodeMonitoredServices(q *reform.Querier, nodeID string) ([]string, error) {
+	pmmAgents, err := FindPMMAgentsRunningOnNode(q, nodeID)
+	if err != nil {
+		return nil, err
+	}
+
+	var serviceIDs []string
+	for _, pmmAgent := range pmmAgents {
+		agents, err := FindAgents(q, AgentFilters{PMMAgentID: pmmAgent.AgentID})
```
ademidoff: **N+1 queries, credential decryption, and duplicate IDs, all to read `ServiceID`.**

* `FindAgents` runs a `FindAgentByID` existence probe (`managed/models/agent_helpers.go:241`) before its real query, once per pmm-agent.
* `FindPMMAgentsRunningOnNode` calls `DecryptAgent` on every row (`managed/models/agent_helpers.go:423`), pulling encrypted passwords through the encryption layer inside the migration transaction just to test `agent.ServiceID != nil`.
* The result carries duplicates straight into the operator-facing warning — the test run logged `it still monitors services [5bf3df6d-361d-42dd-9dd2-3d0b3573ab24 5bf3df6d-361d-42dd-9dd2-3d0b3573ab24]` because two exporters point at the same service.

One query replaces the helper:

```sql
SELECT DISTINCT service_id FROM agents
 WHERE pmm_agent_id IN (SELECT agent_id FROM agents WHERE runs_on_node_id = $1 AND agent_type = 'pmm-agent')
   AND service_id IS NOT NULL
```

managed/models/node_helpers.go
```diff
@@ -334,3 +342,119 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
...
+		}
+		expected[name] = struct{}{}
+	}
+
+	if _, ok := expected[haNodeID]; !ok {
+		l.WithField("ha_peers", haPeers).Warn("PMM_HA_PEERS doesn't list this node, skipping the removal of stale HA nodes.")
+		return nil
+	}
+
+	nodes, err := FindNodes(q, NodeFilters{})
```
ademidoff: **Loads the whole `nodes` table to keep a handful of rows.**

Every HA replica boot materializes every monitored Node inside the migration transaction, only for the loop below to discard everything without `IsPMMServerNode`. On an install with thousands of Nodes that's a full scan plus full row hydration per replica start.

`NodeFilters` (`managed/models/node_helpers.go:92`) has no `IsPMMServerNode` field; adding one — or issuing `q.SelectAllFrom(NodeTable, "WHERE is_pmm_server_node")` directly here — makes the cost proportional to the replica count instead of the inventory size.

documentation/docs/install-pmm/install-HA-clustered.md
ademidoff: This is stated unconditionally, but the cleanup is skipped in several configurations: `PMM_HA_PEERS` built from bare IPs (`10.244.1.7:9761,...`), an empty peer list, a list that doesn't include the local pod, or any entry the parser can't read a name from (a trailing comma is enough). In those cases the Nodes stay in Inventory and an operator will read this bullet as a broken promise.

Suggest qualifying it — the Nodes are removed when `PMM_HA_PEERS` carries resolvable node names — and mentioning that a WARN is logged when the cleanup is skipped, so there's something to grep for.

ademidoff (commented):
Review of the HA stale-node cleanup, against `b2cbb72`. Findings are inline; the first four are the ones I'd act on before merge, the rest are hardening, efficiency and test/doc notes.

The two headline items: the pre-HA `pmm-server` Node is classified as stale (either a permanent unactionable warning or, in the tail case, a `PermissionDenied` that aborts the migration and crash-loops the replica), and the cleanup runs at migration altitude where any failure keeps the server from booting.

Verified locally: `make test-models RUN=TestRemoveStaleHANodes` passes, `golangci-lint` is clean on the changed files, and the `PermissionDenied` path was reproduced with a throwaway test (since deleted). Everything else is reasoned from the code and labelled as such.

Nice fix on the unbracketed IPv6 peer in `haPeerNodeName` — that one is closed.

managed/models/node_helpers.go
4nte: Yes. I've Confirmed this against the chart and on a live cluster. `pmm.haPeers` builds the list with
`range until $replicas`, so `replicas: 1` renders one self-inclusive entry
(`PMM_HA_PEERS=pmm-ha-0.monitoring-service.pmm.svc.cluster.local`) - neither early return fires.
On the empty list: it is not reachable from the chart.

managed/models/node_helpers.go
4nte: Fixed the Node side: `defaultPMMServerNodeID` is now a const, and both the `removeNode` ban and the
sweep's skip compare against it, so nothing in that path reads the mutable `PMMServerNodeID` any more.
That turned out to be necessary rather than defensive, because the cleanup now runs after the replica
registers itself, so the var always holds this replica's UUID by then. Covered by
`KeepsPreHAPMMServerNodeAfterSetupRetry` and `RefusesThePreHAPMMServerNode`.

On `RemoveAgent`: I left it as is. The deletion you describe isn't reachable - `migrateDB` runs at
`main.go:896`, before any server starts, so no API call can land in that window, and the only
in-process caller then is the sweep itself, which never touches the local replica's Node.

managed/models/node_helpers.go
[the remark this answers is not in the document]
4nte: @coderabbitai Following up: we did take the skip after all. @ademidoff raised the same finding on
RemoveStaleHANodes, and it turned out to be reachable with a worse consequence than I assumed - a
DB first initialized by a non-HA deployment (on a version carrying is_pmm_server_node) and later
reused by an HA one keeps a Node with the literal pmm-server ID and the flag set. The sweep
selected it, removeNode refused it outright, and the error aborted the migration transaction, so
the replica crash-looped on Could not migrate DB: timeout. Reproduced locally with the fixture
pmm-server-postgresql Service removed.

The skip is in 433b52413, and ac9e496cc keys both it and removeNode's ban off a
defaultPMMServerNodeID const rather than the mutable PMMServerNodeID - which matters now that the
sweep runs after the replica registers itself, so that var already holds the replica's own UUID.
Covered by KeepsPreHAPMMServerNode, KeepsPreHAPMMServerNodeAfterSetupRetry and
RefusesThePreHAPMMServerNode.

The sweep also no longer runs inside the migration transaction, so even an unexpected failure there
is logged and skipped rather than blocking startup.

managed/models/node_helpers.go
4nte: the first three are fixed, and I benchmarked the fourth before deciding against it.

**N+1 probe.** Gone. The helper no longer goes through `FindAgents`, so there is no
`FindAgentByID` existence check per pmm-agent. It is now `FindAgentsOnNode` +
`FindAgentsByPMMAgentIDs` + `FindServices`: three fixed queries instead of `1 + 2N`, and the
count no longer grows with the number of agents on the Node.

**Duplicate IDs.** Fixed. The result goes through `deduplicateStrings`, so the warning names
each Service once, sorted.

**Inside the migration transaction.** No longer true. The sweep moved out of `migrateDB` and
runs from `SetupDB` after the migration commits, with failures logged instead of returned, so
it cannot block startup at all.

**Decryption.** You are right that this path does not need it: it reads only `ServiceID`,
`AgentType` and `AgentID`. I kept it because the two new helpers are exported and general
purpose - `FindAgentsOnNode` in particular answers a question `management/agent.go:129` and
`management/node.go:478` currently hand-roll - and all eight existing agent finders decrypt.
Two that silently did not would be a quiet trap for the next caller who reads `.Password`. It
is free here in any case: `agentEncryption` is nil-guarded per field, and a replica's
`pmm-agent`, `node_exporter`, `vmagent` and `nomad-agent` carry no credentials, so it is a
struct copy. Happy to drop it and note the exception in the doc comments if you prefer.

**The single query.** I benchmarked both, verified to return the same result, with
`-benchtime 200x -count 5` against a local postgres holding 1000 Nodes, ~2000 agents and
1000 Services:

three helpers: ~1.0 ms/op   (median 0.93, range 0.79-1.29)
single query:  ~0.70 ms/op  (median 0.63, range 0.60-0.91)

Roughly 0.3 ms apart, and the ranges overlap. The win comes from two fewer round trips and
from scanning one string instead of hydrating six wide agent rows, not from the database doing
less work - the UNION actually scans `agents` more times.

Those numbers are loopback, so they understate the gap: a `SELECT 1` from a PMM pod to the HA
Postgres service measures 0.03-0.2 ms warm here, and a real cluster with pgbouncer is more like
0.5-2 ms per round trip, which would put the three-query version a few milliseconds behind
rather than 0.3. It is still paid once per scale-down for the whole cluster, not once per
replica: the pods restart one at a time (`OrderedReady`), the first survivor back removes the
Node, and every later replica finds nothing to do and never reaches these queries at all. For
a few milliseconds once per boot I would be trading raw SQL with manual `rows.Scan` back into
the models layer, plus the two helpers.

Your query would also need widening now: per your other comment on the safety check, it covers
only Services whose exporters hang off the Node's pmm-agent, and the helper also has to count
Services attached to the Node and external exporters running on it.

The factor that would actually change this is inventory growth, since all of it is sequential
scans today - and the fix there is indexes on `agents(runs_on_node_id)`, `agents(node_id)` and
`agents(pmm_agent_id)`, which help both versions equally.

Folded into the single-query section: the loopback caveat with the measured pod-to-Postgres latency, the projection to a real cluster (a few milliseconds rather than 0.3), and the once-per-scale-down framing moved to sit right after it so the two numbers are read together. The closing paragraph now points at indexes as the thing that would genuinely change the calculus.

managed/models/agent_helpers.go
```diff
@@ -427,6 +427,49 @@ func FindPMMAgentsRunningOnNode(q *reform.Querier, nodeID string) ([]*Agent, err
...
+	}
+
+	res := make([]*Agent, len(structs))
+	for i, str := range structs {
+		decryptedAgent := DecryptAgent(*str.(*Agent)) //nolint:forcetypeassert
+		res[i] = &decryptedAgent
+	}
+
+	return res, nil
+}
```
maxkondr: the same is achieved by
```
agents, err := FindAgents(ctx, AgentFilters{NodeID: <nodeID>})
```

managed/models/agent_helpers.go
```diff
@@ -427,6 +427,49 @@ func FindPMMAgentsRunningOnNode(q *reform.Querier, nodeID string) ([]*Agent, err
...
+
```
maxkondr: very similar to the prev on:

```
agents, err := FindAgents(ctx, AgentFilters{PMMAgentID: <pmmAgentID>})
```

managed/models/database.go
```diff
@@ -1517,6 +1519,44 @@ func migrateDB(db *reform.DB, params SetupDBParams) error {
...
+// cosmetic, so this runs outside the migration transaction and only logs failures: tidying them up
+// must never keep a replica from starting.
+func removeStaleHANodes(ctx context.Context, db *reform.DB, params SetupDBParams) {
+	if params.HANodeID == "" || params.SetupFixtures == SkipFixtures {
+		return
+	}
+
+	l := logrus.WithFields(logrus.Fields{"component": "ha", "ha_node_id": params.HANodeID})
+
+	nodes, err := StaleHANodes(db.WithContext(ctx), params.HANodeID, params.HAPeers)
```
maxkondr: what does `StaleHANodes` mean? Add, Get, Remove, List, Update,....?

managed/models/node_helpers.go
```diff
@@ -334,3 +355,160 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
 	}
 	return nil
 }
+
+// StaleHANodes returns the PMM Server Nodes of HA replicas that are no longer configured peers,
+// e.g. after a scale-down, and that can be removed without taking a user's monitoring with them.
+// Peers are the source of truth because they are regenerated from the replica count and restart
+// every replica, while a missing memberlist member may just be restarting.
+func StaleHANodes(q *reform.Querier, haNodeID string, haPeers []string) ([]*Node, error) {
```
maxkondr: what does `haNodeID` mean? HA Leader's ID? In HA (Raft) terminology there are: Leaders and Followers. HaNode - is a general term for any node in cluster.

managed/models/node_model.go
```diff
@@ -38,10 +38,13 @@ const (
 	RemoteAzureDatabaseNodeType NodeType = "remote_azure_database"
 )
 
+// defaultPMMServerNodeID is the PMM Server Node ID that never changes at runtime.
+const defaultPMMServerNodeID = "pmm-server"
+
 // PMMServerNodeID is a special Node ID representing PMM Server Node.
 // It takes the value of "pmm-server" in regular non-HA setups and in Active/Passive HA setups,
 // while in Active/Active HA setups it is set to a dynamically generated UUID.
```
maxkondr: does PMM HA support Active-Active mode? I saw Active-Passive only

managed/models/agent_helpers.go
```diff
@@ -427,6 +427,49 @@ func FindPMMAgentsRunningOnNode(q *reform.Querier, nodeID string) ([]*Agent, err
...
+		args[i] = id
+	}
+	structs, err := q.SelectAllFrom(AgentTable, tail, args...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to select Agents started by pmm-agents: %w", err)
+	}
+
+	res := make([]*Agent, len(structs))
+	for i, str := range structs {
+		decryptedAgent := DecryptAgent(*str.(*Agent)) //nolint:forcetypeassert
```
maxkondr: DecryptAgent/EncryptAgent uses heavy algos - I would recommend checking for Querier's context for cancellation first before running these operations - maybe there is no results receiver anymore

managed/models/database.go
```diff
@@ -1297,6 +1297,8 @@ func SetupDB(ctx context.Context, sqlDB *sql.DB, params SetupDBParams) (*reform.
 		return nil, err
 	}
 
+	removeStaleHANodes(ctx, db, params)
```
maxkondr: I doubt about the place where it should be running. I think that HA Leader is a good candidate for this.

managed/models/node_helpers.go
```diff
@@ -334,3 +355,160 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error { //nolint:
...
+		if !ok {
+			// Trusting the rest would treat a partial list as the whole cluster and remove live replicas.
+			l.WithField("peer", peer).Warn("Can't read a node name from a PMM_HA_PEERS entry, skipping the removal of stale HA nodes.")
+			return nil, nil
+		}
+		expected[name] = struct{}{}
+	}
+
+	if _, ok := expected[haNodeID]; !ok {
+		l.WithField("ha_peers", haPeers).Warn("PMM_HA_PEERS doesn't list this node, skipping the removal of stale HA nodes.")
```
maxkondr: log message states `skipping remove` but in fact the whole func is terminated

managed/models/node_helpers.go
ademidoff: `defaultPMMServerNodeID` is the right shape, and `KeepsPreHAPMMServerNodeAfterSetupRetry` covers exactly the ordering I was worried about.

Agreed on `RemoveAgent`. `models.SetupDB` runs from `migrateDB` at `managed/cmd/pmm-managed/main.go:896`, before anything serves traffic, and I confirmed the sweep can't reach the local replica's Node from the other direction too: `StaleHANodes` returns early unless `haNodeID` is in `expected`, and the Node whose name equals it is then skipped. Nothing to change.

managed/models/node_helpers.go
ademidoff: Thanks for checking it against the chart and on a live cluster — `range until $replicas` rendering a self-inclusive entry at `replicas: 1` is exactly what I couldn't confirm from this repo. `ReportsEveryDepartedReplicaWhenScaledToOne` pins it down.

managed/models/node_helpers.go
ademidoff: All three fixes look right: `FindAgentsOnNode` + `FindAgentsByPMMAgentIDs` + `FindServices` is three fixed queries with no existence probe, and `deduplicateStrings` gives the warning one sorted entry per Service.

**Decryption** — keep it. Consistency with the other eight finders is worth more than the struct copy, and two exported finders that silently skipped it would be exactly the trap you describe. No doc-comment exception needed.

**The single query** — your numbers settle it. The widening my other comment asked for would have made that SQL worse anyway, and indexes on `agents(runs_on_node_id)`, `agents(node_id)` and `agents(pmm_agent_id)` are the thing that would actually change the picture. Thanks for measuring instead of taking my word for it.

ademidoff:
Re-reviewed at `34de666`. All 11 findings from [my earlier review](https://github.com/percona/pmm/pull/5738#pullrequestreview-4882152426) are addressed — checked against the code, not just the commit messages.

| Finding | Resolution |
|---|---|
| Pre-HA `pmm-server` Node classified as stale | `managed/models/node_helpers.go:403` skips `defaultPMMServerNodeID` |
| Cleanup can block startup | Sweep moved out of the migration transaction to `managed/models/database.go:1300`; failures logged, never returned |
| `services.node_id` blind spot | `haNodeMonitoredServices` now covers agents on the Node, agents under its pmm-agents, and `FindServices(ServiceFilters{NodeID:...})` |
| Pre-check error not tolerated | `managed/models/node_helpers.go:414-418` logs and keeps the Node |
| Mutable `PMMServerNodeID` | `defaultPMMServerNodeID` const, used by both the ban and the skip |
| Blank `PMM_HA_PEERS` entry disables the sweep | `managed/models/node_helpers.go:374` skips blank entries |
| Scale-to-one may skip cleanup | Confirmed against the chart, plus `ReportsEveryDepartedReplicaWhenScaledToOne` |
| Test passed via an unrelated fixture | `KeepsPreHAPMMServerNode` removes the fixture Service first; `assertNodeExists` uses `require` |
| N+1 / duplicate Service IDs | Three fixed queries + `deduplicateStrings` |
| Full `nodes` scan | `NodeFilters{IsPMMServerNode: new(true)}` |
| Doc bullet overstated the guarantee | Qualified, with a "When PMM keeps a stale Node" note and a log-grep command |

Two things you added that I hadn't asked for and that close real gaps: `RemoveStaleHANode` re-reads the monitored Services *inside* its own transaction, and a transaction per Node means a partial failure can't leave one half-removed.

Verified locally at this SHA:

* `TestStaleHANodes` (10 subtests), `TestRemoveStaleHANode` (6), `TestHAPeerNodeName` (16) and `TestAgentHelpers` (17) all pass.
* `golangci-lint run managed/models/...` flags nothing in the new code — the issues remaining in that package are pre-existing, and none of the flagged lines fall inside this diff.

One note on @maxkondr's suggestions for `FindAgentsOnNode` and `FindAgentsByPMMAgentIDs`, since they interact with one of my findings: `FindAgents(AgentFilters{NodeID: ...})` filters on `node_id` alone, while `FindAgentsOnNode` is `runs_on_node_id = $1 OR node_id = $1` — that widening is what keeps the sweep from cascade-deleting a Service attached to the stale Node. And `AgentFilters{PMMAgentID: ...}` takes a single ID where the new helper takes a set (and adds an existence probe per call). They aren't equivalent substitutions, so I'd keep both helpers.

Unrelated to this PR, but found while running the suite: `TestNodeHelpers/CreateNode/DuplicateMachineID` fails on roughly a quarter of runs in isolation (2 of 8 here). It asserts that `ORDER BY node_id` puts the newly created Node before the literal `GenericNode` fixture, but the new ID is a random UUID, so the order flips depending on its first character. Pre-existing — this branch only touches that file from line 221 down — but worth a ticket.

From my side the findings are closed. Happy to turn this into a formal approval if that's useful.

maxkondr (approved):

managed/models/agent_helpers.go
```diff
@@ -218,8 +218,13 @@ func checkUniqueAgentID(q *reform.Querier, id string) error {
 type AgentFilters struct {
 	// Return only Agents started by this pmm-agent.
 	PMMAgentID string
+	// Return only Agents started by any of these pmm-agents. An empty slice is not a filter.
```
ademidoff, line 221: Two asymmetries in the new filters are worth spelling out in the `FindAgents` doc comment as well, not only on the fields — a new caller reads the function, not the struct:

* **An empty `PMMAgentIDs` matches everything, not nothing.** You've documented it here and guarded the one caller at `managed/models/node_helpers.go:499`, with `ReportsScaledDownReplicaWithNoAgents` to keep it honest, so nothing is broken today. But the guard is load-bearing and lives at the call site, and the next caller that builds the slice with `append` in a loop gets `nil` on the empty path — exactly the shape that silently returns the whole inventory. I don't think it can be fixed inside the filter-struct pattern (nil, empty and unset are indistinguishable), which is precisely why it belongs in the function's own doc comment.
* **`PMMAgentID` probes, `PMMAgentIDs` doesn't.** The singular runs `FindAgentByID` first and returns `NotFound` for an unknown ID; the plural returns an empty result. Someone switching a call from one to the other loses that error without noticing.

Both are one line each — no code change intended.

managed/models/node_helpers_test.go
```diff
@@ -267,3 +285,488 @@ func TestNodeHelpers(t *testing.T) {
...
+
+	t.Run("ReportsNotFoundWhenAlreadyRemoved", func(t *testing.T) {
+		db := setup(t)
+		q := db.Querier
+
+		require.NoError(t, db.InTransaction(func(tx *reform.TX) error {
+			return models.RemoveStaleHANode(tx.Querier, "ha-node-2")
+		}))
+
+		// what a replica sees when another one won the race; RemoveStaleHANodes reads this as
```
ademidoff: Stale name: `RemoveStaleHANodes` was renamed back to unexported `removeStaleHANodes` in f76164cee when the leader service was reverted. Left over from that revert.

ademidoff (commented):
Re-reviewed the delta since `34de666` (6 commits). Nothing blocking — two nits inline, both documentation-only.

Checked in particular that folding the finders into `AgentFilters` kept the semantics: `OnNodeID` still expands to `(runs_on_node_id = $n OR node_id = $n+1)`, so the `services.node_id` fix is intact, and I traced the placeholder arithmetic through all eight filters — `idx` advances correctly past the `IN` list and the two-placeholder `OnNodeID`. Also confirmed the new `q.Context().Err()` check can't panic: reform initialises `Querier.ctx` to `context.Background()` and `clone()` propagates it, so `Context()` is never nil.

Special mention for the leader revert in f76164cee. Trying @maxkondr's suggestion, reproducing the failure on a live 2 -> 3 scale-up, and writing the reason down at `managed/models/database.go:1526-1533` is a better outcome than either taking or refusing the suggestion on argument alone.

Verified locally at this SHA: `go build ./managed/...` clean; `TestFindStaleHANodes` (12), `TestRemoveStaleHANode` (6), `TestHAPeerNodeName` (15), `TestAgentHelpers` (17) and `TestNodeHelpers` (5) all pass; `golangci-lint` flags nothing in this diff. All 11 findings from my first review are still fixed after the refactor — I re-checked each rather than assuming the rename carried them over.

ademidoff (approved):
Great job!

managed/models/agent_helpers.go
4nte: Added doc comments to clarify this ✅

managed/models/node_helpers_test.go
4nte: stale name has been updated

JiriCtvrtka (approved):

PR #5748, opened 2026-08-08
PMM-15310 Detect dead agent connections on server

ademidoff:
Ticket number: PMM-15310

Feature build: SUBMODULES-4520

## Problem

pmm-managed waited for a pong with no deadline anywhere, so a silently dropped connection stayed registered until the kernel's TCP retransmission timeout expired, ~15 minutes later. Until then:

1. `register()` blocked probing the stale channel, so every reconnect attempt was rejected — the agent gave up after its 5s dial timeout and looped.
2. The periodic health-check ping blocked forever and only logged the eventual error, so the server never tore the dead channel down on its own.
3. A stale connection that finally died after the agent had already re-registered could unregister the live connection, leaving the agent connected while pmm-managed reported it disconnected forever.

The client-side keepalive added in PMM-15200 (#5650) exposed this by making the agent reconnect promptly instead of hanging together with the server. Note that a server-side `keepalive.ServerParameters` would not have helped: agents reach pmm-managed through nginx `grpc_pass`, which terminates HTTP/2 and would answer transport PINGs locally over loopback while the agent↔nginx segment is dead.

### Channel

- `SendAndWaitResponseWithContext()` waits for a response until the context is done and returns `ctx.Err()`.
- Both variants share `sendAndWaitResponse()`, which selects on a `done` channel — nil means wait indefinitely, so `SendAndWaitResponse()` keeps its exact previous behavior and no existing call site changes.
- `unsubscribe()` drops an abandoned subscription so the response map does not grow and a late response is discarded rather than reported as coming from an unknown subscriber.

### Registry

- `staleConnectionProbeTimeout` (2s) bounds the probe that tells a live connection apart from a stale one during re-registration. It stays well below pmm-agent's 5s dial timeout so the reconnecting agent can take over.
- `unregister()` takes the connection it belongs to and no-ops if that connection has already been superseded.
- `ping()` honors the caller's context.

### Handler

- `defaultAgentPingTimeout` (5s) bounds each periodic ping.
- After `maxAgentPingFailures` (2) consecutive unanswered pings the connection is considered dead and torn down, so the agent can reconnect. Worst-case server-side detection is ~25s instead of ~15 minutes.
- The agent connection is passed to `unregister()` so a late teardown cannot evict its successor.

## Testing

- `TestUnregister()` covers the connection-scoped unregistration, including the superseded-connection case.
- `TestServerRequestTimeout()` verifies the context-aware wait returns `DeadlineExceeded` and leaves no leaked subscription.
- `go build ./managed/...`, `go test -race` for `managed/services/agents` and `managed/services/agents/channel`, `golangci-lint` and `license-eye` are clean for the changed code.
- Not yet verified on a live server against the blackholed-connection reproducer from the ticket.

- [ ] API Docs updated

- #5650 (PMM-15200) added the client-side keepalive this change complements.


theTibi:
@coderabbitai full review

managed/services/agents/channel/channel.go
```diff
@@ -167,20 +170,50 @@ func (c *Channel) Send(resp *ServerResponse) {
...
+	case resp, ok := <-ch:
+		if !ok {
+			return nil, errors.New("channel is closed")
+		}
+		return resp.Payload, resp.Error
+
+	case <-done:
+		// Drop the subscription so that the map does not grow and a late response is discarded
+		// instead of being reported as coming from an unknown subscriber.
+		c.unsubscribe(id)
```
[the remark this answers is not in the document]
ademidoff: Valid — fixed in e7b68d5.

Confirmed against the code: `unsubscribe` deletes the entry, so a late `Pong` reaches `removeResponseChannel`, finds `ch == nil` and hits `c.l.Errorf("No subscriber for ID %d", id)`. With a bounded ping wait that is now a routine event for a slow-but-alive agent, not an anomaly, so error level is wrong. The code comment I wrote on the cancellation path also claimed the opposite of what happens, which was its own small bug.

`Channel` now tracks abandoned IDs in a separate `abandoned` set: `unsubscribe` records the ID, and `removeResponseChannel` recognizes it, drops it from the set and logs at debug instead. `close()` clears the set along with `responses`, and the ID is removed as soon as the late response arrives, so it does not accumulate.

Regression test added: `TestLateResponseAfterTimeout` in `channel_test.go` abandons request 1, has the agent answer it late, then completes a second request/response — since `runReceiver` processes messages in order, that proves the late response was handled, and the test asserts the abandoned set is empty afterwards.

---

managed/services/agents/handler.go
```diff
@@ -85,13 +94,24 @@ func (h *Handler) Run(stream agentv1.AgentService_ConnectServer) error { //nolin
...
+			err := h.r.ping(pingCtx, agent)
+			cancelPing()
 			if err != nil {
-				l.Errorf("agent %s ping: %v", agent.id, err)
+				pingFailures++
+				l.Errorf("agent %s ping (attempt %d of %d): %v", agent.id, pingFailures, maxAgentPingFailures, err)
+				if pingFailures >= maxAgentPingFailures {
+					disconnectReason = "ping_timeout"
+					h.r.unregister(ctx, agent.id, disconnectReason, agent)
+					return status.Error(codes.DeadlineExceeded, "Agent is not responding to pings.")
```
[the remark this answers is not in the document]
ademidoff: Adopted in e7b68d5 — it does match the convention in `AGENTS.md` ("Use `logrus` with structured fields", `s.l.WithField("key", value).Error("message")`).

```go
l.WithError(err).WithFields(logrus.Fields{
    "agent_id":          agent.id,
    "ping_failures":     pingFailures,
    "max_ping_failures": maxAgentPingFailures,
}).Error("Failed to ping pmm-agent.")
```

Failure counting and the timeout handling are unchanged.

On the severity: 🟠 Major overstates this. It is log formatting with no behavioral effect, and the `Errorf` line I replaced was the pre-existing style of the line at that spot. Worth doing because it is nearly free and now these logs are queryable by `agent_id`, not because anything was broken.

---

managed/services/agents/registry.go
```diff
@@ -268,7 +273,13 @@ func (r *Registry) register(stream agentv1.AgentService_ConnectServer) (*pmmAgen
...
 		// so we return AlreadyExists error. Otherwise we kick the previous connection
 		// and proceed with the new one.
-		err := r.ping(ctx, currentAgent)
+		//
+		// The ping is bounded: a silently dropped connection never produces a pong, and waiting
+		// for one would block this registration until the kernel's TCP retransmission timeout
+		// expires (~15 minutes), rejecting every reconnect attempt in the meantime (PMM-15310).
+		pingCtx, cancelPing := context.WithTimeout(ctx, staleConnectionProbeTimeout)
+		err := r.ping(pingCtx, currentAgent)
+		cancelPing()
```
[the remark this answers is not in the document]
ademidoff: Real hole, fixed in e7b68d5 — though the consequence is worse than described, so worth spelling out.

`register()` holds no lock across probe → kick → insert, so two connections A and B for the same ID can both read the stale connection S. The interleaving you describe (B kicks A, then registers) is survivable on its own: `unregister` removes the entry before returning it, so only one caller ever gets the pointer back and closes `kickChan` — A is cleanly kicked with `Aborted: Kicked.` and its agent reconnects. Churn, not corruption.

The damaging interleaving is the one where the loser inserts last:

1. A probes S, fails, kicks S, then stalls before inserting.
2. B probes S (already gone from the map), kicks — finds nothing — and inserts B. Map: B.
3. A resumes and overwrites: map is A, while **B's handler is still running on a live stream with no registry entry.**

That is exactly the state this PR exists to eliminate, reached by a different path. It is reachable in the ticket's scenario: the agent's dial timeout is 5s while the server spends up to `staleConnectionProbeTimeout` (2s) plus auth/DB work in `register()`, so the agent can abandon attempt N and open attempt N+1 while the server-side registration for N is still in flight.

Both halves are now closed:

- `registry.go` — kicks issued after a failed probe go through a new connection-scoped `kickConn(ctx, conn)`, so a losing probe can never disconnect the connection that replaced the one it probed.
- `registry.go` — `Kick(ctx, id)` keeps ID-only semantics for explicit forced kicks (`KickAll`, agent deletion); both route through a shared `kick()`.
- `registry.go` — the insert is now a check-and-insert under `r.rw`: if another connection registered while we were probing, it is newer than the one we probed, so it wins and this registration is refused with `AlreadyExists` rather than overwriting it. The refused stream returns before `Handler.Run` starts, so it leaves no orphan, and the agent retries. If the winner is itself dead, the bounded ping loop tears it down within ~25s and the next retry succeeds.

Tests in `registry_test.go` — `TestKickConn` covers the probed connection being kicked, a superseded probe leaving the newer connection registered and un-kicked, and 8 goroutines concurrently kicking the same ID under `-race`, asserting only the registered connection is disconnected and no `kickChan` is double-closed.

One gap, stated plainly: those tests exercise the registry primitives that carry the invariant, not two concurrent `register()` calls end to end — that needs a stream and DB harness and belongs in `api-tests/` rather than a unit test. I did not fake it with a test that re-implements `register()`'s tail.

---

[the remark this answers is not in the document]
ademidoff: @coderabbitai Can we resolve the issue?

ademidoff:
@coderabbitai full review

ademidoff:
@coderabbitai configuration

managed/services/agents/channel/channel.go
```diff
@@ -335,19 +386,37 @@ func (c *Channel) subscribe(id uint32) chan Response {
...
+// arrive and must not be mistaken for an unsolicited one.
+func (c *Channel) unsubscribe(id uint32) {
+	c.rw.Lock()
+	defer c.rw.Unlock()
+	// Channel is closed, no subscriptions left
+	if c.responses == nil {
+		return
+	}
+	c.responses[id] = nil
+}
```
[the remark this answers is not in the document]
ademidoff: Correct, and confirmed against the code — thanks. `removeResponseChannel` deletes the entry before returning the channel, so once the publisher has taken it, nothing will ever call `removeResponseChannel` for that ID again and the marker written afterwards survives until `close`. You were also right that the impact is bounded: `Metrics` skips nil entries, so no metric distortion.

Fixed in 328c53cb5, though not quite as suggested. Guarding the write stops the leak but leaves the other half of the same interleaving in place: the response that *did* arrive in time is still discarded and reported as a timeout, which is the mistake the extra check at the top of the `done` case existed to prevent. It only moved the window from "between the two selects" to "between the read and the lock".

So `unsubscribe` now reports whether the request was still tracked, which turns the outcome into a single locked decision:

```go
case <-done:
    if c.unsubscribe(id) {
        return nil, errWaitAborted
    }

    resp, ok := <-ch
    if !ok {
        return nil, errors.New("channel is closed")
    }
    return resp.Payload, resp.Error
```

Losing the race now means the publisher is committed to delivering, so the response is collected rather than dropped. The non-blocking pre-check is gone, so `channel.go` is net a few lines shorter.

That receive cannot hang. `removeResponseChannel` is only reached from `publish` and `cancel`, and both send unconditionally when the channel is non-nil; the channel is buffered with capacity 1 and a single sender, so the send never blocks. If `close` got there first, `unsubscribe` returns false on the `c.responses == nil` guard and the receive returns `!ok`, giving "channel is closed".

Covered by `TestUnsubscribe` in `managed/services/agents/channel/channel_test.go`, driving the primitives directly since the interleaving cannot be forced through the gRPC harness. Verified it fails without the fix — the "does not resurrect" case reports `map[1:<nil>]`, the resurrected entry — and passes with it.

managed/services/agents/handler.go
```diff
@@ -88,9 +94,17 @@ func (h *Handler) Run(stream agentv1.AgentService_ConnectServer) error { //nolin
...
 			if err != nil {
-				l.Errorf("agent %s ping: %v", agent.id, err)
+				// Not fatal: pmm-agent answers Ping from the same loop that runs every other
+				// request, so a long one - a connection check carrying the user's unbounded
+				// --connection-timeout, or gathering software versions - delays the pong while
+				// the connection is perfectly alive. The channel carries no other signal that
+				// tells the two apart, so a dead connection is reclaimed when the agent
+				// reconnects and takes over its own registration instead. See PMM-15310.
+				l.WithError(err).WithField("agent_id", agent.id).Warn("Failed to ping pmm-agent.")
 			}
```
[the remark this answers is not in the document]
ademidoff: Declining this one — it asks for the teardown that was in the branch until 8a1332f76, which a live server showed disconnects healthy agents.

Two corrections to the premise.

**It is not indefinite.** When a blackholed peer never reconnects, the kernel eventually gives up on the socket the server keeps writing pings into, `runReceiver` fails, `Requests()` closes, and the `req == nil` path at `managed/services/agents/handler.go:118-126` unregisters the connection. That is bounded by `tcp_retries2`, the ~16 minutes the reporter measured in PMM-15310, and it is the behaviour that shipped for years before #5650. So the exposure is a stale registry entry for an agent that is not there anyway, with a bound.

**The teardown's cost is measured, not hypothetical.** pmm-agent answers `Ping` inline in the same loop that serves every other request (`agent/client/client.go:440-511`), and a connection check carries the user's `--connection-timeout`, which is an unbounded `*time.Duration` (`admin/commands/management/add_mysql.go:129`) plus a 1s margin (`managed/services/agents/connection_checker.go:251`). With the threshold in place, a `--connection-timeout=60s` check against an unroutable host disconnected a live, working agent after 29s:

```
07:33:45 warning Ping timed out while pmm-agent is still sending data.
07:33:55 error   Failed to ping pmm-agent.  ping_failures=1
07:34:05 error   Failed to ping pmm-agent.  ping_failures=2
07:34:05 info    Disconnecting client: ping_timeout.
```

Widening the deadline cannot fix that, because the worst case is user input with no upper bound. I also tried gating the failure count on received traffic, and it does not work either: on a quiet agent `pmm_managed_agents_messages_sent_total` and `..._received_total` were both exactly 32, because every message the server receives is a pong to its own ping. The gate bought one 10s window and then failed anyway.

So the trade is a bounded stale entry for an absent agent, against disconnecting present ones. Recovery for the case the ticket is about does not need it: the bounded probe in `register` reclaims the entry in ~2s once the agent reconnects, verified end to end —

```
20:12:16.179 Connected pmm-agent            (agent's first attempt after the blackhole)
20:12:18.317 Failed to ping ... context deadline exceeded
20:12:18.320 pmm-agent ... is kicked
20:12:18.332 Two-way communication channel established in 2.15s
```

against a stock server, where the same blackhole leaves the agent looping on `context canceled` and never recovering.

The reasoning is in 8a1332f76, and `~/Projects/pmm-15310-tests/blackhole-test.sh` reproduces both runs if you want to check. Happy to revisit if you can point at a path where the `req == nil` reclamation does not fire.

managed/services/agents/registry.go
```diff
@@ -268,18 +273,32 @@ func (r *Registry) register(stream agentv1.AgentService_ConnectServer) (*pmmAgen
...
 			return nil, status.Errorf(codes.AlreadyExists, "pmm-agent with ID %q is already connected.", agentMD.ID)
 		}
 
 		l.Warningf("Failed to ping pmm-agent with ID %q: %v", agentMD.ID, err)
-		r.Kick(ctx, agentMD.ID)
+		r.kickConn(ctx, currentAgent)
 		l.Warningf("pmm-agent with ID %q is kicked.", agentMD.ID)
 	}
 	r.rw.Lock()
 	defer r.rw.Unlock()
```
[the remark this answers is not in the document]
ademidoff: Valid observation, but pre-existing and out of scope here — deferring rather than skipping.

`r.db.InTransactionContext` has been inside `r.rw.Lock()` in `register` since well before this branch; my diff added the check-and-insert guard into that critical section without extending the lock's scope. Two things bound the impact today: the block is inside `if r.haService.Params().Enabled`, so a non-HA deployment does no database I/O under the lock at all, and the write lock is only contended by registration, unregistration and `get`.

The reason I would rather not fix it in this PR is testability. The change makes an agent visible in `Registry.agents` before its `is_connected` row commits, and the rollback has to remove the entry only if it is still the registered one. Every behavioural claim in this PR is backed by a live run against a real agent, and I cannot exercise an HA path that way locally — shipping an untested semantic change to HA registration alongside verified fixes would weaken the part that is actually verified.

Worth noting the follow-up is now cheap, because this PR added the primitive it needs. `unregister` takes the owning connection and no-ops when it has been superseded (`managed/services/agents/registry.go:389-410`), so the rollback is:

```go
if err := persist(ctx, agent); err != nil {
    r.unregister(ctx, agent.id, "register_failed", agent)
    return nil, err
}
```

with no risk of evicting a connection that replaced this one in the meantime. Before this branch that rollback could not be written safely at all.

Happy to do it as its own ticket, where it can get an HA environment and a proper look at the ordering, rather than riding on a connection-recovery fix.

managed/services/agents/channel/channel_test.go
```diff
@@ -198,6 +198,119 @@ func TestServerRequest(t *testing.T) {
...
+			defer close(timedOut)
+
+			ctx, cancel := context.WithTimeout(t.Context(), 100*time.Millisecond)
+			defer cancel()
+
+			// The agent never answers this ping, exactly as it would on a silently dropped
+			// connection. Waiting for it must not block forever. See PMM-15310.
+			resp, err := ch.SendAndWaitResponseWithContext(ctx, &agentv1.Ping{})
+			assert.Nil(t, resp)
+			require.ErrorIs(t, err, context.DeadlineExceeded)
```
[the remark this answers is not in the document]
ademidoff: Right about the contract, but this is already handled — keeping the current shape.

`t.FailNow` must run on the test goroutine, agreed. The failure mode it causes here is the test *hanging* rather than failing, because `Goexit` skips the `close(timedOut)` that the test goroutine is waiting on. That is why the assertions sit inside an inner function whose `defer` closes the channel (`managed/services/agents/channel/channel_test.go:205-222`): `Goexit` runs deferred calls, so the waiter is always released. The `t.Errorf` half of `require` is goroutine-safe and still records the failure, so the test fails, it just does not stop early.

I verified that rather than assuming it. Pointing the assertion at the wrong error deliberately:

```
--- FAIL: TestServerRequestTimeout
    Error: Target error should be in err chain:
           expected: "unexpected EOF"
           in chain: "context deadline exceeded"
FAIL  github.com/percona/pmm/managed/services/agents/channel  4.181s
ELAPSED=7s
```

7 seconds, with the real error, instead of a 10-minute package timeout.

Plumbing results out to the test goroutine is cleaner in isolation, but it collides with two things. `testifylint`'s `require-error` rule mandates `require` for error assertions, so `assert` is not an option — that rule is what produced this shape in the first place. And `require` inside `connect` is the established pattern in this file: 12 such calls across it, nearly all pre-existing (`managed/services/agents/channel/channel_test.go:168` in `TestServerRequest`, `:257` in `TestAgentClosesStream`, and so on). Restructuring only the two new tests would leave the file inconsistent, and restructuring all of them is a test refactor unrelated to PMM-15310.

So the current version is a deliberate compromise: satisfies the linter, cannot hang, and does not fork the file's conventions. If you would rather have the channel-plumbing version applied to the whole file, that is a reasonable cleanup — just better as its own change.

managed/services/agents/registry.go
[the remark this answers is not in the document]
ademidoff: Filed as PMM-15328 — https://perconadev.atlassian.net/browse/PMM-15328 — and linked to PMM-15310.

Two notes from writing it up. The same pattern is on the disconnect path too, at `managed/services/agents/registry.go:396-430` in `unregister`, so the ticket covers both sites rather than just `register`. And the contention list is wider than registration: `Registry.get` at `managed/services/agents/registry.go:556` is the first call in every agent-directed operation (actions, jobs, state updates, connection checks, service-info), and `Registry.Collect` at `:576` means Prometheus scrapes stall as well.

The ticket records three options with tradeoffs, and prefers shortening the critical section and rolling back through the connection-scoped `unregister` this PR added, since that is the piece that makes a safe rollback expressible at all.

4nte (approved):

managed/services/agents/channel/channel.go
```diff
@@ -165,22 +175,62 @@ func (c *Channel) Send(resp *ServerResponse) {
...
 func (c *Channel) SendAndWaitResponse(payload agentv1.ServerRequestPayload) (agentv1.AgentResponsePayload, error) { //nolint:ireturn
+	return c.sendAndWaitResponse(nil, payload)
+}
+
+// SendAndWaitResponseWithContext sends request to pmm-agent and blocks until the response is
+// available, the channel is closed, or ctx is done, whichever happens first.
+// On ctx expiration the subscription is dropped and ctx.Err() is returned.
+// It is no-op once channel is closed (see Wait).
+func (c *Channel) SendAndWaitResponseWithContext(ctx context.Context, payload agentv1.ServerRequestPayload) (agentv1.AgentResponsePayload, error) { //nolint:ireturn
+	resp, err := c.sendAndWaitResponse(ctx.Done(), payload)
```
maxkondr: rather unusual approach to pass context's channel to func instead of the context itself.
usually the context is passed and it may be used by fun's logic not only listen for closed context but creating a child contexts as well. With the proposed solution it becomes impossible.

managed/services/agents/channel/channel.go
```diff
@@ -165,22 +175,62 @@ func (c *Channel) Send(resp *ServerResponse) {
...
-	resp, ok := <-ch
-	if !ok {
-		return nil, errors.New("channel is closed")
-	}
 
-	return resp.Payload, resp.Error
+	select {
+	case resp, ok := <-ch:
+		if !ok {
+			return nil, errors.New("channel is closed")
```
maxkondr: move static error to `var` sections (santinel error)

managed/services/agents/channel/channel.go
```diff
@@ -335,19 +385,46 @@ func (c *Channel) subscribe(id uint32) chan Response {
...
-	if ch == nil {
+	ch, ok := c.responses[id]
+	if !ok {
 		c.l.Errorf("No subscriber for ID %d", id)
 		return nil
 	}
 	delete(c.responses, id)
+
+	if ch == nil {
+		c.l.Debugf("Response for ID %d arrived after its sender stopped waiting for it.", id)
```
maxkondr, line 436: this loge message is in wrong place. this func knows nothing about received messages. Its main and only one purpose - remove a particular response channel from internal map. 
Why there is a need to remove it - completely out of scope of this func

managed/services/agents/channel/channel.go
```diff
@@ -165,22 +175,62 @@ func (c *Channel) Send(resp *ServerResponse) {
...
+// dropped connection should use SendAndWaitResponseWithContext instead.
 func (c *Channel) SendAndWaitResponse(payload agentv1.ServerRequestPayload) (agentv1.AgentResponsePayload, error) { //nolint:ireturn
+	return c.sendAndWaitResponse(nil, payload)
+}
+
+// SendAndWaitResponseWithContext sends request to pmm-agent and blocks until the response is
+// available, the channel is closed, or ctx is done, whichever happens first.
+// On ctx expiration the subscription is dropped and ctx.Err() is returned.
+// It is no-op once channel is closed (see Wait).
+func (c *Channel) SendAndWaitResponseWithContext(ctx context.Context, payload agentv1.ServerRequestPayload) (agentv1.AgentResponsePayload, error) { //nolint:ireturn
```
maxkondr: why a yet another func? why not to adjust the existing `SendAndWaitResponse` ? Otherwise it creates a possibility to not use context and the same issue may appear again

managed/services/agents/channel/channel.go
```diff
@@ -335,19 +385,46 @@ func (c *Channel) subscribe(id uint32) chan Response {
 	return ch
 }
 
+// unsubscribe marks the request as abandoned: its sender is gone, but a response may still
+// arrive and must not be mistaken for an unsolicited one.
+//
+// It reports whether the request was still tracked. False means the publisher took the entry
+// first and is committed to delivering a response, so there is nothing left to mark: writing
+// the marker anyway would leave an entry behind that nothing ever clears.
+func (c *Channel) unsubscribe(id uint32) bool {
```
maxkondr: `removeResponseChannel` should be enough - why new func is required?

managed/services/agents/channel/channel.go
```diff
@@ -165,22 +175,62 @@ func (c *Channel) Send(resp *ServerResponse) {
...
+		// select picks a ready case at random, so the response may already have been delivered
+		// when done fired, and it may land while we are giving up. Marking the request
+		// abandoned settles that: it succeeds only while nothing has taken the subscription,
+		// and once taken the publisher is committed to sending, so the response is imminent.
+		// Reporting a timeout for a response that did arrive would be wrong.
+		if c.unsubscribe(id) {
+			return nil, errWaitAborted
+		}
+
+		resp, ok := <-ch
```
maxkondr: in this case it should be enough
```suggestion
        removeResponseChannel(id)
        close(ch)
		resp, ok := <-ch
```

but on the other hand, if context is already canceled - it means the caller is not longer interested in result, so there is no big sense in trying to read and return response even if it is really received.

managed/services/agents/handler.go
```diff
@@ -88,9 +94,17 @@ func (h *Handler) Run(stream agentv1.AgentService_ConnectServer) error { //nolin
 	for {
 		select {
 		case <-ticker.C:
-			err := h.r.ping(ctx, agent)
+			pingCtx, cancelPing := context.WithTimeout(ctx, defaultAgentPingTimeout)
+			err := h.r.ping(pingCtx, agent)
+			cancelPing()
```
maxkondr: this implementation (but the prev one as well) suffered from big problem - this loop is blocked until the Pong is received. On high-load there may be a significant flow between agent<->server. Now all requests processing from agent are (`case req := <-agent.channel.Requests():`) are blocked. Ads result it may block `Channel.runReceiver` and Pong will never be received -> ping response timeout timer will fire.

This is not a truly asynchronous processing that produces a potential problems

managed/services/agents/channel/channel.go
ademidoff: Agreed, and done in c412338db — there is one `SendAndWaitResponse(ctx, payload)` now.

For the record on why the pair existed: routing the ctx-less variant through `context.Background()` internally made `contextcheck` fire at **24** call sites (21 via `sendActionRequest`, 2 in `RestartJob`), because the linker sees a context being dropped on the way to the channel. Rather than nolint those, I had hidden the wait behind a private `done <-chan struct{}`. Your point stands though — that left a method a future caller could pick and reintroduce the unbounded wait, and the sentinel plumbing (`errWaitAborted`, the wrapper pair) existed only to support it. Collapsing removed ~25 lines net.

What the call sites do now:

- **Threaded for real** — every `ActionsService` method already declared a `context.Context` and ignored it (`_ context.Context`), so passing it into `sendActionRequest` was free. Same for `AgentService.Logs`. `CheckConnectionToService`, `GetInfoFromService` and `sendSetStateRequest` already had one in hand. `sendSetStateRequest` is the interesting one: its caller wraps it in `context.WithTimeout(ctx, stateChangeTimeout)` (5s) and the channel was silently ignoring that deadline. It honours it now, and the existing error path re-queues the update.
- **`context.TODO()`** — the job (`StartMy{SQL,MongoDB}{Backup,RestoreBackup}Job`, `StopJob`, `restartSystemService`), version (`GetVersions`) and `PBMSwitchPITR` entry points take no context at all, and threading one in means changing the `jobsService`/`versioner` interfaces in `services/backup` and `services/versioncache` plus their mocks — and it would newly make a backup abort when the API request that started it goes away. That is a separate change with its own testing, so those pass `context.TODO()` and wait exactly as they did before. Two `//nolint:contextcheck` remain in `RestartJob` where a real context is in scope but cannot be handed over yet; happy to file a follow-up ticket for the threading.

Worth noting these calls are no longer able to hang forever regardless: the stale connection is detected and kicked within ~2s, and `Channel.close()` unblocks every waiter with `channel is closed`.

managed/services/agents/channel/channel.go
ademidoff: Done.

managed/services/agents/channel/channel.go
ademidoff: I removed that one, it really looked weird.

managed/services/agents/handler.go
ademidoff: You are right, and the consequence in the pre-PR code was worse than a fired timer — fixed in 9aa64d563 by giving the ping its own goroutine.

Walking the cycle you describe: `Run`'s select was the only reader of `channel.Requests()`, and the same select waited for the pong. Parked in a ping it drained nothing, `agentRequestsCap` is 32, and once that buffer filled `runReceiver` blocked on `c.requests <- ...` — the one goroutine that could publish the pong being waited for. Before this branch `ping` had no deadline, so **nothing broke that cycle**: the handler stayed parked until the stream itself died, and that agent stopped being processed entirely while looking connected. The 5s deadline I added turned it into a warning plus a re-drain; your point is that the coupling should not be there at all, and that is the better fix.

It now looks like this:

```go
go h.state.runStateChangeHandler(ctx, agent)
go h.runPingHandler(ctx, agent)
```

`runPingHandler` exits on `ctx.Done()` or `agent.kickChan`, exactly like `runStateChangeHandler` next to it. `kickChan` is only ever closed (`registry.go`), never sent to, and `state.go` already has a second receiver on it, so an extra one is safe. The deadline stays, for a different reason now: without it a ping on a blackholed socket parks ~15 minutes holding a subscription and observing nothing.

What I deliberately did **not** change: request handling stays serialized per agent. `stateChanged` runs a transaction and can force a VictoriaMetrics reload, `QANCollect` writes through to ClickHouse, and the handlers assume they are not racing each other for one agent. Making those concurrent is a real change to the model with ordering consequences — happy to file it as its own ticket if you want it, but I would not fold it in here.

Verified on a throwaway stack (pmm-agent 3.9.0, server built from this branch): the blackhole check still recovers — probe fails 19:47:37.356 `context deadline exceeded`, `is kicked` the same millisecond, one reconnect attempt, `Connected: true` — the busy-agent check still shows 0 `ping_timeout` disconnects, pings hold a clean 10s cadence, and there is no goroutine leak: after a kick/reconnect cycle plus a 60s block, `runPingHandler` sits at 2 for 2 connected agents, matching `runStateChangeHandler`, with total goroutines at 80 versus 82 before. What I could not synthesize is the 33-message burst itself, so the stall and its fix are reasoned from the code rather than reproduced.

managed/services/agents/channel/channel.go
ademidoff: Fair, and it applies to the line that was already there — `No subscriber for ID` has the same problem, it just predates me. Restructured in ff2fb77eb.

The map function now only answers what it can actually know, and says nothing:

```go
func (c *Channel) removeResponseChannel(id uint32) (chan Response, subscription) {
	...
	if c.responses == nil    { return nil, subscriptionClosed }
	if _, ok := ...; !ok     { return nil, subscriptionUnknown }
	delete(c.responses, id)
	if ch == nil             { return nil, subscriptionAbandoned }
	return ch, subscriptionWaiting
}
```

Deciding what that means belongs to the delivery path, which is reached from `runReceiver` and therefore does know a message arrived, so `deliver` owns the switch and both log lines. `cancel` and `publish` also stop duplicating the take-and-send.

On why the states are needed at all rather than one nil check: they are four genuinely different situations and two of them must stay quiet. An abandoned request logged as unknown puts `No subscriber for ID` on the **error** log for every late response after a timeout — 25 of them in one 10 minute run here — and a closed channel logged as unknown does the same for any response racing a disconnect. That is the whole reason I touched this function in the first place.

Both of those are now tested rather than asserted in a comment. `TestDeliver` covers all four states, and I checked it earns its place by breaking the code on purpose: folding `subscriptionClosed` into `subscriptionUnknown` fails only `stays silent once the channel is closed`, and folding `subscriptionAbandoned` in fails only `reports an abandoned request at debug level`. Worth noting the pre-existing `TestLateResponseAfterTimeout` passes under that second break — the log distinction genuinely had no coverage before.

managed/services/agents/channel/channel.go
ademidoff: Because they do opposite things to the map, but I think the name was doing the damage — `unsubscribe` reads like the inverse of `subscribe`, i.e. exactly "removeResponseChannel by another name". Renamed to `abandon` in 590f86878.

- `removeResponseChannel` **drops** the entry. Called by the publisher, which is about to deliver.
- `abandon` **leaves a marker** in its place — `c.responses[id] = nil`. Called by the waiter that gave up.

That marker is the entire point, and it is what `removeResponseChannel` cannot give you. Drop the entry on timeout instead and the late response that follows lands in `subscriptionUnknown`, so it is logged as `No subscriber for ID` at **error** level. That is not hypothetical: in the live run yesterday the busy-agent scenario produced 25 late responses in ten minutes, all of them expected. They are the difference between an agent whose request loop is momentarily busy and a genuine protocol violation — a response to an ID the server never sent, which is what `TestUnexpectedResponseIdFromAgent` covers, and which we do want on the error log.

You are right that the timeout **race** could be settled either way now that `removeResponseChannel` reports state; the waiter could check for `subscriptionWaiting` and otherwise collect from the channel it already holds. So the bool is not what justifies a second function. The marker is. What the bool buys is marker hygiene: writing one after the publisher already took the entry would leave an entry that no later response can ever clear, since the only response was already delivered. I re-checked that this is really tested by deleting the guard — `TestAbandon/does_not_resurrect_an_entry_the_publisher_already_took` fails with `map[1:<nil>]` left behind.

Cost of the marker: one map entry per abandoned request, cleared when the response arrives, and dropped wholesale by `close`. So it is bounded by the connection's own lifetime, and `Metrics` skips markers so they never show up as queued responses. In the live run all 25 cleared, and both queue-length gauges ended at 0.

Pure rename plus doc, no behaviour change.

managed/services/agents/channel/channel.go
ademidoff: Two separate things here, and you are right about the second one being the interesting question.

**On the suggestion.** I cannot take it as written, because `close(ch)` from the waiting side races the publisher — only the publisher may close that channel. `deliver` closes it itself for an error response, so both sides can end up closing:

```
waiter:    removeResponseChannel(id) -> subscriptionUnknown   (publisher already took it)
publisher: ch <- resp; close(ch)      (resp.Error != nil)
waiter:    close(ch)                  -> panic
```

I put your two lines behind a 2000-iteration race between `deliver` and that waiter, and it panicked with `close of closed channel` on all four runs. The mirror interleaving — waiter closes first, publisher then sends — is `send on closed channel`. The intent works, though: closing is what makes the follow-up read non-blocking. The blocking read plus the bool gets the same guarantee without a receiver closing a channel it does not own, since a false result means the publisher is already committed to sending.

**On whether to read at all.** This is the part worth arguing. It is not about being polite to a caller that walked away — it is that `select` picks a ready case at random, so with the response already sitting in the buffer, "ctx is done" and "response available" are both ready and the outcome would be a coin toss. Reading makes the observable behaviour the same whichever case wins: if a round trip completed, its result is returned; if it did not, `ctx.Err()`.

The caller that makes this concrete is the stale-connection probe in `Registry.register`. A pong landing at 1.999s against a 2s deadline proves the connection is alive, and returning `DeadlineExceeded` there means kicking an agent that answered. Rare, yes — but it is the difference between a correct decision and an arbitrary one, and the alternative discards work that is already done for nothing gained.

Dropping the read is a one-liner (`c.abandon(id); return nil, ctx.Err()`) with no leak, since an entry the publisher already took is gone from the map and the buffered response is simply collected by nobody.

What your comment did catch is that the doc comment claimed `ctx.Err()` flatly and said nothing about the delivered-response case. Fixed in 643468399, along with a note on the branch itself about which caller the choice matters for.

managed/services/agents/handler.go
```diff
@@ -81,18 +87,16 @@ func (h *Handler) Run(stream agentv1.AgentService_ConnectServer) error { //nolin
 	// run pmm-agent state update loop for the current agent.
 	go h.state.runStateChangeHandler(ctx, agent)
 
+	// Ping from its own goroutine. The loop below is the only reader of Requests(), so a ping
+	// waiting for its pong in there stops the queue from draining, and once the queue is full
+	// channel.runReceiver blocks - the very goroutine that has to deliver that pong. Keeping the
+	// two apart also keeps a slow request from delaying a ping. See PMM-15310.
+	go h.runPingHandler(ctx, agent)
```
maxkondr, line 95: no much sense in this additional goroutine - now a Ping is sent asynchronously (), but Pong is still read by `runReceiver` together with all other requests and responses from pmm-agent. Conceptually there is an issue of mixing control and data plane into one stream.

managed/services/agents/versioner.go
```diff
@@ -185,7 +186,7 @@ func (s *VersionerService) GetVersions(pmmAgentID string, softwareList []Softwar
 	}
 
 	request := &agentv1.GetVersionsRequest{Softwares: softwareRequest}
-	response, err := agent.channel.SendAndWaitResponse(request)
+	response, err := agent.channel.SendAndWaitResponse(context.TODO(), request)
```
maxkondr, line 189: unlimited wait?

JiriCtvrtka (approved):

managed/services/agents/handler.go
ademidoff: The pong does still come through `runReceiver` with everything else — but that is not what the goroutine is for. Before it, the two were in a cycle:

```
Run's select  -- parked in ping, waiting for the pong -->  runReceiver
runReceiver   -- blocked on c.requests <- (queue full, 32) -->  Run's select
```

Neither side could move, and nothing outside could break it; pre-branch, with no deadline on the ping, that was permanent for the life of the stream. With the ping moved out, `Run` keeps draining `Requests()` unconditionally, so the queue cannot stay full and `runReceiver` always gets to the pong.

What remains is what you are pointing at: a slow request handler — `QANCollect` writing through to ClickHouse, `stateChanged` forcing a VictoriaMetrics reload — still stops the draining and can starve a pong. The difference is that such a handler is not waiting on the pong, so it finishes and the queue drains. A bounded delay, not a deadlock. I have reworded the comment in c0ac8e345 to say that, since it clearly did not say it before.

On mixing control and data plane in one stream: agreed, and it is not fixable at this level. The pong is a message on the same ordered stream as the data, so head-of-line blocking is inherent — on the agent side too, where `Ping` is served from the same loop as every other request, which is why a ping failure here is only logged and never disconnects anyone. A transport-level keepalive does not substitute for it either: behind nginx, `grpc_pass` terminates HTTP/2 and answers PING frames itself over loopback while the agent-to-nginx segment is dead. Doing this properly means a separate control channel, i.e. a protocol change. Happy to write that up as its own ticket if you want it on the record.

Re-verified live after the change: blackhole recovery still kicks the stale connection and reconnects, the busy-agent case still reports 0 `ping_timeout` disconnects, pings hold a 10s cadence, and `runPingHandler` tracks connected agents exactly across a kick, a 60s block, a SIGKILL and a restart — no leaked goroutines.

managed/services/agents/versioner.go
ademidoff: Yes — unlimited on the context, deliberately unchanged from before this PR, where the call had no context at all. `context.TODO()` is there to mark it rather than to hide it.

What does bound it now is the connection. When the stale one is kicked — within ~2s of the agent's next registration attempt — the handler returns, the stream ends, and `Channel.close()` releases every waiter with `channel is closed`. That is the case from the ticket, and it is the one that used to hang for the kernel's TCP timeout. `TestAgentClosesStream` and `TestAgentClosesConnection` cover the release path.

What is left unbounded is "connection healthy, agent simply slow to answer", and I would argue that is correct here specifically. Gathering software versions runs commands on the monitored host; it is one of the operations named in the ping comment in `handler.go` as a legitimately long request, and the same reasoning as the busy-agent case applies — the server cannot tell a slow agent from a dead one by elapsed time, so an arbitrary deadline here would fail version gathering on loaded hosts.

Threading a real context through is doable and smaller than for the job entry points: `GetVersions` appears in two interfaces (`versioncache`, `backup/deps.go`) plus two mocks, and a context is available one level above both callers — `Run(ctx)` in `versioncache` and `CheckSoftwareCompatibilityForService(ctx, ...)` in the compatibility service. Worth noting it would not answer your question for the background caller though: `Run`'s context lives as long as the service, so that path stays unbounded either way. Only the compatibility check would pick up a real deadline, and only if its caller set one.

So there are two separable pieces of work: propagate the context (mechanical, needs mock regen), and decide whether version gathering deserves a hard timeout and what it should be (a policy call, and a behaviour change for slow hosts). Both are out of scope for a keepalive fix, but I am happy to file them — say which you want and I will raise it. The same `context.TODO()` question applies to the job entry points and `PBMSwitchPITR`, so one ticket can cover the lot.

PR #5753, opened 2026-08-10
PMM-15204: Dynamic RTA CSV export columns

mattiasimonato:
Ticket number: PMM-15204

Feature build: SUBMODULES-4518

The export hardcoded 14 columns, so fields the backend added were silently missing. Columns now come from the API response itself:

- Each row is flattened, hoisting `mongoDbPayload` (and any future engine payload) to the top level.
- Headers default to the snake_case form of the API field name, so they match the wire. An override map covers the six that differ: `operation_id`, `elapsed_exec_time_sec`, `service`, `user_name`, `data_capture_time`, `raw_query`.
- Order is pinned for the 14 agreed columns; anything unrecognised is appended under its API name.
- Columns are collected across **all** rows and passed as `columnHeaders`. This replaces `useKeysAsHeaders`, which read only the first row's keys and dropped fields unset there.
- Non-primitives are serialised, because `export-to-csv` throws on object values, so a future message-typed field would have crashed the Export button rather than just being missed.

This adds two columns that were previously dropped, `service_id` and `query_text`. It also fixes `Exclude` used where `Omit` was meant in `QueryData`, a no-op on object types that left the raw duration on the type and on every row.


ui/apps/pmm/src/pages/rta/overview/export/exportRtaQueriesToCsv.ts
```diff
@@ -12,41 +27,88 @@ export const sanitizeCsvCell = (value: string): string => {
...
+  'queryRawJson',
+].map(toCsvHeader);
+
+const columnRank = (column: string): number => {
+  const index = CSV_COLUMN_ORDER.indexOf(column);
+
+  return index === -1 ? CSV_COLUMN_ORDER.length : index;
+};
+
+const isPlainObject = (value: unknown): value is Record<string, unknown> =>
```
fabio-silva: @mattiasimonato this is something we can send to utils

ui/apps/pmm/src/pages/rta/overview/export/exportRtaQueriesToCsv.ts
```diff
@@ -12,41 +27,88 @@ export const sanitizeCsvCell = (value: string): string => {
...
+        collect(value);
+
+        return;
+      }
+
+      row[toCsvHeader(key)] = toCsvValue(value);
+    });
   };
+
+  collect(query as unknown as Record<string, unknown>);
```
fabio-silva: I think creating an inner function and then calling it, whilst not doing anything else, is way more verbose that just doing the loop right away, we can cut the middle man here

fabio-silva (changes requested):

ui/apps/pmm/src/pages/rta/overview/export/exportRtaQueriesToCsv.ts
mattiasimonato: Moved to `ui/apps/pmm/src/utils/object.utils.ts`
https://github.com/percona/pmm/pull/5753/changes/740306f596aa59ddffc3dc5993f75167ab393567

ui/apps/pmm/src/pages/rta/overview/export/exportRtaQueriesToCsv.ts
mattiasimonato: One thing though: `collect` is recursive (it calls itself on nested objects), which is what flattens the payload, so an inline loop would leave `mongoDbPayload` as an object and `export-to-csv` throws on those. 
Lifted it to a module-level recursive helper instead, so `mapQueryToCsvRow` is now a one-liner.
https://github.com/percona/pmm/pull/5753/changes/d163ba3ce5190a31f69c595785b0a92e89b7d96d

fabio-silva (approved):

matejkubinec (approved):

theTibi:
@coderabbitai full review

PR #5755, opened 2026-08-10
PMM-15280 Provision a Grafana service account for SEP

yyyyyyyan:
## Summary

With `PMM_ENABLE_SEP` set, pmm-server now provisions the Grafana service account SEP authenticates users against, and publishes its token to `/srv/sep-secrets` as two files named after the canonical SEP settings names (`AUTH__PROVIDER__GRAFANA__SERVICE_ACCOUNT_TOKEN` and `PMM__API_KEY`, same value in both).

This replaces the manual minting step SEP's feature-build harness performs today, which calls `POST /graph/api/serviceaccounts` with `admin:admin` basic auth from outside the container — it breaks as soon as the Grafana admin password changes, and it requires whoever provisions SEP to hold a Grafana admin credential.

Everything here is a no-op unless `PMM_ENABLE_SEP` is set. Docker only, matching #5700.

**Base branch is `PMM-15238-expose-pg-to-sep` (#5700), not `main`** — that branch adds the `PMM_ENABLE_SEP` passthrough this one extends. Please merge #5700 first.

### How the account is created

Directly in Grafana's own PostgreSQL database over the local socket, the same way `postgres-sep` provisions the `sep` role as the local superuser. Grafana exposes no credential-free path to create a service account — checked on 12.4.5: `grafana cli admin` offers only `reset-admin-password`, `data-migration`, `secrets-migration`, `secrets-consolidation` and `flush-rbac-seed-assignment`, and `conf/provisioning/` covers only `access-control`, `alerting`, `dashboards`, `datasources`, `plugins` and `sample`. `POST /api/serviceaccounts` therefore needs an Admin credential, which pmm-server does not reliably have once an operator has changed the admin password.

### Two deliberate design points

**It runs as a supervisord one-shot, not from the entrypoint.** Grafana is not a running process at any point during the entrypoint, and on a first boot its schema does not exist at all — the `grafana` database is created by the `pmm-init` playbook and migrated by Grafana itself, both strictly after the entrypoint has exec'd supervisord. The helper waits for `/api/health` to report `"database": "ok"`, the same readiness signal pmm-managed uses.

**The token is validated every start and replaced only when it is missing, empty, or rejected** — not re-minted unconditionally. SEP reads the secrets directory once, at its own container start, so re-minting on every pmm-server start would revoke the token a still-running SEP is holding, and a plain `docker restart pmm-server` would leave SEP 401-ing until it too was restarted.

A freshly minted token that Grafana rejects is fatal and is **never published**, so a future change to Grafana's schema or token encoding surfaces as a failed `sep-provision` rather than as silent SEP auth failure.

### Contract for the SEP side

| Element | Value |
|---|---|
| Directory | `/srv/sep-secrets` (the `pmm-sep-secrets` volume); SEP mounts it read-only with `SECRETS_DIR` pointing at it |
| Filenames | `AUTH__PROVIDER__GRAFANA__SERVICE_ACCOUNT_TOKEN`, `PMM__API_KEY` — same value |
| Modes | directory setgid `2770`, files `0640`; each file's group is inherited from the directory, so it is group 0 whatever uid PMM runs as |
| SEP-side requirement | the SEP container must join group 0 (`group_add: ["0"]`, or `user: "1001:0"`) — PMM cannot `chgrp` to SEP's own group. Leave both settings names unset there rather than empty: an empty value counts as supplied and outranks the mounted file |
| Service account | name `pmm-sep`, login `sa-<orgId>-pmm-sep`, org role Admin, one token named `pmm-sep` |
| Disabled | with `PMM_ENABLE_SEP` unset both files are removed; the Grafana service account is left in place |

## Tested

There is no unit-test layer for the container's shell/ansible surface and this change touches no Go code, so verification is a matrix run against a built image. Because the server image needs the full RPM pipeline, I ran everything against `perconalab/pmm-server-fb:PR-4500-0357c3b` — which ships **percona-grafana 12.4.5 on the embedded PostgreSQL** and already carries this PR's base branch — with the helper and `sep.ini` layered in. Commands were run verbatim; every result below was observed.

- [x] **Mechanism gate on the real target.** The token encoding reproduces an API-minted token exactly on percona-grafana 12.4.5: value `glsa_<secret>_<checksum>`, checksum = `crc32("glsa_"+secret)` little-endian, `api_key.key` = `pbkdf2_hmac("sha256", secret, checksum, 10000, 50)`. The `user`/`org_user`/`api_key` column sets and the `UQE_user_login` / `UQE_org_user_org_id_user_id` / `UQE_api_key_org_id_name` indexes all match what the SQL assumes. `api_key.role` is stored as `Viewer` even when the API request asks for Admin — effective authority comes from `org_user.role`.
- [x] **Fresh start under real supervisord.** `sep-provision` reaches `exit status 0; expected`; both files appear `0640 pmm:root` in a `0770` directory; the token returns 200 from `/api/access-control/user/permissions` and `"role":"Admin"` from `/graph/api/serviceaccounts/search` through nginx.
- [x] **Restart idempotency.** `docker restart` with Grafana state intact keeps the same token value, `api_key` rows stay at 1, one `user` row, and the token still authenticates without SEP being restarted.
- [x] **Grafana state reset**, using the real two-volume topology: destroyed the container and its `/srv` volume while keeping the secrets volume, then started fresh. The stale token was rejected, replaced, and the replacement validated by an actual HTTP call.
- [x] **Stale token and deleted account.** Deleting the `api_key` row, and separately deleting the whole service account through the Grafana API, both lead to a re-mint that authenticates; no accumulation.
- [x] **Disabled.** On a fresh volume with `PMM_ENABLE_SEP` unset: no service account, no secrets directory, an empty `sep-provision` log, `readyz` 200 and the container healthy. Disabling after enabling removes both token files and leaves an unrelated file in the directory untouched and the Grafana account in place.
- [x] **Guards.** `PMM_HA_ENABLE`, `PMM_DISABLE_BUILTIN_POSTGRES`, `GF_DATABASE_HOST` and `GF_DATABASE_URL` each produce the warning and exit 0 without writing anything.
- [x] **Secrecy.** The token secret appears in no process's `argv`, the container environment, `docker logs`, `docker inspect`, or `sep-provision.log` — checked with the needle staged in a file so the probe itself could not create a match, and with a control confirming the probe detects a real hit.
- [x] **Permissions contract.** uid 1001 in group 0 reads both files; the same uid with gid 1001 gets `Permission denied`; uid 1002 cannot read them at all.
- [x] **Access control.** With `PMM_ENABLE_ACCESS_CONTROL=1` the validator endpoint still returns 200 and the role still reads `Admin`, so no token is revoked for a configuration reason.
- [x] **Failure paths.** A corrupted token encoding (simulating a future Grafana change) is fatal, exits 1, and publishes nothing. Grafana never becoming ready is fatal and exits 1. A failing one-shot exits once and stays `EXITED` — the pre-review configuration respawned it ~82 times in 8 seconds, which the review caught.
- [x] **Volume-ownership trap.** A named volume mounted at a path present in the image owned `1000:0 0770` inherits that ownership and is writable; at a path absent from the image it is created `root:root` and writes are denied. This is what the `post-build.yml` mountpoint task exists for. On a real build the one-line check for it is `docker run --rm --entrypoint /bin/ls <image> -ldn /srv/sep-secrets`, which should print `1000 0` — that belongs to the Feature Build item below, not to this one.
- [x] shellcheck clean; `bash -n` clean; the three changed playbooks parse; `docker compose config` succeeds.

**The matrix above was run against the pre-review revision** (`16c2db83`). Reviewing the helper against the two contracts it documents turned up five defects, fixed in `c5d0c344`: the published files took the group of whichever uid pmm-server runs as rather than the directory's, so `group_add: ["0"]` only worked while that uid's primary group was 0; the reuse check read an unreachable Grafana as a rejected token and re-minted, revoking a live SEP token; `install -d` on an existing directory exited 0 only because the mode it asked for matched the image; a `psql` failure in the secondary role repair aborted before the valid token was republished; and the Grafana-not-on-embedded-Postgres guard left token files it can no longer revalidate. The decision paths were re-verified directly — keep, genuine rejection, inconclusive answer, dropped connection, first boot, disabled, and both guards — and the setgid ownership model was reproduced end to end on `oraclelinux:9-slim`, including that Docker carries the bit into a fresh named volume. **The image rows still need re-running on a Feature Build**, since only a real build produces the setgid mountpoint.

Not run here, and needing a Feature Build or the SEP side:

- [ ] **The shipped image path.** The ansible copy tasks, the supervisord registration and the `/srv/sep-secrets` mountpoint surviving into the image were reviewed and reasoned about but not executed — `VOLUME ["/srv"]` prevents baking the mountpoint into a derived image, so only a real build exercises them. **A Feature Build is required before merge.**
- [ ] **End to end with SEP.** This ships only the PMM half. `sidecar/pmm-fb/compose.yaml` on percona/SEP branch `pmm` still passes `SEP_GRAFANA_TOKEN` and declares no shared volume, `SECRETS_DIR` or `group_add`, so the paired stack cannot consume these files yet. A mounted file already outranks the value SEP derives from `SEP_GRAFANA_TOKEN`, so the two channels can coexist during migration. 

## Known limitations

- **A failed provisioning run is not retried until the container restarts.** A one-shot cannot bound its own restarts under supervisord — `startretries` only bounds restarts out of `BACKOFF`, which `startsecs = 0` never reaches — so the choice is between no retry and an unbounded respawn loop. Failure is loud (`EXITED` with a non-zero status, plus a `FATAL` line in `/srv/logs/sep-provision.log`) and the readiness budget is 600s to absorb a slow first boot. 
- **The `pmm-sep` service account name is this feature's namespace.** Re-provisioning adopts an existing `sa-<org>-pmm-sep` account and deletes every token on it, plus any `pmm-sep`-named token elsewhere in the org. That is what makes the state-reset path idempotent, but a manually created account of the same name would be adopted and its tokens replaced.
- **Grafana's schema and token encoding are consumed, not modified**, and are internal to Grafana. The scheme survived 11.6.4 → 12.4.5 with an identical column set and encoding. Treat a change to `%define grafana_version` in `build/packages/rpm/server/SPECS/grafana.spec` as a trigger to re-run the mechanism gate.
- With `PMM_ENABLE_SEP` unset the residue in stock PMM is an empty `/srv/sep-secrets` directory in the image (which must pre-exist for volume ownership to be inherited), a `sep-provision` program that `supervisorctl status` shows as `EXITED` after each boot, an empty `/srv/logs/sep-provision.log`, and — for users of this repo's `docker-compose.yml` — a `pmm-sep-secrets` named volume created unconditionally, because compose has no conditional mounts. None carries data, a service account or any behaviour, but the acceptance criterion's "nothing about stock PMM changes" is these four things wide.
- `docker-compose.dev.yml` is not updated; the dev stack builds from a published dev-container image that does not carry this helper, so the volume would have nothing to deliver.

## Related

- prerequisite: #5700 (PMM-15238)
- Sibling: PMM-15316 delivers SEP's `SECRET_KEY` and database credentials over this same channel

### DIFF
--- a/build/ansible/pmm-docker/post-build.yml
+++ b/build/ansible/pmm-docker/post-build.yml
@@ -70,13 +70,16 @@
     # Must exist in the image: a named volume mounted at a path the image lacks is
     # created root:root, which pmm-server cannot write to. Group-writable like the
     # other /srv directories, so an arbitrary runtime uid in group 0 can still write.
+    # Setgid so the token files written here belong to group 0 whatever that uid's
+    # primary group is - the SEP container joins group 0 to read them. Docker carries
+    # the bit into a fresh named volume mounted at this path.
     - name: Create the SEP secrets mountpoint
       file:
         path: /srv/sep-secrets
         state: directory
         owner: pmm
         group: root
-        mode: 0770
+        mode: 02770
 
     - name: Create empty log directory for nginx
       file:
--- a/build/ansible/pmm-docker/post-build.yml
+++ b/build/ansible/pmm-docker/post-build.yml
@@ -67,6 +67,16 @@
         - directory
       no_log: true
 
+    # Must exist in the image: a named volume mounted at a path the image lacks is
+    # created root:root, which pmm-server cannot write to.
+    - name: Create the SEP secrets mountpoint
+      file:
+        path: /srv/sep-secrets
+        state: directory
+        owner: pmm
+        group: root
+        mode: 0750
+
     - name: Create empty log directory for nginx
       file:
         path: /var/log/nginx
--- a/build/ansible/pmm-docker/post-build.yml
+++ b/build/ansible/pmm-docker/post-build.yml
@@ -68,14 +68,15 @@
       no_log: true
 
     # Must exist in the image: a named volume mounted at a path the image lacks is
-    # created root:root, which pmm-server cannot write to.
+    # created root:root, which pmm-server cannot write to. Group-writable like the
+    # other /srv directories, so an arbitrary runtime uid in group 0 can still write.
     - name: Create the SEP secrets mountpoint
       file:
         path: /srv/sep-secrets
         state: directory
         owner: pmm
         group: root
-        mode: 0750
+        mode: 0770
 
     - name: Create empty log directory for nginx
       file:
--- a/build/ansible/roles/grafana/files/grafana-sep
+++ b/build/ansible/roles/grafana/files/grafana-sep
@@ -1,0 +1,216 @@
+#!/bin/bash
+#
+# Provisions the Grafana service account SEP authenticates with, and publishes its
+# token to a secrets directory the SEP container mounts. One file per canonical SEP
+# settings name, mode 0640, owned by pmm's primary group, so SEP must join group 0.
+#
+# Everything below is a no-op unless PMM_ENABLE_SEP is set. When it is unset again,
+# the token files this script wrote are removed on the next start; the Grafana
+# service account itself is left in place.
+#
+# Runs under supervisord rather than from the entrypoint because Grafana is neither
+# running nor migrated - on a first boot its schema does not exist at all - until
+# long after the entrypoint has exec'd supervisord.
+#
+# The token is validated against Grafana on every start and replaced only when
+# Grafana rejects it, because SEP reads the secrets directory once at its own start:
+# re-minting unconditionally would revoke the token a running SEP is still holding.
+
+set -o errexit
+set -o pipefail
+
+declare SECRETS_DIR="${SECRETS_DIR:-/srv/sep-secrets}"
+declare GRAFANA_URL="http://127.0.0.1:3000"
+declare POSTGRES_BIN_DIR="${POSTGRES_BIN_DIR:-/usr/pgsql-14/bin}"
+declare SA_NAME="pmm-sep"
+declare TOKEN_NAME="pmm-sep"
+declare READY_TIMEOUT=300
+declare READY_INTERVAL=5
+declare -a TOKEN_FILES=(AUTH__PROVIDER__GRAFANA__SERVICE_ACCOUNT_TOKEN PMM__API_KEY)
+
+is_enabled() { [ "$1" = "1" ] || [ "$1" = "true" ]; }
+
+psql_grafana() {
+    "$POSTGRES_BIN_DIR/psql" -X -q -v ON_ERROR_STOP=1 -U postgres -h /run/postgresql -d grafana "$@"
+}
+
+# Same readiness signal pmm-managed waits on: /api/health reporting the database as ok.
+wait_for_grafana() {
+    local waited=0
+
+    while [ "$waited" -lt "$READY_TIMEOUT" ]; do
+        if curl -sS --max-time 5 "$GRAFANA_URL/api/health" 2> /dev/null |
+            grep -q '"database": *"ok"'; then
+            return 0
+        fi
+        sleep "$READY_INTERVAL"
+        waited=$((waited + READY_INTERVAL))
+    done
+
+    return 1
+}
+
+# Proves the token by an authenticated call to the access-control endpoint SEP's
+# org-role lookups go through. The header is fed on stdin rather than passed with
+# -H so the token never appears in any process's argv.
+grafana_authenticates() {
+    local code
+
+    code=$(printf 'header = "Authorization: Bearer %s"\n' "$1" |
+        curl -sS -K - -o /dev/null -w '%{http_code}' \
+            "$GRAFANA_URL/api/access-control/user/permissions") || return 1
+
+    [ "$code" = "200" ]
+}
+
+# Prints the token value, its api_key.key hash, and the user row's salt, rands and
+# uid, one per line. The encoding reproduces Grafana's own: the checksum is
+# crc32("glsa_" + secret) serialised little-endian, and the stored hash is
+# util.EncodePassword(secret, checksum).
+generate_token_material() {
+    python3 - <<'PY'
+import binascii
+import hashlib
+import secrets
+import string
+import zlib
+
+alnum = string.ascii_letters + string.digits
+secret = "".join(secrets.choice(alnum) for _ in range(32))
+checksum = binascii.hexlify(
+    zlib.crc32(("glsa_" + secret).encode()).to_bytes(4, "little")
+).decode()
+
+print(f"glsa_{secret}_{checksum}")
+print(hashlib.pbkdf2_hmac("sha256", secret.encode(), checksum.encode(), 10000, 50).hex())
+print("".join(secrets.choice(alnum) for _ in range(10)))
+print("".join(secrets.choice(alnum) for _ in range(10)))
+print("".join(secrets.choice(string.ascii_lowercase + string.digits) for _ in range(14)))
+PY
+}
+
+# Administers Grafana through the database it owns, the way postgres-sep provisions
+# the sep role as the local superuser: Grafana exposes no credential-free path to
+# create a service account, and pmm-server holds no Grafana admin credential once an
+# operator has changed the admin password.
+#
+# psql only interpolates :'var' when reading a script, hence stdin rather than -c.
+# \gset is what carries the org and service-account ids between statements; the
+# ON CONFLICT clauses target UQE_user_login and UQE_org_user_org_id_user_id, and
+# DO UPDATE rather than DO NOTHING is what makes RETURNING fire on re-provisioning.
+provision_service_account() {
+    local hashed="$1" salt="$2" rands="$3" uid="$4"
+
+    psql_grafana -v sa_name="$SA_NAME" -v token_name="$TOKEN_NAME" -v hashed="$hashed" \
+        -v salt="$salt" -v rands="$rands" -v uid="$uid" <<'SQL'
+BEGIN;
+
+SELECT id AS org_id FROM org ORDER BY id LIMIT 1
+\gset
+
+INSERT INTO "user" (version, login, email, name, password, salt, rands, company, org_id,
+                    is_admin, email_verified, theme, created, updated, help_flags1,
+                    is_disabled, is_service_account, uid, is_provisioned)
+VALUES (0, 'sa-' || :'org_id' || '-' || :'sa_name', 'sa-' || :'org_id' || '-' || :'sa_name',
+        :'sa_name', '', :'salt', :'rands', '', :'org_id',
+        false, false, '', now(), now(), 0, false, true, :'uid', false)
+ON CONFLICT (login) DO UPDATE SET updated = now(), is_disabled = false
+RETURNING id AS sa_id
+\gset
+
+INSERT INTO org_user (org_id, user_id, role, created, updated)
+VALUES (:'org_id', :'sa_id', 'Admin', now(), now())
+ON CONFLICT (org_id, user_id) DO UPDATE SET role = 'Admin', updated = now();
+
+-- Also matched on (org_id, name) so a token orphaned from an earlier account cannot
+-- collide with UQE_api_key_org_id_name and abort the insert below.
+DELETE FROM api_key
+WHERE service_account_id = :'sa_id' OR (org_id = :'org_id' AND name = :'token_name');
+
+-- Grafana stores Viewer in api_key.role for every service-account token whatever role
+-- the request asked for; the effective authority is org_user.role above.
+INSERT INTO api_key (org_id, name, key, role, created, updated, service_account_id, is_revoked)
+VALUES (:'org_id', :'token_name', :'hashed', 'Viewer', now(), now(), :'sa_id', false);
+
+COMMIT;
+SQL
+}
+
+# Written through a temporary file so SEP never reads a half-written secret.
+publish_token() {
+    local token="$1" name tmp
+
+    for name in "${TOKEN_FILES[@]}"; do
+        tmp=$(mktemp "$SECRETS_DIR/.${name}.XXXXXX")
+        printf '%s' "$token" > "$tmp"
+        chmod 0640 "$tmp"
+        mv "$tmp" "$SECRETS_DIR/$name"
+    done
+}
+
+remove_token_files() {
+    local name
+
+    for name in "${TOKEN_FILES[@]}"; do
+        rm -f "${SECRETS_DIR:?}/$name"
+    done
+}
+
+if ! is_enabled "$PMM_ENABLE_SEP"; then
+    remove_token_files
+    exit 0
+fi
+
+if is_enabled "$PMM_HA_ENABLE" || is_enabled "$PMM_DISABLE_BUILTIN_POSTGRES" ||
+    [ -n "$GF_DATABASE_URL" ] || [ -n "$GF_DATABASE_HOST" ]; then
+    echo "WARNING: ignoring PMM_ENABLE_SEP, Grafana is not using the embedded PostgreSQL." >&2
+    exit 0
+fi
+
+if ! install -d -m 0750 "$SECRETS_DIR" 2> /dev/null || [ ! -w "$SECRETS_DIR" ]; then
+    echo "FATAL: $SECRETS_DIR is not writable for $(id -un)." >&2
+    echo "Please make sure the volume mounted there is owned by uid $(id -u) and gid $(id -g) and try again." >&2
+    exit 1
+fi
+
+if ! wait_for_grafana; then
+    echo "FATAL: Grafana did not become ready within ${READY_TIMEOUT}s, cannot provision the SEP service account." >&2
+    exit 1
+fi
+
+declare TOKEN=""
+if [ -s "$SECRETS_DIR/${TOKEN_FILES[0]}" ]; then
+    TOKEN=$(< "$SECRETS_DIR/${TOKEN_FILES[0]}")
+fi
+
+if [ -n "$TOKEN" ] && grafana_authenticates "$TOKEN"; then
+    echo "The Grafana service account token for SEP is still valid, keeping it."
+    publish_token "$TOKEN"
+    exit 0
+fi
+
+echo "Provisioning the Grafana service account for SEP..."
+
+declare MATERIAL
+MATERIAL=$(generate_token_material)
+
+declare -a PARTS
+mapfile -t PARTS <<< "$MATERIAL"
+if [ ${#PARTS[@]} -ne 5 ]; then
+    echo "FATAL: could not generate the Grafana service account token." >&2
+    exit 1
+fi
+
+TOKEN="${PARTS[0]}"
+provision_service_account "${PARTS[1]}" "${PARTS[2]}" "${PARTS[3]}" "${PARTS[4]}"
+
+# Validated before publishing so a half-working credential never reaches SEP, and so a
+# future change to Grafana's schema or token encoding fails loudly here instead of
+# silently breaking SEP's authentication.
+if ! grafana_authenticates "$TOKEN"; then
+    echo "FATAL: Grafana rejected the service account token just provisioned for SEP." >&2
+    exit 1
+fi
+
+publish_token "$TOKEN"
+echo "Published the Grafana service account token for SEP to $SECRETS_DIR."
--- a/build/ansible/roles/grafana/files/grafana-sep
+++ b/build/ansible/roles/grafana/files/grafana-sep
@@ -25,7 +25,9 @@ declare GRAFANA_URL="http://127.0.0.1:3000"
 declare POSTGRES_BIN_DIR="${POSTGRES_BIN_DIR:-/usr/pgsql-14/bin}"
 declare SA_NAME="pmm-sep"
 declare TOKEN_NAME="pmm-sep"
-declare READY_TIMEOUT=300
+# Generous because a failed run is not retried: the one-shot cannot bound its own
+# restarts under supervisord, so waiting longer here is what absorbs a slow first boot.
+declare READY_TIMEOUT=600
 declare READY_INTERVAL=5
 declare -a TOKEN_FILES=(AUTH__PROVIDER__GRAFANA__SERVICE_ACCOUNT_TOKEN PMM__API_KEY)
 declare TMP_FILE=""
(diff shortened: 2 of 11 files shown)


build/ansible/roles/grafana/files/grafana-sep
```diff
@@ -0,0 +1,249 @@
...
+# -H so the token never appears in any process's argv.
+grafana_authenticates() {
+    local code
+
+    code=$(printf 'header = "Authorization: Bearer %s"\n' "$1" |
+        curl -sS -K - --max-time 10 -o /dev/null -w '%{http_code}' \
+            "$GRAFANA_URL/api/access-control/user/permissions") || return 1
+
+    [ "$code" = "200" ]
+}
```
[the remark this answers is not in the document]
yyyyyyyan: The token does reach the request. `printf 'header = "Authorization: Bearer %s"\n' "$1"` substitutes `%s` with the function's first argument, and `curl -K -` reads that line from stdin as a config directive — so the header is built from the passed token. The literal `%s` is the format placeholder, not a redacted value.

Captured against a local listener using the construction from line 69 verbatim:

```
GET /api/access-control/user/permissions HTTP/1.1
Host: 127.0.0.1:18099
User-Agent: curl/8.21.0
Accept: */*
Authorization: Bearer glsa_TESTSECRET_deadbeef
```

Feeding the header on stdin instead of through `-H` is the point of the construction: `printf` is a shell builtin, so the token never lands in any process's `argv` — which is what the comment on line 65 records, and what the secrecy check in the PR description was verifying.

Both call sites depend on this succeeding, so neither is silently dead: the keep-existing path at line 274 only logs "still valid, keeping it" on a 200, and the post-mint check at line 299 treats anything else as fatal and publishes nothing.

theTibi:
@coderabbitai full review

yyyyyyyan:
Note on encryption: the published token is plaintext on the volume, and has to be. SEP sends it to Grafana as a bearer credential, so it must read the raw value; encrypting the files would only move the secret to whatever key SEP could reach, which is the same volume. Grafana's own database never holds the token — only the PBKDF2 hash (`api_key.key`) it would have stored for an API-minted one.

The protection is therefore filesystem-only: `0640`, group 0 from the setgid directory. That is the posture PMM already ships for `/srv/.postgres_password` (`0600 pmm:root`) and `/srv/pmm-encryption.key`, one group-read wider so a second container can read it. Whoever can mount `pmm-sep-secrets`, or is root on the host, holds an Admin-role Grafana token — and that volume outlives both containers, so it also appears in volume backups.

cc @ademidoff

build/ansible/roles/grafana/files/grafana-sep
```diff
@@ -0,0 +1,276 @@
...
+  AND u.login = 'sa-' || ou.org_id || '-' || :'sa_name'
+  AND ou.role <> 'Admin';
+SQL
+}
+
+# Prints, one per line and in the order the caller unpacks them: the token value, its
+# api_key.key hash, and the user row's salt, rands and uid. The encoding reproduces
+# Grafana's own - crc32("glsa_" + secret) little-endian as the checksum,
+# util.EncodePassword(secret, checksum) as the stored hash.
+generate_token_material() {
```
ademidoff: We can't rely on the presence of Python in PMM, it won't be available in the future. Getting rid of Python is a tech debt item because of many concerns, where an increased attack surface is just one of them.

I suggest we code a little Go tool instead, if it's necessary at all (see below).

build/ansible/roles/grafana/files/grafana-sep
```diff
@@ -0,0 +1,276 @@
...
+}
+
+# Administers Grafana through the database it owns, the way postgres-sep provisions the
+# sep role as the local superuser: Grafana exposes no credential-free path to create a
+# service account, and once an operator changes the admin password pmm-server holds none.
+#
+# psql only interpolates :'var' when reading a script, hence stdin rather than -c. The
+# ON CONFLICT clauses target UQE_user_login and UQE_org_user_org_id_user_id, and DO
+# UPDATE rather than DO NOTHING is what makes RETURNING fire on re-provisioning.
+provision_service_account() {
```
ademidoff, line 104: We can't create tokens by directly writing to Grafana database. First, Grafana is known to frequently change the underlying implementation of its auth layer. Second, PMM itself uses an API client to do the same. This implementation seems to be fragile and hard to maintain.

Therefore, a much more straightforward way to do it would be to delegate it to PMM (managed), where token creation is already implemented.

build/ansible/roles/grafana/files/grafana-sep
```diff
@@ -0,0 +1,276 @@
...
+-- the request asked for; the effective authority is org_user.role above.
+INSERT INTO api_key (org_id, name, key, role, created, updated, service_account_id, is_revoked)
+VALUES (:'org_id', :'token_name', :'hashed', 'Viewer', now(), now(), :'sa_id', false);
+
+COMMIT;
+SQL
+}
+
+# Written through a temporary file so SEP never reads a half-written secret.
+publish_token() {
```
ademidoff, line 143: We have a number of tickets where the ask is to stop storing the keys (or secrets) in plain text on disk, which is why doing this way only increases the tech debt.

Let's consider storing it in PMM Settings instead.

ademidoff (changes requested):

ademidoff:
WRT branch naming: let's try to give a meaningful name to each branch. This really helps both people and AI to skim through the branch names without opening Jira tickets just to get an idea about the goal. For example, this branch could be named `PMM-15280-provision-sa-for-sep` or a longer one would also be fine - `PMM-15280-provision-service-account-for-sep`.

It's actually missed in the [CONTRIBUTING](https://github.com/percona/pmm/blob/main/CONTRIBUTING.md) guide, I will note down to mention it there.

build/ansible/roles/grafana/files/grafana-sep
yyyyyyyan: Replaced in dd230bf17 with the Go tool you suggested: `managed/cmd/pmm-sep-token`. It falls out of `make -C managed release` with the rest of `./cmd/...` and installs to `/usr/sbin` alongside `pmm-managed-init` and `pmm-encryption-rotation`, so the packaging change is two spec lines. Output encoding is unchanged — same checksum and same `api_key.key` hash for a given secret.

build/ansible/roles/grafana/files/grafana-sep
yyyyyyyan: That path isn't reachable from here as things stand. `Client.CreateServiceAccount` opens by pulling the caller's credential out of the request context:

https://github.com/percona/pmm/blob/bacbbc7c9f811cc4b3f097ab70713ddd162906ea/managed/services/grafana/client.go#L661-L665

and that helper reads the *incoming* gRPC metadata, refusing the call outright when there is no Authorization or cookie header:

https://github.com/percona/pmm/blob/bacbbc7c9f811cc4b3f097ab70713ddd162906ea/managed/utils/auth/token.go#L55-L65

`createServiceAccount`, `createServiceToken` and every other Grafana call in that client do the same. It works for pmm-agent registration because an authenticated operator's credential is being forwarded through the request — pmm-managed acts on the caller's behalf, never its own.

At container start there is no inbound request to borrow from, and pmm-managed holds no Grafana credential of its own; nothing under `managed/` reads `GF_SECURITY_ADMIN_PASSWORD`. Delegating therefore means first giving pmm-managed a standing Grafana admin credential — the dependency this ticket set out to remove, and one that breaks again the moment an operator rotates that password.

The one credential-free local path Grafana does offer is [`grafana cli`](https://github.com/percona/pmm/blob/bacbbc7c9f811cc4b3f097ab70713ddd162906ea/build/ansible/roles/grafana/files/change-admin-password#L5), which we already use — but its only relevant subcommand resets the admin password, so it would clobber the operator's.

The fragility you're pointing at is real wherever the code lives. What bounds it here is that the freshly minted token is validated against Grafana's own endpoint before anything is published:

https://github.com/percona/pmm/blob/dd230bf17cce708cb0667ce2f12dcbcf1b23320d/build/ansible/roles/grafana/files/grafana-sep#L244-L250

so a change to Grafana's auth schema or token encoding fails at provisioning time rather than silently breaking sign-in later. That bounds the damage; it doesn't make the coupling good.

So the question I'd put back to you: do you want pmm-managed to hold a standing Grafana credential? That is a larger change than this PR and I'd rather scope it separately, but it is the only thing that unlocks the delegation you're describing.

build/ansible/roles/grafana/files/grafana-sep
yyyyyyyan: The token has to be readable by SEP, which runs as a separate container, before SEP can authenticate to anything. That is what rules out PMM Settings specifically: Settings is read through the PMM API, and the credential that authenticates to the PMM API is this token. SEP would have to hold the token in order to fetch the token.

Encryption has the same shape one level down. SEP has to decrypt what it reads, so the key has to reach SEP, and the only channel that reaches SEP is the volume the token is already on. That relocates the secret rather than protecting it, and it protects nothing against the reader that matters, since whoever can read the volume gets both halves.

On the mode: these land at `0640` with group 0 from the setgid directory, already tighter than the file guarding every encrypted secret in PMM today:

https://github.com/percona/pmm/blob/bacbbc7c9f811cc4b3f097ab70713ddd162906ea/managed/utils/encryption/encryption.go#L197-L199

`/srv/.postgres_password` is `0600`; these are one group-read wider, which is exactly what lets a second container read them at all.

What is true, and worth stating plainly: whoever can mount that volume, or is root on the host, holds an Admin-role Grafana token, and it will appear in volume backups. That is a property of handing a credential across a container boundary at all, not of writing it as a file — any channel SEP can read unattended has the same reader set. If there is one with a smaller reader set that still closes in a single `compose up`, I'll take it.

build/ansible/pmm-docker/post-build.yml
```diff
@@ -67,6 +67,20 @@
...
 
+    # Must exist in the image: a named volume mounted at a path the image lacks is
+    # created root:root, which pmm-server cannot write to. Group-writable like the
+    # other /srv directories, so an arbitrary runtime uid in group 0 can still write.
+    # Setgid so the token files written here belong to group 0 whatever that uid's
+    # primary group is - the SEP container joins group 0 to read them. Docker carries
+    # the bit into a fresh named volume mounted at this path.
+    - name: Create the SEP secrets mountpoint
+      file:
+        path: /srv/sep-secrets
```
ademidoff, line 78: We can have a more generic name for a dedicated directory, so we don't end up creating `/srv/sep-x`, `/srv/sep-y` - it is much easier to maintain when you have one.

How about `/srv/sep`? We could put the token in there.

build/ansible/roles/grafana/files/grafana-sep
ademidoff: Agreed, we'll come back to it in the next phase.

yyyyyyyan:
Closing this. Provisioning the service account by writing Grafana's `user`, `org_user` and `api_key` rows directly is the wrong shape, and [the thread on that](https://github.com/percona/pmm/pull/5755#discussion_r3817092857) is where it became clear.

The reasoning that led here held together under one premise, and the premise was the thing worth questioning. The ticket's first acceptance criterion scoped the minting to pmm-server, "using no credential from outside its own trust boundary". Under that constraint Grafana really does offer nothing: `grafana cli admin` covers only password reset and migrations, file provisioning covers datasources, dashboards, alerting, access-control and plugins but not accounts, and `POST /api/serviceaccounts` needs an Admin credential. The database write follows from it. What went unasked is why pmm-server had to be the one minting — SEP is a long-running service, so it can wait for Grafana, retry, and call the same API PMM already uses for its own service accounts:

https://github.com/percona/pmm/blob/27b145b7891b335a32eebf0ecf1b0cb7cafca840/managed/services/grafana/client.go#L880-L930

That dissolves the constraint rather than engineering around it.

**What replaces it.** SEP mints its own token through `POST /api/serviceaccounts`, authenticating with `GF_SECURITY_ADMIN_USER` / `GF_SECURITY_ADMIN_PASSWORD` and falling back to the `admin`/`admin` default, retrying with backoff until Grafana answers. It stores the token, which stays valid across admin-password changes. No PMM-side code.

**One limitation worth recording for that work**, alongside the known case where the admin user bootstrap is disabled: the admin credential is needed whenever SEP holds no *valid* token, which is not only first boot. If Grafana's volume is wiped or restored to a state predating the service account, the stored token dies and SEP has to mint again — and if the admin password was changed through `change-admin-password` rather than through the environment, the value SEP holds is stale and the mint fails. The same applies on a first boot where the password is changed before SEP has minted. Setting `GF_SECURITY_ADMIN_PASSWORD` and giving SEP the same value avoids both; relying on the default means SEP can only mint until the password is first changed.

**What survives, and where it moved.** The secrets directory itself — #5762 now owns it. SEP's `SECRET_KEY` is generated by PMM and the database password is the one `postgres-sep` provisions, so neither can be minted by the consumer and both still travel PMM → SEP as files. The mode and group discussion on [the storage thread](https://github.com/percona/pmm/pull/5755#discussion_r3817105859) applies to that channel rather than to this one.

**Loose ends.** Consolidating on `/srv/sep` ([here](https://github.com/percona/pmm/pull/5755#discussion_r3822495764)) goes to #5762, which is the PR that creates the directory now. #5768 gates the container healthcheck on provisioning completing, and will be reworked to wait on the one provisioner that remains.

marcuscruz-percona:
Really enjoyed reading this one. The verification matrix in the description is unusually thorough, and the two design points — supervisord one-shot rather than entrypoint, and validate-before-replace rather than unconditional re-mint — are both the right calls and well argued. The mechanism gate against percona-grafana 12.4.5 in particular is the kind of evidence that makes a schema-coupled change reviewable at all.

I reviewed the diff against `PMM-15238-expose-pg-to-sep` on two axes: repo conventions, and PMM-15280's acceptance criteria. Verified `AC1`, `AC3`, `AC4`, `AC5` and `AC7` hold against the code — the `ON CONFLICT` + `DELETE` pair really does bound both rows to one, `publish_token`'s mktemp(0600) → chmod 0640 → mv is correct and picks up group 0 from the setgid directory, and the token reaches curl through a `printf` builtin piped to `-K -`, so it is in no process's argv and no environment.

Two things I would like to see addressed before merge, then some smaller ones.

## Must-fix

**1. `managed/services/server/logs_test.go` will fail once this runs in the container.**

`api-tests/server/logs_test.go` was updated, but there is a second expected-files list in the Go unit test. `logs.go:147` globs `/srv/logs/*.log`, `sep.ini` writes `/srv/logs/sep-provision.log`, and `commonExpectedFiles` at `managed/services/server/logs_test.go:36` does not list it, while line 197 asserts equality:

```go
sort.Strings(actual)
assert.Equal(t, commonExpectedFiles, actual)
```

This passes on a dev machine because there is no `/srv/logs` to glob, so it only shows up under `make env`. Adding `"sep-provision.log"` to that slice, in sorted position, should be all it needs.

**2. The published files' group is never verified, and one path produces the wrong one silently.**

The permissions model is right, and `post-build.yml` shipping `/srv/sep-secrets` as `pmm:root 02770` is exactly what makes a fresh named volume inherit group 0. The gap is the fallback:

```bash
if [ ! -d "$SECRETS_DIR" ]; then
    install -d -m 2770 "$SECRETS_DIR" 2> /dev/null || true
fi
```

`install -d` takes the creator's effective gid, not 0, unless the parent is itself setgid. So if the directory arrives from anywhere other than the image mountpoint — a bind mount, or a volume created at that path before this change — the files land in pmm's primary group. The `-w` check on the next line still passes, so provisioning reports success and SEP gets `EACCES` even with `group_add: ["0"]` set correctly on its side. Given `AC2` names readability by the SEP runtime user, a `stat -c %g "$SECRETS_DIR"` check that warns or FATALs on a non-zero gid would close it cheaply, and would also catch a future change to the post-build task.

## Smaller things

**3. `PMM_SEP_SECRETS_DIR` is not registered in the env-var parser.** `#5700` added `case "PMM_ENABLE_SEP", "PMM_SEP_POSTGRES_PASSWORD"` at `managed/utils/envvars/parser.go:123` so these names do not reach the `default:` branch, which appends `"unknown environment variable " + env`. This third name is not in that list. It is only a log warning and the value is a path rather than a secret, so nothing leaks — but since the override has no caller and the ticket makes the directory the contract rather than a knob, dropping `PMM_SEP_SECRETS_DIR` entirely would resolve this and remove an untested code path in one edit.

**4. The SEP-capable guard now exists twice, with different conditions.** `entrypoint.sh:158` warns on `PMM_HA_ENABLE || PMM_DISABLE_BUILTIN_POSTGRES`; `grafana-sep:199-201` adds `GF_DATABASE_URL` and `GF_DATABASE_HOST`. Warning rather than failing is consistent with the entrypoint, so no concern there — but the two conditions have already drifted, and `PMM-15316` will presumably need the same predicate a third time. Worth extracting while there are only two copies.

**5. A killed run can leave the token behind.** `trap cleanup EXIT` does not fire on an untrapped `SIGTERM`, so a `docker stop` landing between the `mktemp` and the `mv` in `publish_token` leaves a `.AUTH__PROVIDER__GRAFANA__SERVICE_ACCOUNT_TOKEN.XXXXXX` file at 0600 group 0 in the directory SEP mounts. `remove_token_files` iterates `TOKEN_FILES`, so the disabled path will not clear it either, and the token stays readable after `PMM_ENABLE_SEP` is unset. Either glob the dotfiles in cleanup, or mktemp outside the shared directory and `mv` in.

**6. `AGENTS.md` / `build/AGENTS.md`.** The "Maintaining This Document" trigger list includes changes to `docker-compose.yml`, and this also adds a supervisord program and a role file. Worth a line in each.

## Notes, not asks

- `READY_INTERVAL` is doing two jobs — the Grafana readiness poll interval at line 27 and `grafana_auth_status`'s retry backoff at line 77. They happen to want the same 5s today; a second constant would keep them independent.
- `provision_service_account "${PARTS[1]}" "${PARTS[2]}" "${PARTS[3]}" "${PARTS[4]}"` puts the field mapping in a comment only. Not worth restructuring a 5-line contract over, just noting it as the spot a future reader misreads.
- `is_enabled` is now defined identically in `entrypoint.sh`, `postgres-sep` and `grafana-sep`. Probably the correct trade for three standalone scripts with no shared lib, but the third copy dropped the comment the first one carries.
- `AC6`'s "nothing about stock PMM changes" is not literally met — the `sep-provision` program, the image directory, the unconditional compose volume and the new archive entry. Your Known limitations section already enumerates all four, which I think is the right resolution; flagging only so it is a deliberate call on the record rather than an oversight.
- `PMM_ENABLE_SEP` is absent from `documentation/docs/` on the base branch too, so `AC8`'s documentation requirement is really a gap for the series rather than something this PR introduced. The `.env.example` block covers the operator-facing contract well in the meantime.

Two things I checked and want to explicitly withdraw as concerns, in case they come up from another reviewer. The `POSTGRES_BIN_DIR:-/usr/pgsql-14/bin` default is correct here and should not be changed to match `postgres-sep`'s `:?` form: `entrypoint.sh:148-155` scopes that export to a subshell on purpose, so a supervisord child cannot inherit it. And `sep.ini` omitting `stopsignal`/`stopwaitsecs` is right for a one-shot that has already exited, notwithstanding the three siblings that set them.

The Feature Build item still stands as the blocker for the image-path rows, per your own checklist.

PR #5765, opened 2026-08-12
[DOCS] PMM-15302 Show ssh key tab only in AMI deployment

matejkubinec:
Ticket number: PMM-15302

Documentation for https://github.com/percona/pmm/pull/5744


catalinaadam (approved):

documentation/docs/install-pmm/install-pmm-server/deployment-options/virtual/ssh.md
```diff
@@ -13,8 +16,6 @@ To configure SSH access:
 2. Enter your public key in the **SSH Key** field.
 3. Click **Apply changes**.
```
travagliad, line 19: Should we mention that applying changes here will replace the existing SSH Key? I think it's important to note that only 1 SSH Key is allowed to exist, and that applying changes here won't disconnect whoever is connected, but it might be risky because applying a wrong SSH Key will remove the previous working one, so users should keep a working session until they confirm new key works.

documentation/docs/install-pmm/install-pmm-server/deployment-options/virtual/ssh.md
```diff
@@ -13,8 +16,6 @@ To configure SSH access:
 2. Enter your public key in the **SSH Key** field.
```
travagliad:
```suggestion
2. Enter your public key in the **SSH key** field.
```

documentation/docs/install-pmm/install-pmm-server/deployment-options/virtual/ssh.md
matejkubinec: @catalinaadam Davi made a good point, updated the file with mention about the singular ssh key + warning

travagliad (approved):

PR #5767, opened 2026-08-12
PMM-15308 Trim whitespace from dashboard titles

theTibi:
Ticket number: PMM-15308

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/4525

## Summary

- **Bug**: panel and dashboard titles carried leading/trailing whitespace — the reported `"InnoDB Buffer Pool Data "`, plus **38 more across 14 dashboards** (e.g. `Nodes_Overview.json` has 8 *leading*-space titles like `" Top Disk Latency"`).
- **Root cause**: nothing validated titles. The `Dashboards → Check` job already runs `cleanup-dash.py --check-only`, but its `CLEANUPERS` list only normalised `editable`/`time`/`timezone`/`refresh`/`id`, so untrimmed titles always passed.
- **Fix**: trim the 39 titles, and add a `trim_titles` cleanuper so the existing CI gate catches regressions — the trim-on-publish the reporter asked for.

## Notes for reviewers

**1. The `--check-only` reporter needed a `title:` case too.** It hardcodes per-field comparisons, so adding `trim_titles` alone would have made CI exit 1 while printing *no reason*. `walk_titles` is shared by the cleanuper and the reporter so the two cannot drift.

**2. Titles were trimmed in place, not via `cleanup-dash.py` write mode.** `--check-only` compares *parsed objects*, while write mode re-serialises with `sort_keys`/`indent=4`. Running it over the tree would reformat **6 unrelated dashboards wholesale** (~26k lines of indentation churn): `MongoDB_Cluster_Summary`, `MongoDB_Collections_Overview`, `MongoDB_ReplSet_Summary`, `PMM_Health`, `PostgreSQL_Instance_Summary`, `PostgreSQL_Top_Queries`. That drift is pre-existing and left alone here — worth its own ticket, since the CI error message points devs at exactly that command.

**3. One behaviour-visible change.** `MySQL_User_Details.json` `panels[0]` is a `text` spacer whose title is literally `" "`; it becomes `""`. Confirmed as intended.

**4. First Python tests in the repo.** `test_cleanup_dash.py` is stdlib `unittest` — no new dependency. Happy to drop it and the CI step if you'd rather not set the precedent.

**5. The whole-tree gate, and how it is scoped.** `TestRepoDashboards` runs `--check-only` over all 76 dashboards, so a dirty dashboard fails the job no matter which file the PR touched. That breadth is deliberate — it is what forces a whole-tree fix when a cleanuper is added, exactly as this PR had to do — but the workflow triggers on `paths: dashboards/**`, so left unscoped it would gate unrelated PRs (a `pmm-app` change, a `yarn.lock` bump) on every dashboard in the tree. The `Test the cleanup script` step therefore runs only when the PR changes a dashboard JSON file, `cleanup-dash.py` / `test_cleanup_dash.py`, or `dashboards.yml` itself — whoever edits the gate should see it run, and that is the one place the widened scope is deliberate. It runs **after** the per-dashboard check so a whole-tree failure never hides the step summary and `::error file=` annotations for the PR's own dashboards. On failure it prints the tool's own per-file report — every dirty field as `"raw" -> "cleaned"` — and sorts failures into three buckets with different remedies: *needs cleanup* (run the cleanuper), *failed without naming a field* (a cleanuper has no case in the `--check-only` reporter; write mode still fixes the file), and *could not be checked at all* (the tool crashed; the cleanuper hits the same error). Each bucket is capped at 10 entries so a systemic fault does not bury the explanation. The trigger matches any `dashboards/misc/test_*.py`, so a PR adding a test module actually runs it. `AGENTS.md` and `dashboards/dashboards/AGENTS.md` document the new command.

**6. Drive-by: the `check` job was fail-open.** `git diff … | sort -u > changed_dashboards.txt || true` returns `sort`'s exit status, so an unresolvable base SHA yielded an empty file and the job reported success with no dashboard checked at all. It now fails with an annotation. Pre-existing, fixed here because note 5 doubled the blast radius (two skipped checks instead of one). The workflow also now triggers on changes to itself — before this, a PR fixing only `dashboards.yml` merged without ever running it.

## Test plan

- [x] `cleanup-dash.py --check-only` fails on exactly 14 dashboards before the trim, passes on **all 76** after
- [x] All 76 dashboards still parse as JSON; **0** untrimmed titles remain
- [x] Diff is title-only — **0** changed dashboard lines that are not `"title":`
- [x] 20 tests pass in a clean `ubuntu:24.04` / Python 3.12.3 container (mirrors CI runner), no pip installs; also green on Python 3.9.6 / 3.13.7 / 3.14.3, so the suite is not tied to the runner's minor version
- [x] **Negative-tested**: reverting the tool → 12 errors + 4 failures; reverting the dashboards → fails listing all 39 titles
- [x] Failure paths verified in throwaway repos — crash (missing `time`, malformed JSON, non-dict root), cleanuper with no reporter case incl. unrelated stderr noise, missing dashboard tree, mixed dirty+broken, and the printed remediation command actually runs and makes the gate pass: **19/19 assertions**, plus a second harness of **24** covering the three-bucket split, exit-code reporting, output capping and per-field reporter coverage (positive *and* negative probes)
- [x] The `check` job's changed-files step executed verbatim against a real / bogus / empty base SHA → **0 / 1 / 1**, each failure carrying a `::error::` annotation; the `tooling` output resolves to **1** for a dashboard / `cleanup-dash.py` / `test_*.py` / `dashboards.yml` change and **0** for a `pmm-app`-only or lockfile-only change
- [x] Write-mode round-trip clean; `plugin.json`'s 68 dashboard paths all resolve
- [ ] Not run: live PMM server render check — change is a pure whitespace trim with no query or schema impact

## Related

- Follow-up candidate: normalise the 6 reformat-prone dashboards so `cleanup-dash.py <file>` is safe to run as CI suggests.


dashboards/misc/test_cleanup_dash.py
```diff
@@ -0,0 +1,192 @@
...
+    def dashboards(self):
+        for dp, _, fns in os.walk(DASH_DIR):
+            for fn in sorted(fns):
+                if fn.endswith('.json'):
+                    yield os.path.join(dp, fn)
+
+    def test_all_dashboards_pass_check_only(self):
+        failures = [p for p in self.dashboards()
+                    if run_cli('--check-only', p).returncode != 0]
+        self.assertEqual(failures, [], f'{len(failures)} dashboard(s) fail the gate')
```
mattiasimonato: This test quietly changes the CI policy. The step in `.github/workflows/dashboards.yml:105` checks only the dashboards a PR touched, on purpose. This one checks all 76, so from now on any PR that touches `dashboards/**` (the QAN plugin, a lockfile bump) is gated on every dashboard in the tree.

I'm fine with that as a policy, but two problems: it isn't mentioned in the PR description, and the test only keeps the exit code, so a failure prints `['path/a.json'] != []` with no reason and none of the "run this command" hint the real step gives. And per your own note 2, that command reformats 6 dashboards wholesale.

`test_no_untrimmed_titles_remain` right below already gives you the repo-wide title guard, in process, with a clear message. I'd drop this one, or print the stdout of the failures.

mattiasimonato (approved):
One comment below on the repo-wide test. Not blocking, but I'd like it sorted before merge.

matejkubinec (approved):

dashboards/misc/test_cleanup_dash.py
theTibi: Fixed in 96188794f — kept the test, took both of your points.

**Scoping.** The suite no longer runs on every `dashboards/**` PR. `Get changed dashboard files` now sets a `tooling` output, and `Test the cleanup script` runs only when the PR changes a dashboard JSON, `cleanup-dash.py` / `test_*.py`, or `dashboards.yml` itself. A QAN plugin change or a lockfile bump is not gated on the tree any more. Two related bits: the step moved *after* `Run cleanup check per dashboard`, so a whole-tree failure can't bury the step summary and `::error file=` annotations for the PR's own dashboards; and I added `dashboards.yml` to the trigger paths plus the `tooling` pattern, because whoever edits the gate should see it run — that is the one place the widened scope is deliberate.

**The failure message.** It keeps the tool's own per-file report now, and splits the three failure kinds because they need different remedies:

```
1 of 77 dashboard(s) need cleanup:
Dashboard: dashboards/dashboards/Insight/Advanced_Data_Exploration.json
  title: "  Dirty Title  " -> "Dirty Title"
Fix the reported fields with `python3 dashboards/misc/cleanup-dash.py <file>`, then review
the diff: write mode re-serialises the whole file, so some dashboards pick up unrelated
reformatting.
1 of 77 dashboard(s) could not be checked at all:
Dashboard: dashboards/dashboards/Insight/broken.json
  json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2
The cleanuper cannot fix these -- write mode hits the same error. Repair the dashboard JSON,
or harden cleanup-dash.py.
```

That is real output from a dirtied copy of the tree, not a mock-up. Your point about note 2 is why the remediation line carries the reformatting caveat rather than just the command. A third bucket covers "exited non-zero but named no field", i.e. a cleanuper with no matching case in the `--check-only` reporter — write mode still fixes those, so it does not send anyone JSON-hunting. Each bucket caps at 10 dashboards with `... and N more`, since a systemic fault would otherwise bury the one line that says what to do.

I kept `test_all_dashboards_pass_check_only` alongside `test_no_untrimmed_titles_remain` because they fail on different things: the in-process one only sees titles, while this one runs the real CLI and catches any cleanuper regression, a broken reporter case, or a crash. Both now name the offending file — parse errors in the in-process tests were raising a path-less `JSONDecodeError`, fixed with `subTest`.

**Description.** Notes 5 and 6 now spell out the whole-tree gate, how it is scoped, and the drive-by fix to the fail-open `|| true` in the changed-files step (it returned `sort`'s status, so an unresolvable base SHA meant a green job with nothing checked).

Verified: 20/20 pass on the current tree; the changed-files step run verbatim against a real / bogus / empty base SHA gives 0 / 1 / 1 with annotations; the gate resolves `tooling=1` for this commit and `0` for a `pmm-app`-only or lockfile-only change.

PR #5781, opened 2026-08-17
PMM-15019 Add agent-env-vars to inventory change agent mongodb-exporter.

JiriCtvrtka:
PMM-15019

FB: https://github.com/Percona-Lab/pmm-submodules/pull/4574

Adds --agent-env-vars to pmm-admin inventory change agent mongodb-exporter. Until now the env
var names could only be set when adding the agent. New API field:
ChangeMongoDBExporterParams.environment_variable_names.

Omitting the flag keeps the current list, --agent-env-vars="" clears it, passing names replaces
the whole list.

Behavior change: add mongodb used to store the names on all four MongoDB agents, now only on
mongodb_exporter. Only agents that pmm-agent starts as a separate process can receive env
vars — the QAN and RTA agents run inside pmm-agent and never read those rows, so it was dead
data. Existing rows are untouched.

Also in here, same mechanism, all small:

MONGODB_URI is rejected with an error at request time in AddMongoDBExporter, ChangeMongoDBExporter, and add mongodb; pmm-agent also skips any name it sets itself as a second line of defense, since it could otherwise override the DSN
--agent-env-vars added to inventory add agent mongodb-exporter (the API field already existed)
duplicate names are collapsed
ToAPIAgent never filled MongoDBExporter.environment_variable_names — bug since PMM-14375
lowercase names (e.g. https_proxy) are now accepted, not just [A-Z_][A-Z0-9_]*
Grandfathering: environment_variable_names is a full-replace field — a caller changing one
unrelated setting must resend the whole list. Because of that, a name already stored before a rule
existed (or under since-tightened rules — the MONGODB_URI reservation, the character/prefix
rules) is let through unchanged on ChangeMongoDBExporter/ChangeAgent, instead of blocking the
whole update. Only the exact, already-stored value is grandfathered (case-sensitively for the
MONGODB_URI check — mongodb_uri and MONGODB_URI are different env vars); submitting a
different reserved/invalid name is still rejected. This check reads the agent's row locked
inside the same transaction as the update, so it can't be raced by a concurrent change to the same
agent.

Validation and secret handling (server-side, not just pmm-admin): names are now validated by
the API itself via a shared utils/envvars policy — same character rule, plus a count/length
bound and a PMM_AGENT_ prefix denylist, since that prefix is pmm-agent's own configuration and
secrets namespace (e.g. PMM_AGENT_SERVER_PASSWORD) and must never be resolvable for pass-through
to an exporter. Values resolved from pmm-agent's own environment are also now fully redacted in
debug logs, rather than only masking DSN-shaped values. Rejected names consistently return a
gRPC InvalidArgument status, not an unadorned error.


JiriCtvrtka:
@copilot review

admin/commands/inventory/add_agent_mongodb_exporter.go
```diff
@@ -75,12 +76,18 @@ type AddAgentMongodbExporterCommand struct {
...
 }
 
 // RunCmd executes the AddAgentMongodbExporterCommand and returns the result.
 func (cmd *AddAgentMongodbExporterCommand) RunCmd() (commands.Result, error) {
 	customLabels := commands.ParseKeyValuePair(&cmd.CustomLabels)
 
+	agentEnvVars, err := commands.ValidateEnvironmentVariableNames(cmd.AgentEnvVars)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
JiriCtvrtka: Declining. wrapcheck is deliberately disabled in .golangci.yml ("we do not use wrapping everywhere"), and the errors from ValidateEnvironmentVariableNames/ParseEnvironmentVariableNames already name both the offending value and the rule — wrapping produces "failed to validate environment variable names: invalid environment variable name: X", which stutters in CLI output. The adjacent commands.ReadFile calls in the same function also return bare, so wrapping only these two would be inconsistent. The TLS-file reads in the change command are wrapped, because there the underlying error doesn't say which file failed — that's the distinction this codebase applies.

utils/envvars/envvars.go
```diff
@@ -0,0 +1,104 @@
...
+	result := make([]string, 0, len(names))
+	seen := make(map[string]struct{}, len(names))
+
+	for _, name := range names {
+		name = strings.TrimSpace(name)
+
+		if _, ok := grandfathered[name]; !ok {
+			if err := ValidateName(name); err != nil {
+				return nil, err
+			}
```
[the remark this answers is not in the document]
JiriCtvrtka: Is it resolved?

agent/agents/supervisor/supervisor.go
```diff
@@ -844,19 +845,58 @@ func (s *Supervisor) processParams(agentID string, agentProcess *agentv1.SetStat
...
-			s.l.Warnf("Environment variable %s not found in pmm-agent environment for agent %s", varName, agentID)
+			l.Warn("Environment variable not found in pmm-agent environment.")
 			continue
 		}
 
-		processParams.Env = append(processParams.Env, fmt.Sprintf("%s=%s", varName, value))
-		s.l.Debugf("Resolved environment variable %s for agent %s", varName, agentID)
+		env = append(env, fmt.Sprintf("%s=%s", varName, value))
+		resolvedNames = append(resolvedNames, varName)
+		l.Debug("Resolved environment variable.")
```
ademidoff: This log statement doesn't seem to be useful inside of a loop. I'd just remove it, otherwise let's restore the one we had before this change.

documentation/docs/use/commands/pmm-admin/inventory.md
```diff
@@ -158,6 +161,32 @@ You can also use `pmm-admin list` to see agents alongside their services.
 - `--custom-labels`
 :   Custom user-assigned labels in `key=value,key=value` format
 
+- `--agent-env-vars` (`mongodb-exporter` only)
```
ademidoff: @JiriCtvrtka Let's not commit this, we need a separate PR against the base DOC branch.

ademidoff (commented):
Overall: I think the grandfathering machinery is more structure than the problem needs, and that it can collapse into `Agent.SetEnvironmentVariableNames`.

The argument: **every name grandfathering preserves is a name pmm-agent discards anyway.** Only two classes get through — `MONGODB_URI`, skipped by the supervisor's `reserved` map (`agent/agents/supervisor/supervisor.go:883`), and names failing `envvars.ValidateName` (dashes, `PMM_AGENT_*`), skipped at `agent/agents/supervisor/supervisor.go:876`. There is no third class. On top of that the path is unreachable from pmm-admin, which validates strictly with `grandfathered == nil` (`admin/commands/inventory/change_agent_mongodb_exporter.go:134`), so a legacy name fails client-side before a request is ever sent.

`Agent.SetEnvironmentVariableNames` already builds the grandfathered set from the row `ChangeAgent` reads inside the same transaction, and already knows `a.AgentType`. Moving the `MONGODB_URI` rule there would let you delete `managed/services/management/common/mongodb_env_vars.go`, `FindAgentByIDForUpdate`, `executeAgentChangeChecked`, the `SELECT ... FOR UPDATE`, and ~85 lines of hand-rolled test setup.

Independently of that redesign, there are five correctness issues worth fixing either way (marked **bug** below). The rest are nits.

managed/services/converters.go
```diff
@@ -329,6 +329,12 @@ func ToAPIAgent(q *reform.Querier, agent *models.Agent) (inventoryv1.Agent, erro
 		exporter.EnableAllCollectors = agent.MongoDBOptions.EnableAllCollectors
 		exporter.EnableDiagnosticDataHistograms = agent.MongoDBOptions.EnableDiagnosticDataHistograms
 
+		envVarNames, err := agent.GetEnvironmentVariableNames()
```
ademidoff, line 338: **bug** — this turns a previously-ignored field into a hard failure of `GetAgent` **and** `ListAgents`.

Before this change a mongodb_exporter row whose `environment_variables` column can't be JSON-decoded (hand-edited DB, failed migration, older writer) just left the field unset. Now `ToAPIAgent` returns `failed to unmarshal shared environment variable names`, so `GetAgent` fails for that agent and `ListAgents` fails for the **entire inventory** — the user loses the whole list because of one unrelated column on one row.

Worse, the row can't be repaired through the API either: the same read in `managed/models/agent_model.go:471` aborts before writing, so any non-empty new list fails too.

Suggest logging and treating an undecodable value as empty here, rather than propagating.

managed/services/management/common/mongodb_env_vars.go
```diff
@@ -0,0 +1,78 @@
...
+// It lives in this cross-service package (rather than services/inventory, where the check
+// originated) so both the inventory API and ManagementService.addMongoDB can apply the same check
+// without either service importing the other.
+func ValidateMongoDBExporterEnvVarNames(names []string, grandfathered map[string]struct{}) error {
+	for _, name := range names {
+		trimmed := strings.TrimSpace(name)
+		if _, ok := grandfathered[trimmed]; ok {
+			continue
+		}
+		if _, ok := mongoDBExporterReservedEnvVars[strings.ToUpper(trimmed)]; ok {
```
ademidoff, line 56: **bug** — the `strings.ToUpper` makes this check case-insensitive, which rejects names that are genuinely distinct and contradicts this file's own premise.

`--agent-env-vars=mongodb_uri` (or `Mongodb_Uri`) now fails with `environment variable name "mongodb_uri" is set by pmm-agent for mongodb_exporter` — which isn't true. pmm-managed sets only `MONGODB_URI` (`managed/services/agents/mongodb.go:68`), and pmm-agent's own guard is a case-**sensitive** map lookup (`agent/agents/supervisor/supervisor.go:883`), so `mongodb_uri` could never have collided.

The doc comment right above (lines 41-43) says a stored `mongodb_uri` "is a different OS environment variable from MONGODB_URI" — and then the check treats them as the same. Net effect: a valid pass-through name is unusable, with a misleading error. Drop the `ToUpper`.

managed/services/management/common/mongodb_env_vars.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/percona/pmm/managed/models"
+)
+
+// mongoDBExporterReservedEnvVars are environment variable names pmm-agent's supervisor always
+// sets itself for mongodb_exporter (see mongodbExporterConfig in managed/services/agents/mongodb.go).
+// A user-selected name here would never take effect: the supervisor skips it to avoid overriding
+// the computed value, silently, on the agent side. Rejecting it here instead gives the caller an
+// actionable error at request time.
+var mongoDBExporterReservedEnvVars = map[string]struct{}{
```
ademidoff, line 35: **bug** — `HTTP_AUTH` is missing from this list.

pmm-managed injects it into the mongodb_exporter environment for pmm-agent < 2.28.0: `ensureAuthParams` appends `HTTP_AUTH=pmm:<agent password>` at `managed/services/agents/agents.go:139`.

So today `--agent-env-vars=HTTP_AUTH` passes this check, is stored, is sent to pmm-agent, and is then silently dropped by the supervisor's `reserved` skip (`agent/agents/supervisor/supervisor.go:883`) — exactly the "user-selected name never takes effect, silently" failure this list exists to convert into an actionable request-time error. The comment above points only at `mongodbExporterConfig` and misses the second injection site.

utils/envvars/envvars.go
```diff
@@ -0,0 +1,105 @@
...
+
+		if _, ok := seen[name]; ok {
+			continue
+		}
+		seen[name] = struct{}{}
+
+		result = append(result, name)
+	}
+
+	if len(result) > MaxNames {
```
ademidoff: **bug** — `MaxNames` (and `MaxNameLength` at line 47) aren't grandfathered, so an agent that already stores more than 32 names can never have its list edited again.

Take an agent created before this PR with 40 stored names — there was no server-side limit then. The owner resends 39 of them to drop one: `NormalizeNamesAllowing` grandfathers each individual name past `ValidateName`, then this check rejects the whole request with `too many environment variable names: 39 (max 32)`. The only edit the server accepts is truncating to ≤32.

The grandfathering path the PR builds elsewhere doesn't cover the count and length bounds it introduces at the same time.

managed/services/inventory/agents.go
```diff
@@ -1815,9 +1847,36 @@ func convertMetricsResolutions(mrs *common.MetricsResolutions) *models.ChangeMet
...
+	ctx context.Context,
+	agentID string,
+	params *models.ChangeAgentParams,
+	check func(current *models.Agent) error,
+) (inventoryv1.Agent, error) {
 	var agent inventoryv1.Agent
 
 	err := as.db.InTransactionContext(ctx, nil, func(tx *reform.TX) error {
+		if check != nil {
+			current, err := models.FindAgentByIDForUpdate(tx.Querier, agentID)
```
ademidoff, line 1872: **bug** — this `SELECT ... FOR UPDATE` is held across `CheckConnectionToService` at line 1891, a network round-trip to the monitored database.

`pmm-admin inventory change agent mongodb-exporter <id> --agent-env-vars=X --username=u` now holds an exclusive lock on the agent row for the full MongoDB dial. If the target is unreachable, that's the whole dial timeout — blocking every other write to that agent and pinning a pooled Postgres connection for the duration.

`executeAgentChange` (every other agent type) takes no such lock, so the transaction shape is now inconsistent between sibling paths. If the check has to stay here, validate and release before the connection check rather than holding through it.

managed/services/management/common/mongodb_env_vars.go
```diff
@@ -0,0 +1,78 @@
...
+// full-replace, so an exporter that already stores a reserved name (e.g. from before this check
+// existed) must still be able to resend it while changing other, unrelated names. Pass a nil
+// grandfathered when there is no existing agent to grandfather, such as on Add.
+// Grandfathered matches case-sensitively (unlike the reserved-name check below, which is
+// case-insensitive on principle): a stored "mongodb_uri" is a different OS environment variable
+// from "MONGODB_URI" and must not grandfather a switch to the latter.
+// It lives in this cross-service package (rather than services/inventory, where the check
+// originated) so both the inventory API and ManagementService.addMongoDB can apply the same check
+// without either service importing the other.
+func ValidateMongoDBExporterEnvVarNames(names []string, grandfathered map[string]struct{}) error {
```
ademidoff, line 50: This is the core of my "more structure than needed" comment (see the review summary).

Enumerate what grandfathering actually lets through:
- `MONGODB_URI` — skipped by the supervisor's `reserved` check (`agent/agents/supervisor/supervisor.go:883`)
- names failing `envvars.ValidateName` (dashes, `PMM_AGENT_*`) — skipped at `agent/agents/supervisor/supervisor.go:876`

There is no third case. So this cross-package `common` helper, `FindAgentByIDForUpdate`, `executeAgentChangeChecked`, the closure plumbing in `ChangeMongoDBExporter` and ~150 lines of test exist to keep storing names that provably have zero runtime effect.

`Agent.SetEnvironmentVariableNames` already grandfathers from the row `ChangeAgent` reads in the same transaction and already has `a.AgentType`. Putting the `MONGODB_URI` rule there removes the row lock, the new service-layer path and this whole package.

admin/commands/inventory/change_agent_mongodb_exporter.go
```diff
@@ -119,6 +126,17 @@ func (cmd *ChangeAgentMongodbExporterCommand) RunCmd() (commands.Result, error)
 		}
 	}
 
+	// Parse environment variable names if provided
+	var agentEnvVars []string
+	if cmd.AgentEnvVars != nil {
+		var err error
+
+		agentEnvVars, err = commands.ValidateEnvironmentVariableNames(*cmd.AgentEnvVars)
```
ademidoff, line 134: Related to the summary: this validates strictly with no grandfathered set, so the server-side grandfathering path is unreachable through the only sanctioned client.

Concretely — agent stores the legacy name `krb5-ktname`, user runs `pmm-admin inventory change agent mongodb-exporter <id> --agent-env-vars=krb5-ktname,KRB5_CONFIG`. `commands.ValidateEnvironmentVariableNames` → `envvars.NormalizeNames` (with `grandfathered == nil`) fails client-side with `invalid environment variable name: krb5-ktname`, and no request is ever sent. Same story for a stored `MONGODB_URI`.

So `FindAgentByIDForUpdate`, `executeAgentChangeChecked`, `MongoDBExporterEnvVarNamesGrandfathered` and the `pg_stat_activity`-polling concurrency test serve raw API callers only. Either pass the current names through from the client, or accept that the server path only needs to be as elaborate as the API contract requires.

managed/services/management/common/mongodb_env_vars.go
```diff
@@ -0,0 +1,78 @@
...
+// names, normalized for ValidateMongoDBExporterEnvVarNames's grandfathered set. Agent must be the
+// row the caller intends to update, read within the same transaction as that update, so the names
+// it grandfathers cannot go stale before the update actually applies them.
+func MongoDBExporterEnvVarNamesGrandfathered(agent *models.Agent) (map[string]struct{}, error) {
+	existing, err := agent.GetEnvironmentVariableNames()
+	if err != nil {
+		return nil, err
+	}
+
+	grandfathered := make(map[string]struct{}, len(existing))
```
ademidoff: These six lines are a verbatim duplicate of the grandfathered-map construction in `managed/models/agent_model.go:471-476` — get names, `make` a map sized to `len`, insert `strings.TrimSpace(name)`.

Two copies of the same normalization rule for the same field means a future change to it (case folding, trimming something else) has to land in two packages or the two grandfathered sets silently diverge. One of them should call the other.

managed/models/agent_helpers.go
```diff
@@ -348,6 +348,28 @@ func FindAgentByID(q *reform.Querier, id string) (*Agent, error) {
 	return new(DecryptAgent(*agent)), nil
 }
 
+// FindAgentByIDForUpdate finds Agent by ID and locks its row (SELECT ... FOR UPDATE) for the
+// duration of the caller's transaction. Use this instead of FindAgentByID when the caller reads
+// the row to decide whether a subsequent write in the same transaction is allowed: without the
+// lock, a concurrent transaction could commit a conflicting write in between, so the decision
+// would be based on data that is no longer current by the time it is acted on.
+func FindAgentByIDForUpdate(q *reform.Querier, id string) (*Agent, error) {
```
ademidoff, line 356: This is a byte-for-byte copy of `FindAgentByID` (line 335) apart from the ` FOR UPDATE` tail: same empty-ID guard, same `reform.ErrNoRows` → `codes.NotFound` mapping with an identical message, same `//nolint:forcetypeassert` cast, same `DecryptAgent` wrap.

Any future fix to the not-found mapping or to decryption has to be applied twice. An unexported `findAgentByID(q, id string, forUpdate bool)` with two thin wrappers removes the copy.

managed/services/management/mongodb.go
```diff
@@ -63,6 +64,16 @@ func (s *ManagementService) addMongoDB(ctx context.Context, req *managementv1.Ad
 			return err
 		}
 
+		// This is a new agent, so there is no existing agent to grandfather.
+		err = common.ValidateMongoDBExporterEnvVarNames(req.EnvironmentVariableNames, nil)
```
ademidoff: This is pure input validation of `req`, but it runs inside the transaction — after the node lookup and after `AddNewService` has already written rows.

`AddService` with `environment_variable_names: ["MONGODB_URI"]` inserts a service row, converts it, resolves the metrics mode, and only then rejects the request and rolls all of it back. Since the check depends on nothing but `req`, it belongs at the top of `addMongoDB` (or in the request-validation layer).

managed/services/inventory/agents.go
```diff
@@ -1815,9 +1847,36 @@ func convertMetricsResolutions(mrs *common.MetricsResolutions) *models.ChangeMet
...
+	return as.executeAgentChangeChecked(ctx, agentID, params, nil)
+}
+
+// executeAgentChangeChecked behaves like executeAgentChange, but if check is non-nil, it is called
+// with the agent's current row, locked (SELECT ... FOR UPDATE) for the rest of the transaction,
+// before the change is applied. Use this when a check's outcome depends on the row's current state
+// and must not be decided from a read taken outside this transaction: a concurrent change could
+// otherwise commit in between that read and this write, making the decision stale by the time it
+// takes effect.
+func (as *AgentsService) executeAgentChangeChecked( //nolint:ireturn
```
ademidoff, line 1859: Two efficiency notes on this wrapper:

1. A checked change reads the same row twice: this function issues `SELECT ... FOR UPDATE` (line 1869), then `models.ChangeAgent` issues a second, unlocked `FindAgentByID` for the identical row (`managed/models/agent_helpers.go:1235`) — two round-trips and two `DecryptAgent` passes per change.
2. `checkEnvVarNames` is installed whenever the wrapper message is present (line 416), so `{"environment_variable_names": {}}` — the documented "remove all" request — still takes the row lock and builds the grandfathered map in order to validate an empty list.

Passing the already-locked row into `ChangeAgent`, or moving the check into `SetEnvironmentVariableNames`, removes both.

managed/services/management/mongodb_env_vars_test.go
```diff
@@ -0,0 +1,208 @@
...
+	t.Cleanup(func() { uuid.SetRand(nil) })
+
+	ctx := logger.Set(context.Background(), t.Name())
+	sqlDB := testdb.Open(t, models.SetupFixtures, nil)
+	t.Cleanup(func() {
+		assert.NoError(t, sqlDB.Close())
+	})
+	db := reform.NewDB(sqlDB, postgresql.Dialect, reform.NewPrintfLogger(t.Logf))
+
+	cc := &mockConnectionChecker{}
```
ademidoff: Nit: this block (and the one at line 155) hand-rolls ~40 lines of mock/teardown boilerplate that `setup(t)` at `managed/services/management/agent_test.go:44` already provides — it constructs all eight mocks, calls `.Test(t)` on each, registers an eight-line `AssertExpectations` cleanup, opens testdb and calls `NewManagementService`.

The tests need `db` for `models.FindAgents`, but `s.db` is reachable in-package — `managed/services/inventory/agents_test.go` does exactly that with `as.db.Querier`.

As written it's ~85 duplicated lines, and adding a new mock to `NewManagementService` now means wiring it into three places.

managed/services/management/mongodb_env_vars_test.go
```diff
@@ -0,0 +1,208 @@
...
+// Environment variable names are only meaningful for agents that pmm-agent starts as a separate
+// process, because they are resolved into SetStateRequest.AgentProcess.env_variable_names. Among
+// the MongoDB agents only mongodb_exporter is such a process; the QAN and RTA agents are built-in
+// agents running inside pmm-agent. This test pins that only the exporter row stores the names, so
+// we do not persist data no one reads.
+func TestAddMongoDBStoresEnvVarNamesForExporterOnly(t *testing.T) {
+	uuid.SetRand(&tests.IDReader{})
+	t.Cleanup(func() { uuid.SetRand(nil) })
+
+	ctx := logger.Set(context.Background(), t.Name())
```
ademidoff: Nit: `t.Context()` here — AGENTS.md (Testing) says to prefer it over `context.Background()`, and line 148 in this same file already does (`logger.Set(t.Context(), t.Name())`).

utils/envvars/envvars.go
```diff
@@ -0,0 +1,105 @@
...
+
+// ValidateName reports whether name is an acceptable environment variable name for pass-through:
+// syntactically valid, within length bounds, and outside pmm-agent's own reserved namespace.
+func ValidateName(name string) error {
+	if name == "" {
+		return errors.New("environment variable name cannot be empty")
+	}
+
+	if len(name) > MaxNameLength {
+		return fmt.Errorf("environment variable name %q is too long (max %d characters)", name, MaxNameLength)
```
ademidoff: Style nit, take it or leave it: `%q` here (and at line 56, and `managed/services/management/common/mongodb_env_vars.go:60`) double-escapes once the error lands in a logrus field — the new test output shows `error="environment variable name \"PMM_AGENT_SERVER_PASSWORD\" is reserved..."`. `'%s'` reads better in logs and still delimits the value.

documentation/docs/use/commands/pmm-admin/inventory.md
```diff
@@ -68,12 +68,15 @@ Only the flags you specify are updated — all other settings remain unchanged.
 
 When you change connection-affecting parameters (username, password, TLS settings, etc.), PMM verifies the new settings by connecting to the database before saving them. If the connection fails (for example, wrong credentials), the command returns an error and **no changes are applied**. Use `--skip-connection-check` to bypass this verification (see [Connection and authentication](#connection-and-authentication)).
 
+`--agent-env-vars` is **not** part of this verification: the connection check is performed by pmm-agent itself, not by the exporter, so it does not read or depend on the names you are adding. However, the check and the rest of the change are applied together as a single operation: if you combine `--agent-env-vars` with a connection-affecting flag and the connection check fails, the whole change — including the environment variable names — is discarded, not just saved. Once a change is saved, its effect on the exporter's environment only shows after the exporter restarts.
```
ademidoff: Two things in this paragraph:

1. "the whole change — including the environment variable names — is discarded, not just saved" is self-contradictory. I think it means "not just the connection-affecting part"?
2. "its effect on the exporter's environment only shows after the exporter restarts" reads as though the operator has to restart something. In fact `filter`/`proto.Equal` (`agent/agents/supervisor/supervisor.go:432`) sees the changed `env_variable_names` and pmm-agent restarts the exporter automatically on the next SetState. As phrased, a reader will go looking for a restart that already happened.

managed/services/management/common/mongodb_env_vars.go
JiriCtvrtka: The comment doesn't contradict the code — lines 40-41 say the reserved check is case-insensitive on principle, unlike grandfathering, which is deliberate. You're right that mongodb_uri could never collide on the agent side (supervisor.go:883 is a case-sensitive lookup), so this is a policy choice, not a bug: I'd rather reject a name differing only in case from the one we compute than have two near-identical vars in the exporter's environment. Rewording the error so it stops claiming pmm-agent sets that exact name.

managed/services/management/common/mongodb_env_vars.go
JiriCtvrtka: ensureAuthParams only appends HTTP_AUTH when agentVersion.Less(v2_28_00) (managed/services/agents/mongodb.go:87). Adding it to the reserved list unconditionally would reject it for every ≥2.28.0 agent, where nothing sets it — a false rejection for the common case to cover a pre-2.28.0 client. It's also not silent: supervisor.go:884 logs a warning when it skips. Fixing the comment to mention the second injection site instead of the list.

managed/services/inventory/agents.go
JiriCtvrtka: I don't think this changes the lock profile. models.ChangeAgent issues q.Update(row) at managed/models/agent_helpers.go:1494, in the same transaction, before CheckConnectionToService at agents.go:1891 — that UPDATE takes an exclusive row lock held to commit. So every agent type already holds an exclusive lock on the row across the full dial; the FOR UPDATE only acquires it two statements earlier. "validate and release before the connection check" isn't achievable inside a single transaction. Side benefit: it also closes the lost-update race between the read at :1235 and the write at :1494 that the other Change* paths still have.

managed/services/management/common/mongodb_env_vars.go
JiriCtvrtka: The size argument is fair and I'm open to it. But the enumeration is missing the case the machinery exists for — see my reply on the pmm-admin comment: MONGODB_URI does reach the server from pmm-admin. Let's settle that first, then decide the shape.

admin/commands/inventory/change_agent_mongodb_exporter.go
JiriCtvrtka: Holds for ValidateName grandfathering — krb5-ktname does die client-side. Not for MONGODB_URI: it's a valid POSIX name with no PMM_AGENT_ prefix, so NormalizeNames passes it, pmm-admin sends it, and the server-side grandfathering is exactly what stops the update being rejected. So "Same story for a stored MONGODB_URI" isn't right, and the path is reachable from pmm-admin.

managed/services/management/common/mongodb_env_vars.go
JiriCtvrtka: Agreed, they're identical. Having MongoDBExporterEnvVarNamesGrandfathered be the one caller.

managed/models/agent_helpers.go
JiriCtvrtka: Not a byte-for-byte copy: FindAgentByID uses q.Reload, this uses q.SelectOneFrom plus a type assertion, and there's no forcetypeassert in the original. Shared surface is the empty-ID guard and the ErrNoRows mapping. Factoring those out anyway.

managed/services/management/mongodb.go
JiriCtvrtka: Agreed — it's pure req validation, nothing depends on the transaction. Moving it above InTransactionContext.

managed/services/inventory/agents.go
JiriCtvrtka: Both correct. Skipping the check when the list is empty is a one-liner. Not passing the locked row into ChangeAgent — that needs a signature change across every caller for one extra SELECT; noting it as a TODO.

managed/services/management/mongodb_env_vars_test.go
JiriCtvrtka: Switching both blocks to setup(t) and reaching s.db in-package.

managed/services/management/mongodb_env_vars_test.go
JiriCtvrtka: Yes — AGENTS.md:382. Fixing.

utils/envvars/envvars.go
JiriCtvrtka: Fair, switching to '%s'.

documentation/docs/use/commands/pmm-admin/inventory.md
JiriCtvrtka: Right on both. The sentence is garbled, and filter/proto.Equal at supervisor.go:432 does compare EnvVariableNames, so pmm-agent restarts the exporter itself — which also contradicts "Changes take effect immediately without restarting the agent" seven lines up. Per your earlier comment this file is coming out of this PR, so I'll carry both fixes into the DOC branch PR.

utils/envvars/envvars.go
```diff
@@ -0,0 +1,112 @@
...
+
+		result = append(result, name)
+	}
+
+	// An agent may already store more than MaxNames names: this bound did not exist before it was
+	// introduced alongside the rest of this policy. Applying it unconditionally would leave such a
+	// list uneditable forever, since the field is full-replace — the owner could not even remove a
+	// name without first truncating to MaxNames and losing the rest. Take the stored count as the
+	// effective bound instead, so an oversized list can only shrink toward MaxNames, never grow.
+	limit := max(MaxNames, len(grandfathered))
```
ademidoff: The bound never converges, so the comment above it doesn't describe what the code does. `limit` is derived from the *stored* count, not from how many of the submitted names are actually grandfathered — so an oversized list can be replaced wholesale with entirely new names at the same oversized count, indefinitely.

Ran it against this branch: 40 legacy names stored, 40 brand-new valid names submitted, none of them grandfathered → `err=<nil>`, 40 names stored again. The same 40 through `NormalizeNames` are correctly rejected with `too many environment variable names: 40 (max 32)`.

Deriving the limit from how many of the *submitted* names are found in `grandfathered` gives the shrink-only behaviour the comment promises.

managed/models/agent_model.go
```diff
@@ -457,13 +458,26 @@ func (a *Agent) GetEnvironmentVariableNames() ([]string, error) {
...
+// (e.g. from before this validation existed, or under since-tightened rules) are carried forward
+// as-is: this is a full-replace field, so a caller resending an existing name alongside a new one
+// must not be rejected because of a name it did not intend to change.
 func (a *Agent) SetEnvironmentVariableNames(names []string) error {
 	if len(names) == 0 {
 		a.EnvironmentVariables = nil
 		return nil
 	}
 
+	grandfathered, err := a.GrandfatheredEnvironmentVariableNames()
```
ademidoff: A row whose `environment_variables` JSON can't be decoded still can't be repaired by sending a new list, which contradicts the recovery path documented at `managed/services/converters.go:337`.

Any non-empty update reaches `GrandfatheredEnvironmentVariableNames` → `GetEnvironmentVariableNames` → `failed to unmarshal shared environment variable names`, and fails. Only an empty list clears the row, because it short-circuits at `managed/models/agent_model.go:466` before grandfathering — and nothing tells the caller that.

Through `ChangeMongoDBExporter` it's worse: the closure at `managed/services/inventory/agents.go:419` returns that raw error before `ChangeAgent` runs, so the client sees `Unknown` rather than the `InvalidArgument` this PR standardises everywhere else.

Treating an undecodable column as "no grandfathered names" rather than an error would make the documented recovery real — it's the same reasoning as the fix in `managed/services/converters.go`.

utils/envvars/envvars.go
```diff
@@ -0,0 +1,112 @@
...
+		return nil, nil
+	}
+
+	result := make([]string, 0, len(names))
+	seen := make(map[string]struct{}, len(names))
+
+	for _, name := range names {
+		name = strings.TrimSpace(name)
+
+		if _, ok := grandfathered[name]; !ok {
```
ademidoff: The grandfathered lookup runs before `ValidateName`, so a stored empty or whitespace-only name grandfathers the empty string and lets it be written back permanently.

A row written through the raw API before this PR (nothing validated it then) can hold `["  "]`. `GrandfatheredEnvironmentVariableNames` trims that to `""` and puts it in the set; a later update containing `""` or `"   "` trims to `""`, matches, skips the `cannot be empty` check and is appended to the result.

Ran it: `NormalizeNamesAllowing([]string{"   ", "KRB5_CONFIG"}, map[string]struct{}{"": {}})` returns `["" "KRB5_CONFIG"]` with a nil error. The row then keeps `""` for good, `ToAPIAgent` reports it, and pmm-agent logs `Skipping invalid environment variable name` on every state update. Skipping the empty string when building the grandfathered set is enough.

managed/models/agent_helpers.go
```diff
@@ -1233,6 +1263,14 @@ func ChangeAgent(q *reform.Querier, agentID string, params *ChangeAgentParams) (
 		}
 	}
 
+	// An empty slice removes all environment variable names, SetEnvironmentVariableNames treats it the same as nil.
+	if params.EnvironmentVariableNames != nil {
+		err = row.SetEnvironmentVariableNames(*params.EnvironmentVariableNames)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
```
ademidoff: Blanket-mapping every `SetEnvironmentVariableNames` failure to `InvalidArgument` mislabels the ones that aren't the caller's fault, and `err.Error()` drops the `%w` chain.

It can fail three ways: a genuinely bad name (`InvalidArgument` is right), a `json.Marshal` failure, and the unmarshal failure in the thread above. The last two are server-side problems reported to the client as "the argument you sent is invalid", so a caller keeps retrying with different names and never succeeds. Same construct at `managed/models/agent_helpers.go:1042`.

Wrapping with `status.Error(..., err.Error())` also discards the wrapped error, so `errors.Is`/`errors.As` stop matching upstream.

managed/services/management/mongodb.go
```diff
@@ -104,14 +117,14 @@ func (s *ManagementService) addMongoDB(ctx context.Context, req *managementv1.Ad
 		mongodb.MongodbExporter = agent.(*inventoryv1.MongoDBExporter) //nolint:forcetypeassert
 
 		if req.QanMongodbProfiler {
+			// Built-in agent: no EnvironmentVariableNames, see the mongodb_exporter comment above.
```
ademidoff, line 120: Right call going forward, but it leaves the rows that are already there.

Every MongoDB service added before this PR has `environment_variables` populated on its `qan_mongodb_profiler` / `qan_mongodb_mongolog` / `rta_mongodb` rows. Nothing reads them (only `mongodbExporterConfig` calls `GetEnvironmentVariableNames`), no API type exposes them for those agent types, and `ChangeAgentParams.EnvironmentVariableNames` is only ever populated by `ChangeMongoDBExporter` — so there is no supported path to clear them, and they persist forever.

A forward-only migration nulling the column for non-exporter agent types would finish the job this PR starts.

managed/services/converters.go
```diff
@@ -329,6 +330,17 @@ func ToAPIAgent(q *reform.Querier, agent *models.Agent) (inventoryv1.Agent, erro
...
 		exporter.EnableDiagnosticDataHistograms = agent.MongoDBOptions.EnableDiagnosticDataHistograms
 
+		// A stored value that cannot be decoded (hand-edited row, failed migration, an older
+		// writer) must not fail the conversion: ListAgents calls ToAPIAgent once per row, so
+		// propagating the error would drop the entire inventory listing because of one unrelated
+		// column on one agent. Report the field as empty and log instead; sending a new list
+		// repairs the row.
+		envVarNames, err := agent.GetEnvironmentVariableNames()
+		if err != nil {
+			logrus.Warnf("Ignoring undecodable environment variable names of agent %s: %s.", agent.AgentID, err)
```
ademidoff: This brings a global-logger dependency into `managed/services` that the package didn't have — the `logrus` import is added purely for this call.

AGENTS.md (Logging) asks for `*logrus.Entry` with structured fields; this is the package-level logger with the agent id interpolated into the message, so on a busy server the warning lands in the global stream with no request context. Threading an entry in, or returning the decode failure for the caller to log, keeps it consistent with the rest of the service layer.

managed/services/management/common/mongodb_env_vars.go
```diff
@@ -0,0 +1,63 @@
...
+// sets itself for mongodb_exporter (see mongodbExporterConfig in managed/services/agents/mongodb.go).
+// A user-selected name here would never take effect: the supervisor skips it to avoid overriding
+// the computed value on the agent side. Rejecting it here instead gives the caller an actionable
+// error at request time.
+//
+// HTTP_AUTH is deliberately absent even though ensureAuthParams (managed/services/agents/agents.go)
+// also injects it: it does so only for pmm-agent older than 2.28.0, so listing it here would reject
+// a name that nothing sets for every current agent. For those old agents the supervisor still skips
+// it and logs a warning, which is the lesser of the two failures.
+var mongoDBExporterReservedEnvVars = map[string]struct{}{
```
ademidoff: Separate point from the HTTP_AUTH thread, which I'm not reopening: the issue is that this list is a hand-maintained copy of what `mongodbExporterConfig` actually sets.

`managed/services/agents/mongodb.go:69` builds `"MONGODB_URI=" + ...`; those keys are the authoritative set, and this restates one of them by hand. Add a second computed variable to the exporter config later and the API accepts it as a user-selected name while pmm-agent skips it and only warns — the exact failure this check exists to prevent.

Deriving the set from the config builder, or a shared constant both sides use, fixes it at the mechanism instead of relying on the two staying in sync.

managed/models/agent_helpers.go
```diff
@@ -1135,8 +1164,9 @@ type ChangeValkeyOptions struct {
 // ChangeAgentParams contains parameters that can be changed for all Agent types.
 type ChangeAgentParams struct {
 	// Common fields for all agents
-	Enabled      *bool              // true - enable, false - disable, nil - no change
-	CustomLabels *map[string]string // empty map - remove all custom labels, non-empty - change, nil - no change
+	Enabled                  *bool              // true - enable, false - disable, nil - no change
+	CustomLabels             *map[string]string // empty map - remove all custom labels, non-empty - change, nil - no change
+	EnvironmentVariableNames *[]string          // empty slice - remove all environment variable names, non-empty - change, nil - no change
```
ademidoff: Inline trailing comment. AGENTS.md: "Don't inline comments (`code // comment`); put comments on separate lines — `//nolint` is the only exception." The two lines above are pre-existing and only realigned by this diff, so this is the one new instance.

agent/agents/process/process.go
```diff
@@ -84,8 +89,20 @@ type Params struct {
...
+		for _, name := range p.ResolvedEnvNames {
+			resolved[name] = struct{}{}
+		}
+
 		printEnv := make([]string, len(p.Env))
 		for i, item := range p.Env {
+			name, _, ok := strings.Cut(item, "=")
+			if ok {
+				if _, isResolved := resolved[name]; isResolved {
+					printEnv[i] = name + "=" + logger.RedactString(item, extensionsv1.RedactType_REDACT_TYPE_FULL)
```
ademidoff: `RedactString` ignores its first argument for `REDACT_TYPE_FULL` — `utils/logger/protobuf.go:169` returns the masked constant unconditionally.

Passing `item` reads as though the value shapes the output, the way it genuinely does on the DSN branch at `agent/agents/process/process.go:106`, and invites someone later to swap in a partial mask on the assumption that it does. Passing `""`, or the constant directly, says what actually happens.

ademidoff (commented):
Re-checked the follow-up commits against head `37160bc7`. Everything from the last round is either fixed or answered — the `FOR UPDATE` rebuttal in particular holds: `models.ChangeAgent` at `managed/services/inventory/agents.go:1883` updates the row before `CheckConnectionToService` at `managed/services/inventory/agents.go:1894` in the same transaction, so the exclusive lock was already held across the dial.

What's left below. The first three are behavioural and I'd want them before merge; the rest are smaller.

managed/services/management/common/mongodb_env_vars.go
ademidoff: Settled below — you're right that `MONGODB_URI` reaches the server from pmm-admin. It doesn't change the shape question, and re-reading `utils/envvars/envvars.go:85` sharpened it: grandfathering skips the whole of `ValidateName`, so a stored `PMM_AGENT_SERVER_PASSWORD` is carried forward too. Pre-PR pmm-admin's `^[A-Z_][A-Z0-9_]*$` accepted that name and nothing checked it server-side, so those rows are reachable through the sanctioned CLI, not just hand-edited DBs. On an older pmm-agent it resolves and hands pmm-agent's own server password to mongodb_exporter's environment.

Proposal: grandfather shape, never authority.

- Keep grandfathering in `Agent.SetEnvironmentVariableNames` for the pattern, length and count rules. That's the case that genuinely locks a row out, it's self-contained in the model, and it costs no plumbing.
- Never grandfather `MONGODB_URI` or the `PMM_AGENT_` namespace. Both are inert on a pmm-agent with this PR (`agent/agents/supervisor/supervisor.go:877` and `agent/agents/supervisor/supervisor.go:883`) and harmful on an older one — the exact override you describe at `agent/agents/supervisor/supervisor.go:856`. Preserving them serves nobody.

That makes the reserved check depend only on `req`, which is what removes the structure: `executeAgentChangeChecked`, `FindAgentByIDForUpdate`, the `checkEnvVarNames` closure and `ConcurrentClearInvalidatesAnInFlightGrandfatheredReplace` all go away, and `managed/services/inventory/agents.go:424` becomes the same one-liner as the Add path at `managed/services/inventory/agents.go:361`.

Cost: an owner of a pre-PR row storing `MONGODB_URI` gets one actionable error and drops the name. That's the correct outcome for a name that does nothing, not a regression.

If you'd rather keep the PR's footprint as it is, the minimum I'd still want is the `PMM_AGENT_` prefix check running unconditionally in `NormalizeNamesAllowing` — two lines, and it's the half of this that isn't a taste question.

admin/commands/inventory/change_agent_mongodb_exporter.go
ademidoff: Conceded — `MONGODB_URI` passes `NormalizeNames` and reaches the server, so the path isn't unreachable and my "same story" line was wrong.

Two things it doesn't rescue. The parity comment at `admin/commands/base.go:202` is still false in the other direction: a stored `krb5-ktname` is rejected client-side and would be accepted server-side by grandfathering. pmm-admin can't replicate server policy because it doesn't know the stored list, so please reword it to say the server applies additional rules and may both accept names rejected here and reject names accepted here.

And on the substance: reachable or not, every name grandfathering preserves is one a current pmm-agent discards anyway, so the machinery only serves agents older than this PR — where what it preserves is exactly what you don't want preserved. Continued in the thread on `managed/services/management/common/mongodb_env_vars.go:50`.

utils/envvars/envvars.go
JiriCtvrtka: Confirmed, and reproduced exactly as you describe: 40 stored, 40 brand-new names
submitted, err=<nil>.

Fixed by counting the submitted names actually found in `grandfathered` (`carried`)
rather than `len(grandfathered)`. The same case now fails with
"too many environment variable names: 40 (max 32)", while a genuine 40->39 shrink
still passes. Comment above it rewritten to describe what the code does.

Added two tests: wholesale swap at the same oversized count, and dropping one stored
name while adding a new one (length unchanged, so also rejected).

managed/models/agent_model.go
JiriCtvrtka: Right, the documented recovery path didn't work.

`GrandfatheredEnvironmentVariableNames` now treats an undecodable column as "nothing
to carry forward" and returns an empty set. Since it can no longer fail I dropped its
error return entirely, which also removes the raw-error path in the
ChangeMongoDBExporter closure - so the Unknown-instead-of-InvalidArgument leak is gone
by construction rather than by mapping.

Verified: sending a new list against a `{"not":"an array"}` column now succeeds and
repairs the row.

utils/envvars/envvars.go
JiriCtvrtka: Reproduced verbatim - `["" "KRB5_CONFIG"]` with a nil error.

Fixed in two places rather than one. The set builder skips blank entries as you
suggested, and `NormalizeNamesAllowing` also refuses to treat "" as grandfathered
whatever set it's handed. The function is exported, so I'd rather the invariant hold
at the enforcement point than depend on every caller building its set correctly.

managed/models/agent_helpers.go
JiriCtvrtka: Added `envvars.InvalidNameError`; a new `environmentVariableNamesError` helper maps
only that to InvalidArgument and returns anything else wrapped, so errors.As keeps
matching upstream. Applied at both sites (1042 and 1270).

Fixing the thread above removes the unmarshal case anyway, but I didn't want the
mapping resting on "json.Marshal of a []string can't fail" - this way it's correct by
construction.

managed/services/management/mongodb.go
JiriCtvrtka: Added migration 119, nulling `environment_variables` where `agent_type <> 'mongodb_exporter'`.

I checked every reader before writing it: only `mongodbExporterConfig` and the
MongoDBExporter branch of `ToAPIAgent` call `GetEnvironmentVariableNames`, so the data
is inert on those rows rather than actively wrong - happy to drop this into a follow-up
if you'd rather not carry a data migration in this PR.

Flagging that this is the one change I have not been able to exercise: no Postgres
available locally, so every testdb-backed test is skipped here. It needs a run against
a live server before merge.

managed/services/converters.go
JiriCtvrtka: Switched to a component-scoped `*logrus.Entry` (`logrus.WithField("component",
"converters")`) with the agent id as a structured field instead of interpolated into
the message.

I didn't thread an entry through ToAPIAgent: it has 25 call sites and this is one
warning. The component-entry shape is what the rest of pmm-managed uses (about a dozen
in main.go alone), so it stays consistent without the churn. Say the word if you'd
still prefer the parameter.

managed/services/management/common/mongodb_env_vars.go
JiriCtvrtka: Agreed on fixing it at the mechanism. The set now lives as
`agents.MongoDBExporterReservedEnvVars`, declared next to the `env` slice in
mongodbExporterConfig that actually populates it, and the `MONGODB_URI` literal is a
constant both sides share. The validator consumes it; no import cycle.

The HTTP_AUTH rationale is preserved in the function doc.

managed/models/agent_helpers.go
JiriCtvrtka: Moved above the field.

agent/agents/process/process.go
JiriCtvrtka: Passing "" now, with a one-line comment noting that REDACT_TYPE_FULL substitutes a
constant and the value plays no part - unlike the DSN branch below it.

ademidoff:
OK, it looks much better now.

One item from the earlier round is still open — the `PMM_AGENT_` namespace is still grandfathered. Details in the thread on `managed/services/management/common/mongodb_env_vars.go:50`.

managed/services/management/common/mongodb_env_vars.go
ademidoff: `4db12ed2d` covers the empty string and the count, but not the `PMM_AGENT_` namespace: `NormalizeNamesAllowing` still skips the whole of `ValidateName` for a grandfathered name, so the prefix ban at `utils/envvars/envvars.go:55` is bypassed.

Against this branch:

```
PMM_AGENT_ grandfathered: err=<nil> result=["PMM_AGENT_SERVER_PASSWORD" "KRB5_CONFIG"]
PMM_AGENT_ fresh:         err=environment variable name 'PMM_AGENT_SERVER_PASSWORD' is reserved for pmm-agent's own configuration and cannot be selected
```

So a row that stored that name before this PR keeps it across every edit, and on a pmm-agent older than this PR it resolves and lands in mongodb_exporter's environment.

This is the "minimum I'd still want" from my earlier comment: run the `ReservedPrefix` check unconditionally, before the grandfathered lookup, so grandfathering forgives shape but never authority. Whether the larger reshape lands is still your call and I'm happy to leave that for a follow-up, but I'd like the prefix check in this PR.

admin/commands/inventory/change_agent_mongodb_exporter.go
ademidoff: Also still open from the last round: the comment at `admin/commands/base.go:202` is unchanged. A stored `krb5-ktname` is rejected client-side and accepted server-side by grandfathering, so "a name rejected here would also be rejected there" isn't true. Reword only, no code change.

utils/envvars/envvars.go
ademidoff: Confirmed on the new code: the wholesale swap now fails with `too many environment variable names: 40 (max 32)`, and a genuine 40 → 39 shrink still succeeds. Counting `carried` rather than the stored size is the right derivation, and the rewritten comment matches the behaviour.

managed/models/agent_model.go
ademidoff: Confirmed — sending a new list against a `{"not":"an array"}` column repairs the row.

Dropping the error return rather than mapping it was the better call: it removes the raw-error path out of the closure instead of relabelling it, so there's no longer a way for that error to reach a client at all.

utils/envvars/envvars.go
ademidoff: Confirmed: `NormalizeNamesAllowing([]string{"   ", "KRB5_CONFIG"}, map[string]struct{}{"": {}})` now returns `environment variable name cannot be empty`.

Agreed on enforcing it at the exported function rather than only in the set builder.

managed/models/agent_helpers.go
ademidoff: Confirmed. Only a name the caller sent maps to `InvalidArgument` now, and everything else surfaces as itself with the chain intact.

Agreed on not resting the mapping on "json.Marshal of a []string can't fail" even though fixing the thread above removes the unmarshal case.

managed/services/management/mongodb.go
ademidoff: Ran it, since you couldn't.

119 is the next free number — main's highest is 118. I seeded a database at 118 with `environment_variables` populated on `mongodb_exporter`, `qan-mongodb-profiler-agent`, `qan-mongodb-mongolog-agent`, `rta-mongodb-agent` and a `mysqld_exporter` control row, then migrated to 119. Only the exporter row kept its names; the other four were cleared, and the exporter's list came back unchanged. I checked the readers separately too: only `managed/services/agents/mongodb.go:76` and `managed/services/converters.go:343` call `GetEnvironmentVariableNames`, so no other agent type loses anything it uses.

On moving it to a follow-up: I'd keep it here now that it's verified. This PR is what makes those rows permanently unreachable, so splitting it means shipping a release where the only way to clear them is by hand.

One nit: the comment names the types with underscores (`qan_mongodb_profiler`), but the stored `agent_type` values are hyphenated (`qan-mongodb-profiler-agent`). The SQL doesn't reference them so it's cosmetic, but in a migration comment describing which rows are affected I'd make them match.

managed/services/converters.go
ademidoff: Component entry is fine, no parameter needed. Threading one through 25 call sites for a single warning isn't worth it, and the structured `agent_id` was the part that mattered.

managed/services/management/common/mongodb_env_vars.go
ademidoff: Confirmed, and `TestImports` still passes, so `managed/cmd/pmm-managed/packages.dot` isn't stale.

One note rather than a request: `management/common` now imports `managed/services/agents`, so the line in this file about living here "without either service importing the other" holds only for direct imports — `TestImports` checks direct imports only, so `inventory` → `common` → `agents` satisfies the layering constraint transitively. Sharing a leaf map is fine; the comment just overstates the isolation now.

managed/models/agent_helpers.go
ademidoff: Confirmed.

agent/agents/process/process.go
ademidoff: Confirmed. The comment carries the reason, which is what was missing.

managed/services/management/common/mongodb_env_vars.go
JiriCtvrtka: Confirmed, and reproduced your exact output. Fixed: `validateNotReserved` is split out
of the shape rules and now runs unconditionally, before the grandfathered lookup -
grandfathering forgives shape, never authority, as you put it.

Traced the full path while fixing it. Reachable through the supported CLI on
v3.6.0-v3.9.1: pmm-admin's old `^[A-Z_][A-Z0-9_]*$` accepts PMM_AGENT_SERVER_PASSWORD
and nothing checked it server-side. The shipped agent then resolves it with a bare
os.LookupEnv - no guard at all, since the ValidateName and reserved checks in
resolveEnvVariableNames are new in this PR - and pmm-agent does read that name from its
own environment (agent/config/config.go:394). Narrow chain (needs a pre-PR row, an
older agent, and an env-var deployment), but a real one, and the fix is two lines.

Three tests: reserved name rejected when grandfathered, same case-insensitively, and
KRB5-KTNAME still carried forward so grandfathering isn't gutted.

Leaving MONGODB_URI and the larger reshape for the follow-up, as you offered. The
argument doesn't transfer - overriding MONGODB_URI breaks the exporter's connection
rather than exposing a credential, and the DSN is user-controlled already.

managed/services/management/common/mongodb_env_vars.go
JiriCtvrtka: Fair - the comment overstated it. Reworded: the isolation is between the two services
only, and this package does import services/agents for the reserved set, so inventory
reaches agents transitively.

JiriCtvrtka:
> OK, it looks much better now.
> 
> One item from the earlier round is still open — the `PMM_AGENT_` namespace is still grandfathered. Details in the thread on `managed/services/management/common/mongodb_env_vars.go:50`.

Thanks, PMM_AGENT_ is fixed now, details in the thread on mongodb_env_vars.go.

utils/envvars/envvars.go
```diff
@@ -0,0 +1,170 @@
...
+
+// NormalizeNames trims and validates each name, collapses duplicates keeping the first occurrence,
+// and checks the resulting list's length bound. Both pmm-admin and the API call this rather than
+// validating and deduplicating separately, so a name repeated or padded with whitespace does not
+// count twice against the limit, and stored names are always deduplicated.
+func NormalizeNames(names []string) ([]string, error) {
+	return NormalizeNamesAllowing(names, nil)
+}
+
+// NormalizeNamesAllowing behaves like NormalizeNames, but skips ValidateName for any (already
```
ademidoff, line 104: This doc comment drifted with `2a48517c9`: it no longer skips `ValidateName` for a grandfathered name, only the shape rules — `validateNotReserved` runs unconditionally and the empty string is excluded.

The inline comment inside the loop states the new rule well, but this is the one a caller reads. Something like "skips the shape rules for any (already trimmed) name present in grandfathered; pmm-agent's own namespace and the empty string are rejected regardless" would match the code.

managed/models/agent_model.go
```diff
@@ -457,13 +458,21 @@ func (a *Agent) GetEnvironmentVariableNames() ([]string, error) {
 	return names, nil
 }
 
-// SetEnvironmentVariableNames encodes shared environment variable names.
+// SetEnvironmentVariableNames encodes shared environment variable names. Names already stored
```
ademidoff, line 461: Same drift here: "carried forward as-is" is no longer true for the two names that are never carried forward — the `PMM_AGENT_` namespace and a blank entry.

Worth a clause, since this is the doc for the exported setter and the exception is the security-relevant half of the rule.

ademidoff (commented):
Verified `2a48517c9` and `980100697` against `a69cbbf35`.

The reserved-namespace fix does what it says: a grandfathered `PMM_AGENT_SERVER_PASSWORD` is now rejected, case-insensitively, while `krb5-ktname` and an over-long stored name are still carried forward — so authority is never forgiven and shape still is. The error keeps its `*InvalidNameError` wrapper, so it still maps to `InvalidArgument`. Your `agent/config/config.go:394` citation checks out as well.

Agreed on leaving `MONGODB_URI` out of it: overriding it redirects the exporter's connection rather than exposing a pmm-agent secret, and whoever controls pmm-agent's environment already controls that process. Different failure, fair to defer.

The migration test passes, and the `node_exporter` NULL row is a better control than what I ran locally. Build, the envvars/models/process/supervisor suites, the inventory and management agent tests, and golangci-lint with `--new-from-rev` are all clean on my side.

Three doc-drift nits below, none blocking.

admin/commands/base.go
```diff
@@ -195,29 +197,12 @@ func ParseDisableCollectors(collectors []string) []string {
 	return disableCollectors
 }
 
-// ValidateEnvironmentVariableNames validates environment variable names.
+// ValidateEnvironmentVariableNames validates environment variable names. Names are trimmed and
+// duplicates are collapsed, keeping the first occurrence. This check is client-side only: the API
+// applies the same policy server-side (see utils/envvars), so a name rejected here would also be
```
ademidoff, line 202: Third time on this one, so anchoring it where the fix goes rather than in the old thread on `admin/commands/inventory/change_agent_mongodb_exporter.go:134`.

"A name rejected here would also be rejected there" is wrong in both directions. A stored `krb5-ktname` is rejected here and accepted server-side by grandfathering; `MONGODB_URI` passes here and is rejected server-side by the reserved check in `managed/services/management/common/mongodb_env_vars.go`. pmm-admin can't replicate server policy because it doesn't know the stored list.

Reword only — say the server applies additional rules and may both accept names rejected here and reject names accepted here.

ademidoff (approved):
Approving. Everything from the earlier rounds is fixed and verified on my side — the count bound, the undecodable column, the empty-string and reserved-namespace grandfathering, the error mapping, the logger, the reserved set, and migration 119 (run against a live Postgres, plus your own test for it).

The three doc-comment nits in my previous review are non-blocking; take them or leave them before merge.

Still worth a follow-up ticket for the deferred work: `MONGODB_URI` grandfathering and collapsing the check into `Agent.SetEnvironmentVariableNames`.

4nte (approved):

PR #5798, opened 2026-08-20
PMM-7 Fix GRANT statement when provisioning the PMM database.

JiriCtvrtka:
Ticket number: PMM-7

Feature build: SUBMODULES-0

## How to reproduce

The bug only triggers when the `pmm-managed` role does not exist yet, which in Docker means
a **volume that has never been used before**. On an existing volume the code path is skipped
entirely, so reusing a volume from a previous run will not reproduce it.

```bash
docker volume create pmm-repro-data
docker run -d --name pmm-repro -v pmm-repro-data:/srv percona/pmm-server:3.8.0

# about 5 seconds later
docker exec pmm-repro grep -A1 'syntax error' /srv/logs/postgresql14.log
```

### Expected output (the bug)

`/srv/logs/postgresql14.log`:

```
ERROR:  syntax error at or near "$1" at character 34
STATEMENT:  GRANT ALL PRIVILEGES ON DATABASE $1 TO $2
ERROR:  relation "schema_migrations" does not exist at character 16
```

`/srv/logs/pmm-managed.log` shows the retry loop that hides the failure — the second pass
finds the role already created and succeeds, so startup completes about a second late and
nothing surfaces to the user:

```
09:57:56.452  Migrating database...
09:57:56.596  WARN Failed to migrate database: failed to grant privileges to user pmm-managed
              on database pmm-managed: pq: syntax error at or near "$1" at column 34 (42601).
09:57:57.597  Migrating database...
09:57:57.853  Database migration completed.
```

### Confirming the consequence

The `grafana` database is provisioned by Ansible and gets its grants correctly, which makes
for a direct comparison inside the same container:

```bash
docker exec pmm-repro bash -c 'export PGPASSWORD=$(cat /srv/.postgres_password); \
  psql -U postgres -h 127.0.0.1 -d postgres \
  -c "SELECT datname, datacl FROM pg_database WHERE datname IN ('"'"'pmm-managed'"'"','"'"'grafana'"'"') ORDER BY datname;"'
```

```
   datname   |                          datacl
-------------+-----------------------------------------------------------
 grafana     | {=Tc/postgres,postgres=CTc/postgres,grafana=CTc/postgres}
 pmm-managed |                                    <-- NULL, no grants at all
```

Why it still works today:

```bash
docker exec pmm-repro bash -c 'export PGPASSWORD=$(cat /srv/.postgres_password); \
  psql -U postgres -h 127.0.0.1 -d "pmm-managed" \
  -c "SELECT has_database_privilege('"'"'pmm-managed'"'"','"'"'pmm-managed'"'"','"'"'CREATE'"'"') AS db_create, \
             has_schema_privilege('"'"'pmm-managed'"'"','"'"'public'"'"','"'"'CREATE'"'"') AS schema_create;"'
```

```
 db_create | schema_create
-----------+---------------
 f         | t
```

The role has no `CREATE` on the database, but creates its tables anyway through
`nspacl = {postgres=UC/postgres,=UC/postgres}` — the entry with the empty grantee is `PUBLIC`
holding `USAGE` and `CREATE` on schema `public`. That default was removed in PostgreSQL 15.

### Not reproducible on restart

```bash
docker restart pmm-repro
docker exec pmm-repro grep -c 'GRANT ALL PRIVILEGES ON DATABASE' /srv/logs/postgresql14.log
```

Still `1` — the count from the first boot. The second boot migrates in a single pass (82 ms,
no warning). This is why the bug has gone unnoticed.

### Verifying the fix

Build the patched binary and swap it into a container on a brand new volume:

```bash
cd managed && CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build \
  -ldflags "-X 'github.com/percona/pmm/version.Version=3.8.0'" \
  -o /tmp/pmm-managed-fixed ./cmd/pmm-managed

docker volume create pmm-fix-data
docker create --name pmm-fix -v pmm-fix-data:/srv percona/pmm-server:3.8.0
docker cp /tmp/pmm-managed-fixed pmm-fix:/usr/sbin/pmm-managed
docker start pmm-fix
```

Result: zero `GRANT` syntax errors, migration completes on the first pass in 341 ms with no
retry and no warning, and the grants are in place:

```
 pmm-managed | {=Tc/postgres,postgres=CTc/postgres,"\"pmm-managed\"=CTc/postgres"}
```

```
 db_create
-----------
 t
```

### Cleanup

```bash
docker rm -f pmm-repro pmm-fix
docker volume rm pmm-repro-data pmm-fix-data
```

## Problem

On the first start over an empty `/srv`, PostgreSQL logs:

```
ERROR:  syntax error at or near "$1" at character 34
STATEMENT:  GRANT ALL PRIVILEGES ON DATABASE $1 TO $2
```

`initWithRoot` passed the database name and the role name as bind parameters. PostgreSQL
does not accept bind parameters in place of identifiers, so the statement fails to parse and
the role never receives any privileges on the database.

This only happens when the role is genuinely missing, which in Docker is exactly once per
volume: `post-build.yml` wipes `/srv` at image build time, and `entrypoint.sh` re-runs
`initdb` creating only the `postgres` superuser. Every later start finds the role and skips
this code path entirely.

The failure has been invisible so far because `migrateDB` retries the whole setup every
second, and on the second pass the role already exists and `initWithRoot` is not called at
all. The missing privileges do no harm on PostgreSQL 14, where `PUBLIC` still holds `CREATE`
on schema `public` — but that stops being true from PostgreSQL 15 on, at which point a fresh
install would spin in the retry loop until `Could not migrate DB: timeout`.

## Changes

- Quote identifiers into the statement with `pq.QuoteIdentifier` instead of binding them.
- Move `GRANT` out of the `countRoles == 0` branch. It is idempotent, and running it on both
  paths means a role that already exists but was provisioned without privileges gets them too.
- Quote passwords with `pq.QuoteLiteral` in `CREATE USER` / `ALTER USER`. They were being
  interpolated into bare single quotes, so a password containing an apostrophe broke the SQL.
- Also quote the `CREATE DATABASE` identifier the same way, so the whole function is consistent.
- Split the DB/role provisioning out of `initWithRoot` into `provisionDBAndRole(db, params)`.
  `initWithRoot` opens its own connection and reads `/srv/.postgres_password`, so nothing in it
  was reachable from a unit test; the new function takes a `*sql.DB` and is covered by sqlmock.

## Tests

New `TestProvisionDBAndRole` in `managed/models/database_init_test.go` covers the fresh-install
path, the existing-role path, and quoting of identifiers and passwords with special characters.
Verified that reverting just the `GRANT` line makes the first two subtests fail with
`could not match actual sql: "GRANT ALL PRIVILEGES ON DATABASE $1 TO $2"`.

`gofmt` and `go vet` are clean. `golangci-lint` was not run locally (the installed binary is
built for another architecture) and is left to CI.

- [ ] API Docs updated — n/a, no API changes.


theTibi:
@coderabbitai full review

PR #5804, opened 2026-08-20
feat: support assumed IAM roles for AWS RDS

fergalhk:
[PMM-15389](https://perconadev.atlassian.net/browse/PMM-15389)

[Feature build](https://github.com/Percona-Lab/pmm-submodules/pull/4548#top)

# Why?

Long-lived AWS access keys are currently required to operate PMM in a multi-account AWS architecture. The use of access keys, rather than STS credentials assumed via another role, is broadly considered to deviate from [AWS best practice](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html):

> As a best practice, use temporary security credentials (such as IAM roles) instead of creating long-term credentials like access keys. Before creating access keys, review the alternatives to long-term access keys.

# What?

This PR introduces the ability for Percona components to assume an IAM role using ambient credentials. This functionality is already [supported by RDS exporter](https://github.com/percona/rds_exporter/blob/915aceb/config/config.go#L25).

[PMM-15389]: https://perconadev.atlassian.net/browse/PMM-15389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


theTibi:
@coderabbitai full review

theTibi:
@coderabbitai this two tickets could be related: https://perconadev.atlassian.net/browse/PMM-12958 and https://perconadev.atlassian.net/browse/PMM-13045

managed/services/management/rds.go
```diff
@@ -171,12 +186,37 @@ func (s *ManagementService) DiscoverRDS(ctx context.Context, req *managementv1.D
...
-			regInstances, err := discoverRDSRegion(ctx, cfg, region)
+			regCfg := cfg
+			if req.AwsRoleArn != "" {
+				// STS is partition-scoped, so the role is assumed in the region being scanned.
+				regCfg.Region = region
+				regCfg.Credentials = assumeRoleProvider(regCfg, req.AwsRoleArn)
+
+				// Resolve now: once the RDS call wraps this, the outer error names RDS and the
+				// assumption failure is no longer distinguishable from any other region error.
+				_, err := regCfg.Credentials.Retrieve(ctx)
```
4nte: Since assumeRoleProvider is constructed for every region, it's cache will remain cold and every region will produce an additional network request to the AWS API. That means 29 additional requests instead of just one.

Based on AWS docs, session tokens from regional AWS STS endpoints are [valid in all AWS Regions](https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-regions.html#:~:text=Session%20tokens%20from%20regional%20AWS,that%20are%20enabled%20by%20default.), meaning it's enough to obtain the session token once and re-use it for all consequent regions in the discovery.

Looking at the commit history I see that you already attempted this optimization and that the default region was selected via `slices.Min(regions)` which would return `af-south-1`, however, this is an **opt-in** region, which means that if the target AWS account doesn't have this region enabled (most will not), the request will fail. Instead I think it's acceptable to use any "AWS default region" as a default region, such as `us-east-1`.

managed/services/management/rds.go
```diff
@@ -536,3 +583,10 @@ func (s *ManagementService) addRDS(ctx context.Context, req *managementv1.AddRDS
 	}
 	return res, nil
 }
+
+// assumeRoleProvider returns a credentials provider that assumes roleARN using the
+// credentials and region already resolved in cfg. The provider is cache-wrapped so the SDK
+// refreshes the assumed credentials before they expire.
+func assumeRoleProvider(cfg aws.Config, roleARN string) aws.CredentialsProvider { //nolint:ireturn
+	return aws.NewCredentialsCache(stscreds.NewAssumeRoleProvider(sts.NewFromConfig(cfg), roleARN))
```
4nte, line 581: This cache will remain cold since the caller constructs it for every region, instead of per-partition.

managed/services/agents/rds.go
```diff
@@ -83,6 +84,7 @@ func rdsExporterConfig(pairs map[*models.Node]*models.Agent, redactMode redactMo
 			Instance:               node.InstanceID,
 			AWSAccessKey:           exporter.AWSOptions.AWSAccessKey,
 			AWSSecretKey:           exporter.AWSOptions.AWSSecretKey,
+			AWSRoleArn:             exporter.AWSOptions.AWSRoleARN,
```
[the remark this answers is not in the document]
4nte: minimum supported pmm-agent version that supports assumed roles is `3.4.0-0`

4nte (changes requested):

documentation/docs/install-pmm/install-pmm-client/connect-database/aws.md
```diff
@@ -101,7 +101,7 @@ The `AmazonRDSforPMMPolicy` is now added to your IAM user.
```
ademidoff, line 101: @4nte @fergalhk 

Folks, the docs cannot be merged with this PR. We are merging PRs well before the release which is why we don't want them to show up before PMM is released. 

May I ask you to extract them to a separate PR while restoring the .md files from `main`, please?

I know, it's inconvenient, but that's what it is :(

ademidoff (changes requested):

documentation/docs/install-pmm/install-pmm-client/connect-database/aws.md
fergalhk: No problem @ademidoff - I'll pull these into their own PR.

managed/services/management/rds.go
fergalhk: > Based on AWS docs, session tokens from regional AWS STS endpoints are [valid in all AWS Regions](https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-regions.html#:~:text=Session%20tokens%20from%20regional%20AWS,that%20are%20enabled%20by%20default.), meaning it's enough to obtain the session token once and re-use it for all consequent regions in the discovery.

The tradeoff is that an outage in the region we're using for STS (likely us-east-1) will prevent us from acquiring the token from STS, even if the database is in a different region. As this is a control plane component this feels like an acceptable tradeoff, but flagging for awareness.

I'll make this change.

documentation/docs/install-pmm/install-pmm-client/connect-database/aws.md
fergalhk: #5838

theTibi:
@coderabbitai full review

managed/services/management/rds.go
```diff
@@ -536,3 +573,35 @@ func (s *ManagementService) addRDS(ctx context.Context, req *managementv1.AddRDS
...
+
+// stsDefaultRegion maps an AWS partition to a region always enabled in that partition,
+// used to reach STS once for a role instead of once per scanned region. A session
+// token issued in this region is valid across every region in the partition.
+var stsDefaultRegion = map[string]string{
+	"aws":        "us-east-1",
+	"aws-cn":     "cn-north-1",
+	"aws-us-gov": "us-gov-west-1",
+	"aws-iso":    "us-iso-east-1",
+}
```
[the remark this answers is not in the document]
fergalhk: This was a conscious decision, because [we don't support `aws-iso-b`](https://github.com/percona/pmm/blob/6729e04bf33bb1546605fbaa4bf16c575df7c5b3/managed/utils/validators/validators.go#L84-L107). Happy to update if reviewers feel differently.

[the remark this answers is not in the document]
4nte: @coderabbitai This was a conscious decision, aws-iso-b is not supported today.

PR #5805, opened 2026-08-20
Improve pnpm setup and linking

fabio-silva:
- Ditch swc plugin in favor of regular React plugin for Vite (improved performance);
- Fix pnpm link to local peak-ui, with instructions


mattiasimonato (approved):

docker-compose.dev.yml
```diff
@@ -3,7 +3,7 @@ services:
   pmm-server:
     profiles:
       - pmm
-    image: "${PMM_SERVER_IMAGE:?Fatal: .env file is not provided}"
+    image: '${PMM_SERVER_IMAGE:?Fatal: .env file is not provided}'
```
matejkubinec, line 6: Any reason for this change?

matejkubinec (approved):

docker-compose.dev.yml
fabio-silva: good point, must have been from an old format. Will revert it in another PR (I merged this one without realizing your comment).

PR #5808, opened 2026-08-20
PMM-15361 Widen the healthcheck grace to 90s

yyyyyyyan:
A cold `docker compose up` marks `pmm-server` unhealthy for a few seconds before it becomes healthy, so anything gated on `depends_on: {pmm-server: {condition: service_healthy}}` aborts when compose polls inside that window. `pmm-client` in our own `docker-compose.yml` is gated that way.

## What happens

`--start-period=25s` stops absorbing probe failures before a cold boot reaches `readyz`. Three failures at 4s intervals then mark the container unhealthy at roughly 37s, and it recovers a few seconds later.

Measured on a cold boot against an empty volume, fast local storage:

```
t=0s   starting
t=34s  UNHEALTHY        <- 4 failed probes after the start period
t=34s  readyz first pass
t=38s  healthy
```

With a dependent gated on `condition: service_healthy`, three cold runs gave two passes and one failure:

```
dependency failed to start: container pmm-server is unhealthy
```

A separate measurement on slower hardware put the cold start near 90s, which widens the window and loses the race far more often than one in three.

## The change

`--start-period` 25s → 90s, with a comment recording what the number is sized against.

A healthy boot is unaffected — the start period ends at the first passing probe, so nothing waits longer than it does today. The cost is that a genuinely wedged server reports unhealthy after about 102s rather than thirty-seven seconds. This opened at 180s; review preferred keeping that verdict tight enough to still catch a startup regression, and 90s clears every cold boot measured here.

## Notes

Present since the EL9 image landed in 2023 (`7a6862e18`); not a recent regression. It surfaced during PMM-15205 work because that stack briefly carried a 720s start period for unrelated reasons, which masked it.

The comment matters as much as the value here. The 720s above went unexamined for months because its stated rationale had stopped being the real one, and it was then reverted without anyone re-deriving what it was for.

## Tested

- [x] Cold boot timings above, captured from `docker inspect` health transitions on `perconalab/pmm-server-fb:PR-4500-7dcd575`.
- [x] Three cold `docker compose up` runs with a dependent gated on `service_healthy`, before the change: 2 pass, 1 fail.
- [ ] Not re-run against an image built from this branch — the change is to the `HEALTHCHECK` instruction itself, so it needs a build to exercise. Worth confirming in a Feature Build that the instruction lands as intended and that the same gate passes repeatedly.


build/docker/server/Dockerfile.el9
```diff
@@ -49,6 +49,10 @@ VOLUME [ "/srv" ]
 
 EXPOSE 8080 8443
 
-HEALTHCHECK --interval=4s --timeout=2s --start-period=25s --retries=3 CMD curl -sf http://127.0.0.1:8080/v1/server/readyz
+# Sized on a cold boot against an empty volume, measured at 34s to first pass readyz
+# on fast local storage and near 90s on slower hardware. At 25s the container was
+# marked unhealthy before it was ready, and a dependent gated on service_healthy
+# aborted whenever compose polled inside that window.
```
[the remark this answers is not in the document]
yyyyyyyan: Both correct — fixed in 042630ba5.

The 25s was the grace period ending, not the unhealthy mark. The probe run I measured this from shows the distinction: four failures after the grace period, then `unhealthy` at t=34s, `readyz` passing at the same tick, `healthy` at 38s. So the comment was describing the wrong event.

The second point is the one that would have aged worse. Written as "At 25s", one line above `--start-period=180s`, it reads as current configuration rather than as what was replaced. The comment now says "the previous 25s grace period".

build/docker/server/Dockerfile.el9
```diff
@@ -49,6 +49,11 @@ VOLUME [ "/srv" ]
 
 EXPOSE 8080 8443
 
-HEALTHCHECK --interval=4s --timeout=2s --start-period=25s --retries=3 CMD curl -sf http://127.0.0.1:8080/v1/server/readyz
+# Sized on a cold boot against an empty volume: 34s to first pass readyz on fast
+# local storage, near 90s on slower hardware. The previous 25s grace period ended
+# before that, so three failed probes marked the container unhealthy around 34s,
+# and a dependent gated on service_healthy aborted whenever compose polled during
+# the few seconds before it recovered.
+HEALTHCHECK --interval=4s --timeout=2s --start-period=180s --retries=3 CMD curl -sf http://127.0.0.1:8080/v1/server/readyz
```
ademidoff:
```suggestion
HEALTHCHECK --interval=4s --timeout=2s --start-period=90s --retries=3 CMD curl -sf http://127.0.0.1:8080/v1/server/readyz
```

We have always tried to keep the this value as low as possible to make sure we optimize the container startup time. A higher-than-necessary window could lead to us not being able to capture a regression responsible for the slow start.

I agree that bringing SEP that depends on a healthy start of PMM requires us to review the stance, but making this value too relaxed is probably not a good option either. I guess we need to find middle ground.

@yurkovychv What is your take on this?

yyyyyyyan:
**Corroborating measurement from a second machine**

Hit this on the PMM+SEP paired feature-build stack, where a SEP side-car is gated on `depends_on: {pmm-server: {condition: service_healthy}}`.

- Cold `docker compose up -d`: pmm-server reaches healthy in **~72s** (1m12.350s wall clock on the timed run; other runs in the same range).
- With the shipped `--start-period=25s --interval=4s --retries=3`, compose aborts the dependent **every time**: `dependency failed to start: container … is unhealthy`. Not intermittent — every attempt, across two feature-build tags (`PR-4500-7dcd575` and `PR-4500-73298a4`).
- pmm-server converges to healthy shortly after, so the container itself is fine; the verdict just lands ~37s in, by which point the dependent is already gone.
- Overriding `start_period` to 300s in that stack's compose makes `up -d` succeed in one invocation. **72s sits comfortably inside the 180s proposed here.**

Worth it for the framing: the 34s on fast local storage is 3s *under* the 37s threshold, so it reads as a flake; ~72s is 35s *over*, so it fails deterministically. Same defect — the two numbers bracket it better than either alone does.

build/docker/server/Dockerfile.el9
yurkovychv: @ademidoff you are right. By keeping the healthcheck in reasonable timerange we can identify performance degradation. Due to this approach we've been able to identify issues in the past

ademidoff: @yyyyyyyan Would the value proposed above work for SEP?

yyyyyyyan: @ademidoff I ran more tests and yes, 90s clears it - the SEP stack cold-boots pmm-server to healthy in ~72s, so a 90s grace absorbs that with 18s to spare and the container never leaves `starting`. Taking the suggestion would not break the case that prompted this.

Two things I'd weigh before landing on it:

1. The first is what a wider start period actually costs, because I think it's less than it looks. The grace doesn't run for its full duration — it ends at the **first passing probe**. Measured against a live daemon: with a 120s start period, a probe that succeeds at t=8s and then starts failing puts the container `unhealthy` at t=12s. So once a server has been ready even once, it's back on the normal `--interval=4s --retries=3` budget whatever the start period says, and a startup regression that still reaches `readyz` is caught in ~12s at 25s, at 90s and at 180s alike. What a wider grace relaxes is only the verdict on a container that has **never** been ready: ~102s at 90s, ~192s at 180s, against ~34s today.
2. The second is margin. We have three observations — 34s on fast NVMe, ~72s on the SEP stack, near 90s on slower hardware — not an upper bound. A 90s grace sizes the budget at the slowest one measured, so that case survives only on the ~12s the three post-grace retries buy. That's structurally the same shape as the bug this fixes: 25s against a 34s boot was also nearly enough, and it still lost one cold run in three. The 72s is the loaded figure too, measured with a dependent and its database co-starting, which is the situation `condition: service_healthy` exists for.

So I'd prefer to meet at **120s**: it clears every measurement with ~40s of headroom instead of ~12s, and still flags a wedged server inside two minutes rather than three. If the goal is "much lower than 180s", 120s gets most of that without sizing the grace at the slowest boot on record.

If you'd both still rather keep it tighter, I'm ok keeping 90s as it does clear the case I measured!!

ademidoff: > The grace doesn't run for its full duration

Actually it does in case there is a regression causing PMM to be unable to start within the grace period. Therefore, I'd rather keep it at 90s for now.

yyyyyyyan: Agreed on the mechanism — a server that never reaches readyz does burn the whole grace, which makes the grace the verdict time. That was the trade I was weighing against margin, and I'm happy to land on your side of it.

Applied as suggested in 61c266b3e. The comment above the instruction is unchanged: it records the measurements rather than the chosen value, so it still reads correctly at 90s.

build/docker/server/Dockerfile.el9
```diff
@@ -49,6 +49,11 @@ VOLUME [ "/srv" ]
 
 EXPOSE 8080 8443
 
-HEALTHCHECK --interval=4s --timeout=2s --start-period=25s --retries=3 CMD curl -sf http://127.0.0.1:8080/v1/server/readyz
+# Sized on a cold boot against an empty volume: 34s to first pass readyz on fast
+# local storage, near 90s on slower hardware. The previous 25s grace period ended
+# before that, so three failed probes marked the container unhealthy around 34s,
+# and a dependent gated on service_healthy aborted whenever compose polled during
+# the few seconds before it recovered.
+HEALTHCHECK --interval=4s --timeout=2s --start-period=90s --retries=3 CMD curl -sf http://127.0.0.1:8080/v1/server/readyz
```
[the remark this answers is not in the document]
yyyyyyyan: Good catch on the blind spot — an image-level `HEALTHCHECK` genuinely does stop at the first Compose file that replaces it, and that is worth checking rather than assuming. I went through both files, and they land differently.

`qa-integration/pmm_psmdb_diffauth_setup/docker-compose-pmm-psmdb.yml:70-78` is the real one. That override drops `start_period` altogether, so with `interval: 3s` and `retries: 20` the verdict lands about 60s after container start — inside the 72–90s cold boot measured for this ticket — and `psmdb` at line 65 is gated on `condition: service_healthy`. That stack can still abort, and nothing this PR does reaches it.

`codeceptjs-e2e/docker-compose-clickhouse.yml:33-38` I don't think is exposed. `start_period: 60s` with `interval: 5s` and `retries: 10` puts its verdict near 110s, which is more headroom than the image's own 90s + 3×4s ≈ 102s. It also already carries a comment describing this exact failure, so it looks like it was widened for the same reason and is ahead of us rather than behind.

I swept this repo for the same shape while I was at it: `docker-compose.yml`, `api-tests/docker-compose.yml` and `qan-api2/docker-compose.yml` each have a `service_healthy` dependent on `pmm-server`, and none of them override the image healthcheck — so the change does reach every in-repo consumer.

The psmdb fix lives in percona/pmm-qa, so it can't ride along here. Tracking it as a follow-up rather than leaving this thread open.

yyyyyyyan:
Add a start period to the pmm-server healthcheck override in the psmdb diffauth stack.

Deferred from the review of this PR. The `--start-period=90s` landing here is an image-level setting, so it does not reach a Compose file that replaces the healthcheck — and `qa-integration/pmm_psmdb_diffauth_setup/docker-compose-pmm-psmdb.yml:70-78` in percona/pmm-qa is one of those. It sets `interval: 3s`, `timeout: 2s` and `retries: 20` with no `start_period`, so the unhealthy verdict lands around 60s while the cold boot measured for PMM-15361 runs 72–90s. The `psmdb` service at line 65 waits on `condition: service_healthy`, so it aborts.

The remedy is to give that override `start_period: 90s`, or drop the override and let it inherit the image's, then re-run the stack cold against an empty volume. It belongs in percona/pmm-qa, which is why it isn't part of this PR.

Raised by CodeRabbit on this PR. The second file it cited, `codeceptjs-e2e/docker-compose-clickhouse.yml:33-38`, already allows about 110s before it gives a verdict and needs nothing.

build/docker/server/Dockerfile.el9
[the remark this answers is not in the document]
yyyyyyyan: No need, thanks — PMM work is tracked in Jira rather than GitHub issues, and the follow-up is already queued from the deferral comment on this PR, so a GitHub issue would just be a second record of the same thing.

Resolving here.

PR #5811, opened 2026-08-21
PMM-15326: Stop logging the SEP URL and token values

plebioda:
Ticket number: PMM-15326

Feature build: N/A - not user-facing. This only changes what is written to the log at trace level; no API, UI, metric or dashboard surface is touched.

## What

`ParseEnvVars` no longer writes the values of `PMM_SEP_URL` and `PMM_SEP_TOKEN` to the log. Both are logged by name only.

## Why

OM (PMM-15326) adds `--sep-url` and `--sep-token` as kingpin flags with `.Envar("PMM_SEP_URL")` / `.Envar("PMM_SEP_TOKEN")`, so both land in the process environment that `ParseEnvVars` walks - it is called with `os.Environ()` in `pmm-managed-init`, and with the server's environment in `services/server`.

That function traces every variable it sees **before** the switch that decides what to do with it:

```go
logrus.Tracef("ParseEnvVars: %#q: k=%#q v=%#q", env, k, v)
```

`env` is the whole original `KEY=value` assignment, so the value is written three times over. The `case` that skips SEP's variables is forty lines below that line, which means a bearer token for SEP's API was logged on the way to being ignored.

This is reachable in a real deployment: `PMM_TRACE` is a live (if deprecated) environment variable, so an operator turning trace on to debug something unrelated writes the token into the log.

`PMM_SEP_URL` is redacted alongside the token because a URL can carry credentials in its userinfo, and a rule that relies on nobody ever writing one that way is not a rule.

## For reviewers: this is scoped to OM's two variables, and it is not the whole problem

The trace line predates SEP - it has been there since pmm-managed moved into this repo (`50219e654`, PMM-10206) - and every secret that reaches it is logged. I tested each one that flows through `ParseEnvVars`:

| variable | after this PR |
| --- | --- |
| `PMM_SEP_TOKEN` | redacted |
| `PMM_SEP_URL` | redacted |
| `PMM_SEP_POSTGRES_PASSWORD` | still logged |
| `PMM_CLICKHOUSE_PASSWORD` | still logged |
| `PMM_POSTGRES_DBPASSWORD` | still logged |
| `AWS_SECRET_KEY` | still logged |
| `AWS_ACCESS_KEY` | still logged |

All five of the remaining ones are named in the same `switch` as "not a setting, skip", and all five have their values traced before they get there. `PMM_SEP_POSTGRES_PASSWORD` is the sharpest case: it sits two lines from the `PMM_SEP_URL` case this PR adds.

I deliberately did not widen the fix, because this PR belongs to OM and that leak is not OM's. **The call is yours**, and there are three shapes:

1. **Widen `secretEnvVars`** to cover all seven. A few more map entries in a file this PR already touches, and the redaction becomes true in general rather than for OM's two variables.
2. **Leave it here and open a follow-up ticket.** What this PR does today.
3. **Invert the default** - log the key at trace level and never the value, for everything. Strictly safest and simpler than a denylist, but it removes a debugging aid that people may rely on, so it is a PMM-team decision rather than something to slip into an OM PR.

Happy to do 1 or 3 in this PR if you would rather not carry a follow-up. The comment on `secretEnvVars` states the current scope explicitly so the map is not misread as a complete list of secrets.

## Why the guard is a `continue` and not a switch case

An earlier revision of this also added `case "PMM_SEP_URL", "PMM_SEP_TOKEN":` to the switch, to keep the list of "not a server setting" keys complete in one place. Codecov caught that it was unreachable - it was the whole of the missing patch coverage - so it is gone.

That makes the guard's `continue` do two jobs, which is worth knowing when reading it: it redacts, and it classifies. These keys now have no case in the switch, so without the guard they would fall through to the `default` branch and hit the unknown-variable warning at `parser.go:350`, which appends `env` - the whole `KEY=value` assignment - at a level operators actually read. That would be a worse leak than the trace line this PR set out to fix.

`TestSEPSecretsAreNotLogged` asserts no warnings alongside the redaction, so both jobs are covered.

## Tests

`TestSEPSecretsAreNotLogged` captures the standard logger at trace level, since `Tracef` is where the value leaked, and asserts that neither the token nor a password placed in the URL's userinfo appears in any log entry - while both keys still appear by name, which is what keeps the skip auditable.

## Related work

First of seven stacked PRs splitting #5795 ("OpenManager initial implementation") into area-owned reviews, all targeting `PMM-15299-open-manager`. This one is independent of the rest and can be reviewed and merged on its own. The others will be linked here as they open.

### DIFF
--- a/managed/utils/envvars/parser.go
+++ b/managed/utils/envvars/parser.go
@@ -123,6 +123,9 @@ func ParseEnvVars(envs []string) (*models.ChangeSettingsParams, []error, []strin
 		case "PMM_ENABLE_SEP", "PMM_SEP_POSTGRES_PASSWORD":
 			// skip env variables consumed by the entrypoint to expose postgres to SEP
 			continue
+		case "PMM_SEP_URL", "PMM_SEP_TOKEN":
+			// skip SEP connection settings, parsed separately
+			continue
 		case "PERCONA_TELEMETRY_DISABLE":
 			// skip the Pillars telemetry environment variable
 			continue
@@ -330,6 +333,12 @@ var secretEnvVarMarkers = []string{"PASSWORD", "SECRET", "TOKEN", "KEY"}
 
 // redactSecretEnvVar replaces the value of a credential-bearing environment variable.
 func redactSecretEnvVar(key, value string) string {
+	if key == "PMM_SEP_URL" {
+		// A URL can carry credentials in its userinfo, and none of the markers below
+		// match "URL".
+		return "<redacted>"
+	}
+
 	for _, marker := range secretEnvVarMarkers {
 		if strings.Contains(key, marker) {
 			return "<redacted>"
--- a/managed/utils/envvars/parser_test.go
+++ b/managed/utils/envvars/parser_test.go
@@ -20,7 +20,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/percona/pmm/managed/models"
 )
@@ -278,6 +281,66 @@ func TestEnvVarValidator(t *testing.T) {
 	})
 }
 
+// TestSEPSecretsAreNotLogged pins the reason the SEP variables are a special case
+// rather than two more entries in the skip list: their values must not reach a log
+// at any level. The trace line that used to leak them carried the whole original
+// assignment, forty lines before the switch that meant to ignore these keys.
+//
+// Not parallel. It raises the standard logger's level and installs a global hook,
+// both of which are process-wide.
+func TestSEPSecretsAreNotLogged(t *testing.T) {
+	const (
+		token    = "bearer-value-that-must-not-be-logged"
+		password = "userinfo-password-that-must-not-be-logged"
+	)
+
+	logger := logrus.StandardLogger()
+
+	// Captured before NewGlobal, which appends to the global logger's hooks. Reset
+	// only clears the entries a hook has collected, so without restoring the set the
+	// hook would outlive this test and keep collecting for the rest of the binary.
+	// The slices are copied too, because Add appends to them.
+	previousHooks := logrus.LevelHooks{}
+	for level, hooks := range logger.Hooks {
+		previousHooks[level] = append([]logrus.Hook(nil), hooks...)
+	}
+	previousLevel := logger.GetLevel()
+
+	hook := test.NewGlobal()
+	// Tracef is where the value leaked.
+	logger.SetLevel(logrus.TraceLevel)
+
+	t.Cleanup(func() {
+		logger.SetLevel(previousLevel)
+		logger.ReplaceHooks(previousHooks)
+	})
+
+	envs := []string{
+		"PMM_SEP_TOKEN=" + token,
+		// Credentials in the userinfo are why the URL is redacted alongside the token.
+		"PMM_SEP_URL=https://sep:" + password + "@sep.example.com:8000",
+	}
+
+	gotEnvVars, gotErrs, gotWarns := ParseEnvVars(envs)
+	assert.Equal(t, &models.ChangeSettingsParams{}, gotEnvVars)
+	assert.Nil(t, gotErrs)
+	assert.Nil(t, gotWarns)
+
+	entries := hook.AllEntries()
+	require.Len(t, entries, 2)
+
+	for _, entry := range entries {
+		line, err := entry.String()
+		require.NoError(t, err)
+		assert.NotContains(t, line, token)
+		assert.NotContains(t, line, password)
+	}
+
+	// Logged by name, which is what keeps the skip auditable at trace level.
+	assert.Contains(t, entries[0].Message, "PMM_SEP_TOKEN")
+	assert.Contains(t, entries[1].Message, "PMM_SEP_URL")
+}
+
 func TestRedactSecretEnvVar(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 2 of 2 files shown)


plebioda:
@coderabbitai review

managed/utils/envvars/parser_test.go
```diff
@@ -277,3 +280,52 @@ func TestEnvVarValidator(t *testing.T) {
...
+
+	hook := test.NewGlobal()
+	logger := logrus.StandardLogger()
+
+	previousLevel := logger.GetLevel()
+	logger.SetLevel(logrus.TraceLevel) // Tracef is where the value leaked.
+	t.Cleanup(func() {
+		logger.SetLevel(previousLevel)
+		hook.Reset()
+	})
```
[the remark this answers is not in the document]
plebioda: Fixed - the finding is correct. `test.NewGlobal()` calls `logrus.AddHook(hook)` (hooks/test/test.go:22) and `Hook.Reset()` only does `t.Entries = make([]logrus.Entry, 0)` (test.go:87-91), so the hook stayed installed on the standard logger for the rest of the test binary.

The set is now captured before `NewGlobal` and restored with `logger.ReplaceHooks(previousHooks)` in `t.Cleanup`. The slices are copied as well as the map, because `LevelHooks.Add` appends to them and a shallow map copy would share the backing arrays.

Verified rather than assumed: a temporary test counting `logrus.StandardLogger().Hooks` after this one ran reported 1 hook remaining before the change and 0 after.

managed/utils/envvars/parser_test.go
```diff
@@ -277,3 +280,52 @@ func TestEnvVarValidator(t *testing.T) {
...
+	const (
+		token    = "bearer-value-that-must-not-be-logged"
+		password = "userinfo-password-that-must-not-be-logged"
+	)
+
+	hook := test.NewGlobal()
+	logger := logrus.StandardLogger()
+
+	previousLevel := logger.GetLevel()
+	logger.SetLevel(logrus.TraceLevel) // Tracef is where the value leaked.
```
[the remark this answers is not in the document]
plebioda: Fixed. `logger.SetLevel(logrus.TraceLevel) // Tracef is where the value leaked.` was indeed an inline comment, and AGENTS.md permits those only for `//nolint`. Moved to the preceding line.

managed/utils/envvars/parser.go
```diff
@@ -82,6 +98,20 @@ func ParseEnvVars(envs []string) (*models.ChangeSettingsParams, []error, []strin
...
+			// have no case in the switch below, so without this branch they would reach
+			// the trace line (which carries `env`, the whole original assignment) and
+			// then fall through to the unknown-variable warning, which appends `env` too.
+			// Either one would put a bearer token, or a URL with credentials in it, into
+			// the log - the warning at a level operators actually read.
+			//
+			// So the continue is doing two jobs: it redacts, and it classifies these as
+			// "not a server setting". TestSEPSecretsAreNotLogged covers both, by
+			// asserting no warnings alongside the redaction.
+			logrus.Tracef("ParseEnvVars: %#q: value not logged", k)
```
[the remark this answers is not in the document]
plebioda: Declining this one, with a reason.

`parser.go` uses `logrus.Tracef` twice and `logrus.WithField` zero times. The line immediately below the one you are commenting on is:

```go
logrus.Tracef("ParseEnvVars: %#q: k=%#q v=%#q", env, k, v)
```

Converting only the new line to a structured field would make it the single structured call in a function that logs with format strings throughout, which reads worse than either consistent choice. AGENTS.md asks to "match surrounding code in the file you edit", and that is what the current form does.

The general point about structured logging is fair, and `ParseEnvVars` would be better for it - but as a conversion of the whole function, not one line of it, and not inside a PR about redacting two credentials. Happy to open that as a follow-up if it is wanted.

On the test half of the suggestion: it currently asserts against `entry.String()`, the rendered line, rather than `entry.Message`. That is deliberate - it checks what would actually be written to the log, which is the property the PR is about, and it stays valid whether the call is `Tracef` or `WithField`.

managed/utils/envvars/parser.go
```diff
@@ -45,6 +45,22 @@ const (
...
+//
+// This is NOT the full set of secrets this function walks. PMM_SEP_POSTGRES_PASSWORD,
+// PMM_CLICKHOUSE_PASSWORD, PMM_POSTGRES_DBPASSWORD and AWS_SECRET_KEY are all named in
+// the switch below as "not a setting, skip", and all of them reach the trace line first,
+// so their values are still logged. That predates SEP and is deliberately left alone
+// here rather than fixed in passing.
+var secretEnvVars = map[string]bool{
+	"PMM_SEP_URL":   true,
+	"PMM_SEP_TOKEN": true,
+}
```
4nte: There is another mechanism for redaction (in this very file) is already implemented and landed on main recently: https://github.com/percona/pmm/blob/main/managed/utils/envvars/parser.go#L333

This PR branched before it landed so you need to rebase.

managed/utils/envvars/parser.go
4nte, line 323: Add skip case for SEP env vars, so they don't get flagged (and logged) as unknown
```go
case "PMM_SEP_URL", "PMM_SEP_TOKEN":
      // skip SEP connection settings, parsed separately
      continue
```

4nte (changes requested):

managed/utils/envvars/parser.go
plebioda: Done - rebased onto the updated PMM-15299-open-manager, which now carries `redactSecretEnvVar` ported from main's ba6720641 (without the unrelated ClickHouse datasource feature that commit shipped alongside). PMM_SEP_URL/PMM_SEP_TOKEN now reuse that mechanism instead of the separate `secretEnvVars` map.

This PR's own commit is now fully absorbed into the base (f1c5fb7a6 / 674a9144), so the branch has nothing left to merge. Closing as superseded rather than merging an empty diff.

managed/utils/envvars/parser.go
plebioda: Added, with your exact wording - see the switch case in the base commit (f1c5fb7a6, PMM-15299-open-manager). 

Also found the same workaround had been independently duplicated in #5795 (pom-inventory) - three of its own commits reimplemented this exact map-based fix. That branch is rebased onto the same base now too, so its copy drops out and both PRs end up sharing one fix instead of two.

plebioda:
This PR was auto-closed by GitHub since the fix landed directly in PMM-15299-open-manager (commits 674a9144 and f1c5fb7a6), so every branch based on it inherits it rather than needing its own merge. Replied to both of @4nte's comments above with the details.

Also found the same secret-redaction workaround had been independently duplicated in #5795; that's been rebased onto the same base so it no longer carries its own copy either.

PR #5815, opened 2026-08-21
PMM-15326: Authorize the OM routes by role

plebioda:
Ticket number: PMM-15326

Feature build: N/A - authorization rules only; the routes they cover are added by the API and service PRs in the same stack.

## What

Role rules for OM's routes in `managed/services/grafana/auth_server.go`, plus cases in `TestResolveRule` for each one.

Reading is viewer. The writes are named individually:

| route | role | why |
| --- | --- | --- |
| `GET /v1/om/**` | viewer | reading the estate |
| `POST /v1/om/inventory/runs:trigger` | editor | see below |
| `DELETE /v1/om/inventory/hosts/{id}` | admin | destructive to that row's history |
| `DELETE /v1/om/inventory/services/{id}` | admin | same |
| `PUT /v1/om/inventory/config` | admin | changes the sweep schedule for the whole deployment |
| `DELETE /v1/om/inventory/config/overrides/{key}` | admin | same |

## Please look at this: a viewer can trigger a topology collection

`POST /v1/om/topology/runs:collect` resolves to **viewer**, and the test asserts it.

The reasoning: a topology collection only reads PMM's own inventory and VictoriaMetrics. It runs no code on database hosts, takes nothing from the caller, and writes only OM's own run history. So it is closer to a refresh of a page than to a write.

The counter-argument is that it is still a `POST` that costs work and adds rows, and that "viewer" for anything that mutates state may not be a line you want crossed. **If you would rather it were editor, say so and it changes in one line.** I would rather have that decided here than have it noticed later as an accident.

## Why the refresh is editor and not admin

`POST /v1/om/inventory/runs:trigger` runs SEP's fixed probe payload on hosts SEP already covers. Nothing the caller supplies is executed, and no database is touched. The per-host refresh is the button beside a row that answers "I just fixed this, is it healthy now" - putting that behind admin would put the routine question behind the rarest role.

## Why the five gRPC write methods are named in `rules` rather than `methodRules`

`methodRules` is keyed `"METHOD /path"`, and a gRPC method name carries no verb. So `/om.v1.OmService/DeleteInventoryHost` would walk past every write rule and land on the `"/om."` prefix at viewer. The five write methods are therefore named exactly in `rules`, beside the agent and RTA endpoints that do the same thing.

In a normal deployment the gateway reaches pmm-managed over HTTP, so nothing uses these. They exist so the gRPC surface is not a way around the HTTP roles.

## Two details in the table worth a second look

**The trailing slash on the id-bearing paths is not load-bearing.** `nextPrefix` strips one as its own step, so the slash-less spelling catches the same requests. It is kept because it reads as "everything under here" rather than as a path.

**There are two rules for the config route on purpose.** The first names the route that exists; `DELETE /v1/om/inventory/config/` is a backstop, because `resolveRule` walks longest prefix first and anything else that ever appears under `config/` would otherwise fall through to the viewer rule at `/v1/om`. A DELETE anywhere under an app's configuration should default to admin, not to readable.

## Tests

`TestResolveRule` gains a case per route rather than a sample, because whether a write rule applies depends on how `resolveRule` walks prefixes, and that is not obvious from reading the rule table. The thing that would silently fall through to viewer is a write path with no rule at all, so each is asserted rather than trusted.

## Related work

One of seven stacked PRs splitting #5795 ("OpenManager initial implementation") into area-owned reviews. This one targets `PMM-15299-open-manager` directly, depends on nothing else in the stack, and can be reviewed and merged on its own - the rules are inert until the routes exist.

### DIFF
--- a/managed/services/grafana/auth_server.go
+++ b/managed/services/grafana/auth_server.go
@@ -65,6 +65,22 @@ var rules = map[string]role{
 	"/server.v1.":               admin,
 	"/qan.v1.CollectorService.": viewer,
 	"/qan.v1.QANService.":       viewer,
+	// Reading OM over gRPC is a viewer's business; the writes are named exactly below.
+	//
+	// The HTTP methodRules cannot cover these: that table is keyed "METHOD /path" and a
+	// gRPC method name carries no verb, so /om.v1.OmService/DeleteInventoryHost would walk
+	// past every write rule and land here on viewer. A gRPC method name is unambiguous on
+	// its own, which is why these sit in `rules` rather than being method-qualified, the
+	// same as the agent and RTA endpoints above.
+	//
+	// The gateway reaches pmm-managed over HTTP, so in a normal deployment nothing uses
+	// these. They exist so the gRPC surface is not a way around the HTTP roles.
+	"/om.": viewer,
+	"/om.v1.OmService/TriggerInventoryRefresh":       editor,
+	"/om.v1.OmService/DeleteInventoryHost":           admin,
+	"/om.v1.OmService/DeleteInventoryService":        admin,
+	"/om.v1.OmService/UpdateInventoryConfig":         admin,
+	"/om.v1.OmService/DeleteInventoryConfigOverride": admin,
 
 	"/v1/alerting":                    viewer,
 	"/v1/alerting/rules":              editor,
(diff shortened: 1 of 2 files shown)


plebioda:
@coderabbitai review

4nte:
> The counter-argument is that it is still a POST that costs work and adds rows, and that "viewer" for anything that mutates state may not be a line you want crossed. If you would rather it were editor, say so and it changes in one line.

I think this semantically matches to the viewer role. The fact that it mutates state is a technical detail and I don't think that's enough to naively categorize it as a privileged operation.

On the other side, if the concern is that triggering this action excessively could lead to a denial of service, I'd recommend implementing some kind of a throttling mechanism that would prevent this from happening regardless what user role is.

4nte (approved):

PR #5816, opened 2026-08-21
PMM-15326: Implement the OM service in pmm-managed

plebioda:
Ticket number: PMM-15326

Feature build: N/A on its own - this serves the OM API but nothing renders it until the UI PRs land. Happy to produce one alongside the UI branches if that is more useful than one per PR.

## What

`managed/services/om` implements `OmService`, and `cmd/pmm-managed/main.go` wires it in - registered on the gRPC server, on the HTTP gateway, and started on a ticker under the HA leader service.

Two halves, sharing a gRPC facade.

**Topology.** Derives a MongoDB view from PMM's own data. Two sources feed it: `inventorySource` reads services and nodes from `managed/models`, `metricsSource` evaluates a fixed set of queries against VictoriaMetrics. Facts from each source are merged by a documented precedence (`facts.go`), and the merge records which source won each field and how old it was. `projection.go` turns merged facts into the emitted document; `store.go` persists the run and snapshot through the models added in #5812.

**Inventory.** Proxies SEP's `om_inventory` app - hosts, services, runs, config - so that no browser path needs a SEP bearer. pmm-managed holds the credential; the Grafana session authorises the caller.

## What this PR is not: a topology graph

The document is a **grouped service inventory**, and the wording throughout now says so. Services are grouped by `(environment, cluster-or-replication-set label)` and each row is stamped with a `process_role`; `ClusterType` is the one thing inferred about a group's shape. What it does *not* do is reconstruct a shard map - a sharded cluster appears as a flat list of mongos, configsvr and shardsvr rows that share a label, not as shards with their members.

That is deliberate for this PR rather than an omission, and `schema_version` is how a later build grows `replica_sets` / `shards` without a retraction. Calling it "reconstructs replica sets and sharded clusters", as an earlier draft of the package comment did, would have been the version worth objecting to.

## The one structural thing to know before reading

`sep_client.go` owns everything about talking to SEP: base URL, token, HTTP client, timeout, status mapping. Callers ask it for an app handle (`.app(module)`) whose paths are relative to `/api/apps/<module>`, and hold that handle rather than each other.

So `probeSource` is now only a `factSource` - it collects on-host facts for the topology merge - and the inventory handlers hold their own handle instead of reaching through the probe source for its transport. The next SEP-backed consumer (actions, when they land) asks for a third handle rather than copying the transport or hanging its verbs off a type called `probeSource`.

Dependency direction is one-way and worth checking if you disagree with the layout: nothing on the topology side references any inventory symbol.

## Behaviours that are decisions, not accidents

**Reads never collect.** `GetTopology` serves memory, then the stored snapshot, and never runs a collection. Collection happens on a 30s ticker on the HA **leader**, which is where it belongs: a pass persists a run and then prunes shared history. `TriggerTopologyCollection` refuses on a non-leader with `FailedPrecondition`, naming the leader when it knows it.

**A cold estate answers, and says so.** No memory, no snapshot, leader has not ticked yet: the response is an empty document with `stale = true` and `observed_at` unset - not an error. An empty document therefore means "nothing collected yet", never "your estate is empty", and the proto comment says that where a client will read it. Staleness is recomputed against the clock on every read rather than frozen when the document was built, so a restored snapshot is reported by its real age.

**SEP is optional.** With no `PMM_SEP_URL`, the probe source reports itself disabled and the document is built from PMM's own inventory and metrics alone. OM degrades rather than failing.

**The kingpin flags point at SEP, not at one of its apps.** `--sep-url` is the base URL; each consumer appends its own `/api/apps/<module>` path.

**The document is schema-versioned** (`schemaVersion = 1`), so a stored snapshot written by an older build is recognisable rather than misread. It starts at 1 because nothing has shipped - there is no stored snapshot anywhere that predates this, so there is nothing for a higher number to be compatible with.

**`staleAfter` is 5 minutes**; past that a fact is reported stale rather than silently trusted.

**Per-source outcomes are recorded on the run.** This is the part I would most like a second opinion on: it exists so that a thin document is distinguishable from a broken collector. A run that produced little because a source failed looks different from one that produced little because there is little to see.

**Cluster rows carry a server-issued opaque id.** `Cluster.id` is derived deterministically from `(environment, cluster, replication_set)`. Clients must not parse it: two clusters can and do share a display label - two generations of a sandbox both called `sharded-cluster` - and the label is the thing that may change without the row being a different cluster.

**Provenance, one field of it.** `TopologyService.observed_at` carries the age of the newest live field on that row, so one lagging service in an otherwise fresh fleet is visible. Deliberately deferred: a per-field `map<string, string> field_sources`. The merge does record which source won each field, and the run receipts hold it, but nothing consumes it yet and it is the expensive half on the wire. Worth adding when something reads it, not before.

**Facts stay keyed by PMM service id.** A naturally per-host attribute - `repo_reachable`, `executor_reachable` - has nowhere to go in the merge, and that is a known cost written down in `facts.go` rather than an oversight. Host attributes live on `/v1/om/inventory/hosts` instead. The alternative, widening `Fact` with an optional node key so the merge becomes `(node|service, field)`, forces a decision about whether a host fact attaches to every service on that node or to a new slot, and that is bigger than this PR. What it buys by waiting: the copies cannot silently disagree, because there are no copies.

**`applyHealth` is a named no-op.** `package om` is documented as the topology *and health* API, and `projection.go` says the document reports observations rather than verdicts. Both can be true only if there is a place the health rules are supposed to go, so there is one: a pipeline stage called from `collect` after `buildDocument`, plus a reserved `ClusterHealth` enum. It does nothing today. It exists so that the first health rule does not land as another `if` on `exporter_up` inside the projection.

## Reviewing this

~6.1k lines, of which ~2.4k are tests. The parts carrying the most judgement, in order: `facts.go` (the precedence and staleness rules), `projection.go` (what the document actually says), `service.go` (the read/collect split and the leader rules), then `inventory.go` (error mapping and the config-write validation).

`inventory.go` bounds what a config write may contain - depth and shape - because those values are passed through to SEP. Worth a look from that angle specifically. Its 401 mapping is also deliberate: a 401 from SEP is pmm-managed's credential failing, not the caller's session expiring, and the two must not be reported the same way.

`service_test.go` and `store_test.go` cover the lifecycle, which previously had no tests at all: restore-then-serve never reaching `collect`, the cold-start answer, concurrent reads not each starting a pass, the non-leader refusal both with and without a known leader, `Aborted` while a run is in flight, which sources are in the slice and in what order, persist failure still returning the in-memory document, protojson round-trip fidelity, a restored snapshot being re-staled against the clock, and pruning keeping exactly `runHistory` rows.

## Related work

Fourth of seven stacked PRs splitting #5795 ("OpenManager initial implementation") into area-owned reviews.

- #5812 - the topology tables and models this persists through
- #5813 - the OM API this implements (**this PR's base**)
- #5815 - the role rules for these routes
- #5811 - redaction of the SEP URL and token in logs

Unlike some of the others, this base is a **real dependency**: the service will not compile without the generated code in #5813, and it needs the models from #5812.

Two independent architectural reviews were run against this branch; the six items they agreed on are settled in `2ac39a7a4` - the leader-only read path (the one actual bug), the SEP client extraction, cluster identity and type, the health seam, per-service provenance, and the fact-grain decision above - together with the lifecycle tests that keep the first of them from regressing.

### DIFF
--- a/managed/cmd/pmm-managed/main.go
+++ b/managed/cmd/pmm-managed/main.go
@@ -73,6 +73,7 @@ import (
 	hav1beta1 "github.com/percona/pmm/api/ha/v1beta1"
 	inventoryv1 "github.com/percona/pmm/api/inventory/v1"
 	managementv1 "github.com/percona/pmm/api/management/v1"
+	omv1 "github.com/percona/pmm/api/om/v1"
 	rtav1 "github.com/percona/pmm/api/realtimeanalytics/v1"
 	serverv1 "github.com/percona/pmm/api/server/v1"
 	userv1 "github.com/percona/pmm/api/user/v1"
@@ -95,6 +96,7 @@ import (
 	managementgrpc "github.com/percona/pmm/managed/services/management/grpc"
 	"github.com/percona/pmm/managed/services/minio"
 	"github.com/percona/pmm/managed/services/nomad"
+	"github.com/percona/pmm/managed/services/om"
 	"github.com/percona/pmm/managed/services/qan"
 	"github.com/percona/pmm/managed/services/realtimeanalytics"
 	"github.com/percona/pmm/managed/services/scheduler"
@@ -234,6 +236,10 @@ type gRPCServerDeps struct {
 	versionCache              *versioncache.Service
 	vmdb                      *victoriametrics.Service
 	vmalert                   *vmalert.Service
+
+	// Built in main so its collection timer can be registered as an HA leader service.
+	// runGRPCServer only exposes it.
+	omService *om.Service
 }
 
 // runGRPCServer runs gRPC server until context is canceled, then gracefully stops it.
@@ -323,6 +329,8 @@ func runGRPCServer(ctx context.Context, deps *gRPCServerDeps) {
 	userv1.RegisterUserServiceServer(gRPCServer, user.NewUserService(deps.db, deps.grafanaClient))
 
 	hav1beta1.RegisterHAServiceServer(gRPCServer, ha.NewHAServer(deps.ha))
+
+	omv1.RegisterOmServiceServer(gRPCServer, deps.omService)
 
 	// Register RTA service with in-memory store
 	rtaStore := realtimeanalytics.NewStore()
@@ -430,6 +438,8 @@ func runHTTP1Server(ctx context.Context, deps *http1ServerDeps) {
 		backupv1.RegisterRestoreServiceHandler,
 
 		dumpv1beta1.RegisterDumpServiceHandler,
+
+		omv1.RegisterOmServiceHandler,
 
 		rtav1.RegisterRealtimeAnalyticsServiceHandler,
 
@@ -677,6 +687,16 @@ func main() { //nolint:gocognit,maintidx,cyclop
 
 	kingpin.Version(version.FullInfo())
 	kingpin.HelpFlag.Short('h')
+
+	// Where SEP is, not where any one of its apps is: OM's on-host facts come from
+	// the om_inventory app today and the actions apps will come from the same SEP, so
+	// each consumer appends its own /api/apps/<module> path. Optional -- with no URL
+	// the probe source reports itself disabled and the document is built from PMM's own
+	// inventory and metrics alone.
+	sepURLF := kingpin.Flag("sep-url", "Base URL of SEP, e.g. http://127.0.0.1:8000").
+		Envar("PMM_SEP_URL").String()
+	sepTokenF := kingpin.Flag("sep-token", "Bearer token for SEP's API").
+		Envar("PMM_SEP_TOKEN").String()
 
 	victoriaMetricsURLF := kingpin.Flag("victoriametrics-url", "VictoriaMetrics base URL").Envar("PMM_VM_URL").
 		Default(models.VMBaseURL).String()
@@ -1177,6 +1197,20 @@ func main() { //nolint:gocognit,maintidx,cyclop
 		return nil
 	}))
 
+	// Where SEP is, optional. Empty means OM builds its document from PMM's own inventory
+	// and metrics alone and records the probe source as disabled.
+	omService := om.New(db, v1.NewAPI(vmClient), logrus.WithField("component", "om"))
+	omService.WithProbeSource(*sepURLF, *sepTokenF)
+
+	// Leader-only, like every other periodic writer here. A collection persists a run and
+	// its snapshot and then prunes the shared history, so running it on every node of an
+	// HA cluster would have each node writing runs and pruning the others' -- and the
+	// pruning is what makes that destructive rather than merely wasteful.
+	haService.AddLeaderService(ha.NewContextService("om", func(ctx context.Context) error {
+		omService.Run(ctx)
+		return nil
+	}))
+
 	wg.Go(func() {
 		runGRPCServer(ctx,
 			&gRPCServerDeps{
@@ -1208,6 +1242,7 @@ func main() { //nolint:gocognit,maintidx,cyclop
 				templatesService:          alertingService,
 				versionCache:              versionCache,
 				vmalert:                   vmalert,
+				omService:                 omService,
 				vmClient:                  &vmClient,
 				vmdb:                      vmdb,
 			})
--- a/managed/cmd/pmm-managed/main.go
+++ b/managed/cmd/pmm-managed/main.go
@@ -1199,7 +1199,7 @@ func main() { //nolint:gocognit,maintidx,cyclop
 
 	// Where SEP is, optional. Empty means OM builds its document from PMM's own inventory
 	// and metrics alone and records the probe source as disabled.
-	omService := om.New(db, v1.NewAPI(vmClient), logrus.WithField("component", "om"))
+	omService := om.New(db, v1.NewAPI(vmClient), haService, logrus.WithField("component", "om"))
 	omService.WithProbeSource(*sepURLF, *sepTokenF)
 
 	// Leader-only, like every other periodic writer here. A collection persists a run and
--- a/managed/cmd/pmm-managed/main.go
+++ b/managed/cmd/pmm-managed/main.go
@@ -1201,6 +1201,7 @@ func main() { //nolint:gocognit,maintidx,cyclop
 	// and metrics alone and records the probe source as disabled.
 	omService := om.New(db, v1.NewAPI(vmClient), haService, logrus.WithField("component", "om"))
 	omService.WithProbeSource(*sepURLF, *sepTokenF)
+	prom.MustRegister(om.NewMetricsCollector(omService))
 
 	// Leader-only, like every other periodic writer here. A collection persists a run and
 	// its snapshot and then prunes the shared history, so running it on every node of an
--- a/managed/services/om/facts_test.go
+++ b/managed/services/om/facts_test.go
@@ -1,0 +1,244 @@
+// Copyright (C) 2023 Percona LLC
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package om
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMergeFacts(t *testing.T) {
+	t.Parallel()
+
+	now := time.Date(2026, 8, 11, 12, 0, 0, 0, time.UTC)
+
+	t.Run("precedence decides, not call order", func(t *testing.T) {
+		t.Parallel()
+
+		// The probe is listed after metrics for version, so it loses however late it runs.
+		merged := mergeFacts([]SourceResult{
+			{Source: sourceProbe, Facts: []Fact{
+				{Service: "s1", Field: fieldVersion, Value: "7.0.1", Source: sourceProbe},
+			}},
+			{Source: sourceMetrics, Facts: []Fact{
+				{Service: "s1", Field: fieldVersion, Value: "7.0.39", Source: sourceMetrics, ObservedAt: new(now)},
+			}},
+		}, defaultPrecedence)
+
+		require.Contains(t, merged, "s1")
+		assert.Equal(t, "7.0.39", merged["s1"][fieldVersion].Value)
+		assert.Equal(t, sourceMetrics, merged["s1"][fieldVersion].Source)
+	})
+
+	t.Run("the fallback source wins when the preferred one is silent", func(t *testing.T) {
+		t.Parallel()
+
+		merged := mergeFacts([]SourceResult{
+			{Source: sourceInventory, Facts: []Fact{
+				{Service: "s1", Field: fieldEndpoint, Value: "host:27017", Source: sourceInventory},
+			}},
+			{Source: sourceMetrics, Facts: []Fact{}},
+		}, defaultPrecedence)
+
+		assert.Equal(t, "host:27017", merged["s1"][fieldEndpoint].Value)
+		assert.Equal(t, sourceInventory, merged["s1"][fieldEndpoint].Source)
+	})
+
+	t.Run("a source not listed for a field cannot set it", func(t *testing.T) {
+		t.Parallel()
+
+		// exporter_up names metrics alone: reachability is not the probe's to assert.
+		merged := mergeFacts([]SourceResult{
+			{Source: sourceProbe, Facts: []Fact{
+				{Service: "s1", Field: fieldExporterUp, Value: 1.0, Source: sourceProbe},
+			}},
+		}, defaultPrecedence)
+
+		assert.NotContains(t, merged["s1"], fieldExporterUp)
+	})
+
+	t.Run("probe-only fields are reserved and merge untouched", func(t *testing.T) {
+		t.Parallel()
+
+		merged := mergeFacts([]SourceResult{
+			{Source: sourceProbe, Facts: []Fact{
+				{Service: "s1", Field: fieldInstalledVersion, Value: "7.0.40", Source: sourceProbe},
+			}},
+			{Source: sourceMetrics, Facts: []Fact{
+				{Service: "s1", Field: fieldInstalledVersion, Value: "nonsense", Source: sourceMetrics},
+			}},
+		}, defaultPrecedence)
+
+		assert.Equal(t, "7.0.40", merged["s1"][fieldInstalledVersion].Value)
+	})
+
+	t.Run("a nil value never shadows a real one", func(t *testing.T) {
+		t.Parallel()
+
+		merged := mergeFacts([]SourceResult{
+			{Source: sourceMetrics, Facts: []Fact{
+				{Service: "s1", Field: fieldVersion, Value: nil, Source: sourceMetrics},
+			}},
+			{Source: sourceProbe, Facts: []Fact{
+				{Service: "s1", Field: fieldVersion, Value: "7.0.39", Source: sourceProbe},
+			}},
+		}, defaultPrecedence)
+
+		assert.Equal(t, "7.0.39", merged["s1"][fieldVersion].Value)
+	})
+
+	t.Run("the freshest series wins when one service has several", func(t *testing.T) {
+		t.Parallel()
+
+		// A replica-set reconfiguration leaves the superseded mongodb_members_self behind,
+		// and it survives in the lookback window under the same service_id. Measured in
+		// the sandbox: PRIMARY at 0s, SECONDARY at 7500s, PRIMARY at 68477s. Rank cannot
+		// separate them -- they are all the metrics source -- so recency has to.
+		fresh := now
+		stale := now.Add(-7500 * time.Second)
+		ancient := now.Add(-68477 * time.Second)
+
+		for _, order := range [][]Fact{
+			{
+				{Service: "s1", Field: fieldState, Value: "SECONDARY", Source: sourceMetrics, ObservedAt: new(stale)},
+				{Service: "s1", Field: fieldState, Value: "PRIMARY", Source: sourceMetrics, ObservedAt: new(fresh)},
+				{Service: "s1", Field: fieldState, Value: "PRIMARY", Source: sourceMetrics, ObservedAt: new(ancient)},
+			},
+			// The same facts the other way round: the answer must not depend on the order
+			// VictoriaMetrics happened to return them in.
+			{
+				{Service: "s1", Field: fieldState, Value: "PRIMARY", Source: sourceMetrics, ObservedAt: new(fresh)},
+				{Service: "s1", Field: fieldState, Value: "SECONDARY", Source: sourceMetrics, ObservedAt: new(stale)},
+			},
+		} {
+			merged := mergeFacts([]SourceResult{{Source: sourceMetrics, Facts: order}}, defaultPrecedence)
+			assert.Equal(t, "PRIMARY", merged["s1"][fieldState].Value)
+			assert.Equal(t, fresh, *merged["s1"][fieldState].ObservedAt)
+		}
+	})
+
+	t.Run("a higher-precedence source still wins over a fresher low-precedence one", func(t *testing.T) {
+		t.Parallel()
+
+		// Recency only breaks ties. It must not let the probe outrank metrics for version
+		// just by having looked more recently.
+		merged := mergeFacts([]SourceResult{
+			{Source: sourceMetrics, Facts: []Fact{
+				{Service: "s1", Field: fieldVersion, Value: "7.0.39", Source: sourceMetrics, ObservedAt: new(now.Add(-time.Hour))},
+			}},
+			{Source: sourceProbe, Facts: []Fact{
+				{Service: "s1", Field: fieldVersion, Value: "7.0.1", Source: sourceProbe, ObservedAt: new(now)},
+			}},
+		}, defaultPrecedence)
+
+		assert.Equal(t, "7.0.39", merged["s1"][fieldVersion].Value)
+	})
+
+	t.Run("provenance survives the merge", func(t *testing.T) {
+		t.Parallel()
+
+		observed := now.Add(-90 * time.Second)
+		merged := mergeFacts([]SourceResult{
+			{Source: sourceMetrics, Facts: []Fact{
+				{Service: "s1", Field: fieldState, Value: "PRIMARY", Source: sourceMetrics, ObservedAt: new(observed)},
+			}},
+		}, defaultPrecedence)
+
+		held := merged["s1"][fieldState]
+		require.NotNil(t, held.ObservedAt)
+		assert.Equal(t, observed, *held.ObservedAt)
+
+		age, datable := held.age(now)
+		assert.True(t, datable)
+		assert.Equal(t, 90*time.Second, age)
+	})
+}
+
+func TestFieldSetFreshness(t *testing.T) {
+	t.Parallel()
+
+	now := time.Date(2026, 8, 11, 12, 0, 0, 0, time.UTC)
+	maxAge := 30 * time.Second
+
+	newFields := func(age time.Duration) fieldSet {
+		observed := now.Add(-age)
+		return fieldSet{
+			now:    now,
+			maxAge: maxAge,
+			fields: map[string]MergedField{
+				// Volatile: only meaningful while fresh.
+				fieldExporterUp: {Value: 1.0, Source: sourceMetrics, ObservedAt: new(observed)},
+				fieldState:      {Value: "PRIMARY", Source: sourceMetrics, ObservedAt: new(observed)},
+				// Durable: describes the service, not the moment.
+				fieldVersion: {Value: "7.0.39", Source: sourceMetrics, ObservedAt: new(observed)},
+				fieldEdition: {Value: "Community", Source: sourceMetrics, ObservedAt: new(observed)},
+			},
+		}
+	}
+
+	t.Run("fresh volatile facts read through", func(t *testing.T) {
+		t.Parallel()
+
+		fields := newFields(5 * time.Second)
+		assert.InDelta(t, 1.0, *fields.f64(fieldExporterUp), 0.001)
+		assert.Equal(t, "PRIMARY", fields.str(fieldState))
+		assert.Zero(t, fields.staleVolatile())
+	})
+
+	t.Run("stale volatile facts are dropped, durable ones are kept", func(t *testing.T) {
+		t.Parallel()
+
+		// This is the case that made a stopped database report UP: the sample survives in
+		// the long lookback window, and only the age says it is not about now.
+		fields := newFields(10 * time.Minute)
+		assert.Nil(t, fields.f64(fieldExporterUp))
+		assert.Empty(t, fields.str(fieldState))
+		assert.Equal(t, "7.0.39", fields.str(fieldVersion), "a version does not go stale")
+		assert.Equal(t, "Community", fields.str(fieldEdition))
+		assert.Equal(t, 2, fields.staleVolatile())
+	})
+
+	t.Run("an undatable fact is never stale", func(t *testing.T) {
+		t.Parallel()
+
+		// Inventory is current by definition and carries no observed_at.
+		fields := fieldSet{
+			now:    now,
+			maxAge: maxAge,
+			fields: map[string]MergedField{
+				fieldEndpoint: {Value: "host:27017", Source: sourceInventory},
+			},
+		}
+		assert.Equal(t, "host:27017", fields.str(fieldEndpoint))
+	})
+
+	t.Run("a missing or mistyped field reads as absent", func(t *testing.T) {
+		t.Parallel()
+
+		fields := fieldSet{
+			now:    now,
+			maxAge: maxAge,
+			fields: map[string]MergedField{fieldVersion: {Value: 42, Source: sourceMetrics}},
+		}
+		assert.Empty(t, fields.str(fieldVersion))
+		assert.Empty(t, fields.str(fieldVendor))
+		assert.Nil(t, fields.f64(fieldCPUUsage))
+		assert.False(t, fields.truthy(fieldIsMongos))
+	})
+}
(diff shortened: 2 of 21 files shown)


plebioda:
@coderabbitai review

managed/cmd/pmm-managed/main.go
```diff
@@ -324,6 +331,14 @@ func runGRPCServer(ctx context.Context, deps *gRPCServerDeps) {
...
 	hav1beta1.RegisterHAServiceServer(gRPCServer, ha.NewHAServer(deps.ha))
 
+	omSvc := om.New(deps.db, v1.NewAPI(*deps.vmClient), logrus.WithField("component", "om"))
+	omSvc.WithProbeSource(deps.sepURL, deps.sepToken)
+	omv1.RegisterOmServiceServer(gRPCServer, omSvc)
+
+	// Refresh the topology document on a timer, so the run history exists even when
+	// nobody is looking at the page.
+	go omSvc.Run(ctx)
+
```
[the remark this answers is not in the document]
plebioda: Fixed, and you were right that this mattered.

`go omSvc.Run(ctx)` ran on every node, and a collection persists a run plus its snapshot and then calls `PruneOmTopologyRuns`. The pruning is what makes it destructive rather than merely wasteful: on an HA cluster each node would be deleting the others' history.

OM is now registered like every other periodic writer here - `checks`, `telemetry`, `scheduler`, `versionCache`, `cleaner`:

```go
haService.AddLeaderService(ha.NewContextService("om", func(ctx context.Context) error {
    omService.Run(ctx)
    return nil
}))
```

One thing worth recording about *where*: registering from `runGRPCServer` via `deps.ha` would have raced. `runGRPCServer` is launched in `wg.Go(...)` at main.go:1208, while `AddLeaderService("cleaner")` is at 1257 and `haService.Run(ctx)` at 1263 - so a registration from inside the goroutine could land after the HA service had already started iterating. The service is therefore constructed in `main` beside the others and passed into `gRPCServerDeps` as `omService`, with `runGRPCServer` reduced to `omv1.RegisterOmServiceServer(gRPCServer, deps.omService)`. The `sepURL`/`sepToken` fields on the deps struct are gone, since the construction that used them moved.

I have not gated `GetTopology`'s stale-cache rebuild or `TriggerTopologyCollection` on leadership. Those are request-scoped: a user asking for the page on a follower should get an answer rather than an error, and both paths write through the same transaction as the timer. If you think a follower should refuse to write at all, that is a broader decision about OM's HA story than this PR settles, and I would rather it were explicit.

managed/services/om/facts_test.go
```diff
@@ -0,0 +1,244 @@
...
+	t.Run("precedence decides, not call order", func(t *testing.T) {
+		t.Parallel()
+
+		// The probe is listed after metrics for version, so it loses however late it runs.
+		merged := mergeFacts([]SourceResult{
+			{Source: sourceProbe, Facts: []Fact{
+				{Service: "s1", Field: fieldVersion, Value: "7.0.1", Source: sourceProbe},
+			}},
+			{Source: sourceMetrics, Facts: []Fact{
+				{Service: "s1", Field: fieldVersion, Value: "7.0.39", Source: sourceMetrics, ObservedAt: new(now)},
```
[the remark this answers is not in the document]
plebioda: This one is a false positive - the package compiles.

`new(now)` here and `new("127.0.0.1")` in `projection_test.go` are Go 1.26's `new(expr)`, which allocates and initialises from a typed expression. The repo is on `go 1.26.5` (go.mod), and both predate this PR:

```
$ go vet ./managed/services/om/     # clean
$ go test ./managed/services/om/... # ok
```

Worth knowing the exact rule, because it is easy to test wrongly: `new(expr)` needs a *typed* expression, so `new(now)` compiles while `new(false)` does not - `false` is an untyped constant. It also needs the module's `go` directive at 1.26; I initially reproduced a failure only because a scratch module had `go 1.22.2`.

No change made. Happy to reconsider if you are seeing an actual build failure somewhere I am not.

managed/services/om/inventory.go
```diff
@@ -0,0 +1,780 @@
...
+	probe, err := s.inventoryProbe()
+	if err != nil {
+		return nil, err
+	}
+
+	limit := req.GetLimit()
+	if limit <= 0 {
+		limit = defaultInventoryRunLimit
+	}
+	query := url.Values{"limit": []string{strconv.FormatInt(int64(limit), 10)}}
```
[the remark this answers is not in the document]
plebioda: Fixed. The limit was bounded below but not above, and the value is forwarded to SEP verbatim rather than to our own database - so an unbounded page was something a caller could ask the inventory app for and then wait on through this proxy.

Added `maxInventoryRunLimit = 200` beside the existing default, and the clamp is now a `switch` over both ends.

`ListTopologyRuns` reads our own tables with `defaultRunLimit` and the same lower-bound-only guard. I have left it, because the ceiling there protects a local query rather than a proxied one and the trade is different; say the word if you would rather both were symmetric.

managed/services/om/inventory.go
```diff
@@ -0,0 +1,780 @@
...
+}
+
+// freshnessToProto projects the freshness block.
+func freshnessToProto(f sepFreshness) *omv1.InventoryFreshness {
+	return &omv1.InventoryFreshness{
+		FirstSeenAt:         optionalTimestamp(f.FirstSeenAt),
+		LastAttemptAt:       optionalTimestamp(f.LastAttemptAt),
+		LastSuccessAt:       optionalTimestamp(f.LastSuccessAt),
+		FailingSince:        optionalTimestamp(f.FailingSince),
+		ConsecutiveFailures: int32(f.ConsecutiveFailures), //nolint:gosec
```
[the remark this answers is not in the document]
plebioda: Fixed. The `//nolint:gosec` was silencing the linter rather than the wrap, which is the worse of the two outcomes.

Replaced with a saturating `clampInt32`, so the column stays monotonic: a reader can learn "very many" but never "minus two billion". Practically it needs more than 2^31 consecutive failures to reach, but the clamp is three lines and it removes the suppression, which is the part I would rather not leave in.

managed/services/om/inventory.go
```diff
@@ -0,0 +1,780 @@
...
+// an unset `optional` scalar entirely, even under EmitUnpopulated. A silently absent
+// nullable field is a column that reads as "not collected" when the truth may be
+// "collected, and it was empty".
+func observedString(observed map[string]any, key string) *string {
+	value, ok := observed[key].(string)
+	if !ok || value == "" {
+		return nil
+	}
+	return &value
+}
```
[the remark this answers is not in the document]
plebioda: Fixed - and it was worse than you flagged, because the comment also still said "wrapper" after the wrapper-to-`optional` conversion earlier in this stack. So it was stale *and* self-contradicting.

Rewritten to say what the code does and why: empty and absent are deliberately merged, because for what these read - a config path, an argv, an OS name - an empty string carries no more information than no string at all, and collapsing them keeps the column from showing a blank cell a reader has to interpret. The note about protojson dropping unset `optional` scalars stays, since that is what makes the omission reachable.

I kept the behaviour rather than letting an empty string travel, for the reason above. Push back if you think one of those fields has a meaningful empty value.

managed/services/om/projection_test.go
```diff
@@ -0,0 +1,323 @@
...
+		assert.False(t, isStale(doc, projectionNow))
+	})
+}
+
+func TestInventoryEndpoint(t *testing.T) {
+	t.Parallel()
+
+	service := mongoService("s1", "mongo-1", "cl1", "rs0", "prod")
+	service.Port = pointer.ToUint16(27017)
+	service.Address = new("127.0.0.1")
```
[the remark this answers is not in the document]
plebioda: Same as the `facts_test.go` comment - false positive. `new("127.0.0.1")` is Go 1.26's `new(expr)`; `models.Service.Address` is `*string` and this assigns a `*string`. `go vet` is clean and the tests pass, and the line predates this PR. No change made.

managed/services/om/service.go
```diff
@@ -0,0 +1,485 @@
...
+func (s *Service) readInventory() ([]*models.Service, map[string]*models.Node, string, time.Duration, error) {
+	var (
+		services   []*models.Service
+		nodesByID  map[string]*models.Node
+		originNode string
+		maxAge     time.Duration
+	)
+
+	serviceType := models.MongoDBServiceType
+	errTX := s.db.InTransaction(func(tx *reform.TX) error {
```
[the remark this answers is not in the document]
plebioda: Fixed, and this was the sharpest finding of the eight.

It is also a direct follow-on from a fix I had just made on the schema PR: I had moved the *write* path to `InTransactionContext` and stopped there, leaving every read on plain `db.InTransaction`. So the guideline was half-applied, which is arguably worse than not at all, because it looks done.

All four read paths are now context-bound, with the context threaded from the handlers:

- `readInventory(ctx)` - from `collect(ctx)`
- `restore(ctx)` - from `restoreOnce(ctx)`, from `GetTopology(ctx, ...)`
- `listRuns(ctx, limit)` and `getRun(ctx, runID)` - the two handlers no longer discard their context with `_ context.Context`

There is no plain `s.db.InTransaction(` left in the package.

managed/services/om/sources.go
```diff
@@ -0,0 +1,372 @@
...
+
+	for serviceID, v := range values {
+		fact := Fact{Service: serviceID, Field: signal.valueField, Value: v, Source: sourceMetrics}
+		// Dated by the matching age query, so a value read over a day-long window still
+		// knows how old it is.
+		if age, ok := ages[serviceID]; ok {
+			fact.ObservedAt = r.observedAt(age)
+		}
+		r.result.Facts = append(r.result.Facts, fact)
+	}
```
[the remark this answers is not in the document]
plebioda: Fixed. This was a genuinely subtle catch and the analysis is exactly right.

`ages` kept the smallest age per service while the value query chose its winner independently - the largest sample under `reduceMax`, whichever arrived last otherwise. So the two need not come from the same series, and for `metricReplicationLag` a node reporting one current series and one hours-stale series with larger lag would hand back the stale maximum stamped as current, which `fieldSet.live` then reads as a statement about now.

`ages` now returns the *oldest* age per service, so a mixed pair reads as stale. Wrong in the direction that shows rather than the direction that misleads.

Two things worth noting:

The label and presence facts are unaffected. Each is dated with its own series' age inside the `each` callback, because each comes from exactly one series - only the reduced value fact had the mismatch. I have said so in the doc comment so the asymmetry does not read as an oversight.

`metricsSource` had no test file at all, which is why changing freshest to oldest broke nothing and told me nothing. There is now `sources_test.go` with `TestReducedValueIsDatedByTheOldestSeries`, built on a two-line fake of the `victoriaMetricsClient` interface: one service, two lag series at 30s and 4h, a reduced value of 900, asserting the fact carries the 4-hour age. I verified it fails against the old `age < held` and passes against the new one, so it is a regression test rather than a restatement.

managed/services/om/inventory.go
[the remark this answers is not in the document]
plebioda: Done - `ListTopologyRuns` is now bounded too.

You are right that the local-database boundary changes the impact rather than removing the need, so I have taken the suggestion rather than the offer to leave it asymmetric.

The two ceilings do differ, and for a reason worth stating:

- `maxInventoryRunLimit = 200` is a judgement about what is reasonable to proxy to SEP.
- `maxRunLimit = runHistory` (100) is not a judgement at all. `runHistory` is how many runs the history keeps, so asking for more than that cannot return more rows - a larger limit is only a larger query. Tying the constant to `runHistory` rather than picking a second number means the two cannot drift apart if the retention changes.

Also worth noting for the record: I have now installed golangci-lint v2.12.2 locally, matching the version pinned in `Makefile.include` and CI, and ran it with `--new-from-rev` over this PR's diff - 0 issues. Together with `license-eye` and `make format && git diff --exit-code`, that is the full `Checks` gate reproduced locally, which is what I should have been doing from the start rather than discovering the license-header and lint failures one CI run at a time.

plebioda:
Ported two commits from the integration branch, rebased onto #5813's new commit (this branch needs it to compile -- the generated `since`/`until` fields live there):

- [`c6de44aab`](https://github.com/percona/pmm/commit/c6de44aab) forwards `since`/`until` to SEP as RFC3339 and rejects an inverted range as `InvalidArgument` before the call reaches SEP -- companion to #5813, backing the OM UI's Inventory-page date-range filter.
- [`0fa9b9d6b`](https://github.com/percona/pmm/commit/0fa9b9d6b) is a one-line cleanup found while porting: `maxInventoryRunLimit` was 200, but the proto already validates `limit <= 100` before a request reaches this handler and SEP enforces the same `le=100` independently, so the 200 clamp was dead code. Aligned to 100, matching what the comment on the constant already says the ceiling is for.

Branch rebased onto the updated #5813 tip (real dependency, not just ordering), hence the force-push. `go build ./...`, `go vet` and `TestInventoryRuns*` all clean.

managed/services/om/sources.go
```diff
@@ -0,0 +1,445 @@
...
+		return result
+	}
+
+	ids := make([]string, 0, len(services))
+	for _, service := range services {
+		ids = append(ids, service.ServiceID)
+	}
+
+	run := &metricsRun{src: s, result: &result, covered: make(map[string]bool, len(ids))}
+	for batch := range slicesChunk(ids, queryBatch) {
```
4nte: Can we use https://pkg.go.dev/slices#Chunk here from stdlib?

managed/services/om/inventory_client.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	if resp.StatusCode >= http.StatusBadRequest {
+		return sepStatusError(resp)
+	}
+	if out == nil || resp.StatusCode == http.StatusNoContent {
+		return nil
+	}
+	if err := json.NewDecoder(resp.Body).Decode(out); err != nil { //nolint:noinlineerr
```
4nte: `noinlineerr` is enabled on purpose per our `.golangci.yml`.
Please remove the supression and break this up in separate lines to satisfy the linter.

managed/services/om/probe_source.go
```diff
@@ -0,0 +1,279 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("GET %s: unexpected status %s", url, resp.Status)
+	}
+
+	answer := []probeService{}
+	if err := json.NewDecoder(resp.Body).Decode(&answer); err != nil { //nolint:noinlineerr
+		return nil, fmt.Errorf("GET %s: failed to decode the response: %w", url, err)
+	}
+	return answer, nil
```
4nte: Don't suppress `noinlineerr`

managed/services/om/inventory_test.go
```diff
@@ -0,0 +1,840 @@
...
+}
+
+// newSEPStubSeq serves one canned answer per request, in order.
+func newSEPStubSeq(t *testing.T, code int, bodies ...string) *sepStub {
+	t.Helper()
+
+	stub := &sepStub{bodies: bodies}
+	stub.server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		call := stubCall{method: r.Method, path: r.URL.Path, query: r.URL.RawQuery}
+		if raw, err := io.ReadAll(r.Body); err == nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+	if req.Failing != nil {
+		query.Set("failing", strconv.FormatBool(req.GetFailing()))
+	}
+	if req.Executor != nil {
+		query.Set("executor", strconv.FormatBool(req.GetExecutor()))
+	}
+
+	hosts := []sepHost{}
+	call := inventoryCall{method: http.MethodGet, path: "hosts", query: query}
+	if err := probe.call(ctx, call, &hosts); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+// GetInventoryHost returns one host.
+func (s *Service) GetInventoryHost(ctx context.Context, req *omv1.GetInventoryHostRequest) (*omv1.GetInventoryHostResponse, error) {
+	probe, err := s.inventoryProbe()
+	if err != nil {
+		return nil, err
+	}
+
+	host := sepHost{}
+	call := inventoryCall{method: http.MethodGet, path: inventoryPath("hosts", req.GetNodeId())}
+	if err := probe.call(ctx, call, &host); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+// for rows left behind when a node was replaced -- restarting a pmm-agent runs
+// `setup --force`, which mints a new node ID, so OM gains a row and keeps the old one.
+func (s *Service) DeleteInventoryHost(ctx context.Context, req *omv1.DeleteInventoryHostRequest) (*omv1.DeleteInventoryHostResponse, error) {
+	probe, err := s.inventoryProbe()
+	if err != nil {
+		return nil, err
+	}
+
+	call := inventoryCall{method: http.MethodDelete, path: inventoryPath("hosts", req.GetNodeId())}
+	if err := probe.call(ctx, call, nil); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+	if nodeID := req.GetNodeId(); nodeID != "" {
+		query.Set("node_id", nodeID)
+	}
+	if req.Failing != nil {
+		query.Set("failing", strconv.FormatBool(req.GetFailing()))
+	}
+
+	services := []sepService{}
+	call := inventoryCall{method: http.MethodGet, path: "services", query: query}
+	if err := probe.call(ctx, call, &services); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+// GetInventoryService returns one service.
+func (s *Service) GetInventoryService(ctx context.Context, req *omv1.GetInventoryServiceRequest) (*omv1.GetInventoryServiceResponse, error) {
+	probe, err := s.inventoryProbe()
+	if err != nil {
+		return nil, err
+	}
+
+	service := sepService{}
+	call := inventoryCall{method: http.MethodGet, path: inventoryPath("services", req.GetServiceId())}
+	if err := probe.call(ctx, call, &service); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+
+// DeleteInventoryService forgets one service.
+func (s *Service) DeleteInventoryService(ctx context.Context, req *omv1.DeleteInventoryServiceRequest) (*omv1.DeleteInventoryServiceResponse, error) {
+	probe, err := s.inventoryProbe()
+	if err != nil {
+		return nil, err
+	}
+
+	call := inventoryCall{method: http.MethodDelete, path: inventoryPath("services", req.GetServiceId())}
+	if err := probe.call(ctx, call, nil); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+	if since != nil {
+		query.Set("since", since.AsTime().UTC().Format(time.RFC3339Nano))
+	}
+	if until != nil {
+		query.Set("until", until.AsTime().UTC().Format(time.RFC3339Nano))
+	}
+
+	runs := []sepRun{}
+	call := inventoryCall{method: http.MethodGet, path: "runs", query: query}
+	if err := probe.call(ctx, call, &runs); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+// GetInventoryRun returns one refresh.
+func (s *Service) GetInventoryRun(ctx context.Context, req *omv1.GetInventoryRunRequest) (*omv1.GetInventoryRunResponse, error) {
+	probe, err := s.inventoryProbe()
+	if err != nil {
+		return nil, err
+	}
+
+	run := sepRun{}
+	call := inventoryCall{method: http.MethodGet, path: inventoryPath("runs", req.GetRunId())}
+	if err := probe.call(ctx, call, &run); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+	body := map[string]any{"node_ids": nodeIDs}
+
+	accepted := struct {
+		RunID     string   `json:"run_id"`
+		Status    string   `json:"status"`
+		StartedAt *string  `json:"started_at"`
+		Scope     []string `json:"scope"`
+	}{}
+	call := inventoryCall{method: http.MethodPost, path: "runs", body: body}
+	if err := probe.call(ctx, call, &accepted); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+// GetInventoryConfig returns the inventory app's configuration.
+func (s *Service) GetInventoryConfig(ctx context.Context, _ *omv1.GetInventoryConfigRequest) (*omv1.GetInventoryConfigResponse, error) {
+	probe, err := s.inventoryProbe()
+	if err != nil {
+		return nil, err
+	}
+
+	settings := []sepSetting{}
+	call := inventoryCall{method: http.MethodGet, path: "config"}
+	if err := probe.call(ctx, call, &settings); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+		return nil, err
+	}
+	err = validateConfigValues(req.GetValues())
+	if err != nil {
+		return nil, err
+	}
+
+	applied := []sepSetting{}
+	call := inventoryCall{method: http.MethodPatch, path: "config", body: req.GetValues().AsMap()}
+	if err := probe.call(ctx, call, &applied); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory.go
```diff
@@ -0,0 +1,829 @@
...
+func (s *Service) DeleteInventoryConfigOverride(
+	ctx context.Context, req *omv1.DeleteInventoryConfigOverrideRequest,
+) (*omv1.DeleteInventoryConfigOverrideResponse, error) {
+	probe, err := s.inventoryProbe()
+	if err != nil {
+		return nil, err
+	}
+
+	call := inventoryCall{method: http.MethodDelete, path: inventoryPath("config", req.GetKey())}
+	if err := probe.call(ctx, call, nil); err != nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

managed/services/om/inventory_client.go
```diff
@@ -0,0 +1,250 @@
...
+			"PMM's credential for the inventory app was rejected (%s): %s", resp.Status, detail)
+	default:
+		return status.Errorf(codes.Internal, "the inventory app answered %s: %s", resp.Status, detail)
+	}
+}
+
+// sepErrorDetail extracts something readable from the app's error body.
+func sepErrorDetail(resp *http.Response) string {
+	var envelope sepError
+	if err := json.NewDecoder(resp.Body).Decode(&envelope); err != nil || envelope.Detail == nil { //nolint:noinlineerr
```
4nte: Don't suppress noinlineerr

4nte:
We need to be able to toggle OM, I suggest introducing a new field to settings:

 ```go
 type Settings struct {
    // ...
	OpenManager struct {
		Enabled *bool `json:"enabled"`
		// + other OM specific settings
	} `json:"open_manager"`

```
And an env var `PMM_ENABLE_OM` override, which takes precedence over whatever was previously set in the settings  (env var always wins).

4nte:
OM doesn't register a Prometheus collector, where agents, inventory, ha, advisor and backups all do. So the per-run counters OM already computes end up only in om_topology_runs, which prune caps at roughly 50 minutes, and the run summary is logged at Info.

Since the numbers already exist, a small collector looks close to free, and it'd make a degraded pass visible on PMM Health instead of just in a log line. Rough starting point:

- pmm_managed_om_runs_total{status} - is it collecting at all
- pmm_managed_om_document_age_seconds - is what we serve actually current
- pmm_managed_om_services{state="up"|"total"|...} - did the pass see the estate

managed/services/om/inventory_client.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/probe_source.go
plebioda: Fixed. `err` was already in scope here, so this is a plain reassignment (`err = ...`) followed by its own check, rather than the inline `if err := ...; err != nil` noinlineerr was flagging.

managed/services/om/inventory_client.go
plebioda: Fixed. No `err` was in scope yet here, so this declares it on its own line and checks it separately, rather than inline.

managed/services/om/inventory_test.go
plebioda: Fixed. Same treatment: `raw, err := io.ReadAll(...)` on its own line, checked separately.

managed/services/om/sources.go
plebioda: Done - swapped for stdlib `slices.Chunk` and deleted the hand-rolled `slicesChunk` it was reimplementing field for field.

plebioda:
@4nte re: the Prometheus collector suggestion -

Added `MetricsCollector` in `om_metrics.go`, following the ha/inventory/backup shape (manual `prom.Desc` + `prom.MustNewConstMetric`, `Describe` delegating to `DescribeByCollect`), registered next to `omService`'s own construction in main.go:

- `pmm_managed_om_runs_total{status}` - completed runs by outcome, tallied by three atomic counters on `Service` and incremented once per run.
- `pmm_managed_om_document_age_seconds` - age of the newest observation in the document currently served. Deliberately the same age `Snapshot.Stale` is already computed from, so that existing signal becomes alertable rather than only renderable. Unset until the estate has been observed at least once.
- `pmm_managed_om_services{state}` - estate size versus how much answered, from the last pass's Summary.

Verified live against the dev stack: `runs_total` advanced on the next collection tick, `document_age_seconds` tracked real elapsed time, and `services{state}` matched `GetTopology`'s own summary (14/14, 4 clusters).

plebioda:
@4nte re: the OM toggle -

Filed as its own ticket: PMM-15360 ("OM: global on/off switch in PMM"). It needs a real design pass of its own - default state during preview, where it lives in Settings, whether a deployment can lock it out entirely, and how the switch reaches the OM Inventory app on SEP's side of this boundary are all still open questions there. Keeping it out of this PR rather than half-building it here, since any partial toggle I add now would likely conflict with what that ticket settles on.

plebioda:
Resolved 16 stale threads above (@4nte's noinlineerr suppressions + the slices.Chunk suggestion) - these were already fixed in code by a prior commit (all 15 suppressions removed, swapped for stdlib slices.Chunk), the threads just never got marked resolved.

4nte (approved):

4nte:
`om` declares `victoriaMetricsClient` in `deps.go` but the package isn't in
`.mockery.yaml`, so no `mock_victoria_metrics_client_test.go` is generated and the tests
hand-roll their own doubles instead.

We should add `services/om: victoriaMetricsClient` to keep it consistent.

PR #5817, opened 2026-08-21
PMM-15326: Add the OM UI plugin

plebioda:
Ticket number: PMM-15326

Feature build: N/A on its own - this package is not mounted anywhere until the navigation PR stacked on top of it lands. A build covering both is more useful than one per PR; happy to produce one.

## What

`@sep/plugins-om`, the OpenManager plugin: four pages over pmm-managed's `/v1/om`, plus the shared components, hooks and types behind them.

- **Overview** - the topology document rolled up by environment and cluster.
- **Services** - one row per monitored MongoDB service, joined against the inventory estate.
- **Hosts** - the page a host with no database appears on, which no other OM page can show: it has no service to be listed through.
- **Inventory** - the collection run history and the app's runtime settings.

## Two things worth knowing before reading

**The plugin depends on `@sep/api` nowhere.** Every page reads PMM's own origin, so no request it makes needs a SEP bearer - pmm-managed proxies SEP's estate at `/v1/om/inventory` and the Grafana session authorises the call. That is the whole reason this is a PMM plugin rather than a SEP one.

**The types are hand-written, not generated.** OM is not in PMM's checked-in OpenAPI client set, and `types.ts` says so at the top. The fields are typed `?: T | null` because the wire form is a proto3 `optional` scalar: protojson omits an unset one rather than emitting null, so a missing key is what arrives, and every reader here uses `== null`, which catches both.

## Two conventions the tables depend on, and they differ on purpose

`cpu_usage_percent` and `connections_free_percent` are **-1 when not measured**, never absent and never 0 - zero CPU is a real reading, so the numeric sentinel is what keeps "idle" and "unknown" apart in a numeric column.

`replication_lag_seconds` and `oplog_window_seconds` are **absent when they do not apply** - a router and a standalone have no replica-set oplog. That is a different statement from -1's "we could not measure it".

Both are documented in `types.ts` and mirrored in `om.proto`. If you would rather one convention covered both cases, that is worth saying now rather than after the columns ship.

## The lockfile is split on purpose

`ui/pnpm-lock.yaml` here carries **only** this package's own importer block. The three lines adding `'@sep/plugins-om'` to the `apps/pmm` importer land in the navigation PR, with the `package.json` that declares the dependency - otherwise `pnpm install --frozen-lockfile` fails on both branches: this one would declare a dependent that does not declare the dep, and that one would be missing its own entry. Both branches were checked against `--frozen-lockfile` after the split, and again after the review round below.

**What that block declares changed in review.** It now carries `date-fns` and `@vitejs/plugin-react`, and no longer carries `@vitejs/plugin-react-swc`. `date-fns` is the monorepo's date library and already a dependency of `apps/pmm`; `@vitejs/plugin-react` is what `main` moved to. Both changes are scoped to this package - the rest of the tree is still on `-swc` because this branch's base predates that change on `main`.

## Reviewing this

~6.3k lines, of which ~900 are tests and 580 are `types.ts`. The parts carrying the most judgement: `topologyHooks.ts` and `inventoryHooks.ts` (what is fetched, and when), `topology.ts` and `inventory.ts` (the roll-ups, the joins and the freshness helpers), then the four pages. `api.ts` is the transport both halves share.

## Review round 1 (2026-08-25)

Fourteen inline comments from @fabio-silva across thirteen threads, and nine from Copilot. Twelve of thirteen taken, and all nine, across six commits (`4c395dc` → `504d7a7`); each is answered in its own thread.

| commit | what |
| --- | --- |
| `8a38150` | Services no longer reads a *pending* estate as an empty one; Overview keeps Sync in its error state; both refresh buttons derive "running" from every run rather than `runs[0]`; the Hosts table is keyed on `node_id` |
| `c700ad7` | `pom.v1` → `om.v1`; the inventory types are exported; the dead `useOmTopologyRun` is deleted; `aria-label` on the config switches |
| `4149c92` | `isRefreshActive` folded into `isRunActive`; query keys uppercased; both triggers drop bodies they do not need; the `timedelta` wire contract documented |
| `428084b` | `@vitejs/plugin-react`; `format.ts` onto `date-fns` |
| `e1d2c3c` | `hooks.ts` split into `api.ts` / `topology.ts` / `topologyHooks.ts`; `refreshAll()` / `refreshHosts()` replace `mutate(undefined)`; components converted to arrow functions |
| `fc5e93a` | the estate queries follow their own refreshes, so Services no longer shows pre-refresh rows for a minute |

**One thing held back**, and it is the open question on this PR: the suggestion to change `!= null` to `!== null` throughout. Proto3 `optional` fields are dropped by protojson rather than nulled - they sit in a synthetic oneof, which `unpopulatedFieldRanger` skips - so under `!==` the cluster roll-up admits `undefined` into `Math.max` and reports `NaN` for max replication lag and min oplog window on every row. Reasoning and references are in the thread on `ProbeValue.tsx`.

**Two changes worth a second look**, both deliberate but both visible:

- `formatTimestamp` moved from `toLocaleString()` to a fixed `yyyy-MM-dd HH:mm:ss` in local time. Run timestamps are compared down a column, so a stable format beats a familiar one - but it is a rendering change and easy to revert.
- The compact duration formatter was kept rather than replaced, and renamed `formatCompactDuration`. date-fns emits `2 days 3 hours` where these cells need `2d 3h`; the rename also ends a collision with date-fns' own export.

## Related work

Sixth of seven stacked PRs splitting #5795 ("OpenManager initial implementation") into area-owned reviews.

- #5811 - `PMM_SEP_URL` / `PMM_SEP_TOKEN` redaction
- #5812 - the OM topology tables and models
- #5813 - the OM API this reads
- #5815 - the role rules for these routes
- #5816 - the service that serves them

This PR targets `PMM-15299-open-manager` directly and depends on nothing else in the stack at compile time - the plugin's own vitest suite runs without the app declaring the dependency, because pnpm picks packages up by workspace glob rather than by dependent.

**One thing the reviewer should know:** `PMM-15299-open-manager` currently conflicts with `main`, and `ui/pnpm-lock.yaml` is one of the seven conflicting files - because the same pnpm toolchain migration (PMM-15288) landed on `main` as #5728 and on this branch via its own line. That conflict is not this PR's and does not affect merging into the base, but reconciling it later will touch this file.


plebioda:
@coderabbitai review

ui/packages/plugins/om/src/components/ConfigForm.tsx
```diff
@@ -0,0 +1,341 @@
...
+  // Re-seed from the effective values whenever they change, including after a write:
+  // what the server settled on wins over what was typed. Without this the boxes would
+  // keep showing a submission the app may have coerced or rejected.
+  useEffect(() => {
+    setDrafts(
+      Object.fromEntries(
+        editable.map((setting) => [setting.key, String(setting.value)])
+      )
+    );
+  }, [editable]);
```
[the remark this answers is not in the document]
plebioda: Fixed. `setting.value` is `unknown`, so `String(null)` really did seed the field with the literal text `"null"`, `isValid` accepted it as non-empty, and a save would have written that string back.

Now `setting.value == null ? '' : String(setting.value)`, with a comment saying why, since `String()` is the obvious thing to reach for.

ui/packages/plugins/om/src/components/ConfigForm.tsx
```diff
@@ -0,0 +1,341 @@
...
+  const render = (setting: OmInventorySetting) => (
+    <SettingField
+      key={setting.key}
+      setting={setting}
+      draft={drafts[setting.key] ?? ''}
+      resetting={reset.isPending}
+      onChange={(next) =>
+        setDrafts((current) => ({ ...current, [setting.key]: next }))
+      }
+      onReset={() => reset.mutate(setting.key)}
```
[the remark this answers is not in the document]
plebioda: Fixed. `reset.isError` now renders its own `Alert` beside `update.isError`. A failed reset was silent while the field kept the overridden value, which is the worst combination - the reader would have concluded the reset worked.

ui/packages/plugins/om/src/components/HealthBadge.tsx
```diff
@@ -0,0 +1,55 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import { Chip } from '@percona/percona-ui';
```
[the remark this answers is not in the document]
plebioda: Superseded by the MUI answer below - `@percona/peak-ui` is not needed here, and I have not added it as a dependency.

ui/packages/plugins/om/src/components/HealthBadge.tsx
```diff
@@ -0,0 +1,55 @@
...
+export function RunStatusBadge({ status }: { status: OmTopologyRunStatus }) {
+  return (
+    <Chip
+      size="small"
+      label={RUN_STATUS_LABEL[status] ?? status}
+      color={RUN_STATUS_COLOR[status] ?? 'default'}
+      variant={status === 'RUN_STATUS_RUNNING' ? 'outlined' : 'filled'}
+    />
+  );
+}
```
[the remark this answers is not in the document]
plebioda: Fixed in both files, and the evidence is on your side.

Every other `Chip` in the repo comes from `@mui/material`: `HighAvailabilityBadge.tsx`, `NavItemBadge.tsx` and `SeverityChip.tsx`. There is no other `@percona/percona-ui` Chip anywhere. `ui/AGENTS.md` names MUI and `@percona/peak-ui` as the component libraries and does not mention `percona-ui` at all.

Switched to `import Chip from '@mui/material/Chip'` in `HealthBadge.tsx` and `SnapshotBar.tsx`. Zero risk: both used only `size`, `label`, `color` and `variant`, all standard MUI props, and MUI is already a dependency of the plugin.

`InventoryPage.tsx` keeps `Table`/`MRT_ColumnDef` from `@percona/percona-ui` - that is a real component with no MUI equivalent, and `percona-ui` is used 15 times elsewhere in the tree, so it is not banned, just not the right source for a Chip.

ui/packages/plugins/om/src/components/ProbeValue.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <Box component="span" sx={{ cursor: 'help' }}>
+        {value}
+        <Box
+          component="span"
+          sx={{ color: 'warning.main', ml: 0.5, whiteSpace: 'nowrap' }}
+        >
+          {age == null ? '(stale)' : `(${formatDuration(age)} old)`}
+        </Box>
+      </Box>
+    </Tooltip>
```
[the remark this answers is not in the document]
plebioda: Fixed. The tooltip anchor now takes `tabIndex={0}` and has a `:focus-visible` outline, so `detail` - which carries `last_error` - is reachable without a pointer. MUI opens a Tooltip on focus, but only if the anchor can receive focus, which a bare `<span>` cannot.

ui/packages/plugins/om/src/components/RunEntities.tsx
```diff
@@ -0,0 +1,208 @@
...
+import { formatDuration } from '../format';
+import { Unavailable } from './Unavailable';
+import type { OmInventoryRun } from '../types';
+
+/** How a host was matched, spelled out. The raw values are terse to the point of coy. */
+const RESOLUTION_LABEL: Record<string, string> = {
+  name: 'by node name',
+  address: 'by node address',
+  orphaned: 'no executor',
+};
```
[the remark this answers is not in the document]
plebioda: Fixed, and it was worse than a fallback: the labels never rendered at all.

`RESOLUTION_LABEL` was keyed `name` / `address` / `orphaned`, while `entity.resolution` carries `EXECUTOR_RESOLUTION_NAME` and friends. So every lookup missed and every row showed the raw enum. Now keyed as `Record<OmExecutorResolution, string>` - which makes the compiler enforce completeness - with all four values including `EXECUTOR_RESOLUTION_UNSPECIFIED`.

ui/packages/plugins/om/src/components/SnapshotBar.tsx
```diff
@@ -0,0 +1,64 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import { Stack, Tooltip, Typography } from '@mui/material';
+import { Chip } from '@percona/percona-ui';
```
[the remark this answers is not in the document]
plebioda: Fixed - see the `HealthBadge.tsx` reply. Now `Chip` from `@mui/material`, matching the three existing Chip call sites in the repo.

ui/packages/plugins/om/src/hooks.ts
```diff
@@ -0,0 +1,361 @@
...
+  OmEnvironmentSection,
+  OmProcessRole,
+  OmServiceRow,
+  OmTopologyResponse,
+  OmTopologyRun,
+  OmTopologyRunAccepted,
+  OmTopologyRunStatus,
+} from './types';
+
+const OM_BASE = '/v1/om';
```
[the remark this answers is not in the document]
plebioda: Declining this one. `src/lib/constants.ts` belongs to `apps/pmm`; this is a separate workspace package and cannot import from it.

The in-repo precedent for a plugin is the ATW plugin, which does exactly what OM does - `ui/packages/plugins/atw/src/hooks.ts:50`:

```ts
const ATW_BASE = '/apps/atw';
```

`OM_BASE` is the same shape in the same position. The guideline you quote is about the app, where `src/lib/constants.ts` exists.

ui/packages/plugins/om/src/hooks.ts
```diff
@@ -0,0 +1,361 @@
...
+      message?: string;
+    } | null;
+    // The status has to survive onto the error: a 409 from the trigger is an expected
+    // outcome the button renders differently, not a failure.
+    throw new OmApiError(
+      response.status,
+      body?.message ?? `Request failed with ${response.status}`
+    );
+  }
+  return (await response.json()) as T;
```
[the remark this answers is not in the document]
plebioda: Declining, on the same precedent as the `types.ts` comment - see there for the detail.

Short version: `axios-case-converter` is configured on the app's axios instance in `apps/pmm/src/api/api.ts`. This plugin does its own `fetch` against PMM's origin and never goes through that instance, so nothing converts casing for it. ATW's types are snake_case for the same reason.

ui/packages/plugins/om/src/hooks.ts
```diff
@@ -0,0 +1,361 @@
...
+}
+
+/** Refresh everything the snapshot backs, once a run reaches a terminal status. */
+export function useInvalidateOmTopologySnapshot() {
+  const queryClient = useQueryClient();
+  return () => {
+    queryClient.invalidateQueries({ queryKey: topologyKey });
+    queryClient.invalidateQueries({ queryKey: runsKey });
+  };
+}
```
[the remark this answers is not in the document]
plebioda: Declining. `src/hooks/` is the app's layout; for a plugin package the convention is a flat `hooks.ts`, which is what the ATW plugin has:

```
ui/packages/plugins/atw/src/  ->  AtwApp.tsx  hooks.ts  types.ts  index.ts  ...
```

No `src/hooks/` directory exists in any plugin. OM splitting topology and inventory into `hooks.ts` and `inventoryHooks.ts` already goes one step further than the precedent.

ui/packages/plugins/om/src/HostsPage.tsx
```diff
@@ -0,0 +1,534 @@
...
+      <Stack direction="row" spacing={1}>
+        <Tooltip title="Probe this host now. Dispatches a job and takes tens of seconds; the row updates when it lands.">
+          <Button
+            size="small"
+            disabled={refresh.isPending}
+            onClick={() => refresh.mutate([row.original.node_id])}
+          >
+            Refresh
+          </Button>
+        </Tooltip>
```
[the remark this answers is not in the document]
plebioda: Fixed, on both buttons. MUI disables pointer events on a disabled `ButtonBase`, so neither Tooltip opened while `refresh.isPending` - precisely when a reader wants to know why the button is dead.

Both are now wrapped in `<Box component="span">`, and the comment says the span is load-bearing so nobody tidies it away later.

ui/packages/plugins/om/src/inventoryHooks.ts
```diff
@@ -0,0 +1,372 @@
...
+      const { runs } = await request<{ runs: OmInventoryRun[] }>(
+        `/inventory/runs?limit=${limit}`
+      );
+      return runs ?? [];
+    },
+    refetchInterval: (query) =>
+      isRefreshActive(query.state.data?.[0]?.status)
+        ? REFRESH_POLL_MS
+        : ESTATE_POLL_MS,
+  });
```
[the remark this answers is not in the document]
plebioda: Fixed - good catch, and the reasoning is right.

`data?.[0]?.status` only looked at the newest run. Refreshes are host-scoped, so two can overlap, and a narrow one started later can finish first - leaving `data[0]` terminal while a broader run is still probing. Fast polling stopped and the estate was invalidated early, so the real completion would not show until the idle poll.

Now `(query.state.data ?? []).some((run) => isRefreshActive(run.status))`.

One thing I have left: `useInvalidateEstateOnRefreshEnd(runs?.[0]?.status)` still watches the newest run's edge. The polling fix means the data keeps refreshing while any run is active, so the estate is no longer stale for a poll interval - but the invalidation edge is still newest-run-only. Moving it to "the active set became empty" is the fuller fix and I would rather do it as its own change than fold it in here. Flag it if you would rather it went in now.

ui/packages/plugins/om/src/InventoryPage.tsx
```diff
@@ -0,0 +1,386 @@
...
+    header: 'Started',
+    Cell: ({ row: { original } }) => formatTimestamp(original.start_time),
+  },
+  {
+    accessorFn: (row) => formatRunDuration(row.start_time, row.end_time),
+    id: 'duration',
+    header: 'Duration',
+    Cell: ({ row: { original } }) =>
+      formatRunDuration(original.start_time, original.end_time) || '—',
+  },
```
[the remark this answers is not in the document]
plebioda: Fixed. `accessorFn` returned the formatted string, so the column sorted lexicographically and "9s" landed after "10m".

Extracted `runDurationSeconds()` from `formatRunDuration()` in `format.ts`, so the accessor sorts on the number while `Cell` keeps the formatted value - the pattern `HostsPage.tsx` already used for its age column. `formatRunDuration` now delegates to it, so there is one parse rather than two.

ui/packages/plugins/om/src/OverviewPage.tsx
```diff
@@ -0,0 +1,439 @@
...
+  const regionId = useId();
+  const columns = useColumns();
+  const table = useMaterialReactTable({
+    columns,
+    data: section.clusters,
+    enablePagination: false,
+    enableDensityToggle: false,
+    enableExpanding: true,
+    enableTopToolbar: false,
+    getRowId: (row) => row.cluster_name ?? UNNAMED_ENVIRONMENT,
```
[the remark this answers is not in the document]
plebioda: Fixed. `getRowId` is now `(row, index) => row.cluster_name ?? \`unnamed-${index}\``, so two unnamed clusters in one table no longer share a row id - which matters because MRT keys expansion and selection state on it.

ui/packages/plugins/om/src/ServicesPage.tsx
```diff
@@ -0,0 +1,467 @@
...
+  const rows = useMemo(
+    () =>
+      failingOnly
+        ? joined.filter((row) => row.inventory && isFailing(row.inventory))
+        : joined,
+    [joined, failingOnly]
+  );
+  const failingCount = useMemo(
+    () =>
+      joined.filter((row) => row.inventory && isFailing(row.inventory)).length,
```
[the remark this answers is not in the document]
plebioda: Fixed, and this was the most valuable of the sixteen - it is the same class of bug as the null-versus-absent one earlier in this stack: two different states collapsing into one.

`joinServiceInventory` does `(inventory ?? [])`, so a failed estate query produced an empty map, every row got `inventory: null`, `ProbeValue` reported `not_in_inventory` for all of them, and `failingCount` was 0. A confident wrong answer built out of one failed request.

Three changes, keeping the topology table fully usable as the page's comment always intended:

1. A new `inventory_unavailable` reason in `OmUnavailableReason` / `UNAVAILABLE_PHRASE`, distinct from `not_in_inventory`. That one is a statement about the estate; this one admits the estate could not be read.
2. `ProbeValue` takes `inventoryUnavailable` and picks between the two, so the probe columns stop asserting anything.
3. `failingCount` is `number | null` - null when the estate failed - and `Counts` renders "probe status unavailable" instead of a chip. "0 failing" is a claim, and an empty join is not evidence for it.

Plus a non-blocking `Alert` above the table saying the estate could not be read and that the topology columns are unaffected.

ui/packages/plugins/om/src/types.ts
```diff
@@ -0,0 +1,630 @@
...
+export interface OmInventorySetting {
+  key: string;
+  value: unknown;
+  default_value: unknown;
+  type: string;
+  reload: OmSettingReload;
+  has_override: boolean;
+  is_advanced: boolean;
+  description?: string | null;
+}
```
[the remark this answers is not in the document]
plebioda: Declining, on the in-repo precedent.

The guideline holds where `axios-case-converter` runs - it is configured on the app's axios instance in `apps/pmm/src/api/api.ts`. This plugin does not use that instance: it does its own `fetch` against PMM's origin (which is also why it needs no SEP bearer). Nothing converts casing on this path, so camelCase models would have to be produced by a hand-written mapping layer over every response.

The sibling plugin does the same thing. `ui/packages/plugins/atw/src/types.ts` declares `category_root`, `parent_category`, `parent_category_label`, `category_label` - snake_case throughout, for the same reason.

So OM matches the only other plugin in the tree. If PMM wants plugins to convert at the boundary, that is worth doing to both plugins together rather than making OM the exception - happy to take it on as its own change if that is the direction.

ui/packages/plugins/om/src/inventoryHooks.ts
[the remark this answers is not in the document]
plebioda: Done - the full fix is in.

`useInvalidateEstateOnRefreshEnd` now takes the runs collection rather than a single status, derives `anyActive` with `some(...)`, and fires only on the nonempty -> empty edge of the active set. Both call sites in `HostsPage.tsx` and `InventoryPage.tsx` pass `runs` instead of `runs?.[0]?.status`.

You were right that the half-fix was worse than it looked: with only the polling changed, a narrow run reaching a terminal status would still have invalidated while a broader run was mid-write, refetching rows the older run had not finished - and then never firing again for the run that actually mattered. The doc comment now spells that out so the collection argument does not look incidental.

ui/packages/plugins/om/src/OverviewPage.tsx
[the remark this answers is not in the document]
plebioda: Fixed - and this one was my miss, not a deferral. Your original finding named both collisions and I only addressed one.

`key={section.env_name ?? UNNAMED_ENVIRONMENT}` is now `key={section.env_name ?? \`unnamed-${index}\`}`, matching the `getRowId` fallback, with a comment tying the two together so the next reader sees they are the same problem at two levels.

ui/packages/plugins/om/src/inventoryHooks.ts
```diff
@@ -0,0 +1,381 @@
...
+}
+
+/**
+ * True while a refresh has not reached a terminal status.
+ *
+ * Typed on the union rather than on `string`, because the compiler is the only thing
+ * that would have caught this comparison going stale when the wire values changed.
+ */
+export function isRefreshActive(
```
fabio-silva: what's the difference between this and `isRunActive` ?

ui/packages/plugins/om/vitest.config.ts
```diff
@@ -0,0 +1,46 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import react from '@vitejs/plugin-react-swc';
```
fabio-silva: Let's use `@vitejs/plugin-react'` instead, we're changing that on main as well

ui/packages/plugins/om/src/format.ts
```diff
@@ -0,0 +1,112 @@
+/**
```
fabio-silva, line 1: `date-fns` is available on the monorepo, I think these functions can be replaced with theirs

ui/packages/plugins/om/src/hooks.ts
```diff
@@ -0,0 +1,362 @@
+/**
```
fabio-silva: why is there `hooks` and `inventoryHooks` on the same folder? This file is mostly non-hooks, but only a bunch of utils

ui/packages/plugins/om/src/hooks.ts
```diff
@@ -0,0 +1,362 @@
...
+
+/** Poll interval while a collection run is in flight (ms). */
+const RUN_POLL_MS = 3000;
+// Idle cadence for the document and the run history. pmm-managed collects on its own
+// schedule, so both change without anyone here asking; without a poll the page only
+// ever shows what it happened to load with.
+const SNAPSHOT_POLL_MS = 30000;
+
+const topologyKey = ['om', 'topology'] as const;
+const runsKey = ['om', 'runs'] as const;
```
fabio-silva: if these are static constants, let's uppercase

ui/packages/plugins/om/src/inventoryHooks.ts
```diff
@@ -0,0 +1,381 @@
...
+ * per-row refresh will meet it sooner or later. The caller reads `OmApiError.status`
+ * and says so.
+ */
+export function useRefreshInventory() {
+  const queryClient = useQueryClient();
+  return useMutation<OmInventoryRunAccepted, Error, string[] | undefined>({
+    mutationFn: (nodeIds) =>
+      request<OmInventoryRunAccepted>('/inventory/runs:trigger', {
+        method: 'POST',
+        body: JSON.stringify(nodeIds?.length ? { node_ids: nodeIds } : {}),
```
fabio-silva: I don't see any call to `useRefreshInventory`'s `mutate` with arguments, why is body or even `JSON.stringify()` here?

ui/packages/plugins/om/src/InventoryPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+    <Stack direction="row" alignItems="center" gap={1}>
+      <Tooltip title="Probe every host in the estate and collect what no metric carries">
+        <span>
+          <Button
+            variant="contained"
+            startIcon={
+              running ? <CircularProgress size={16} /> : <PlayArrowIcon />
+            }
+            disabled={running || trigger.isPending}
+            onClick={() => trigger.mutate(undefined)}
```
fabio-silva: what's the `undefined` for?

ui/packages/plugins/om/src/hooks.ts
```diff
@@ -0,0 +1,362 @@
...
+ * VictoriaMetrics is on localhost — so the response already carries a terminal status
+ * rather than a `202` to poll.
+ */
+export function useTriggerOmTopologyRun() {
+  const queryClient = useQueryClient();
+  return useMutation<OmTopologyRunAccepted, Error, void>({
+    mutationFn: () =>
+      request<OmTopologyRunAccepted>('/topology/runs:collect', {
+        method: 'POST',
+        body: '{}',
```
fabio-silva: unnecessary body

ui/packages/plugins/om/src/hooks.ts
```diff
@@ -0,0 +1,362 @@
...
+}
+
+/**
+ * The whole estate from the latest terminal snapshot.
+ *
+ * One request, one document: the API assembles the tree per run, so there is no
+ * per-cluster fetch and nothing to paginate. Keeping the previous page in place stops
+ * the table collapsing to a spinner while a refetch is in flight.
+ */
+export function useOmTopology() {
```
fabio-silva: why is this here instead of on `inventoryHooks`?

ui/packages/plugins/om/src/components/ProbeValue.tsx
```diff
@@ -0,0 +1,125 @@
...
+        reason={
+          inventory.freshness.last_success_at
+            ? 'metric_not_collected'
+            : 'probe_never_succeeded'
+        }
+      />
+    );
+  }
+
+  const failing = inventory.freshness.failing_since != null;
```
fabio-silva, line 79: ### 
```suggestion
  const failing = inventory.freshness.failing_since !== null;
```

fabio-silva: Please change remaining operators like this

ui/packages/plugins/om/src/components/ConfigForm.tsx
```diff
@@ -0,0 +1,348 @@
...
+  return 'text';
+}
+
+/** Parse an edited value back into what the app expects on the wire. */
+function toWireValue(setting: OmInventorySetting, draft: string): unknown {
+  switch (inputKind(setting)) {
+    case 'bool':
+      return draft === 'true';
+    case 'int':
+      return Number(draft);
```
fabio-silva, line 88: according to code, `timedelta` type also returns `int`. What's the result of calling `Number(draft)` if `draft` is a `timedelta`?

ui/packages/plugins/om/src/components/HealthBadge.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+/** Discovery-run status as a chip. */
+export function RunStatusBadge({ status }: { status: OmTopologyRunStatus }) {
+  return (
+    <Chip
+      size="small"
+      label={RUN_STATUS_LABEL[status] ?? status}
+      color={RUN_STATUS_COLOR[status] ?? 'default'}
+      variant={status === 'RUN_STATUS_RUNNING' ? 'outlined' : 'filled'}
```
fabio-silva: there's an util created to check whether status is running or not. Let's either use it anywhere or nowhere

ui/packages/plugins/om/src/components/Metric.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const UNMEASURED = -1;
+
+/**
+ * Render a percentage, or the reason there is none.
+ *
+ * `-1` is the worker's not-measured sentinel; every other value — including `0` —
+ * is a real reading and must render as a number. A falsy check here would report an
+ * idle server as unmonitored.
+ */
+export function Percent({ value }: { value: number }) {
```
fabio-silva: We use arrow functions for React components. Please refactor on all components.

fabio-silva (changes requested):

fabio-silva:
@plebioda please include screenshots

ui/packages/plugins/om/src/inventoryHooks.ts
plebioda: None. Deleted `isRefreshActive` and pointed its seven callers at `isRunActive`, which now says in its docstring that it covers both run kinds - they are different operations but report the same `RunStatus`, so it is one question. `4149c92`.

It has since moved to `api.ts` in the file split below.

ui/packages/plugins/om/vitest.config.ts
plebioda: Done in `428084b` - `@vitejs/plugin-react`, with `@vitejs/plugin-react-swc` dropped from the package. Vite prints the same recommendation itself on every run now.

Only this package: the rest of the tree is still on `-swc` because this branch's base predates that change on `main`, and those are not this PR's to move.

ui/packages/plugins/om/src/format.ts
plebioda: Done in `428084b`. Parsing, validity and differences now come from `parseISO`, `isValid`, `differenceInMilliseconds` and `format`, and `date-fns` is declared on the package.

One helper stayed, renamed `formatCompactDuration` to say why: it is a compact two-unit formatter (`2d 3h`, `4m 12s`) for table cells and chips, and date-fns' `formatDuration` emits `2 days 3 hours` - the short form needs a custom `locale.formatDistance`, which is more code than the arithmetic it replaces. The rename also ends the collision with date-fns' export of the same name, so a file can import both.

Two behaviour changes fell out. Both look like improvements to me, but flagging them in case you disagree:

- `formatTimestamp` moves from `toLocaleString()` to a fixed `yyyy-MM-dd HH:mm:ss` in local time, because run timestamps are read down a column and a locale-varying format is not a column anyone can scan. Covered by a test.
- `formatAge` deliberately does **not** use `formatDistanceToNowStrict`, which rounds to a single unit and answers "about 1 hour ago". These columns have to distinguish `1h 12m` from `1h 58m`.

ui/packages/plugins/om/src/hooks.ts
plebioda: Agreed on both counts, and `hooks.ts` is gone in `e1d2c3c`. Only four of its ten exports were hooks; the rest was the fetch client, the error class, a status predicate and four pure roll-ups.

It now mirrors the shape the inventory side already had:

| file | holds |
| --- | --- |
| `api.ts` | `OmApiError`, `request`, `isRunActive` - shared by both halves |
| `topology.ts` | `toServiceRows` / `toClusterRows` / `toEnvironmentSections` |
| `topologyHooks.ts` | the four topology hooks |

`api.ts` holds the transport because both halves use it: the topology document and SEP's proxied estate are very different sources that arrive over one origin with one auth story and one error envelope. And naming the pair `topologyHooks.ts` / `inventoryHooks.ts` is what stops the question you asked further down from arising again.

ui/packages/plugins/om/src/hooks.ts
plebioda: Done - `TOPOLOGY_KEY` / `RUNS_KEY` (now in `topologyHooks.ts`) and the four in `inventoryHooks.ts`. They were the only lower-camel module constants in either file, three lines below `OM_BASE` and `RUN_POLL_MS`. `4149c92`.

ui/packages/plugins/om/src/inventoryHooks.ts
plebioda: There is one - `HostsPage.tsx`, the per-row Refresh action, which called `refresh.mutate([row.original.node_id])`. That is what the plural on `node_ids` buys: the row action and a full sweep are the same endpoint with a different scope, so the scoped call has to serialise `{node_ids: [...]}`.

The `: {}` fallback was genuinely unnecessary though, for the same reason as your "unnecessary body" note - fixed in `4149c92`, so an unscoped sweep now sends no body at all and the scoped one still sends its ids.

Separately, `e1d2c3c` replaces the raw `mutate` with `refreshAll()` and `refreshHosts(ids)`, so the two scopes are now named at every call site rather than being one optional argument.

ui/packages/plugins/om/src/InventoryPage.tsx
plebioda: It was load-bearing rather than decoration: TanStack's `MutateFunction` is `(variables, options?)` with `variables` positional, so with `TVariables = string[] | undefined` a plain `mutate()` does not compile. The contrast is `SyncButton.tsx`, which does write `trigger.mutate()` because that mutation's variable is `void`.

You are right that it reads as an oversight, though, so it is gone. `useRefreshInventory` now returns `refreshAll()` and `refreshHosts(ids)` beside the mutation state, and the three call sites say which sweep they are asking for. `e1d2c3c`.

ui/packages/plugins/om/src/hooks.ts
plebioda: Right, dropped. `body: "*"` on the method plus grpc-gateway's generated decoder treating `io.EOF` as an empty request (`om.pb.gw.go:138`) means the empty message binds either way. Same fix applied to the inventory trigger, which no longer sends `{}` for an unscoped sweep. `4149c92`.

ui/packages/plugins/om/src/hooks.ts
plebioda: This one reads `/v1/om/topology`, which is pmm-managed's own derivation from PMM inventory and VictoriaMetrics - it touches no host and has nothing to do with SEP's estate.

But the docstring opened with "The whole **estate** from the latest terminal snapshot", and "estate" is this package's word for the *inventory* side everywhere else, so the question was fair. Reworded in `4149c92`; the split you raise above then made it structural, and the hook now lives in `topologyHooks.ts`.

ui/packages/plugins/om/src/components/ProbeValue.tsx
plebioda: Holding this one, and it is worth saying why at length because the second comment asks to generalise it.

`failing_since` is declared `?: string | null` in `types.ts`, so `undefined` is a value it can hold. `undefined !== null` is `true`, which means the suggestion renders a healthy service as failing - the value, plus a `(stale)` badge, plus a tooltip claiming a probe failure.

Narrow point in its favour, which I checked: at *this* line the two are currently equivalent. `failing_since` is a `google.protobuf.Timestamp` (`om.proto:450`), not a proto3 `optional`, so it has presence and protojson emits an explicit `null` for it under `EmitUnpopulated` (set in `managed/cmd/pmm-managed/main.go:387`). So `!== null` would work here today. It works by accident of one field's proto kind, though, and it contradicts the type the file declares - which is why I would rather not.

`oxlintrc.json` also sets `"eqeqeq": ["warn", "always", { "null": "ignore" }]`, so `!= null` is deliberately exempt repo-wide, and `apps/pmm/src` uses it in 20 places.

If the real objection is that `types.ts` makes "absent" and "null" indistinguishable, I agree that is worth fixing - but properly: type each field to the single absence its proto kind actually produces (`?: T` for `optional` scalars, `: T | null` for message fields), with `om.proto` as the authority. Happy to do that here or as a follow-up.

plebioda: This is the one I want to push back on rather than the line above, because applied across the file it is a live bug.

Fields declared proto3 `optional` are **dropped** by protojson even under `EmitUnpopulated` - they sit in a synthetic oneof, and protobuf-go skips those (`encoding/protojson/encode.go:210`: `if m.Has(fd) || fd.ContainingOneof() != nil { continue }`). pmm-managed documents the same behaviour at `managed/services/om/inventory.go:661`, and it is why those fields are `*string` / `*float64` on the Go side at all.

So `optional double replication_lag_seconds = 15;` (`om.proto:129`) arrives as an absent key, and the cluster roll-up is:

```ts
if (service.replication_lag_seconds != null) {
  maxLag = maxLag == null
    ? service.replication_lag_seconds
    : Math.max(maxLag, service.replication_lag_seconds);
}
```

Under `!== null` that guard admits `undefined`, `Math.max(n, undefined)` is `NaN`, and every cluster row on Overview reports `NaN` for max replication lag and min oplog window. Same shape for `state`, `version`, `last_error` and the rest of the `optional` scalars.

Keeping `!= null` as the uniform form for that reason - it is the only one that catches both absences, and both do occur on this wire. I have added a comment at those two guards in `topology.ts` saying so, since they are the ones where it matters.

ui/packages/plugins/om/src/components/ConfigForm.tsx
plebioda: It yields the seconds, so `Number(draft)` is right - but only because of an annotation rather than the type, which is worth recording.

SEP declares those fields as `Annotated[timedelta, PlainSerializer(lambda v: round(v.total_seconds()), return_type=int)]` (`app/core/utils/fields.py:852`), and `dump_field_value` serialises through the annotated adapter, so the configuration API returns `1800` - not `"PT30M"`, which is what pydantic's ISO-8601 default would give. The `type` string comes from `__name__` on the stripped annotation, hence the literal `"timedelta"` this matches on.

Added a comment on `toWireValue` saying so in `4149c92`: drop that serialiser upstream and `Number` yields NaN, every duration field renders invalid, and nothing in either repo would have explained why.

ui/packages/plugins/om/src/components/HealthBadge.tsx
plebioda: Agreed - `isRunActive(status)` now. Ties in with the duplicate predicate above; both in `4149c92`.

ui/packages/plugins/om/src/components/Metric.tsx
plebioda: Done - 30 components across 13 files, in `e1d2c3c`. Hooks and plain helpers keep their function declarations. Landed as its own mechanical commit so it does not sit on top of the behavioural fixes.

Only this package, same as the vite plugin.

ui/packages/plugins/om/src/inventoryHooks.ts
```diff
@@ -0,0 +1,381 @@
...
+  return useQuery<OmInventoryService[]>({
+    queryKey: servicesKey,
+    placeholderData: keepPreviousData,
+    queryFn: async () => {
+      const { services } = await request<{ services: OmInventoryService[] }>(
+        '/inventory/services'
+      );
+      return services ?? [];
+    },
+    refetchInterval: ESTATE_POLL_MS,
```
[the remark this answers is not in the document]
plebioda: Correct, and fixed properly rather than cheaply in `fc5e93a`.

The cheap version - mounting the invalidator on ServicesPage too - would add a runs query to a page that does not otherwise need one, and would still depend on every future page remembering. So the watch is now internal to `useOmInventoryHosts` and `useOmInventoryServices` instead, where there is nothing to forget. It also drives their cadence: three seconds while a sweep is in flight rather than the idle minute, since a refresh writes rows as its dispatches land and a half-written estate should converge rather than sit.

The runs query is keyed, so a page that reads the history itself shares this one rather than opening a second.

ui/packages/plugins/om/src/components/ConfigForm.tsx
```diff
@@ -0,0 +1,348 @@
...
+      <Box sx={{ minWidth: 280, maxWidth: 280 }}>
+        {kind === 'bool' ? (
+          <Stack direction="row" alignItems="center" gap={1}>
+            <Switch
+              size="small"
+              checked={draft === 'true'}
+              onChange={(event) =>
+                onChange(event.target.checked ? 'true' : 'false')
+              }
+            />
```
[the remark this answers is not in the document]
plebioda: Fixed - `aria-label` on the switch input, so the setting's label is announced with the control.

The second half of the comment does not hold, though: there is exactly one `<Switch>` in this file, and line 231 is the `useEffect` that re-seeds the drafts after a write.

plebioda:
Thanks @fabio-silva - review round pushed, `4c395dc` → `fc5e93a`. Replied in each thread; the summary:

**Taken (13 of 15)**

| commit | what |
| --- | --- |
| `8a38150` | Services no longer reads a *pending* estate as an empty one; Overview keeps Sync in its error state; both refresh buttons derive "running" from every run rather than `runs[0]`; Hosts table keyed on `node_id` |
| `c700ad7` | `pom.v1` → `om.v1`; the inventory types are exported; dead `useOmTopologyRun` deleted; `aria-label` on the config switches |
| `4149c92` | `isRefreshActive` deleted in favour of `isRunActive`; query keys uppercased; both triggers drop their unnecessary bodies; the `timedelta` chain documented |
| `428084b` | `@vitejs/plugin-react`; `format.ts` onto `date-fns` |
| `e1d2c3c` | `hooks.ts` split into `api.ts` / `topology.ts` / `topologyHooks.ts`; `refreshAll()` / `refreshHosts()` replace `mutate(undefined)`; 30 components converted to arrow functions |
| `fc5e93a` | the estate queries follow their own refreshes, so Services no longer goes stale for a minute |

**Held back (1)** - the `!= null` → `!== null` change, and specifically the ask to apply it to the remaining operators. Proto3 `optional` fields are dropped by protojson rather than nulled, so under `!==` the cluster roll-up admits `undefined` into `Math.max` and reports `NaN` for max replication lag and min oplog window on every row. Full reasoning in the thread - happy to be argued out of it, and I have offered a version of the underlying fix I do think is right.

**Two things to know before re-reading**

- **The lockfile block changed.** The PR body says this branch carries only its own importer entry, which is still true, but that entry now declares `date-fns` and `@vitejs/plugin-react` and no longer declares `@vitejs/plugin-react-swc`. `pnpm install --frozen-lockfile` re-verified on this branch and on #5818 separately. I will update the body.
- **`formatTimestamp` moved from `toLocaleString()` to a fixed `yyyy-MM-dd HH:mm:ss`.** Deliberate - run timestamps are compared down a column - but it is a visible change and easy to revert if you would rather keep the reader's locale.

Vite plugin and arrow components were applied to this package only; the rest of the tree is still on `-swc` and function declarations, and those are not this PR's to move.

Screenshots still to come.

plebioda:
> @plebioda please include screenshots

Screenshot from 2026-08-25 11-33-03
Screenshot from 2026-08-25 11-32-58
Screenshot from 2026-08-25 11-32-54
Screenshot from 2026-08-25 11-32-49

ui/packages/plugins/om/src/ServicesPage.tsx
```diff
@@ -0,0 +1,497 @@
...
+  const { data, isPending, isError, error } = useOmTopology();
+  // Deliberately not gated on the estate loading or failing. The snapshot is PMM's own
+  // and always available; the estate is a second service that may be unwell, and a
+  // page that blanked when it was would be exactly what proxying it through
+  // pmm-managed was meant to stop.
+  const {
+    data: inventory,
+    isError: inventoryFailed,
+    error: inventoryError,
+  } = useOmInventoryServices();
```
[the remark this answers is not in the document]
plebioda: Correct, and fixed in `8a38150`.

`ProbeValue`'s `inventoryUnavailable?: boolean` is now a three-state `OmEstateStatus` (`'ready' | 'pending' | 'unavailable'`) with a `missingRowReason()` helper, and there is a new `inventory_pending` reason in the vocabulary. All four sites on this page that used to hard-code `not_in_inventory` go through it, so a row with no estate entry says which of the three silences it is.

`failingCount` is `null` whenever the estate has not answered - pending as well as failed - rather than only on error, and the chip reads "reading probe status…" while it is in flight. The `isError` path had been thought about; the pending one had not.

ui/packages/plugins/om/src/InventoryPage.tsx
```diff
@@ -0,0 +1,393 @@
...
+ * Separate from `SyncButton`, which rebuilds pmm-managed's snapshot in about a tenth
+ * of a second and never touches a host. This queues a job per host and takes tens of
+ * seconds, so the app answers `running` and the list below polls until it lands. Two
+ * triggers that different must not look alike, which is most of why they live on
+ * different pages.
+ */
+function RefreshButton() {
+  const { data: runs } = useOmInventoryRuns();
+  const trigger = useRefreshInventory();
+  const running = isRefreshActive(runs?.[0]?.status);
```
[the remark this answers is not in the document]
plebioda: Right, and it contradicted this PR's own reasoning - `useOmInventoryRuns`' poll cadence and `useInvalidateEstateOnRefreshEnd` both scan every run and say in their comments why. The button did not.

Fixed in `8a38150`: `running` now comes from every returned run. Hosts had no active-run guard at all, on either the row action or "Refresh all", so both got the same condition. `fc5e93a` then moved the fold behind `useIsEstateRefreshing()` so neither page repeats it.

ui/packages/plugins/om/src/hooks.ts
```diff
@@ -0,0 +1,362 @@
...
+ *
+ * Read out of the history rather than fetched: a run is small, the list is already in
+ * the cache, and pmm-managed keeps the history in memory anyway — a per-run endpoint
+ * would be a second way to say the same thing.
+ */
+export function useOmTopologyRun(run_id: string | undefined) {
+  const { data, ...rest } = useOmTopologyRuns();
+  return {
+    ...rest,
+    data: run_id ? data?.find((run) => run.run_id === run_id) : undefined,
```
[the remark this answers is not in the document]
plebioda: Correct on both halves - the lookup was bounded by the history page, and `GET /v1/om/topology/runs/{run_id}` does exist for exactly this (`om.proto:868`), so the docstring's claim that a per-run endpoint would be redundant was wrong.

Deleted rather than repointed, in `c700ad7`: the hook had no call sites, only a re-export from the barrel. Whatever needs per-run detail should query that endpoint under its own cache key rather than inherit a hook written against the list.

ui/packages/plugins/om/src/index.ts
```diff
@@ -0,0 +1,114 @@
...
+  OmTopologyResponse,
+  OmTopologyRun,
+  OmTopologyRunAccepted,
+  OmTopologyRunCounts,
+  OmTopologyRunError,
+  OmTopologyRunStatus,
+  OmTopologySnapshotEnvelope,
+  OmTopologySummary,
+  OmUnavailableReason,
+} from './types';
```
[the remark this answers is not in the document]
plebioda: Fixed in `c700ad7`. The barrel now exports the eighteen inventory types alongside the hooks that return them, plus `OmHostFilters` (which lives in `inventoryHooks.ts`) and `OmClusterType`, which had been missed on the topology side too.

ui/packages/plugins/om/src/OverviewPage.tsx
```diff
@@ -0,0 +1,447 @@
...
+
+  if (isError) {
+    return (
+      <Alert severity="error">
+        {/* A 503 here is the expected first-run state, not a fault: the API says so
+            when no discovery has completed, and Sync is the way out. */}
+        {(error as Error)?.message ?? 'Could not load the topology.'}
+      </Alert>
+    );
+  }
```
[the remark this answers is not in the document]
plebioda: Correct, and it matters more than it looks: Overview is the index route, so this is the first thing a fresh installation sees.

Fixed in `8a38150` by mirroring what ServicesPage already does - the error branch keeps `OmHeader` and its `SyncButton`, with the alert below it.

ui/packages/plugins/om/src/HostsPage.tsx
```diff
@@ -0,0 +1,541 @@
...
+          !row.executor.driver_healthy
+      ).length,
+      failing: rows.filter((row) => isFailing(row)).length,
+    }),
+    [rows]
+  );
+
+  const table = useMaterialReactTable({
+    columns,
+    data: rows,
```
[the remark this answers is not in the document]
plebioda: Correct. Fixed in `8a38150` with `getRowId: (row) => row.node_id`, matching the run and cluster tables.

Worth noting the window is wider than "periodically refetched": as of `fc5e93a` this query also drops to a three-second cadence while a refresh is in flight, so rows move more often than they used to.

plebioda:
Ported five commits from the integration branch, implementing Adamo's UI feedback on [PMM-15326 comment 477895](https://perconadev.atlassian.net/browse/PMM-15326?focusedCommentId=477895) plus two bugs a follow-up review found in that work:

- [`34131a5f6`](https://github.com/percona/pmm/commit/34131a5f6) — Inventory: nine quick-filter chips (15m/30m/1h/4h/8h/today/week/month/all) over the server-side date window from #5813/#5816.
- [`669e486c8`](https://github.com/percona/pmm/commit/669e486c8) — Hosts: Not monitored/Monitored/All filter (defaults to All -- tried defaulting narrower first, reverted after a monitored host vanished from the page on load and read as a sync bug), a Reachable/Unreachable chip replacing raw latency, and bulk row selection wired to live Refresh/Forget actions.
- [`598c2c85e`](https://github.com/percona/pmm/commit/598c2c85e) — Overview: P/S badges beside the service name (traced `state` to `replSetGetStatus().stateStr`, unmapped, ten real values -- amber covers the ones outside Adamo's ask, gray for hidden/delayed is flagged as a separate backend gap rather than faked), mongos's own red badge, Member state column removed from both views.
- [`cca306294`](https://github.com/percona/pmm/commit/cca306294) — fixes the Inventory `since` window freezing at click time and silently growing past its own chip label the longer a tab stayed open.
- [`d353526e3`](https://github.com/percona/pmm/commit/d353526e3) — fixes bulk Forget retrying (and re-deleting) hosts that had already succeeded, and Cancel clearing the row selection it shouldn't have touched.

Fast-forward push, no rebase. `tsc --noEmit`, `oxlint`, `oxfmt --check` clean; 66/66 Vitest tests pass.

mattiasimonato (approved):

fabio-silva (approved):

plebioda:
Resolved 3 stale threads above (@fabio-silva's questions on inventoryHooks/InventoryPage/hooks.ts) - each already had a substantive reply from a prior session citing the fix commit, just never marked resolved.

PR #5818, opened 2026-08-21
PMM-15326: Mount OM in the PMM app and its navigation

plebioda:
Ticket number: PMM-15326

Feature build: N/A on its own - this is the last piece, so a build here covers the whole OM feature. Happy to produce one.

## What

Mounts the OM plugin in the PMM app: the sidebar entry with its four children, the route, and the page chrome. Seven files, ~130 lines.

This is deliberately separated from the plugin itself (#5817) because it carries the decisions a plugin reviewer is the wrong person for - where OpenManager appears, what it is called, and who may see it.

## The three decisions in here

**OM is not part of `addSepApps`, and that is deliberate.** Those entries are gated as a group on SEP, and the group is expected to gain a flag gate with real auth. OM is served by pmm-managed over PMM's own inventory and metrics, so hiding it when SEP is off or unreachable would hide a working page. It goes in as its own entry instead.

**The chrome is `OmPage`, not `SepPage`.** `SepPage` holds its children behind `SepAuthGate` until a SEP bearer has been minted from the PMM session, and it fails closed - so a SEP that is down, unconfigured or refusing the exchange would blank a page whose data comes from pmm-managed's own inventory, VictoriaMetrics and stored snapshot. OM has no SEP call on any browser path, so it must not inherit SEP's availability. It used `SepPage` while its backend was a SEP app; #5816 moved the estate behind `/v1/om/inventory`, which is what makes this correct now.

**Admin-only, enforced on the route rather than in the sidebar.** `NavigationProvider` only *hides* the entry for non-admins - the route still matches on direct navigation. So `OmPage` gates it too, using the nav's own predicate: `isPMMAdmin` is `isGrafanaAdmin || orgRole === Admin`, and `roles` (org-role only) cannot express the Grafana-admin half, so it covers the remaining case and `Page` renders its standard unauthorized card.

Worth a second opinion on the label and the icon: the entry reads **OpenManager** with `MongoIcon`, on the grounds that the feature is MongoDB-only today. If the intent is to widen it later, a Mongo icon is a thing to regret.

## The route is a splat

`OmApp` composes its own `<Routes>`, like `AtwApp`, so the mount is `${OM_PATH}/*`. All four pages sit under it. Inventory used to be mounted separately inside `SepPage` because it read SEP's app directly; it does not any more.

## The lockfile line

The three lines adding `'@sep/plugins-om'` to the `apps/pmm` importer are here rather than in #5817, with the `package.json` that declares the dependency. Splitting it that way is what lets `pnpm install --frozen-lockfile` pass on both branches. Verified on each.

## Related work

Last of seven stacked PRs splitting #5795 ("OpenManager initial implementation") into area-owned reviews.

- #5811 - `PMM_SEP_URL` / `PMM_SEP_TOKEN` redaction
- #5812 - the OM topology tables and models
- #5813 - the OM API
- #5815 - the role rules for the OM routes
- #5816 - the OM service in pmm-managed
- #5817 - the OM UI plugin (**this PR's base**)

The base is a real dependency: the router imports `OmApp` from `@sep/plugins-om`, so this does not typecheck without #5817.

**One thing the reviewer should know:** `PMM-15299-open-manager` currently conflicts with `main`, and two of this PR's files - `ui/apps/pmm/package.json` and `src/contexts/navigation/navigation.provider.tsx` - are among the seven conflicting ones, because the same pnpm toolchain migration (PMM-15288) landed on `main` as #5728 and on this branch via its own line. That conflict predates this work and does not affect merging into the base, but reconciling it later will touch these files, `navigation.provider.tsx` most sharply since that is where `addOm()` is wired in.

### DIFF
--- a/ui/apps/pmm/package.json
+++ b/ui/apps/pmm/package.json
@@ -33,6 +33,7 @@
     "@sep/api": "workspace:*",
     "@sep/framework": "workspace:*",
     "@sep/plugins-atw": "workspace:*",
+    "@sep/plugins-om": "workspace:*",
     "@tanstack/react-query": "^5.100.7",
     "axios": "^1.13.5",
     "axios-case-converter": "^1.1.1",
--- a/ui/apps/pmm/src/contexts/navigation/navigation.provider.tsx
+++ b/ui/apps/pmm/src/contexts/navigation/navigation.provider.tsx
@@ -13,6 +13,7 @@ import {
   addUsersAndAccess,
   addHomePage,
   addSepApps,
+  addOm,
 } from './navigation.utils';
 import { useUser } from 'contexts/user';
 import { useAdvisors } from 'hooks/api/useAdvisors';
@@ -93,6 +94,9 @@ export const NavigationProvider: FC<PropsWithChildren> = ({ children }) => {
         // SEP apps mounted as native routes (migration). Shown once the session
         // is established; role/flag gating comes with real auth (Option B).
         items.push(...addSepApps());
+
+        // Served by pmm-managed, so it is not gated with the SEP group.
+        items.push(...addOm());
 
         if (settings.backupManagementEnabled) {
           items.push(NAV_BACKUPS);
--- a/ui/apps/pmm/src/contexts/navigation/navigation.utils.tsx
+++ b/ui/apps/pmm/src/contexts/navigation/navigation.utils.tsx
@@ -1,7 +1,7 @@
 import DarkModeOutlined from '@mui/icons-material/DarkModeOutlined';
 import LightModeOutlined from '@mui/icons-material/LightModeOutlined';
 import MonitorHeartIcon from '@mui/icons-material/MonitorHeart';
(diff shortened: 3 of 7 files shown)


plebioda:
@coderabbitai review

fabio-silva (approved):

ademidoff (approved):

mattiasimonato (approved):

plebioda:
Heads-up @fabio-silva: force-pushed, `d02d613` → `fa4d20e8`, which will have dismissed your approval. Sorry about that.

**Nothing in this PR changed.** Its single commit is identical - `git range-diff` is empty. The rebase was only to sit on the new head of #5817, which took a review round.

One thing that did move underneath it: this branch's three `ui/pnpm-lock.yaml` lines still add `@sep/plugins-om` to the `apps/pmm` importer, but the plugin's own importer block on #5817 now declares `date-fns` and `@vitejs/plugin-react`. `pnpm install --frozen-lockfile` re-checked on both branches separately.

PR #13, opened 2016-09-06
Updates for 1.0.4 release

azhebel:

PR #5824, opened 2026-08-23
PMM-15160 Fix QAN sparkline divide by zero

theTibi:
Ticket number: PMM-15160

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/4540

## Summary

- **Bug**: `qan-api2` panics with `runtime error: integer divide by zero` when a QAN report's time range collapses into a single wall-clock minute. The report fails to load and a full goroutine stack is written to `/srv/logs/qan-api2.log` on every request.
- **Root cause**: both `SelectSparklines` implementations align the period's bounds *down* to the minute and then divide by the number of minutes in it. When both bounds land in the same minute the period becomes `0`, so `amountOfPoints` becomes `0` and `minutesInPoint := timePeriod / 60 / amountOfPoints` divides by zero. Clamping only `amountOfPoints` moves the panic to the next line, because `minutesInPoint` is then `0` and Go panics on `0/0` too.
- **Fix**: clamp the *period* to one minute, which makes both divisors safe from one place, and extract the arithmetic into `models/sparklines.go` so the two copies cannot drift apart again. A degenerate period is reported as a single one-minute point — what the panicking code was trying to compute.

### Two panic paths, not one

The ticket names only `GetReport` → `reporter.go`. `GetMetrics` → `metrics.go` holds an independent copy of the same arithmetic and panics identically; both are fixed and both are verified live below.

### Defects the panic was masking

Once the request stops panicking it completes, and these become visible on the same input:

1. **`+Inf` and `NaN` in the response.** The per-second rates divide by the period length in Go *and*, for the report query, inside the SQL. A zero-length period made them `+Inf`; `Inf/Inf` then made `MainMetricPercent` `NaN`. The generated clients declare these as `float32`, so they cannot decode `"Infinity"` or `"NaN"` at all. The duration is now clamped to one second everywhere it is derived — Go and SQL — via `models.PeriodDuration`.
2. **`GetMetrics` never validated `from <= to`.** `GetReport` has that guard; `GetMetrics` did not, and with `totals: true` the early return is skipped, so a reversed range produced a point for a window that cannot match anything.
3. **Those guards returned unwrapped errors**, so a client mistake surfaced as HTTP 500. They now return `codes.InvalidArgument`.

## Review round

Three further commits address the review. Each finding was checked against the code before acting.

| Area | Change |
|---|---|
| Duplication | One `models.PeriodDuration` replaces four copies of the same clamp, applied to **all five** `PeriodDuration` assignments rather than only the one the ticket reached |
| Structure | `models/sparklines.go` holds the helper, its three constants and its test, which were spread over three files |
| Precondition | The helper aligns both bounds itself and **returns them**, instead of documenting an alignment precondition nothing enforced — and that its own test violated |
| Error codes | The request guards that still returned a bare error now return `InvalidArgument`, so each behaves like the one beside it |
| Tests | Two panic-only tests and a ~20-line fake driver removed; the wiring test covers the same ground with observable results |
| `PercentOfTotal` | Both totals lookups asked for `m_<col>sum`, missing the underscore every query alias carries; the first also read the per-query map instead of the totals map |
| `Rows.Err()` | Five methods ignored it, so a ClickHouse failure part-way through a result set was reported as success |

**A correction worth flagging:** the review suggested moving the minute-alignment prologue into the helper. Done literally that would have changed the query window — the aligned bounds are also used for the bind parameters and the gap-fill grid. The helper now returns them. Computing from raw bounds disagrees with the aligned answer in **178 of 915** sampled cases, so a caller that skipped the prologue got a different series; that is now unreachable.

## Behaviour changes to be aware of

1. **`PercentOfTotal` was silently `0` for every metric `GetMetrics` returns** and now carries real values. This is user-visible in the QAN details panel and should not be mistaken for a regression. Verified against ClickHouse: `61.119 / 110.495 = 0.5531393`, which is what the API now returns. The three regenerated fixtures change **only** by gaining `percentOfTotal`; no line was removed.
2. **`GetMetrics` with a reversed range** previously returned an empty `200` (or, with `totals: true`, a fabricated point); it now returns `400`. `GetReport` and `getFilters` already rejected reversed ranges.
3. **`Rows.Err()` now surfaces errors that were previously swallowed.** A mid-stream ClickHouse failure used to read as success; in the two sparkline methods the gap-fill loop then invented an empty point for every row that never arrived, giving a flat-zero series indistinguishable from real data.

## Test plan

- [x] **Unit tests** — `qan-api2/models/sparklines_test.go` and `rows_error_test.go`, **39 subtests**, no ClickHouse required. Point arithmetic for 13 period shapes; an `MinInt64`/`MaxInt64` corner grid; alignment equivalence (aligning before the call must change nothing); both real `SelectSparklines` pinned to the helper; and a driver that opens successfully then fails on iteration, covering all five `Rows.Err()` sites.
- [x] **Red before, green after** — on `origin/main` the suite fails with `integer divide by zero`; the `Rows.Err()` test fails on this branch before these commits with all five returning `nil`.
- [x] **API tests** — new `api-tests/qan/`, **21 subtests**: **21 pass** on this branch, **13 fail** on `main`, **7 fail** on this branch before these commits.
- [x] **Live on PMM Server 3.9.1**, `qan-api2` swapped under supervisor, verified by binary SHA, exercised both in-container and through PMM's nginx:

  | Request | `main` | this branch |
  |---|---|---|
  | `getReport`, same minute | **500** + panic logged | **200**, 1 point, `time_frame 60` |
  | `getMetrics`, same minute | **500** + panic | **200** |
  | `getFilters`, all-zero main metric | 28 × `"NaN"` | **0** |
  | client mistakes (6 endpoints) | **500** | **400** `InvalidArgument` |
  | divide-by-zero panics logged | **3** | **0** |

- [x] **No regression for normal ranges** — 28 requests captured from both binaries against the repo's own 83,726-row fixture and compared semantically: **21 identical**, and every one of the 7 differences is an intended fix. Byte comparison is invalid here, as protojson randomises map order and tied rows reorder.
- [x] `go build`, `go vet`, `gofmt -l` — clean.
- [x] `LOG_LEVEL=error bin/golangci-lint run -c=.golangci.yml --new-from-rev=origin/main --new` — **0 issues**.
- [x] `go tool license-eye -c .licenserc.yaml header check` — 0 invalid.
- [x] Full `go test ./qan-api2/...` — **4/4 packages green**, including the four ClickHouse-backed tests that previously could not run.

## Notes for reviewers

- **Why the duration floor is 1 second, not 60**: 1 s is the smallest floor that leaves every real duration untouched. A 60 s floor would change existing behaviour for any 1–59 s window that works today — `08:44:50–08:45:10` is 20 s spanning a minute boundary, does not panic on `main`, and a 60 s floor would third its Load. One consequence is recorded on `PeriodDuration`: for a single-instant request the row rates divide by 1 s while the sparkline in the same response reports one whole minute. There is no free fix; the disagreement is left in place deliberately.
- **Two of the new API tests issue raw HTTP** rather than using the generated Swagger client, against the convention in `api-tests/AGENTS.md`. The reason is recorded in the package comment and is a bug in its own right — see the first follow-up below.

## Follow-ups, not in this PR

1. **PMM-15545** — `DropOldPartition` destroys all data for a large `--data-retention`. Once the value exceeds the age of the Unix epoch (20714 days today, moving forward daily), `toYYYYMMDD(now() - toIntervalDay(?))` underflows ClickHouse's unsigned `DateTime` and wraps to ~2106, so every partition compares as older and is dropped. A flag meant to keep more data deletes all of it. Hit while testing this branch.
2. **PMM-15546** — the generated Go client cannot decode any successful QAN report: `json: cannot unmarshal string into Go struct field ...stats.p99 of type float32`. Report responses carry ClickHouse `NaN` both as a deliberate "metric not collected" sparkline sentinel and from aggregating over zero rows. Either the sentinel or the generated field types must change. This is also why no API test covered these endpoints before now.

### DIFF
--- a/qan-api2/models/base.go
+++ b/qan-api2/models/base.go
@@ -25,6 +25,34 @@ import (
 
 const queryTimeout = 30 * time.Second
 
+// sparklinePoints works out how a sparkline period is divided into points, returning
+// how many points to return and how many seconds each one covers. Both bounds must already be
+// aligned to a whole minute, which makes their difference a whole number of minutes.
+//
+// A period of less than a minute -- both bounds inside the same minute, or a reversed
+// range -- leaves no minutes to spread across points, which would make every divisor
+// here zero. Such a period is reported as a single one-minute point instead.
+func sparklinePoints(periodStartFromSec, periodStartToSec int64) (int64, int64) {
+	timePeriod := max(periodStartToSec-periodStartFromSec, secondsPerMinute)
+
+	// If time range is bigger then two hour - amount of sparklines points = 120 to avoid huge data in response.
+	// Otherwise amount of sparklines points is equal to minutes in time range to not mess up calculation.
+	amountOfPoints := int64(optimalAmountOfPoint)
+	// reduce amount of point if period less then 2h.
+	if timePeriod < int64(minFullTimeFrame.Seconds()) {
+		// minimum point is 1 minute
+		amountOfPoints = timePeriod / secondsPerMinute
+	}
+
+	// how many full minutes we can fit into given amount of points.
+	minutesInPoint := timePeriod / secondsPerMinute / amountOfPoints
+	// we need aditional point to show this minutes
+	remainder := (timePeriod / secondsPerMinute) % amountOfPoints
+	amountOfPoints += remainder / minutesInPoint
+
+	return amountOfPoints, minutesInPoint * secondsPerMinute
+}
+
 var sparklinePointAllFields = []string{
 	"point",
 	"timestamp",
--- a/qan-api2/models/reporter.go
+++ b/qan-api2/models/reporter.go
@@ -210,7 +210,7 @@ func (r *Reporter) Select(ctx context.Context, periodStartFromSec, periodStartTo
 	}{
 		PeriodStartFrom:     periodStartFromSec,
 		PeriodStartTo:       periodStartToSec,
-		PeriodDuration:      periodStartToSec - periodStartFromSec,
+		PeriodDuration:      max(periodStartToSec-periodStartFromSec, 1), // the query divides by it
 		Dimensions:          escapeColonsInMap(dimensions),
 		Labels:              escapeColonsInMap(labels),
 		Group:               group,
@@ -319,25 +319,10 @@ func (r *Reporter) SelectSparklines(ctx context.Context, dimensionVal string,
 	group string, column string, isTotal bool,
 ) ([]*qanpbv1.Point, error) {
 	// Align to minutes
-	periodStartToSec = periodStartToSec / 60 * 60     //nolint:mnd
-	periodStartFromSec = periodStartFromSec / 60 * 60 //nolint:mnd
+	periodStartToSec = periodStartToSec / secondsPerMinute * secondsPerMinute
+	periodStartFromSec = periodStartFromSec / secondsPerMinute * secondsPerMinute
 
-	// If time range is bigger then two hour - amount of sparklines points = 120 to avoid huge data in response.
-	// Otherwise amount of sparklines points is equal to minutes in time range to not mess up calculation.
-	amountOfPoints := int64(optimalAmountOfPoint)
-	timePeriod := periodStartToSec - periodStartFromSec
-	// reduce amount of point if period less then 2h.
-	if timePeriod < int64(minFullTimeFrame.Seconds()) {
-		// minimum point is 1 minute
-		amountOfPoints = timePeriod / 60 //nolint:mnd
-	}
-
-	// how many full minutes we can fit into given amount of points.
-	minutesInPoint := (periodStartToSec - periodStartFromSec) / 60 / amountOfPoints //nolint:mnd
-	// we need aditional point to show this minutes
-	remainder := ((periodStartToSec - periodStartFromSec) / 60) % amountOfPoints //nolint:mnd
-	amountOfPoints += remainder / minutesInPoint
-	timeFrame := minutesInPoint * 60 //nolint:mnd
+	amountOfPoints, timeFrame := sparklinePoints(periodStartFromSec, periodStartToSec)
 
 	arg := map[string]any{
 		"dimension_val":     dimensionVal,
@@ -579,7 +564,10 @@ func (r *Reporter) SelectFilters(
 func (r *Reporter) queryFilters(ctx context.Context, periodStartFromSec,
 	periodStartToSec int64, dimensionName, mainMetricName string, tmplQueryFilter *template.Template, queryDimensions, queryLabels map[string][]string,
 ) ([]*customLabel, float32, error) {
-	durationSec := periodStartToSec - periodStartFromSec
+	// A single-instant request would divide the per-second rates below by zero, giving
+	// +Inf -- and then Inf/Inf = NaN where the percentage is worked out. See the note
+	// in services/analytics/profile.go.
+	durationSec := max(periodStartToSec-periodStartFromSec, 1)
 	var labels []*customLabel
 
 	lbacFilter, err := headersToLbacFilter(ctx)
--- a/qan-api2/models/sparklines_test.go
+++ b/qan-api2/models/sparklines_test.go
@@ -54,7 +54,7 @@ func unreachableDB(t *testing.T) *sqlx.DB {
 // sparklineCtx carries incoming gRPC metadata, which headersToLbacFilter requires.
 func sparklineCtx(t *testing.T) context.Context {
 	t.Helper()
-	return metadata.NewIncomingContext(context.Background(), metadata.Pairs("test", t.Name()))
+	return metadata.NewIncomingContext(t.Context(), metadata.Pairs("test", t.Name()))
 }
 
 // sparklineRanges covers every shape of period the point arithmetic has to survive,
--- a/qan-api2/models/sparklines_test.go
+++ b/qan-api2/models/sparklines_test.go
@@ -1,0 +1,271 @@
+// Copyright (C) 2023 Percona LLC
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package models
+
+import (
+	"context"
+	"database/sql"
+	"database/sql/driver"
+	"errors"
+	"io"
+	"math"
+	"testing"
+
+	"github.com/jmoiron/sqlx"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/metadata"
+)
+
+// errNoConnection is returned by unreachableDB instead of ever opening a connection.
+var errNoConnection = errors.New("no database in this test")
+
+// unreachableConnector hands back a *sqlx.DB whose queries always fail. The sparkline
+// point arithmetic runs before any query is issued, so this lets us drive
+// SelectSparklines all the way through that arithmetic without a live ClickHouse:
+// a panic in it surfaces as a panic, while correct arithmetic surfaces as a query error.
+type unreachableConnector struct{}
+
+func (unreachableConnector) Connect(context.Context) (driver.Conn, error) {
+	return nil, errNoConnection
+}
+
+func (unreachableConnector) Driver() driver.Driver { return nil }
+
+func unreachableDB(t *testing.T) *sqlx.DB {
+	t.Helper()
+	db := sqlx.NewDb(sql.OpenDB(unreachableConnector{}), "clickhouse")
+	t.Cleanup(func() { _ = db.Close() })
+	return db
+}
+
+// sparklineCtx carries incoming gRPC metadata, which headersToLbacFilter requires.
+func sparklineCtx(t *testing.T) context.Context {
+	t.Helper()
+	return metadata.NewIncomingContext(context.Background(), metadata.Pairs("test", t.Name()))
+}
+
+// sparklineRanges covers every shape of period the point arithmetic has to survive,
+// including the zero-width range from PMM-15160 that made amountOfPoints 0.
+func sparklineRanges() []struct {
+	name     string
+	from, to int64
+} {
+	const (
+		base   = int64(1750322640) // 2026-06-19T08:44:00Z
+		minute = int64(60)
+		hour   = 60 * minute
+	)
+	return []struct {
+		name     string
+		from, to int64
+	}{
+		// PMM-15160: both endpoints align down to the same minute, so timePeriod == 0.
+		{"same minute", base + 10, base + 50},
+		{"identical timestamps", base, base},
+		{"one minute", base, base + minute},
+		{"two minutes", base, base + 2*minute},
+		// Just inside and just outside the < 2h branch that reduces the point count.
+		{"just under two hours", base, base + 2*hour - minute},
+		{"exactly two hours", base, base + 2*hour},
+		{"twelve hours", base, base + 12*hour},
+		{"thirty days", base, base + 30*24*hour},
+		// Reversed range: must not panic either, however the period is treated.
+		{"reversed", base + hour, base},
+		{"reversed within a minute", base + 50, base + 10},
+	}
+}
+
+func TestReporterSelectSparklinesNeverPanics(t *testing.T) {
+	t.Parallel()
+
+	r := NewReporter(unreachableDB(t))
+	for _, tc := range sparklineRanges() {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			require.NotPanics(t, func() {
+				_, _ = r.SelectSparklines(sparklineCtx(t), "queryid", tc.from, tc.to,
+					nil, nil, "queryid", "load", false)
+			})
+		})
+	}
+}
+
+func TestMetricsSelectSparklinesNeverPanics(t *testing.T) {
+	t.Parallel()
+
+	m := NewMetrics(unreachableDB(t))
+	for _, tc := range sparklineRanges() {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			require.NotPanics(t, func() {
+				_, _ = m.SelectSparklines(sparklineCtx(t), tc.from, tc.to,
+					"", "queryid", nil, nil)
+			})
+		})
+	}
+}
+
+// TestSparklinePoints pins the point arithmetic itself. The panic tests above exercise
+// the real methods but cannot see their return values, because the query fails first.
+func TestSparklinePoints(t *testing.T) {
+	t.Parallel()
+
+	const (
+		base   = int64(1750322640) // 2026-06-19T08:44:00Z, already minute-aligned
+		minute = int64(60)
+		hour   = 60 * minute
+	)
+	for _, tc := range []struct {
+		name          string
+		from, to      int64
+		wantPoints    int64
+		wantTimeFrame int64
+	}{
+		// PMM-15160: a period of less than a minute must still yield one usable point.
+		{"identical bounds", base, base, 1, 60},
+		{"reversed by a minute", base + minute, base, 1, 60},
+		{"reversed by an hour", base + hour, base, 1, 60},
+		// Below two hours, one point per minute.
+		{"one minute", base, base + minute, 1, 60},
+		{"two minutes", base, base + 2*minute, 2, 60},
+		{"one hour", base, base + hour, 60, 60},
+		{"one minute under two hours", base, base + 2*hour - minute, 119, 60},
+		// At and above two hours, capped at the optimal point count.
+		{"exactly two hours", base, base + 2*hour, 120, 60},
+		{"two hours and a minute", base, base + 2*hour + minute, 121, 60},
+		// minutesInPoint > 1 with a non-zero remainder, so the remainder really is divided.
+		{"five hours", base, base + 5*hour, 150, 120},
+		{"seven hours", base, base + 7*hour, 140, 180},
+		{"twelve hours", base, base + 12*hour, 120, 360},
+		{"thirty days", base, base + 30*24*hour, 120, 21600},
+	} {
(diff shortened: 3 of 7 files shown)


qan-api2/models/sparklines_test.go
```diff
@@ -0,0 +1,271 @@
...
+	t.Helper()
+	db := sqlx.NewDb(sql.OpenDB(unreachableConnector{}), "clickhouse")
+	t.Cleanup(func() { _ = db.Close() })
+	return db
+}
+
+// sparklineCtx carries incoming gRPC metadata, which headersToLbacFilter requires.
+func sparklineCtx(t *testing.T) context.Context {
+	t.Helper()
+	return metadata.NewIncomingContext(context.Background(), metadata.Pairs("test", t.Name()))
```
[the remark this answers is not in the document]
theTibi: Done in eee8bc337 — `sparklineCtx` now builds on `t.Context()`, so the context is cancelled at test cleanup.

qan-api2/services/analytics/filters.go
```diff
@@ -34,7 +37,7 @@ func (s *Service) GetFilteredMetricsNames(ctx context.Context, in *qanpb.GetFilt
 	periodStartFromSec := in.PeriodStartFrom.Seconds
 	periodStartToSec := in.PeriodStartTo.Seconds
 	if periodStartFromSec > periodStartToSec {
-		return nil, fmt.Errorf("from-date %v cannot be later then to-date %v", in.PeriodStartFrom, in.PeriodStartTo)
+		return nil, status.Errorf(codes.InvalidArgument, "from-date %v cannot be later then to-date %v", in.PeriodStartFrom, in.PeriodStartTo)
```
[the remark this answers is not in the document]
theTibi: Not taking this one — the suggested form is exactly what `status.Errorf` already does. From `grpc-go`'s `status/status.go`:

```go
// Errorf returns Error(c, fmt.Sprintf(format, a...)).
func Errorf(c codes.Code, format string, a ...any) error {
	return Error(c, fmt.Sprintf(format, a...))
}
```

Same `*status.Error`, same code, same message — it would only add an `fmt` import and a nested call. The repo also prefers `Errorf`: 510 `status.Errorf` vs 246 `status.Error(` call sites (non-vendor).

The `AGENTS.md` rule being cited (`status.Error()` with gRPC codes) contrasts gRPC status codes with bare `fmt.Errorf`/ad-hoc HTTP errors — which is precisely the change these three lines make. `status.Errorf` satisfies it.

JiriCtvrtka (approved):
Only medium thing which worth a follow-up ticket:
api2/models/reporter.go#L560: NaN from 0/0 in MainMetricPercent. Real, and it's the same class of client-decode failure the PR targets, but it's a pre-existing bug on an unchanged line, reachable only with main_metric_name=num_queries_with_errors on an error-free workload.

ademidoff (commented):
Review of the divide-by-zero fix: 13 comments inline, roughly most to least severe.

qan-api2/models/reporter.go
```diff
@@ -584,7 +569,10 @@ func (r *Reporter) SelectFilters(
 func (r *Reporter) queryFilters(ctx context.Context, periodStartFromSec,
 	periodStartToSec int64, dimensionName, mainMetricName string, tmplQueryFilter *template.Template, queryDimensions, queryLabels map[string][]string,
 ) ([]*customLabel, float32, error) {
-	durationSec := periodStartToSec - periodStartFromSec
+	// A single-instant request would divide the per-second rates below by zero, giving
+	// +Inf -- and then Inf/Inf = NaN where the percentage is worked out. See the note
+	// in services/analytics/profile.go.
+	durationSec := max(periodStartToSec-periodStartFromSec, 1)
```
ademidoff: This clamp removes the `Inf/Inf` form of the NaN, but the `0/0` form survives it: when the main metric is zero for every row in the window (`num_queries_with_errors`, say), every `label.mainMetricPerSec` is 0, so `mainMetricPerSec` is 0, the `totals[label.key]` fallback on line 555 is 0 too, and `MainMetricPercent: label.mainMetricPerSec / total` on line 560 is 0/0 = NaN — which the generated client's `float32` (`api/qan/v1/json/client/qan_service/get_filtered_metrics_names_responses.go:788`) still cannot decode, failing the whole getFilters response, i.e. exactly the failure mode this PR is fixing. Guarding `total != 0` the way `makeStats` guards `divider != 0` would close it.

qan-api2/models/reporter.go
```diff
@@ -324,25 +324,10 @@ func (r *Reporter) SelectSparklines(ctx context.Context, dimensionVal string,
...
-	// how many full minutes we can fit into given amount of points.
-	minutesInPoint := (periodStartToSec - periodStartFromSec) / 60 / amountOfPoints //nolint:mnd
-	// we need aditional point to show this minutes
-	remainder := ((periodStartToSec - periodStartFromSec) / 60) % amountOfPoints //nolint:mnd
-	amountOfPoints += remainder / minutesInPoint
-	timeFrame := minutesInPoint * 60 //nolint:mnd
+	periodStartToSec = periodStartToSec / secondsPerMinute * secondsPerMinute
+	periodStartFromSec = periodStartFromSec / secondsPerMinute * secondsPerMinute
+
+	amountOfPoints, timeFrame := sparklinePoints(periodStartFromSec, periodStartToSec)
```
ademidoff: `SelectSparklines` never checks `rows.Err()` after the scan loop — line 448 returns `results, err` with `err` still the nil from `QueryxContext` — so a ClickHouse error mid-stream is reported as success and the gap-fill loop on line 435 fabricates empty points for every row that never arrived, giving a flat-zero sparkline with nothing logged. Same omission in `Reporter.Select` (line 278, whose result feeds `qan-api2/services/analytics/profile.go:145` `results[0]`) and in `Metrics.SelectSparklines` (`qan-api2/models/metrics.go:577`).

qan-api2/services/analytics/profile.go
```diff
@@ -34,9 +37,15 @@ func (s *Service) GetReport(ctx context.Context, in *qanpb.GetReportRequest) (*q
...
+		return nil, status.Errorf(codes.InvalidArgument, "from-date %v cannot be later then to-date %v", in.PeriodStartFrom, in.PeriodStartTo)
 	}
-	periodDurationSec := periodStartToSec - periodStartFromSec
+	// A report may be requested for a single instant, where every per-second rate below
+	// -- and the SQL that divides by this duration -- would otherwise divide by zero and
+	// yield +Inf, which the generated API clients cannot decode into their float fields.
+	// One second is the smallest floor that leaves every real duration untouched; note it
+	// makes these rates disagree with the sparkline, which reports such a period as one
+	// whole minute.
+	periodDurationSec := max(periodStartToSec-periodStartFromSec, 1)
```
ademidoff: This floors the period at 1 s while `sparklinePoints` floors the same period at 60 s, so a single response now carries two different rates for the same data: for 08:44:10–08:44:50 the row's `Load` and the SQL `load` column divide `m_query_time_sum` by 40 while the one sparkline point divides it by 60, and the Load column reads 1.5x the sparkline sitting next to it. The comment acknowledges the disagreement, and since this request used to panic the disagreement is newly user-visible — deriving both from one clamp (rounding the report period up to the same minute grid) would avoid shipping it.

qan-api2/models/reporter.go
```diff
@@ -215,7 +215,7 @@ func (r *Reporter) Select(ctx context.Context, periodStartFromSec, periodStartTo
 	}{
 		PeriodStartFrom:     periodStartFromSec,
 		PeriodStartTo:       periodStartToSec,
-		PeriodDuration:      periodStartToSec - periodStartFromSec,
+		PeriodDuration:      max(periodStartToSec-periodStartFromSec, 1), // the query divides by it
```
ademidoff: Only this `PeriodDuration` got clamped: `qan-api2/models/reporter.go:363`, `qan-api2/models/metrics.go:79`, `qan-api2/models/metrics.go:527` and `qan-api2/models/metrics.go:1223` still assign the raw `periodStartToSec - periodStartFromSec`. No SQL template divides by those today so the inconsistency is invisible, but the moment one does — as `queryReportTmpl` lines 120 and 122 already do — a `from == to` request is back to `SUM(...) / 0` → `+Inf` and an undecodable float on the wire. A single `periodDuration(from, to)` helper used by all five sites is the same argument this PR makes for extracting `sparklinePoints`.

qan-api2/services/analytics/object_details.go
```diff
@@ -36,6 +38,9 @@ func (s *Service) GetMetrics(ctx context.Context, in *qanpb.GetMetricsRequest) (
 		return nil, fmt.Errorf("period_start_to is required: %v", in.PeriodStartTo)
```
ademidoff: The `from > to` guards became `InvalidArgument`, but the nil-timestamp guards right beside them (this line and line 34) still return a bare `fmt.Errorf`, so `GetMetrics` with `period_start_from` omitted is HTTP 500 while the same call with `from > to` is now 400. Same split at `qan-api2/services/analytics/profile.go:34`, `:51` and `:122`, and `qan-api2/services/analytics/filters.go:34` — the PR's own reason for touching profile.go and filters.go ("worse to leave the same guard behaving three different ways") applies to these as well.

qan-api2/services/analytics/object_details.go
```diff
@@ -103,7 +108,8 @@ func (s *Service) GetMetrics(ctx context.Context, in *qanpb.GetMetricsRequest) (
 	// Get totals for given filter
 	totals := totalsList[totalLen-1]
 
-	durationSec := periodStartToSec - periodStartFromSec
+	// See the note in profile.go: a single-instant request must not divide by zero.
+	durationSec := max(periodStartToSec-periodStartFromSec, 1)
```
ademidoff: The `durationSec` this line feeds into `makeMetrics` meets a malformed key at the other end: `qan-api2/services/analytics/object_details.go:192` and `:215` read `mm["m_"+k+"sum"]`, missing the underscore that every `queryMetricsTmpl` alias carries (`m_query_time_sum`), and line 192 additionally reads `mm` where line 215 correctly reads the totals map `t`. `interfaceToFloat32(nil)` returns 0, so `if sum > 0 && totalSum > 0` never fires and `PercentOfTotal` is silently 0 for every metric `GetMetrics` returns. Pre-existing, but this PR rewrites the arguments the function is called with.

qan-api2/services/analytics/object_details.go
```diff
@@ -36,6 +38,9 @@ func (s *Service) GetMetrics(ctx context.Context, in *qanpb.GetMetricsRequest) (
 		return nil, fmt.Errorf("period_start_to is required: %v", in.PeriodStartTo)
 	}
 	periodStartToSec := in.PeriodStartTo.Seconds
+	if periodStartFromSec > periodStartToSec {
```
ademidoff, line 41: This changes `GetMetrics`' public contract: a reversed range used to return 200 with an empty body (or a fabricated point under `totals:true`) and now returns 400. qan-api2 serves already-deployed UIs, so a saved Grafana panel with a mis-set range, or the details panel during a range drag, will render an error where it used to render an empty chart — worth calling out in the PR description and pinning with an api-test, since nothing in `api-tests/` currently exercises any QAN report endpoint.

qan-api2/models/reporter.go
ademidoff: Inline trailing comment: AGENTS.md, Global Development Conventions → Code Style says "Don't inline comments (`code // comment`); put comments on separate lines — `//nolint` is the only exception". Move it to its own line above the field.

qan-api2/services/analytics/profile.go
```diff
@@ -34,9 +37,15 @@ func (s *Service) GetReport(ctx context.Context, in *qanpb.GetReportRequest) (*q
 	periodStartFromSec := in.PeriodStartFrom.Seconds
 	periodStartToSec := in.PeriodStartTo.Seconds
 	if periodStartFromSec > periodStartToSec {
-		return nil, fmt.Errorf("from-date %v cannot be later then to-date %v", in.PeriodStartFrom, in.PeriodStartTo)
+		return nil, status.Errorf(codes.InvalidArgument, "from-date %v cannot be later then to-date %v", in.PeriodStartFrom, in.PeriodStartTo)
 	}
-	periodDurationSec := periodStartToSec - periodStartFromSec
+	// A report may be requested for a single instant, where every per-second rate below
```
ademidoff: This note and the clamp under it are copy-pasted at four sites — here, `qan-api2/models/reporter.go:218`, `qan-api2/models/reporter.go:575` and `qan-api2/services/analytics/object_details.go:112` — stitched together by comments pointing at each other ("See the note in profile.go"). The first two compute the same duration from the same inputs for the same response (the Go-side `Load` and the SQL-side `load` column), so they can drift into two different loads; a `models.PeriodDuration(from, to)` helper mirroring `sparklinePoints` would collapse all four and cover the unclamped siblings at the same time.

qan-api2/services/analytics/object_details.go
ademidoff: With `durationSec` floored at 1, the `durationSec > 0` conjunct on lines 206 and 226 is always true. The PR notes both under "Dead after this change" — dropping them is a two-line change inside the diff's own blast radius, and leaving them in implies a zero-duration path that can no longer occur.

qan-api2/models/base.go
```diff
@@ -25,6 +25,34 @@ import (
 
 const queryTimeout = 30 * time.Second
 
+// sparklinePoints works out how a sparkline period is divided into points, returning
+// how many points to return and how many seconds each one covers. Both bounds must already be
```
ademidoff: This precondition is unenforced, and the PR's own test violates it: `TestSparklinePointsAlwaysUsable` (`qan-api2/models/sparklines_test.go:167`) feeds unaligned and overflowing bounds and asserts behaviour for them. A third caller that skips the `x / secondsPerMinute * secondsPerMinute` prologue and passes `from=T, to=T+119` gets `(1, 60)` back and the gap-fill loop silently drops the second minute's rows. Doing the alignment inside the helper would remove the duplicated prologue from `qan-api2/models/reporter.go:327-328` and `qan-api2/models/metrics.go:505-506` and make the precondition unfalsifiable.

qan-api2/models/base.go
```diff
@@ -25,6 +25,34 @@ import (
...
+// aligned to a whole minute, which makes their difference a whole number of minutes.
+//
+// A period of less than a minute -- both bounds inside the same minute, or a reversed
+// range -- leaves no minutes to spread across points, which would make every divisor
+// here zero. Such a period is reported as a single one-minute point instead.
+func sparklinePoints(periodStartFromSec, periodStartToSec int64) (int64, int64) {
```
ademidoff: The helper lands in `base.go` — the package catch-all holding `queryTimeout` and the sparkline column lists — but depends on `optimalAmountOfPoint`, `minFullTimeFrame` and `secondsPerMinute` declared in `qan-api2/models/metrics.go:37-44`, and is tested from a new `qan-api2/models/sparklines_test.go` that mirrors no source file, so reading one piece of arithmetic means opening three files. A `qan-api2/models/sparklines.go` holding the helper and those three constants would keep the unit, its constants and its test together.

qan-api2/models/sparklines_test.go
```diff
@@ -0,0 +1,271 @@
...
+		{"exactly two hours", base, base + 2*hour},
+		{"twelve hours", base, base + 12*hour},
+		{"thirty days", base, base + 30*24*hour},
+		// Reversed range: must not panic either, however the period is treated.
+		{"reversed", base + hour, base},
+		{"reversed within a minute", base + 50, base + 10},
+	}
+}
+
+func TestReporterSelectSparklinesNeverPanics(t *testing.T) {
```
ademidoff: This test and `TestMetricsSelectSparklinesNeverPanics` drive the real methods against `unreachableConnector`, which fails at the query — so all they can observe is that the arithmetic ran without panicking, which `TestSparklinePoints` asserts directly and `TestSparklineWiring` asserts through the real methods with observable results. Adding the two reversed ranges to `TestSparklineWiring`'s case list would cover the same ground and let both tests and the ~20-line fake driver go away.

theTibi:
@ademidoff thanks — all 13 verified against the code, none were wrong. Addressed in 3 commits.

**Fixed:** #4/#9 (one `PeriodDuration` at all five sites), #11 (`models/sparklines.go` holds the helper, its constants and its test), #12, #13, #5, #8, #10, and #1.

**Two of yours were understated:**
- #2 — five methods ignore `Rows.Err()`, not three: `Metrics.Get` and `Metrics.SelectHistogram` too.
- #6 — fixed here rather than deferred, since this branch already edits `makeMetrics` eleven lines away and that key is what makes the `totalSum > 0` guard dead. Verified against ClickHouse: `61.119 / 110.495 = 0.5531393`, which is what the API now returns. The three regenerated fixtures only gain `percentOfTotal`; no line removed.

**One correction on #12:** moving the prologue into the helper as written would have changed the query window — the aligned bounds are used downstream for the bind parameters and the gap-fill grid. The helper now returns them. Computing from unaligned bounds differs from the aligned answer in 178 of 915 sampled cases, including your `T+119` example (1 point instead of 2).

**#3 — pushing back.** The remedy is `max(x, 60)`, which changes sub-minute windows that work today: `08:44:50–08:45:10` is 20 s spanning a minute boundary, no panic on main, and a 60 s floor thirds its Load. There is no free fix; I left the 1 s floor and documented the disagreement on `PeriodDuration`.

**#7 — api-tests added** (`api-tests/qan/`), pinning the degenerate period, the guards and the 200→400 change: 21 subtests pass on this branch, 13 fail on main, 7 fail on this branch before these commits.

Two follow-ups, filed rather than included here:
- **PMM-15545** — `DropOldPartition` computes its cutoff as `toYYYYMMDD(now() - toIntervalDay(N))`. Once `N` exceeds the age of the Unix epoch (20714 days today, moving forward daily) the subtraction underflows ClickHouse's unsigned `DateTime` and wraps to ~2106, so **every** partition is dropped. `--data-retention=36500` empties the table. Hit while testing this branch.
- **PMM-15546** — the generated Go client cannot decode any successful QAN report: `json: cannot unmarshal string into Go struct field ...stats.p99 of type float32`. Sparkline NaN is a deliberate "not collected" sentinel, so either it or the field types must change. That is why two of the new tests use raw HTTP, recorded in the package comment.

*(Correction to my earlier wording: I first wrote that the retention threshold was ~2800 days. That was wrong — 2900 yields a correct cutoff. The real threshold is the epoch age, ~20714 days.)*

Verified on PMM Server 3.9.1 with `qan-api2` swapped under supervisor: main logs the divide-by-zero panic and returns 500, this branch returns 200 with one 60 s point; `getFilters` goes from 28 `"NaN"` to none; six endpoints go 500 → 400. Full suite, lint, license and vet clean.

PR #5839, opened 2026-08-26
PMM-15362 Disable default-on node_exporter collectors

theTibi:
## Summary

- **Bug**: `pmm-admin inventory change agent node-exporter <id> --disable-collectors=<name>` reports success and stores the value, but for any collector node_exporter enables **by default** (`cpu`, `diskstats`, `meminfo`, `filesystem`, `netdev`, `stat`, …) the metrics keep being exported. Disabling a *non-default* collector (`processes`, `buddyinfo`, `meminfo_numa`) already worked.
- **Root cause**: `nodeExporterConfig` only called `collectors.FilterOutCollectors("--collector.", …)`, which strips the redundant enable flag PMM adds. For a collector node_exporter turns on by itself that is a no-op, so it kept running. The sibling exporters already solve this — `postgresql.go` and `mongodb.go` both follow `FilterOutCollectors` with `collectors.DisableDefaultEnabledCollectors`; `node.go` was simply missing that second call.
- **Fix**: emit `--no-collector.<name>` for the default-on collectors PMM enables, via the same helper.

### Why the list is a whitelist, and deliberately not "every default-on collector"

An **unknown** `--no-collector.<name>` makes node_exporter exit at startup, which would take *all* metrics for that host — far worse than the bug. So `defaultNodeExporterCollectors` contains only collectors `nodeExporterConfig` already passes `--collector.<name>` for. That is self-proving: if the positive flag is already accepted by every agent we drive today, the negation is too. No version gate is needed, unlike `postgresql.go`, whose collectors are *not* passed positively.

`os` was considered and **excluded**: it is default-on and does appear in PMM's LR `collect[]` list, but the collector only exists from node_exporter 1.3.0 (`CHANGELOG.md`, 2021-10-20) while this code still builds configs for agentVersion 2.15.1 — an unprovable risk with total-metric-loss failure mode, for ~2 series VictoriaMetrics already stops ingesting.

### Two related fixes in the same blast radius

- `DisableDefaultEnabledCollectors` now collapses repeated names. Without it, `--disable-collectors=cpu,cpu` produces two `--no-collector.cpu` args and node_exporter refuses to start (`flag 'collector.cpu' cannot be repeated`) — **verified live**, see below. This also protects the mongodb and postgres call sites.
- `inventory change agent node-exporter` now runs `--disable-collectors` through `commands.ParseDisableCollectors`, as the `add` path already does. Previously `--disable-collectors="cpu, meminfo"` stored `" meminfo"` untrimmed, matching nothing — the ticket's "accepted but does nothing" symptom via the ticket's own command. The reported change now shows the parsed value instead of the raw one.

## Ticket

- PMM-15362

## Test plan

Unit:

- [x] `go test ./managed/services/agents/ -run TestNodeExporterConfig` — `LinuxDisabledCollectors` fails on the base commit with exactly the 5 missing `--no-collector.*` flags, passes here. New subtests: `LinuxDisabledCollectorsNotDefaultEnabled`, `LinuxAllDefaultCollectorsDisabled` (also pins that every entry has a matching `--collector.<name>`, and that none collides with the static `--no-collector.*` block), `MacOSDisabledCollectors`.
- [x] `go test ./managed/utils/collectors/...` — new duplicate-input case.
- [x] `go test ./admin/commands/...` — new `DisableCollectorsAreTrimmed` case; fails against the pre-fix file.
- [x] Each new test mutation-verified (revert the code → the test fails).
- [x] `make init` + `bin/golangci-lint run --new-from-rev=origin/main` → 0 issues; `gofumpt`/`goimports`/`gci` clean; `go-sumtype` clean; `check-license` 0 invalid; `go build ./...` clean.

Live, on an isolated PMM 3.10.0 server whose stock `pmm-managed` is built from this PR's base commit, with a registered pmm-agent (node_exporter 1.8.2) — only the `pmm-managed`/`pmm-admin` binaries swapped:

- [x] **Before** — `node_disk_` = 324; run the ticket's command; API stores `["diskstats"]`; `--collector.diskstats` gone, no `--no-collector.diskstats`; `node_disk_` = **324**. Bug reproduced end to end.
- [x] **After** — `--no-collector.diskstats` present on the real exporter cmdline; `node_disk_` = **0**; `node_cpu_seconds_total` = 88 unchanged.
- [x] **Duplicate input** `--disable-collectors=diskstats,diskstats,cpu,diskstats` — exactly one flag each, exporter `Running`, both families 0.
- [x] **Counterfactual** — same input against a build with this fix but *without* the dedupe: exporter status `Done`, agent log repeating `node_exporter: error: flag 'collector.diskstats' cannot be repeated`, listen port walking 42000→42025. The dedupe is load-bearing.
- [x] **Trim** — `--disable-collectors="cpu, meminfo"` stores `['cpu','meminfo']`, both `--no-collector.*` flags present, both families 0.
- [x] **Reversibility** — replacing the disabled list brought `node_disk_` back to 324.
- [x] **Non-default control** — `--disable-collectors=processes`: enable flag removed, **no** `--no-collector.processes` added, `node_processes_` = 0, `node_disk_` still 324.
- [x] Flag acceptance on node_exporter **1.4.0** (`percona/pmm-client:2`) and **1.8.2** (`percona/pmm-client:3`); a bogus name is correctly rejected on both, confirming the check discriminates.

Not covered: darwin (no macOS node — unit + mutation test only); VictoriaMetrics/Grafana ingestion; a live matrix of older agents (1.4.0 exercised as a standalone binary only).

- [ ] **pmm-submodules (FB) PR:** https://github.com/Percona-Lab/pmm-submodules/pull/4547

## Known residuals, deliberately out of scope

1. Roughly 17 default-on collectors PMM neither enables nor disables (`pressure`, `schedstat`, `softnet`, `os`, `dmi`, `udp_queues`, `nvme`, `btrfs`, `cpufreq`, `selinux`, `thermal_zone`, `watchdog`, `rapl`, `fibrechannel`, `powersupplyclass`, `tapestats`, `textfile`) remain silent no-ops for `--disable-collectors`. Only `os` is scraped by PMM; the rest never reach VictoriaMetrics, so the cost is host-side CPU. Handling them safely needs per-collector version gating.
2. Between the exporter restart and the VictoriaMetrics config regenerating, a stale `collect[]=<name>` makes node_exporter answer **HTTP 400** for the whole scrape job. **Now mitigated** (see the review follow-up below): `ChangeNodeExporter` regenerates the scrape config before requesting the state update. Measured on a live PMM 3.9.1 over 20 disable transitions, counting scrapes VM lost to `disabled collector: cpu`: **13 before, 3 after**. It cannot reach zero from the PMM side, because VictoriaMetrics acknowledges `/-/reload` immediately and applies the new scrape config asynchronously, so a short race against the exporter restart remains. Removing it entirely needs node_exporter to stop failing the whole scrape when `collect[]` names a disabled collector — filed as PMM-15547.
3. ~~Five sibling `change agent` commands still pass `--disable-collectors` untrimmed.~~ **Fixed** in the review follow-up: the value is normalised in `models.CreateAgent`/`ChangeAgent` instead, which covers all six commands, the UI and a direct API call.

## Review follow-up (Alex's review)

Four further commits, each verified by mutation (revert the fix, the matching test fails):

- **`9aa3a4e` Collapse the duplicated darwin guard, simplify the dedupe.** `nodeExporterConfig` tested `node.Distro != "darwin"` twice. `FilterOutCollectors` keys on `--collector.<name>` and can never match `--no-collector.<name>`, so the negations can be appended inside the existing guard. Output is byte-identical — confirmed on a live exporter (57 args captured on each build, empty `diff`).
- **`8ccb450` Cover the whitelist converse, duplicate input and the darwin gap.** `LinuxDefaultOnCollectorsAreListed` pins the direction that was missing: every default-on collector `nodeExporterConfig` enables must be in `defaultNodeExporterCollectors`, so adding one later cannot silently re-introduce this bug. Duplicate input is now exercised through `nodeExporterConfig`, where a collision with the static `--no-collector.*` block is also possible. `MacOSDisabledCollectors` no longer asserts that `--no-collector.*` is *never* produced on darwin, which pinned the darwin gap as intended behaviour; it now asserts only that the Linux-only collectors are never negated there, with a TODO.
- **`f9d31db` Normalize disabled collectors in the models layer.** Trimming only in `pmm-admin` left the reported problem reachable from the UI and from a direct `PUT /v1/inventory/agents/<id>`. Normalising in `models.CreateAgent`/`ChangeAgent` covers all six exporter types, the UI and the raw API in one place. The dedupe in `DisableDefaultEnabledCollectors` stays: normalising on write does nothing for rows already stored with a repeated name.
- **`de3ff34` Update the scrape config before restarting node_exporter.** Residual 2 above.

Live verification, on an isolated PMM 3.9.1 with node_exporter 1.8.2, swapping only the `pmm-managed`/`pmm-admin` binaries between this PR's previous head (`f31e609`) and the follow-up:

- [x] **Untrimmed name through the raw API**, the path `pmm-admin`-side trimming never covered. Same `PUT ["cpu", " meminfo"]`: before, stored `["cpu", " meminfo"]` with `--collector.meminfo` still on the cmdline and no negation — a silent no-op; after, stored `["cpu", "meminfo"]`, `--no-collector.meminfo` present, enable flag gone.
- [x] **Duplicates** `["diskstats","diskstats","cpu"]` store deduplicated; the exporter stays `RUNNING` with exactly one flag each on both builds (the existing helper dedupe already prevented the startup failure, so this commit only cleans what is stored).
- [x] **Non-default control** `["processes"]` still emits no `--no-collector.processes`, and restores cpu/diskstats/meminfo.
- [x] **Exporter args byte-identical** across the two builds for the same disabled set.
- [x] **`ForceConfigurationUpdate` failure path**: with VictoriaMetrics stopped, the API still returns 200, the change lands, the exporter keeps running, and exactly one warning is logged against the `ChangeAgent` request.
- [x] `make check-license` 0 invalid; `bin/golangci-lint run --new-from-rev` 0 issues; `gofumpt`, `gci`, `go-sumtype` clean.

Still not covered: darwin (no macOS node — the gap is now a TODO rather than a pinned assertion), and node_exporter versions other than 1.8.2 live.

### Note for the release note

Because `collect[]` is an *exclusive* filter and the scrape config already drops a disabled collector, the metrics had **already** stopped reaching Grafana before this fix. The real gains are that the collector stops executing on the monitored host, a direct `/metrics` scrape no longer exposes it, and the empty-`collect[]` fallback no longer resurrects it.

### DIFF
--- a/admin/commands/inventory/change_agent_node_exporter.go
+++ b/admin/commands/inventory/change_agent_node_exporter.go
@@ -83,10 +83,12 @@ func (cmd *ChangeAgentNodeExporterCommand) RunCmd() (commands.Result, error) {
 	// Parse custom labels if provided
 	customLabels := commands.ParseKeyValuePair(cmd.CustomLabels)
 
+	disableCollectors := commands.ParseDisableCollectors(cmd.DisableCollectors)
+
 	body := &agents.ChangeAgentParamsBodyNodeExporter{
 		Enable:            cmd.Enable,
 		EnablePushMetrics: cmd.PushMetrics,
-		DisableCollectors: cmd.DisableCollectors,
+		DisableCollectors: disableCollectors,
 		ExposeExporter:    cmd.ExposeExporter,
 		LogLevel:          convertLogLevelPtr(cmd.LogLevel),
 	}
@@ -132,8 +134,8 @@ func (cmd *ChangeAgentNodeExporterCommand) RunCmd() (commands.Result, error) {
 			changes = append(changes, "disabled expose exporter")
 		}
 	}
-	if cmd.DisableCollectors != nil {
-		changes = append(changes, fmt.Sprintf("updated disabled collectors: %v", cmd.DisableCollectors))
+	if disableCollectors != nil {
+		changes = append(changes, fmt.Sprintf("updated disabled collectors: %v", disableCollectors))
 	}
 	if cmd.LogLevel != nil {
 		changes = append(changes, fmt.Sprintf("changed log level to %s", *cmd.LogLevel))
--- a/managed/services/agents/node.go
+++ b/managed/services/agents/node.go
@@ -32,6 +32,30 @@ var (
 	v2_28_00 = version.MustParse("2.28.0-0")
 )
 
+// defaultNodeExporterCollectors lists the collectors enabled below that node_exporter also enables
+// on its own. For those, dropping PMM's --collector.<name> flag leaves the collector running, so an
+// explicit --no-collector.<name> is needed as well. This is deliberately not every default-on
+// collector: only a collector nodeExporterConfig already passes --collector.<name> for may be added,
+// because that proves the flag exists in the node_exporter of every pmm-agent this code builds
+// configs for. An unknown flag stops the exporter from starting.
+var defaultNodeExporterCollectors = []string{
+	"bonding",
+	"cpu",
+	"diskstats",
+	"entropy",
+	"filefd",
+	"filesystem",
+	"hwmon",
+	"loadavg",
+	"meminfo",
+	"netdev",
+	"netstat",
+	"stat",
+	"time",
+	"uname",
+	"vmstat",
+}
+
 func nodeExporterConfig(node *models.Node, exporter *models.Agent, agentVersion *version.Parsed) (*agentv1.SetStateRequest_AgentProcess, error) {
 	listenAddress := getExporterListenAddress(node, exporter)
 	tdp := models.TemplateDelimsPair(exporter.ExporterOptions.MetricsPath)
@@ -123,6 +147,14 @@ func nodeExporterConfig(node *models.Node, exporter *models.Agent, agentVersion 
 	}
 
 	args = collectors.FilterOutCollectors("--collector.", args, exporter.ExporterOptions.DisabledCollectors)
+
+	if node.Distro != "darwin" {
+		args = append(args, collectors.DisableDefaultEnabledCollectors(
+			"--no-collector.",
+			defaultNodeExporterCollectors,
+			exporter.ExporterOptions.DisabledCollectors,
+		)...)
+	}
 
 	if exporter.ExporterOptions.MetricsPath != "" {
 		args = append(args, "--web.telemetry-path="+exporter.ExporterOptions.MetricsPath)
--- a/managed/services/agents/node_test.go
+++ b/managed/services/agents/node_test.go
@@ -16,8 +16,10 @@
 package agents
 
 import (
+	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	agentv1 "github.com/percona/pmm/api/agent/v1"
@@ -213,7 +215,7 @@ func TestNodeExporterConfig(t *testing.T) {
 			AgentID:   "agent-id",
 			AgentType: models.NodeExporterType,
 			ExporterOptions: models.ExporterOptions{
-				DisabledCollectors: []string{"cpu", "netstat", "netstat.fields", "vmstat", "meminfo"},
+				DisabledCollectors: []string{"cpu", "diskstats", "netstat", "netstat.fields", "vmstat", "meminfo"},
 			},
 		}
 		agentVersion := version.MustParse("2.15.1")
@@ -228,7 +230,6 @@ func TestNodeExporterConfig(t *testing.T) {
 			Args: []string{
 				"--collector.bonding",
 				"--collector.buddyinfo",
-				"--collector.diskstats",
 				"--collector.entropy",
 				"--collector.filefd",
 				"--collector.filesystem",
@@ -255,6 +256,8 @@ func TestNodeExporterConfig(t *testing.T) {
 				"--no-collector.arp",
 				"--no-collector.bcache",
 				"--no-collector.conntrack",
+				"--no-collector.cpu",
+				"--no-collector.diskstats",
 				"--no-collector.drbd",
 				"--no-collector.edac",
 				"--no-collector.infiniband",
@@ -263,8 +266,10 @@ func TestNodeExporterConfig(t *testing.T) {
 				"--no-collector.ksmd",
 				"--no-collector.logind",
 				"--no-collector.mdadm",
+				"--no-collector.meminfo",
 				"--no-collector.mountstats",
 				"--no-collector.netclass",
+				"--no-collector.netstat",
 				"--no-collector.nfs",
 				"--no-collector.nfsd",
 				"--no-collector.ntp",
@@ -275,6 +280,7 @@ func TestNodeExporterConfig(t *testing.T) {
 				"--no-collector.systemd",
 				"--no-collector.tcpstat",
 				"--no-collector.timex",
+				"--no-collector.vmstat",
 				"--no-collector.wifi",
 				"--no-collector.xfs",
 				"--no-collector.zfs",
@@ -289,6 +295,85 @@ func TestNodeExporterConfig(t *testing.T) {
 		require.Equal(t, expected.Args, actual.Args)
 		require.Equal(t, expected.Env, actual.Env)
 		require.Equal(t, expected, actual)
+	})
+
+	t.Run("LinuxDisabledCollectorsNotDefaultEnabled", func(t *testing.T) {
+		t.Parallel()
+		disabled := []string{"processes", "buddyinfo", "meminfo_numa", "no_such_collector"}
+		node := &models.Node{}
+		exporter := &models.Agent{
+			AgentID:   "agent-id",
+			AgentType: models.NodeExporterType,
+			ExporterOptions: models.ExporterOptions{
+				DisabledCollectors: disabled,
+			},
+		}
+		agentVersion := version.MustParse("2.15.1")
+
+		actual, err := nodeExporterConfig(node, exporter, agentVersion)
+		require.NoError(t, err, "Unable to build node exporter config")
+
+		requireNoDuplicateFlags(t, actual.Args)
+		for _, collector := range disabled {
+			assert.NotContains(t, actual.Args, "--collector."+collector)
+			assert.NotContains(t, actual.Args, "--no-collector."+collector)
+		}
+	})
+
+	t.Run("LinuxAllDefaultCollectorsDisabled", func(t *testing.T) {
+		t.Parallel()
+		node := &models.Node{}
+		exporter := &models.Agent{
+			AgentID:   "agent-id",
+			AgentType: models.NodeExporterType,
+			ExporterOptions: models.ExporterOptions{
+				DisabledCollectors: defaultNodeExporterCollectors,
+			},
+		}
+		agentVersion := version.MustParse("2.15.1")
+
+		actual, err := nodeExporterConfig(node, exporter, agentVersion)
+		require.NoError(t, err, "Unable to build node exporter config")
+
+		requireNoDuplicateFlags(t, actual.Args)
+		for _, collector := range defaultNodeExporterCollectors {
+			assert.Contains(t, actual.Args, "--no-collector."+collector)
+			assert.NotContains(t, actual.Args, "--collector."+collector)
+		}
+
+		enabled, err := nodeExporterConfig(node, &models.Agent{
+			AgentID:   "agent-id",
+			AgentType: models.NodeExporterType,
+		}, agentVersion)
+		require.NoError(t, err, "Unable to build node exporter config")
+
+		for _, collector := range defaultNodeExporterCollectors {
+			assert.Contains(t, enabled.Args, "--collector."+collector,
+				"%q may only be listed if node_exporter is also told to enable it", collector)
+		}
+	})
+
+	t.Run("MacOSDisabledCollectors", func(t *testing.T) {
+		t.Parallel()
+		node := &models.Node{
+			Distro: "darwin",
+		}
+		exporter := &models.Agent{
+			AgentID:   "agent-id",
+			AgentType: models.NodeExporterType,
+			ExporterOptions: models.ExporterOptions{
+				DisabledCollectors: []string{"cpu", "diskstats", "hwmon", "meminfo"},
+			},
+		}
+		agentVersion := version.MustParse("2.15.1")
+
+		actual, err := nodeExporterConfig(node, exporter, agentVersion)
+		require.NoError(t, err, "Unable to build node exporter config")
+
+		requireNoDuplicateFlags(t, actual.Args)
+		for _, arg := range actual.Args {
+			assert.False(t, strings.HasPrefix(arg, "--no-collector."), "unexpected %q on a darwin node", arg)
+		}
 	})
 
 	t.Run("MacOS", func(t *testing.T) {
--- a/managed/utils/collectors/collectors.go
+++ b/managed/utils/collectors/collectors.go
@@ -45,17 +45,25 @@ func FilterOutCollectors(prefix string, args, disabledCollectors []string) []str
 
 // DisableDefaultEnabledCollectors returns CLI arguments to disable default enabled collectors based on input.
 // DefaultCollectors and disabledCollectors should be collector names without prefix.
-// Result will be returned with prefix.
+// Result will be returned with prefix, at most once per collector.
 func DisableDefaultEnabledCollectors(prefix string, defaultCollectors []string, disabledCollectors []string) []string {
 	defaultCollectorsMap := make(map[string]struct{})
 	for _, defaultCollector := range defaultCollectors {
 		defaultCollectorsMap[defaultCollector] = struct{}{}
 	}
 	args := []string{}
+	emitted := make(map[string]struct{})
 	for _, collector := range disabledCollectors {
-		if _, ok := defaultCollectorsMap[collector]; ok {
-			args = append(args, fmt.Sprintf("%s%s", prefix, collector))
+		if _, ok := defaultCollectorsMap[collector]; !ok {
+			continue
 		}
+		// A repeated flag is rejected by the exporters' flag parser, which would stop the
+		// exporter from starting, so a collector listed twice must still yield one argument.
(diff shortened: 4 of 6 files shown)


JiriCtvrtka:
@theTibi I see that I created my PR a several hours after yours. I asked Claude which PR seems more ready, and apparently mine does: https://github.com/percona/pmm/pull/5841.

Could you also ask it to compare the two PRs? Then we can decide which one to close. Thank you.

admin/commands/inventory/change_agent_node_exporter.go
```diff
@@ -83,10 +83,12 @@ func (cmd *ChangeAgentNodeExporterCommand) RunCmd() (commands.Result, error) {
 	// Parse custom labels if provided
 	customLabels := commands.ParseKeyValuePair(cmd.CustomLabels)
 
+	disableCollectors := commands.ParseDisableCollectors(cmd.DisableCollectors)
```
ademidoff, line 86: There is no way back once a collector is disabled: `--disable-collectors=` makes kong produce `[]string{}`, `commands.ParseDisableCollectors` (`admin/commands/base.go:181-195`) returns a nil slice, the swagger body field has no `omitempty` so it serialises as `"disable_collectors": null`, protojson maps both `null` and `[]` back to a nil slice, and `managed/models/agent_helpers.go:1268` treats nil as "no change". So after an operator runs `--disable-collectors=cpu`, no CLI or REST call restores cpu — the only workaround is replacing the list with some other harmless name. The mechanism predates this PR, but it was harmless while disabling a default-on collector did nothing; now that it takes effect, the one-way door is worth closing (the "Reversibility" live test only swapped one non-empty list for another, so it never hit this path).

ademidoff, line 86: Trimming here fixes one of six `change agent` commands, so the untrimmed value the ticket complains about stays reachable: `pmm-admin inventory change agent mysqld-exporter <id> --disable-collectors="heartbeat, global_status"` still stores `" global_status"` (`admin/commands/inventory/change_agent_mysqld_exporter.go:146`; same at `admin/commands/inventory/change_agent_mongodb_exporter.go:158`, `admin/commands/inventory/change_agent_postgres_exporter.go:146`, `admin/commands/inventory/change_agent_proxysql_exporter.go:108`, `admin/commands/inventory/change_agent_valkey_exporter.go:141`), as does the UI or any direct `PUT /v1/inventory/agents/<id>` with a space-padded name, and a row already holding `" meminfo"` from before the upgrade stays a permanent no-op. Trimming in `models.ChangeAgent`/`CreateAgent` (`managed/models/agent_helpers.go`) — or in the inventory service before it builds `ChangeExporterOptions` — would cover all six commands, the UI and the raw API in one place, rather than deferring to PMM-14919.

managed/services/agents/node.go
```diff
@@ -32,6 +32,30 @@ var (
 	v2_28_00 = version.MustParse("2.28.0-0")
 )
 
+// defaultNodeExporterCollectors lists the collectors enabled below that node_exporter also enables
+// on its own. For those, dropping PMM's --collector.<name> flag leaves the collector running, so an
+// explicit --no-collector.<name> is needed as well. This is deliberately not every default-on
+// collector: only a collector nodeExporterConfig already passes --collector.<name> for may be added,
+// because that proves the flag exists in the node_exporter of every pmm-agent this code builds
+// configs for. An unknown flag stops the exporter from starting.
+var defaultNodeExporterCollectors = []string{
```
ademidoff, line 41: This whitelist is a hand-maintained mirror of the `--collector.<name>` block 30 lines below, and only one direction is tested: `LinuxAllDefaultCollectorsDisabled` asserts every entry here has a matching `--collector.<name>`, but nothing asserts the converse — that every positively-enabled collector which node_exporter also enables by default appears here. The 15 entries are exactly right today (percona/node_exporter registers buddyinfo/meminfo_numa/processes as defaultDisabled, and textfile.lr/.mr/.hr plus standard.go/standard.process with isDefaultEnabled=false), so this isn't a live bug; but add `--collector.softnet` (default-on) to the block tomorrow, or bump the bundled node_exporter so an existing collector flips to default-on, and `--disable-collectors=<that name>` is silently a no-op again — exactly what PMM-15362 is about — with no test failing.

managed/services/agents/node.go
```diff
@@ -124,6 +148,14 @@ func nodeExporterConfig(node *models.Node, exporter *models.Agent, agentVersion
 
 	args = collectors.FilterOutCollectors("--collector.", args, exporter.ExporterOptions.DisabledCollectors)
 
+	if node.Distro != "darwin" {
```
ademidoff: Skipping darwin means the ticket's symptom still reproduces on macOS: `FilterOutCollectors` has nothing to strip there (the `--collector.*` block at `managed/services/agents/node.go:73` is Linux-only) and no `--no-collector.<name>` is emitted, so the collector keeps running on the monitored host and a direct `/metrics` scrape still exposes it — the PR's stated goal isn't met for darwin nodes. Every collector in the darwin `hrCollect` list (`managed/services/victoriametrics/scrape_configs.go:297-306`: cpu, diskstats, filesystem, loadavg, meminfo, netdev, time) exists and is default-enabled in darwin node_exporter, so a small darwin-specific whitelist would cover it; only the Linux-only names (bonding, entropy, filefd, hwmon, netstat, stat, vmstat) must not be emitted there. As written, `managed/services/agents/node_test.go:375` asserts that no `--no-collector.` argument is ever produced on darwin, which will make the eventual fix look like a test regression.

ademidoff: `node.Distro != "darwin"` is now tested twice in the same function (also at `managed/services/agents/node.go:73`), so the darwin rule lives in two places and a future change to it has to be made twice or the blocks silently disagree. `FilterOutCollectors("--collector.", ...)` keys on `--collector.<name>` and can never match a `--no-collector.<name>` argument, so appending these args inside the existing block, before the filter call, produces byte-identical output with one guard instead of two.

managed/services/agents/node_test.go
```diff
@@ -291,6 +297,85 @@ func TestNodeExporterConfig(t *testing.T) {
...
+		require.NoError(t, err, "Unable to build node exporter config")
+
+		requireNoDuplicateFlags(t, actual.Args)
+		for _, collector := range disabled {
+			assert.NotContains(t, actual.Args, "--collector."+collector)
+			assert.NotContains(t, actual.Args, "--no-collector."+collector)
+		}
+	})
+
+	t.Run("LinuxAllDefaultCollectorsDisabled", func(t *testing.T) {
```
ademidoff, line 324: The duplicate-input case that motivated the dedupe is only covered in `managed/utils/collectors/collectors_test.go`, never through `nodeExporterConfig` — where a collision with the static `--no-collector.*` block at `managed/services/agents/node.go:82-108` is also possible. `requireNoDuplicateFlags` is already available and this subtest is one line away from covering it (`DisabledCollectors: []string{"diskstats", "diskstats", "cpu"}`); without that, a regression in the wiring at `managed/services/agents/node.go:151` — calling the helper twice, or a future overlap between the whitelist and the static block — ships with green unit tests, which is precisely the `flag 'collector.diskstats' cannot be repeated` failure described in the PR.

managed/services/agents/node_test.go
```diff
@@ -291,6 +297,85 @@ func TestNodeExporterConfig(t *testing.T) {
...
+		t.Parallel()
+		node := &models.Node{}
+		exporter := &models.Agent{
+			AgentID:   "agent-id",
+			AgentType: models.NodeExporterType,
+			ExporterOptions: models.ExporterOptions{
+				DisabledCollectors: defaultNodeExporterCollectors,
```
ademidoff: This aliases the production global into `models.ExporterOptions` while `LinuxDisabledCollectors`, `LinuxDisabledCollectorsNotDefaultEnabled` and `MacOSDisabledCollectors` run in parallel. Nothing mutates the argument today, so it passes; but the moment anyone sorts or filters `disabledCollectors` in place inside `FilterOutCollectors`/`DisableDefaultEnabledCollectors` — a natural way to implement the dedupe, incidentally — the global is permanently reordered for every other parallel subtest and for the running server, producing a failure that reproduces only under `-race`/ordering luck. `slices.Clone(defaultNodeExporterCollectors)` costs nothing.

managed/utils/collectors/collectors.go
```diff
@@ -45,17 +45,25 @@ func FilterOutCollectors(prefix string, args, disabledCollectors []string) []str
...
 // DefaultCollectors and disabledCollectors should be collector names without prefix.
-// Result will be returned with prefix.
+// Result will be returned with prefix, at most once per collector.
 func DisableDefaultEnabledCollectors(prefix string, defaultCollectors []string, disabledCollectors []string) []string {
 	defaultCollectorsMap := make(map[string]struct{})
 	for _, defaultCollector := range defaultCollectors {
 		defaultCollectorsMap[defaultCollector] = struct{}{}
 	}
 	args := []string{}
+	emitted := make(map[string]struct{})
```
ademidoff: Deduping at the last consumer leaves the duplicates everywhere else: `--disable-collectors=cpu,cpu` still round-trips into `agents.exporter_options->disabled_collectors` as `["cpu","cpu"]`, is echoed back by ChangeAgent/ListAgents and shown by `pmm-admin list`. Only `DisableDefaultEnabledCollectors` is immune now — `FilterOutCollectors` happens to be immune because it builds a map, but the next helper that emits one argument per entry reintroduces `flag 'collector.cpu' cannot be repeated`. Deduping in `commands.ParseDisableCollectors` plus `models.ChangeAgent`/`CreateAgent` normalises the stored value once and makes this helper-level guard unnecessary.

ademidoff: `emitted` isn't needed — `defaultCollectorsMap` can answer "have I already emitted this?" by itself: emit, then `delete(defaultCollectorsMap, collector)`. The membership check above then does double duty, the second lookup and insert go away, and the explanatory comment stays attached to the single remaining branch.

ademidoff (commented):
One finding lands in a file this PR doesn't touch, so it goes here: `managed/services/inventory/agents.go:246` — `ChangeNodeExporter` pushes the new agent state without regenerating the VictoriaMetrics scrape config first, so for up to the 3s `updateBatchDelay` (`managed/services/victoriametrics/victoriametrics.go:47`, cleared when StateChanged round-trips into `managed/services/agents/handler.go:272`) VM still sends `collect[]=cpu` while node_exporter has already restarted with `--no-collector.cpu`; percona/node_exporter returns `disabled collector: cpu` as HTTP 400 for the *entire* scrape, so every HR node metric for that host gaps for a few scrapes, not just `node_cpu_*`. That window existed before this PR only for the three non-default collectors PMM enables (processes, buddyinfo, meminfo_numa) — this change extends it to cpu/meminfo/diskstats/filesystem/netdev, i.e. the common case. `prometheusService` in `managed/services/inventory/deps.go:49` already exposes `ForceConfigurationUpdate(ctx)`; calling it before `RequestStateUpdate` closes the window entirely, since dropping a name from `collect[]` early is harmless.

admin/commands/inventory/change_agent_node_exporter.go
theTibi: Confirmed live: `[]` and `null` both return 200 and leave the list unchanged. Not fixed here — proto3 repeated fields have no presence, so `[]` and `null` are indistinguishable by the time they reach the model. `ChangeExporterOptions` already honours nil = no change / empty = clear, so the fix is an API-surface change (FieldMask or wrapper), not a model one.

theTibi: Agreed, wrong layer. Moved to `models.CreateAgent`/`ChangeAgent` in f9d31db, covering all six commands, the UI and a raw PUT. Verified live: `PUT ["cpu", " meminfo"]` was a silent no-op before, now stores `["cpu","meminfo"]` and emits `--no-collector.meminfo`.

managed/services/agents/node.go
theTibi: Added `LinuxDefaultOnCollectorsAreListed` in 8ccb450: every default-on collector we enable must appear in the whitelist. Verified by removing `cpu` from the list.

managed/services/agents/node.go
theTibi: Correct, and the assertion was the worse half. 8ccb450 narrows it to the Linux-only names with a TODO, so a darwin fix won't read as a regression. Not fixing darwin itself — no macOS node to verify against.

theTibi: Fixed in 9aa3a4e. Byte-identical confirmed on a live exporter: 57 args on each build, empty diff.

managed/services/agents/node_test.go
theTibi: Added `LinuxDuplicateDisabledCollectors` in 8ccb450, including a name that collides with the static block. Verified by reverting the dedupe.

managed/services/agents/node_test.go
theTibi: Fixed in 8ccb450 — `slices.Clone`.

managed/utils/collectors/collectors.go
theTibi: Normalised in `models` in f9d31db. Keeping the helper guard: normalising on write does nothing for rows already stored with a repeat. Worth noting the duplicate case was never a startup failure in practice — the helper dedupe already caught it, so this only cleans what's stored.

theTibi: Fixed in 9aa3a4e.

theTibi:
Re the `ChangeNodeExporter` / `ForceConfigurationUpdate` finding: done in de3ff34, gated on `DisableCollectors != nil`, failure logged rather than returned.

It narrows the window but doesn't close it — `/-/reload` is acknowledged immediately and applied asynchronously, so a race against the exporter restart remains. Measured over 20 disable transitions on a live PMM 3.9.1, counting scrapes lost to `disabled collector: cpu`: 13 before, 3 after. Eliminating it needs node_exporter to stop failing the whole scrape when `collect[]` names a disabled collector — filed as PMM-15547.

PR #5840, opened 2026-08-26
PMM-15384 Group the SEP apps under Management

nachodd:
Stacked on #5820 — base branch is `PMM-15359-report-failed-ui-actions`, so review only the last commit.

## What

PMM-15358 opened the SEP routes to every signed-in user, which left **Support diagnostics** and **MySQL Backups** rendering as two loose top-level entries wedged between the inventory divider and the admin-only Inventory/Backups/Configuration block. That placement was deliberate — it kept the administrator's ordering untouched and deferred the grouping to this ticket.

They now sit under a single collapsible **Management** section (icon: `HandymanOutlined`), placed right below Inventory.

## Why these choices

- **Placement — right below Inventory.** Exactly where the flat entries already were, so no pre-existing entry moves for an admin.
- **Divider — unchanged.** `NAV_DIVIDERS.inventory` still opens the block with Inventory itself, so its name still describes what it opens. No `NAV_DIVIDERS` change.
- **No page of its own.** `SidebarNavItem` derives a collapsible's link from `children[0].url` and ignores the parent's, so `NAV_MANAGEMENT` deliberately carries no `url`; the header opens whichever app leads the group.
- **Empty-shell resilience.** `addSection()` returns nothing at all when the child list is empty, rather than a collapsible that expands onto nothing. Callers spread the result, so the section can contribute zero entries if per-app filtering is ever added.

## Acceptance criteria

- [x] Viewer sees the Management section with both SEP apps (visible to Viewer, Editor and Admin — only the in-page write controls stay admin-gated, per PMM-15358).
- [x] Admin sees the same section; every previously reachable entry is still reachable and nothing moved.
- [x] Deep-linking to `/pmm/sep/atw` or `/pmm/sep/mysql-backups` expands the section and marks the correct child active. The sidebar auto-expands by object identity between the resolved active item and the section's `children`, so the test asserts identity, not just the id.
- [x] `cd ui && make lint && make format-check && make test` and `pnpm --filter ui exec tsc --noEmit` are clean.

## Tests

`navigation.provider.test.tsx` is new — the assembled nav tree had no coverage at all. It covers the group's shape for admin/editor/viewer, that the SEP apps no longer appear at top level, that each child keeps its `url`/`matches`/icon, the ordering of the block the section joins, and deep links into both apps. `navigation.utils.test.ts` gains `addSepApps` and `addSection` coverage.

## Test plan

1. Sign in as a Viewer — **Management** appears below the inventory divider with both apps; each opens read-only.
2. Sign in as an Admin — same section, Inventory still directly above it, Backups/Configuration unchanged below.
3. Navigate directly to `/pmm/sep/atw` and to `/pmm/sep/mysql-backups` — the section is expanded and the right child highlighted in both cases.
4. Collapse the sidebar — the section shows its icon with a tooltip, as other collapsibles do.

## Notes

- Label and icon were confirmed with the ticket owner.
- Base `PMM-15359` predates 15358's `!user.isAnonymous` guard around `addSepApps()`, and merging 15358 here would drag ~60 unrelated files into this diff. Rather than wait for the sync, this branch now carries an identical guard (#5840 review), so the eventual conflict is comment-only and a test pins the guard in place.
- Review feedback addressed before opening: replaced an unreachable emptiness guard (and the vacuous test that "covered" it) with the reachable `addSection()` helper, and narrowed a whole-tree ordering assertion to the block this change actually touches.


fabio-silva (approved):

ui/apps/pmm/src/contexts/navigation/navigation.provider.tsx
```diff
@@ -97,11 +97,12 @@ export const NavigationProvider: FC<PropsWithChildren> = ({ children }) => {
...
-      // and holds every unsafe method to administrators, so a non-admin gets a
-      // read-only view with no write control rendered (PMM-15358). Grouping
-      // these under a "Management" section is a follow-up.
+      // SEP apps mounted as native routes, grouped under "Management" and
+      // placed right below Inventory so no pre-existing entry moves. Offered to
+      // every signed-in user, not only admins: SEP's API serves its reads to
+      // any authenticated session and holds every unsafe method to
+      // administrators, so a non-admin gets a read-only view with no write
+      // control rendered (PMM-15358).
       items.push(...addSepApps());
```
mattiasimonato: [#5819](https://github.com/percona/pmm/pull/5819) has the anonymous guard on this exact line in `ui/apps/pmm/src/contexts/navigation/navigation.provider.tsx`:

```ts
if (!user.isAnonymous) {
  items.push(...addSepApps());
}
```

This PR rewrites the comment right above it and leaves the `push` bare, so the sync will conflict here and the `if` is easy to lose while reconciling the prose.

Could we add the same guard on this branch? If both sides carry an identical `if`, the conflict becomes comment-only and the guard survives.

```suggestion
      // SEP apps mounted as native routes, grouped under "Management" and
      // placed right below Inventory so no pre-existing entry moves. Offered to
      // every signed-in user, not only admins: SEP's API serves its reads to
      // any authenticated session and holds every unsafe method to
      // administrators, so a non-admin gets a read-only view with no write
      // control rendered (PMM-15358).
      //
      // Signed-in is the rule, so anonymous is excluded: it has no Grafana
      // session cookie to exchange for a SEP bearer, and the entry would open
      // on SepAuthGate's failure card rather than on the app.
      if (!user.isAnonymous) {
        items.push(...addSepApps());
      }
```

That also makes this test pass now, and `ui/apps/pmm/src/contexts/navigation/navigation.provider.test.tsx` currently covers every role except the one in flight:

```ts
it('offers no Management section to an anonymous session', () => {
  expect(findById(renderNavTree(TEST_USER_ANONYMOUS), 'management')).toBeUndefined();
});
```

Happy to drop this if you'd rather keep [PMM-15384](https://perconadev.atlassian.net/browse/PMM-15384) free of authorization changes and handle it when [PMM-15358](https://perconadev.atlassian.net/browse/PMM-15358) syncs. Rest of the PR looks good.

nachodd: Good call — applied in bf5515a73. Same `if (!user.isAnonymous)` as #5819, with #5819's anonymous paragraph appended verbatim to the rewritten comment, so the sync conflicts on prose only and the guard survives on either side.

Also added a NavigationProvider test asserting `management` is absent for `TEST_USER_ANONYMOUS`, so the guard can't be dropped silently during that reconcile.

mattiasimonato (approved):

PR #5841, opened 2026-08-27
PMM-15362 Disable default-on node_exporter collectors explicitly.

JiriCtvrtka:
Ticket number: PMM-15362

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/4578

## Problem

`pmm-admin inventory change agent node-exporter --disable-collectors=diskstats` reported success,
stored the value in the inventory API and dropped `--collector.diskstats` from the node_exporter
command line — but the metrics kept coming. node_exporter enables `cpu`, `diskstats`, `meminfo`,
`filesystem`, `netdev`, `stat` and friends on its own, so removing the redundant enable flag changes
nothing; they need `--no-collector.<name>`. Disabling collectors that node_exporter defaults to off
(`processes`, `buddyinfo`, `meminfo_numa`) always worked, which is why this went unnoticed.

## Solution

`nodeExporterConfig` now appends `--no-collector.<name>` for every disabled collector that
node_exporter enables by default, the way `mongodb.go` and `postgresql.go` already do via
`collectors.DisableDefaultEnabledCollectors`. `defaultEnabledNodeExporterCollectors` was derived
mechanically from percona/node_exporter (every `registerCollector`/`RegisterCollectorPublic`/
`ReplaceCollector` call in the file set `go list` selects for `GOOS=linux`), not from upstream docs,
so it matches the shipped binary. 14 of those names are in the static `--no-collector.` block above
already, so the flags are appended only when missing — `requireNoDuplicateFlags` guards that.

## Collector names are matched exactly

A name in `DisabledCollectors` is matched against the exact node_exporter collector name and never
stands for a family, which is the behaviour `FilterOutCollectors` already documents. `textfile` is
therefore the upstream base collector alone — the textfile metrics PMM actually collects come from
the separate, default-off `textfile.hr`/`textfile.mr`/`textfile.lr` collectors, so silencing those
means listing them by name, and dropping their `--collector.` flag is what disables them.

Two consequences worth knowing when reviewing or testing this:

- Disabling `textfile` does not stop `node_textfile_*`. That is unchanged by this PR (before it, the
  name was a no-op in both directions) and it is not what `--no-collector.textfile` is for.
- `--no-collector.textfile` is in fact inert under PMM today: the base collector appears in no
  `collect[]` list, and `--collector.textfile.directory` is never set, so it has nothing to read.
  It stays in the list because the list is an inventory of what node_exporter turns on by itself,
  and an unfiltered `/metrics` scrape does run it.

Anything disabled here has to stay in sync with the `collect[]` filter built by
`scrapeConfigsForNodeExporter`, which filters the same names out. Naming a disabled collector in
`collect[]` makes `NewNodeCollector` fail and node_exporter answer the whole resolution endpoint
with HTTP 400 — so a mismatch costs every metric at that resolution, not just one collector. The
same applies in time, not just in content — see "Config update timing" below.

## Config update timing

Because disabling a collector now changes what node_exporter accepts in `collect[]`, the scrape
config and the exporter have to change together. `ChangeNodeExporter` never requested a config
rebuild of its own: the only thing that regenerated it was the `StateChanged` callback, which
arrives as the exporter is already restarting. That left a window of up to `updateBatchDelay` (3s)
where VictoriaMetrics scraped the restarted process with the stale `collect[]` and got a 400 for
the whole resolution endpoint. Default HR is 5s, so the real cost is zero or one missed scrape on
one node, and it self-heals — but it is new behaviour, since before this PR disabling a collector
could never produce a 400.

`ChangeNodeExporter` now forces a synchronous config update before requesting the state update, the
same pattern already used for port changes in `agents.Handler.stateChanged` (PMM-14267), which
makes the ordering deterministic rather than merely self-healing.

## Why this is gated on pmm-agent 3.x

node_exporter's default-enabled set is version dependent, and kingpin exits on an unknown flag — so
emitting these unconditionally would turn a harmless no-op into a node_exporter that refuses to
start, losing every metric for that node. Per the fork's release branches:

| pmm-agent | node_exporter | missing from the list |
| --- | --- | --- |
| 2.31.0 | 1.3.1 | `selinux`, `watchdog` |
| 2.32.0 – 2.44.1 | 1.4.0 | `watchdog` |
| 3.x | 1.8.2 | — |

So the new flags are emitted only for pmm-agent >= 3.0.0 (`version.NodeExporterV1_8`). Older agents
keep the current behaviour: the enable flag is dropped, nothing else changes.

## Not covered

- macOS nodes. Collectors are not tweaked there at all today (the default-on set differs from Linux),
  and that stays true — a test pins it.
- node_exporter ships `--collector.disable-defaults` (since 1.0.0). Passing it, plus the enable flags
  we already list, would make this list, the 26-entry static `--no-collector.` block and the version
  gate all unnecessary. It needs `--collector.os` added first — `scrape_configs.go` requests
  `collect[]=os` for LR but we never enable it explicitly — so it is a separate change.
- `proxysql.go` has the same bug class (`mysql_status`, `mysql_connection_pool`,
  `mysql_connection_list` default to true in proxysql_exporter). It cannot reuse this helper:
  proxysql_exporter uses stdlib `flag`, where negation is `-collect.X=false`. Separate ticket.
- The other exporters need no equivalent to the config-update change. mongodb_exporter and
  postgres_exporter already emit `--no-collector.` and already get a `DisabledCollectors`-filtered
  `collect[]`, but neither fails the scrape: mongodb_exporter feeds `collect[]` into
  `GetRequestOpts`, and postgres_exporter switch-maps the names with no `default`, so unknown ones
  are silently dropped. node_exporter is the only one of the three that answers 400.

## Testing

- `LinuxDisabledDefaultEnabledCollectors` — exact args at pmm-agent 3.0.0: `dmi` gains
  `--no-collector.dmi`, already-disabled `arp` stays single, `netstat.fields` is handled.
- `LinuxDisabledCollectors` — pmm-agent 2.15.1, asserts the gated path is unchanged.
- `LinuxDisabledTextfileCollectors` — exact-name matching: `textfile` gains
  `--no-collector.textfile`, `textfile.hr` only loses its enable flag (it is default-off, so it must
  not gain a `--no-` form), and `textfile.mr`/`textfile.lr` keep collecting.
- `MacOSDisabledCollectors` — darwin with disabled collectors produces no `--no-collector.` flags.
- `ChangeNodeExporterAndRemoveCustomLabels` — asserts the forced config update happens. The mock
  expectation is load-bearing: `setup`'s teardown calls `vmdb.AssertExpectations(t)`, so the test
  fails if `ChangeNodeExporter` stops forcing the update.


managed/services/agents/node.go
```diff
@@ -124,6 +187,22 @@ func nodeExporterConfig(node *models.Node, exporter *models.Agent, agentVersion
 
 	args = collectors.FilterOutCollectors("--collector.", args, exporter.ExporterOptions.DisabledCollectors)
 
+	// Collectors are not tweaked on macOS, where node_exporter enables a different set by default.
+	// Older pmm-agents ship node_exporter builds that do not know all of the flags below and would exit.
+	if node.Distro != "darwin" && agentVersion.IsFeatureSupported(version.NodeExporterV1_8) {
```
4nte, line 192: Heads up on a second-order effect of this, not a blocker.

`ChangeNodeExporter` (`managed/services/inventory/agents.go:238`) never calls `vmdb.RequestConfigurationUpdate()`. The only thing that regenerates the VM scrape config is the `StateChanged` callback at `managed/services/agents/handler.go:272`, and that arrives as the exporter is restarting. So on `pmm-admin inventory change agent node-exporter --disable-collectors=diskstats`:

- t+0: agent restarts node_exporter, `toStarting` emits STARTING, server queues a VM config update
- t+~10ms: new process is bound and serving, now answering `?collect[]=...diskstats...` with `400 disabled collector: diskstats`
- t+~3s: `updateBatchDelay` expires, VM config is rewritten without `collect[]=diskstats`

Since the 400 fails the whole resolution endpoint, the node loses every HR metric for that window, not just diskstats. Default HR is 5s (`managed/models/settings.go:206`), so the real cost is zero or one missed scrape on one node.

This is new behaviour: before this PR, disabling a collector could never produce a 400, because the exporter kept it enabled. I checked whether it can wedge and it can't, so I don't think it should hold the merge: `toStarting` emits STARTING before `exec`, `stateChanged` requests a config update on every status change and not just port changes, and `SendActualStatuses` (`agent/client/client.go:274`) re-fires on every reconnect, so a dropped connection at the wrong moment still recovers.

If you want it deterministic rather than merely self-healing, `ForceConfigurationUpdate(ctx)` before `RequestStateUpdate` in `ChangeNodeExporter` closes it, which is the same pattern already used for port changes in `handler.go:262-268` (PMM-14267).

4nte (approved):

managed/services/inventory/agents.go
```diff
@@ -243,6 +244,16 @@ func (as *AgentsService) ChangeNodeExporter(ctx context.Context, agentID string,
...
 	}
+
+	// Changing the disabled collectors changes the set of collectors node_exporter accepts in
+	// collect[] parameters, and it rejects the whole scrape with 400 for an unknown one. Force an
+	// immediate synchronous config update so that VictoriaMetrics never scrapes the restarted
+	// exporter with the stale collector list, the same way port changes are handled in
+	// agents.Handler.stateChanged (PMM-14267).
+	if err := as.vmdb.ForceConfigurationUpdate(ctx); err != nil {
+		return nil, fmt.Errorf("failed to force configuration update: %w", err)
+	}
```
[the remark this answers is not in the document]
JiriCtvrtka: Thanks — the premise is right, but I don't think the severity is, and I'd rather not
diverge from the pattern this was asked to follow.

Correct that `executeAgentChange` commits before the forced update, so a failure returns
before `RequestStateUpdate` and the agent isn't told to restart.

It doesn't stay unapplied, though. `handler.go:97` calls `RequestStateUpdate` when a
pmm-agent connects, and `state.go:185` retries state updates on a backoff (PMM-15228),
so a failed force update is picked up on the next reconnect or retry. The window also
needs `configAndReload` to actually fail to open at all.

More to the point, this is the same tradeoff `handler.go:262-272` already makes for port
changes (PMM-14267), and that is the pattern this change was explicitly asked to follow
in https://github.com/percona/pmm/pull/5841#discussion_r3893006451. Making only this call
site retryable would leave the two force-update paths inconsistent while fixing half of a
property neither has. If we want that, it should be its own ticket covering both.


What do you think?

managed/services/inventory/agents.go
```diff
@@ -243,6 +244,17 @@ func (as *AgentsService) ChangeNodeExporter(ctx context.Context, agentID string,
 	if !ok {
 		return nil, unexpectedAgentTypeError(agent)
 	}
+
+	// Changing the disabled collectors changes the set of collectors node_exporter accepts in
+	// collect[] parameters, and it rejects the whole scrape with 400 for an unknown one. Force an
+	// immediate synchronous config update so that VictoriaMetrics never scrapes the restarted
+	// exporter with the stale collector list, the same way port changes are handled in
+	// agents.Handler.stateChanged (PMM-14267).
+	err = as.vmdb.ForceConfigurationUpdate(ctx)
```
ademidoff: **Re-enable direction opens a symmetric HTTP 400 window.**

Forcing the VM config update before the exporter restarts fixes the *disable* direction, but it inverts the ordering on *re-enable*. With `DisabledCollectors=["cpu"]` the exporter is running with `--no-collector.cpu`; clearing the list rebuilds the scrape config synchronously, so `collect[]=cpu` is back in the HR job immediately while `RequestStateUpdate` is still async. The next HR scrape (5s by default) hits the un-restarted exporter, `NewNodeCollector` returns `disabled collector: cpu` (`collector/collector.go:114`) and node_exporter answers **400 for the entire endpoint** — every HR metric for that node is lost, not just cpu.

If the pmm-agent is disconnected, `RequestStateUpdate` silently returns (`managed/services/agents/state.go:89-93`) and the 400s continue until it reconnects and restarts the exporter. Before this PR no `--no-collector.` flags existed, so a 400 was impossible in either direction.

managed/services/inventory/agents.go
```diff
@@ -243,6 +244,17 @@ func (as *AgentsService) ChangeNodeExporter(ctx context.Context, agentID string,
...
+	if err != nil {
```
ademidoff: **Returning here aborts after the DB change is already committed.**

`executeAgentChange` has committed `DisabledCollectors` by this point. If `ForceConfigurationUpdate` fails — VictoriaMetrics mid-restart, or the `victoriametrics -promscrape.config.dryRun` validator killed because the gRPC client context was cancelled (`exec.CommandContext(ctx, ...)`, `managed/services/victoriametrics/victoriametrics.go:298`) — we return early and `as.state.RequestStateUpdate` on the next line never runs.

Result: inventory says the collector is disabled, node_exporter keeps running with the old args until some unrelated event triggers a state update, and the caller is told the operation failed even though it was applied. The old async path (`RequestConfigurationUpdate` from `stateChanged`) could never block the state update this way. Log-and-continue, or at minimum still calling `RequestStateUpdate`, would be safer.

managed/services/inventory/agents.go
```diff
@@ -243,6 +244,17 @@ func (as *AgentsService) ChangeNodeExporter(ctx context.Context, agentID string,
...
+		return nil, fmt.Errorf("failed to force configuration update: %w", err)
```
ademidoff: **`status.Error()` rather than a bare `fmt.Errorf` on the API surface.**

`AGENTS.md`, "PMM-specific choices": *API errors (Go): `status.Error()` with gRPC codes — not ad-hoc HTTP errors in service layers*. `ChangeNodeExporter` is the inventory API service layer (reached from `managed/services/inventory/grpc/agents_server.go:245`), so this reaches pmm-admin/UI as `codes.Unknown` with no actionable code.

Worth noting too that the wrapped error may already be a status error — `validateConfig` returns `status.Error(codes.Aborted, ...)` at `managed/services/victoriametrics/victoriametrics.go:306` — and `%w`-wrapping destroys that code. (`managed/services/agents/handler.go:268` does the same thing, but that is the agent stream handler, not an API method.)

managed/services/inventory/agents.go
```diff
@@ -243,6 +244,17 @@ func (as *AgentsService) ChangeNodeExporter(ctx context.Context, agentID string,
 	if !ok {
 		return nil, unexpectedAgentTypeError(agent)
 	}
+
+	// Changing the disabled collectors changes the set of collectors node_exporter accepts in
```
ademidoff: **The force is unconditional, unlike the PMM-14267 precedent cited here.**

`managed/services/agents/handler.go:262-272` gates the force on `portsChanged` and otherwise uses the async path. This one runs on every `ChangeNodeExporter` call: `models.ChangeExporterOptions` documents `DisabledCollectors []string // nil = no change` (`managed/models/agent_helpers.go:1065`), so a request that only flips `Enable`, sets custom labels, or changes log level or metrics resolutions still pays a full synchronous rebuild — `populateConfig` runs a DB transaction over the whole inventory, `validateConfig` forks an external `victoriametrics` binary, then an HTTP reload, all inside the gRPC request.

For an agent with push metrics enabled the server-side VM config does not contain that target at all (its scrape config ships to the client's vmagent in the same `SetStateRequest`), so the force is pure added latency. Suggest gating on "`DisabledCollectors` actually changed".

managed/services/inventory/agents.go
ademidoff: **This path bypasses the serializing `reloadCh` goroutine.**

`victoriametrics.Service` holds no mutex (`managed/services/victoriametrics/victoriametrics.go:58-70`). `RequestConfigurationUpdate` funnels through a single `reloadCh` consumer, but `ForceConfigurationUpdate` calls `updateConfiguration` directly on the caller's goroutine. Two concurrent `ChangeNodeExporter` calls (or one plus the background reload loop, or plus `handler.stateChanged`'s port-change force) each read `oldCfg` in `configAndReload` (`victoriametrics.go:334`), then both `os.WriteFile` the same `scrapeConfigPath` (`:365`). If the second one's reload fails, its deferred `restore` writes back the `oldCfg` it captured *before* the first write (`:348`), permanently discarding the first caller's change and reloading VM with a stale config.

Pre-existing since PMM-14267, but this change moves the unsynchronized path from a rare event (port change) onto a routine API call.

managed/services/agents/node.go
ademidoff: **For most of this list the flag changes nothing that is scraped — while creating the 400 hazard.**

Every collector PMM names in `hrCollect`/`mrCollect`/`lrCollect` (cpu, diskstats, filesystem, loadavg, meminfo, netdev, time, netstat, stat, vmstat, filefd, hwmon, bonding, entropy, uname, os, …) is already dropped from `collect[]` by `FilterOutCollectors` (`managed/services/victoriametrics/scrape_configs.go:265`, `:278`, `:307`), and node_exporter's filtered handler only runs the collectors named there (`collector/collector.go:122-125`) — so the metric already stops without any flag. Adding `--no-collector.X` for those names changes nothing scraped, but it is precisely what makes `NewNodeCollector` return 400 when the flag set and `collect[]` are momentarily out of sync, in both directions.

Restricting the new flags to collectors that appear in no `collect[]` list (dmi, pressure, rapl, nvme, selinux, watchdog, …), or taking the `--collector.disable-defaults` route the PR description already identifies, would deliver the same user-visible result with zero 400 risk — and would make the `ForceConfigurationUpdate` change in `managed/services/inventory/agents.go` unnecessary.

managed/services/agents/node.go
```diff
@@ -32,6 +33,68 @@ var (
...
+// which is why they are appended only when missing.
+//
+// Entries are exact node_exporter collector names, the same way DisabledCollectors is matched everywhere
+// else, so a name here never stands for a family of collectors. In particular "textfile" is the upstream
+// base collector alone: the textfile metrics PMM actually collects come from the separate, default-off
+// "textfile.hr"/"textfile.mr"/"textfile.lr" collectors, and silencing those means listing them by name so
+// that FilterOutCollectors drops their "--collector." flag. Disabling a collector must stay in sync with
+// scrapeConfigsForNodeExporter, which filters the same names out of "collect[]" - naming a disabled
+// collector there makes node_exporter answer the whole resolution endpoint with HTTP 400.
+var defaultEnabledNodeExporterCollectors = []string{
```
ademidoff, line 47: **Lower version bound only, so this list can rot into a hard failure.**

The gate is `agentVersion.IsFeatureSupported(version.NodeExporterV1_8)` = `>= 3.0.0-0`, open-ended. node_exporter has removed default-on collectors before (`ntp` upstream), and 1.9+/2.x can remove or flip more. If a future pmm-agent ships a node_exporter where, say, `selinux` or `tapestats` no longer registers, a user who disables it gets `--no-collector.<name>`, kingpin errors on the unknown flag and node_exporter exits — the same catastrophic failure this PR's own rationale describes for 2.x, moved into the future.

Nothing ties these 46 entries to the vendored exporter: no generator, no test comparing against the binary's `--help`, no upper version bound. I verified all 46 against the `pmm-3.9.1` node_exporter clone and the list is exactly right today, which is why the drift would be silent when it happens.

managed/services/agents/node.go
```diff
@@ -32,6 +33,68 @@ var (
...
+	"pressure",
+	"rapl",
+	"schedstat",
+	"selinux",
+	"sockstat",
+	"softnet",
+	"stat",
+	"tapestats",
+	// the upstream base collector only, not PMM's textfile.hr/textfile.mr/textfile.lr
+	"textfile",
```
ademidoff, line 85: **`"textfile"` emits a provably inert flag.**

In the fork the base collector is `cl.ReplaceCollector("textfile", true, NewTextFileCollector)` reading `--collector.textfile.directory` (`percona/perconacollector/textfile.directory.go:38,58`), which pmm-managed never sets, and the base name appears in no `collect[]` list — so `--no-collector.textfile` is a guaranteed no-op.

The cost is UX: `pmm-admin inventory change agent node-exporter <id> --disable-collectors=textfile` now adds a flag to the command line and the user still sees every `node_textfile_*` series, because the real collectors are `textfile.hr/mr/lr`. The comment above and the new test pin this behavior accurately, but the entry itself buys nothing and is one more name that can rot.

managed/services/agents/node.go
```diff
@@ -124,6 +187,22 @@ func nodeExporterConfig(node *models.Node, exporter *models.Agent, agentVersion
...
+			defaultEnabledNodeExporterCollectors,
+			exporter.ExporterOptions.DisabledCollectors,
+		)
+		for _, arg := range disableArgs {
+			// some collectors are already disabled above
+			if !slices.Contains(args, arg) {
+				args = append(args, arg)
+			}
+		}
+	}
```
ademidoff: **Could live inside the existing darwin branch; the dedup is O(n·m).**

`node.Distro != "darwin"` is now tested twice (here and at `managed/services/agents/node.go:112`), so a future change to how macOS is detected has two places to update and one is easy to miss. `FilterOutCollectors` only matches the `--collector.` prefix, so `--no-collector.` args are untouched by it — this block could sit inside the existing `if node.Distro != "darwin"` branch with just the version check added.

The dedup rescans the whole ~60-element `args` slice for each of up to 46 candidates; the 14 overlapping names are a known static set that could be precomputed instead of probed.

ademidoff (commented):
Review of `origin/main...origin/PMM-15362-node-exporter-disable-default-collectors` (5 files).

I cross-checked the collector list against the `pmm-3.9.1` node_exporter clone: the 46 names match the shipped binary's default-enabled Linux set **exactly** (43 `registerCollector(..., defaultEnabled)` plus `filefd`/`rapl`/`sockstat` registered via subsystem consts), and `textfile.hr/mr/lr` plus `standard.go`/`process` are confirmed default-off — the list is correct as of today. I also confirmed in `collector/collector.go:107-118` and `node_exporter.go:88-91` that a `collect[]` naming a flag-disabled collector really does yield HTTP 400 for the whole endpoint.

12 inline comments below. The three worth deciding on before merge:

1. **The re-enable direction is still open.** The forced config update fixes the disable ordering but inverts it when the list is cleared, producing the same whole-endpoint 400 — indefinitely if the agent is offline.
2. **The flags may be unnecessary for most of the list.** For any collector already in a `collect[]` list, `--no-collector.X` changes nothing scraped but is exactly what makes the 400 possible.
3. **No upper version bound on the hard-coded list.** A future node_exporter that drops one of these names turns the flag into a kingpin exit and takes every metric for the node with it.

### One finding that falls outside the diff

**`managed/services/victoriametrics/scrape_configs.go:218` — when every collector at a resolution is disabled, `collect[]` is omitted and node_exporter falls back to the unfiltered handler.**

`mrCollect` is only `["hwmon", "textfile.mr"]`. Disable both and `FilterOutCollectors` returns an empty slice, so `if len(collect) != 0` skips setting `Params` and the MR job scrapes plain `/metrics`; node_exporter takes the `len(filters) == 0` branch (`node_exporter.go:82-86`) and runs every enabled collector. The user asked for fewer metrics and gets the full unfiltered set at MR resolution, duplicated against HR/LR. The new `--no-collector.` flags shrink that set but do not fix the fallback — cpu, diskstats, filesystem, netdev, stat … stay enabled and now stream at MR too. Same for `lrCollect`'s 5 names, and for the darwin 7-name `hrCollect` where this PR deliberately emits no flags.

managed/services/agents/node_test.go
```diff
@@ -291,6 +293,219 @@ func TestNodeExporterConfig(t *testing.T) {
...
+			"--collector.stat",
+			"--collector.textfile.directory.hr=" + pathsBase(agentVersion, "{{", "}}") + "/collectors/textfile-collector/high-resolution",
+			"--collector.textfile.directory.lr=" + pathsBase(agentVersion, "{{", "}}") + "/collectors/textfile-collector/low-resolution",
+			"--collector.textfile.directory.mr=" + pathsBase(agentVersion, "{{", "}}") + "/collectors/textfile-collector/medium-resolution",
+			"--collector.textfile.hr",
+			"--collector.textfile.lr",
+			"--collector.textfile.mr",
+			"--collector.time",
+			"--collector.uname",
+			"--collector.vmstat.fields=^(pg(steal_(kswapd|direct)|refill|alloc)_(movable|normal|dma3?2?)" +
```
ademidoff, line 343: **This pins a tuning flag surviving its collector being disabled.**

`LinuxDisabledDefaultEnabledCollectors` disables `vmstat` but not `vmstat.fields`, so `--collector.vmstat.fields=...` stays on the command line right next to `--no-collector.vmstat`. The same test has to list `"netstat.fields"` explicitly alongside `"netstat"` to get that regex dropped, because `FilterOutCollectors` splits on `=` and treats `--collector.netstat.fields` as a distinct flag name.

Harmless to kingpin, but it means `--disable-collectors=vmstat` leaves dead tuning flags behind while `--disable-collectors=netstat,netstat.fields` does not, and the asymmetry within a single test reads as intentional rather than as a wart. Consider filtering `<name>.`-prefixed tuning flags along with the collector, or at least stating the intent in a comment.

managed/services/agents/node_test.go
```diff
@@ -291,6 +293,219 @@ func TestNodeExporterConfig(t *testing.T) {
...
+			"--web.config.file={{ .TextFiles.webConfig }}",
+		}
+
+		requireNoDuplicateFlags(t, actual.Args)
+		require.Equal(t, expected, actual.Args)
+		// ... while textfile.hr is default-off, so it must not get one
+		require.NotContains(t, actual.Args, "--no-collector.textfile.hr")
+	})
+
+	t.Run("MacOSDisabledCollectors", func(t *testing.T) {
```
ademidoff, line 507: **Missing the dup-flag assertion its two siblings got, and the version gate is untested.**

`LinuxDisabledDefaultEnabledCollectors` (`managed/services/agents/node_test.go:379`) and `LinuxDisabledTextfileCollectors` (`managed/services/agents/node_test.go:473`) both call `requireNoDuplicateFlags(t, actual.Args)`; this one does not, so a future change that starts emitting a `--no-collector.` form on darwin while the `--collector.` form is still present would pass.

Separately, nothing pins the gate boundary: `LinuxDisabledCollectors` uses 2.15.1 and the new tests use 3.0.0, so a regression moving `version.NodeExporterV1_8` to, say, 3.2.0 — or a change in how `Parsed.Less` handles pre-releases, where `3.0.0-dev` currently compares as supported — would fail no test.

managed/services/inventory/agents_test.go
```diff
@@ -106,6 +106,8 @@ func TestAgents(t *testing.T) {
 		})
 
 		t.Run("ChangeNodeExporterAndRemoveCustomLabels", func(t *testing.T) {
+			as.vmdb.(*mockPrometheusService).On("ForceConfigurationUpdate", ctx).Return(nil)
```
ademidoff: **No `.Once()`, and the error branch is untested.**

Without `.Once()`, `AssertExpectations` only proves the call happened at least once — a regression that forces a full synchronous VM rebuild twice per request, or inside a loop over agents, would still pass.

More importantly, nothing exercises the failure branch added at `managed/services/inventory/agents.go:254-256`. A test returning an error from this mock would immediately surface that `RequestStateUpdate` is skipped while the DB change has already been committed.

managed/services/inventory/agents.go
JiriCtvrtka: Blocker, fixed

Correct, and this was the worst of the three. Fixed: the force now logs on failure, falls back to RequestConfigurationUpdate(), and falls through to RequestStateUpdate — the change is committed, so the exporter has to be restarted regardless. Added ChangeNodeExporterDisableCollectorsWithFailingConfigurationUpdate, which is exactly the test you suggested.

managed/services/inventory/agents.go
JiriCtvrtka: Blocker, fixed

Agreed. The force is now gated on the disabled set actually growing: p.DisableCollectors == nil (labels, enable, log level, resolutions) skips it entirely. ChangeNodeExporterAndRemoveCustomLabels now asserts ForceConfigurationUpdate is not called.

managed/services/inventory/agents.go
JiriCtvrtka: Blocker (convention), resolved by the fix above

Right on both counts, including that %w over validateConfig's status.Error(codes.Aborted, …) flattens to codes.Unknown. Moot now: the path no longer returns an error at all, it logs and continues, so the fmt import is gone.

managed/services/inventory/agents.go
JiriCtvrtka: High, fixed

Good catch, this was a real asymmetry. Fixed by the same gating: the force only runs when the disabled set grows. Shrinking or clearing it now falls through to the async update stateChanged triggers after the exporter restarts, which is the correct order for that direction. Pinned by ChangeNodeExporterEnableCollectorsBack.

JiriCtvrtka: Medium, follow-up ticket

Confirmed — no mutex on Service, and configAndReload's deferred restore can write back a stale oldCfg. Pre-existing since PMM-14267, and the gating above shrinks the exposure back to a rare event rather than every ChangeNodeExporter. Filing a separate ticket; serializing ForceConfigurationUpdate through the same consumer belongs there, not here.

managed/services/agents/node.go
JiriCtvrtka: Medium, acknowledged, not fixed

Fair, and thanks for independently verifying all 46 against the 3.9.1 clone. No cheap fix in this PR — a generator or a test diffing against the shipped binary's --help is the real answer, and it needs the exporter available at test time. Tracking it as a follow-up.

managed/services/agents/node.go
JiriCtvrtka: Low, fixed

Done — the block moved inside the existing node.Distro != "darwin" branch, so the macOS check appears once, and the dedup is a set built once from args. sort.Strings at the end makes placement output-identical; the existing tests pass unchanged.

managed/services/agents/node_test.go
JiriCtvrtka: Low, half fixed

First half correct — requireNoDuplicateFlags added to MacOSDisabledCollectors. Second half doesn't hold: the boundary is pinned from both sides already. LinuxDisabledCollectors runs at 2.15.1 and expects no --no-collector. additions, LinuxDisabledDefaultEnabledCollectors runs at 3.0.0 and expects them, so moving NodeExporterV1_8 in either direction fails a test. Only the pre-release ordering of 3.0.0-dev is unpinned.

managed/services/inventory/agents_test.go
JiriCtvrtka: Low, fixed

Both done. .Once() on every expectation, plus three new subtests covering the gate, the failure path, and the re-enable direction.

managed/services/agents/node_test.go
JiriCtvrtka: Low, comment

The asymmetry is FilterOutCollectors' documented whole-flag-name matching, and changing it would touch mysqld/mongodb/postgres too — out of scope here. Added a comment stating the intent so it doesn't read as accidental.

managed/services/agents/node.go
JiriCtvrtka: Low, keeping as-is

Keeping it deliberately. textfile genuinely is default-on, so dropping the entry would make --disable-collectors=textfile a silent no-op — the exact bug class this PR fixes. That it changes nothing scraped is the same property as every other collect[] name, which is the discussion in your other comment; the UX gap is that textfile.hr/mr/lr have to be named separately, and the doc comment covers that.

managed/services/agents/node.go
JiriCtvrtka: Keeping the list. Narrowing to collectors outside collect[] would make "disabled" mean "not requested by our current scrape jobs", which breaks for the empty-collect[] fallback you found, for --expose-exporter, and for the usual reason people disable a collector (it's slow or it hangs — it still runs). It would also couple the flag list to hr/mr/lrCollect membership, so a future addition there silently turns a real disable into a no-op with nothing to catch it.

On the 400 window: it's now handled in both directions — the forced rebuild is gated on the disabled set growing, and re-enabling falls through to the async update after the exporter restarts. The window isn't new either; textfile.hr/mr/lr have had it since before this PR.

--collector.disable-defaults is the better end state but not a swap I want here: os is in lrCollect with no --collector.os flag, so we already rely on node_exporter's defaults and that route would silently drop node_os_info. Worth a follow-up with a full audit of collect[] names against the enable list.

managed/services/inventory/agents.go
```diff
@@ -234,6 +246,19 @@ func (as *AgentsService) ChangeNodeExporter(ctx context.Context, agentID string,
...
+	// not know, so a collector that is being disabled has to leave the scrape config before the
+	// exporter restarts without it. Capture the current set to tell that direction from the opposite
+	// one, where the config must be rebuilt only after the restart.
+	var disablesMoreCollectors bool
+	if p.DisableCollectors != nil {
+		current, err := models.FindAgentByID(as.db.Querier, agentID)
+		if err != nil {
+			return nil, err
+		}
+		disablesMoreCollectors = hasNewDisabledCollectors(current.ExporterOptions.DisabledCollectors, p.DisableCollectors)
```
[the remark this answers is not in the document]
JiriCtvrtka: Confirmed the mechanism — ["cpu"] → ["meminfo"] does re-add cpu to collect[] before the restart. Not taking the two-phase fix though. Skipping the force for mixed changes doesn't help (the exporter then restarts with --no-collector.meminfo while the old config still names meminfo, so the window just moves), and the union config isn't derivable from the DB that updateConfiguration reads — it'd mean writing the union, forcing, then writing the real value, with a bogus intermediate state visible over the API. The window is one agent round trip and self-heals, same as the plain re-enable case we already leave to the async path. Eliminating the class needs the collect[]-name question settled instead, which is being discussed on the other thread.

managed/services/inventory/agents.go
```diff
@@ -215,6 +216,58 @@ func (as *AgentsService) AddNodeExporter(ctx context.Context, p *inventoryv1.Add
...
+// disabled and of what it will have disabled after the restart, so the resulting collect[] names
+// only collectors that both processes know.
+func (as *AgentsService) widenDisabledCollectors(ctx context.Context, agentID string, disabledCollectors []string) error {
+	params := &models.ChangeAgentParams{
+		ExporterOptions: &models.ChangeExporterOptions{
+			DisabledCollectors: disabledCollectors,
+		},
+	}
+
+	_, err := models.ChangeAgent(as.db.Querier, agentID, params)
```
ademidoff: **This commit lands outside `executeAgentChange`'s transaction, and nothing converges it back.**

`executeAgentChange` runs its `models.ChangeAgent` inside `as.db.InTransactionContext` (`managed/services/inventory/agents.go:1917`); the widen writes through `as.db.Querier` directly. It has to — `updateConfiguration` reads the inventory through its own querier, so the union must be committed before the forced rebuild can see it — but that makes the intermediate state durable rather than transient.

If pmm-managed dies between this commit and the real one, or is killed during the forced rebuild, the agent keeps the union set permanently: the next configuration rebuild and the next exporter restart both read it, so collectors the user never asked to disable stay disabled, with nothing to notice or correct it. The window is not instantaneous either — it spans all of `ForceConfigurationUpdate` (an inventory-wide transaction in `populateConfig`, forking the `victoriametrics` binary for the dry-run validation, then the HTTP reload) plus the change transaction. Seconds under load.

I don't think there is a clean fix inside this design, but it seems worth deciding explicitly rather than by omission: reconcile on startup, hold the widened set somewhere other than the agent row, or accept it and say so in the doc comment above, so the next reader does not assume the intermediate state cannot outlive the request.

managed/services/inventory/agents.go
```diff
@@ -234,15 +287,59 @@ func (as *AgentsService) ChangeNodeExporter(ctx context.Context, agentID string,
...
 	if err != nil {
+		if widenedFrom != nil {
+			// the widened set is already committed, so put the agent back the way the caller found it
+			restore := &models.ChangeAgentParams{
+				ExporterOptions: &models.ChangeExporterOptions{
+					DisabledCollectors: widenedFrom,
+				},
+			}
+
+			_, restoreErr := models.ChangeAgent(as.db.Querier, agentID, restore)
```
ademidoff, line 367: **The restore branch is best-effort and has no test.**

If this write fails, the error is logged and the request returns the original `executeAgentChange` error, so the agent stays on the union set — the same end state as the crash case above, and the caller cannot tell the difference. `as.vmdb.RequestConfigurationUpdate()` just below then rebuilds from whatever is stored, so a failed restore also pushes the union into the scrape config asynchronously.

None of the four new subtests makes `executeAgentChange` fail after a widen, so this branch is entirely unexercised — it is most of what codecov flags as uncovered in this file. Covering it may be awkward: a node_exporter has no `ServiceID`, so `AffectsConnection` never reaches the connection check, which leaves a DB-level failure or a `toInventoryAgent` failure as the only triggers. If there is no reasonable way to reach it from a test, that is worth a sentence in the comment — and worth asking how reachable the branch is in production too.

managed/services/inventory/agents.go
```diff
@@ -234,15 +287,59 @@ func (as *AgentsService) ChangeNodeExporter(ctx context.Context, agentID string,
...
+	// node_exporter answers the whole resolution endpoint with HTTP 400 when collect[] names a
+	// collector it does not know, so a collector being disabled has to leave the scrape config before
+	// the exporter restarts without it, while a collector being re-enabled may only enter the config
+	// after that restart. A single request can do both at once, so the configuration is first rebuilt
+	// from the union of the two sets - the collectors that stay enabled whichever process answers the
+	// scrape - and the requested list is applied afterwards. What the union holds back is picked up by
+	// the asynchronous update that agents.Handler.stateChanged triggers once the exporter is back up.
+	var widenedFrom []string
+	if p.DisableCollectors != nil {
+		current, err := models.FindAgentByID(as.db.Querier, agentID)
```
ademidoff: **`previous` is read outside any transaction, so a concurrent change yields a stale union.**

Two `ChangeNodeExporter` calls against the same agent can both read `previous` here before either writes, and each widens from its own snapshot. Say A disables `cpu` and B disables `meminfo`, both starting from an empty list: A commits union `[cpu]`, forces, commits `[cpu]`, and requests the state update; B then commits union `[meminfo]` and forces a rebuild whose `collect[]` names `cpu` again — while the exporter A just restarted has `--no-collector.cpu`. That is exactly the 400 this path exists to prevent, reintroduced through the read-modify-write.

The lost update on the list itself (B's final `[meminfo]` discarding A's `[cpu]`) is inherent to the replacement semantics of `disable_collectors` and predates this PR, so I would not hold the PR for it. The stale-union part is new, though, and it is the same family as the `victoriametrics.Service` serialization you are deferring — fine to fold into that follow-up, but worth naming there explicitly rather than leaving implicit.

ademidoff (commented):
Re-reviewed at `3fa4bfa6`. The reworked ordering looks right to me: I traced the union path end to end and the mixed disable-and-enable case really is covered now. `ChangeNodeExporterMixedCollectorChange` asserting the stored set *during* the forced rebuild is a better test than I would have asked for. And good catch on the repeated-name to repeated-flag crash in `managed/utils/collectors/collectors.go:48-67` — I missed that one entirely.

You were also right and I was wrong on the version gate: `LinuxDisabledCollectors` at 2.15.1 plus `LinuxDisabledDefaultEnabledCollectors` at 3.0.0 do bracket the cutoff from both sides. Agreed on keeping the full list and on keeping `textfile`; `--collector.disable-defaults` silently dropping `node_os_info` settles that direction for me.

Three inline comments on the new widen/restore path, all about the durability of the intermediate state rather than the ordering logic. Plus two things outside the diff:

**Neither deferred item has a ticket yet.** I searched PMM issues created since 2026-09-14 and found nothing for the `ForceConfigurationUpdate`/`reloadCh` serialization or for the collector-list drift guard. Both are reasonable to defer — they just need somewhere to live before this merges, since these threads get resolved away with the PR.

**The empty-`collect[]` fallback is still open** — `managed/services/victoriametrics/scrape_configs.go:218`. `mrCollect` is only `hwmon` + `textfile.mr`, so disabling both leaves `Params` unset, the job scrapes plain `/metrics`, and node_exporter runs *every* enabled collector at that interval. The new `--no-collector.` flags shrink that set but do not remove the fallback. Not a regression from this PR and out of its diff, but it is the one case where a user asks for fewer metrics and gets more, so it is worth a ticket next to the other two.

managed/services/inventory/agents.go
JiriCtvrtka: Medium: Agreed, and taken the third option — it's now written down instead of implied. The doc comment on widenDisabledCollectors says the union is committed rather than staged, that it has to be because ForceConfigurationUpdate reads the inventory through its own querier, and what a crash in that window leaves behind: the agent keeps the union until the next ChangeNodeExporter call for it, with nothing reconciling it on startup. Startup reconciliation is deferred to the same follow-up as the ForceConfigurationUpdate/reloadCh serialization — it's the same missing piece seen from the other side.

managed/services/inventory/agents.go
JiriCtvrtka: Low: Covered — ChangeNodeExporterRestoresWidenedCollectorsOnFailure. The branch turned out to be easier to reach than the comment suggests: an invalid custom-label name makes models.ChangeAgent fail inside executeAgentChange's transaction, after the widen has already committed. The test asserts the forced rebuild happened, the agent is back on the set the caller found, and RequestConfigurationUpdate went out.

One correction on the mechanism: it isn't the missing ServiceID that keeps the connection check away — ChangeNodeExporter hardcodes SkipConnectionCheck: true, which short-circuits before ServiceID is ever looked at. Same conclusion, but worth stating so nobody later tries to reach the branch through a service.

I've also made the best-effort nature explicit in the comment: the caller gets the error from the change it asked for, and a failed restore leaves the same state a crash would — cleared by the next successful change.

managed/services/inventory/agents.go
JiriCtvrtka: Medium: Fixed the part that's new here: the read of the stored set and the write of the union now share one transaction, so two concurrent changes can't both widen from the same snapshot. Your A/B scenario no longer produces a union computed from a set that's already been replaced.

What's left is end-to-end serialization — the union has to be committed before the rebuild reads it, so a change starting after that commit and finishing before the requested list is applied can still widen from a set about to be replaced. That's named explicitly in the doc comment and folded into the ForceConfigurationUpdate/reloadCh follow-up, as you suggested. The lost update on the list itself I'm leaving alone — it's the replacement semantics of disable_collectors and predates this.

ademidoff (approved):

PR #5843, opened 2026-08-27
PMM-15397 Hide Collect pane from read-only sessions

nachodd:
## What

On the **Support diagnostics** incident details page (`/pmm-ui/sep/atw/:incidentId`), a user whose only role is **Viewer** or **Editor** saw the entire **Collect** pane — heading, `Category` / `Subcategory` selects, and the `Snippets` autocomplete — while the execute form underneath was already withheld (`CollectPane.tsx` gates it on `canMutate`). The pane was an interactive dead end, and `CategoryBrowser` fired category/snippet `GET`s that could never lead to an execution.

Now:

- The Collect `Paper` and `CollectPane` are gated on `canMutate` at the page level. `CollectPane`'s own `canMutate` gates are untouched, so the exported component stays safe if it is ever mounted elsewhere.
- The two-pane grid drops to a single column at `md` and above when Collect is withheld, so **Results** spans the full width instead of leaving a dead column.
- An inline `ReadOnlyNotice` from `@sep/framework` sits where the pane was, so the page reads as restricted rather than broken. Same shape as the precedent in `SnippetExecutionAccordion`.
- `ResultsPane`'s empty state no longer says *"Run snippets from the Collect pane…"* to a session that has no Collect pane, following the role-aware copy pattern already in `IncidentListPage`.

## Why

Reachability is not the authorization boundary here (PMM-15358): any signed-in user may open Support diagnostics, and write controls are gated one by one on `canMutate`. The Collect pane was only partially gated.

**UI-only change.** SEP resolves a minimum role of `ADMIN` for every unsafe ATW route, so these sessions already got a 403 from the execution endpoints. No backend, permission, or `deriveCanMutate` change.

## Acceptance criteria covered

- [x] Viewer and Editor see no Collect pane: no heading, no category selects, no snippet picker.
- [x] They see a notice explaining they cannot collect diagnostics, instead of an unexplained missing panel.
- [x] Results still renders for them, full width, no empty column at `md` and above.
- [x] No category or snippet requests are issued for a read-only session.
- [x] Results' empty state does not reference a pane that is not on the page.
- [x] Admin sees the page unchanged: both panes, side by side, Collect fully functional.
- [x] `cd ui && make lint && make test` passes.

## Test plan

Automated — `ui/packages/plugins/atw/tests/`:

- `IncidentWorkspacePage.test.tsx` — Collect present and no notice for a session that may mutate; Collect heading and `Snippets` combobox absent, notice and its sentence present, Results still rendered for a non-admin; no `GET /apps/atw/` for a non-admin.
- `ResultsPane.test.tsx` — the empty state keeps the "Collect pane" wording for an admin and drops it for a read-only session.

Each new negative assertion was confirmed to fail with the source change reverted, so they are real guards rather than vacuous ones.

Ran: `cd ui && make lint` (0 errors, 0 warnings in `@sep/plugins-atw`), `make format-check`, `pnpm check-types`, `make test` (atw 101 passed; full UI suite 484 passed / 13 skipped).

Manual:

1. As an **Admin**, open an incident — both panes render, a snippet still executes.
2. Set the user's org role to **Viewer**, reload the same incident URL — no Collect pane, notice shown, Results full width, no category requests in DevTools → Network.
3. Repeat as **Editor** — identical.
4. Back to **Admin** — the pane returns.

image


fabio-silva (approved):

nachodd:

yyyyyyyan (approved):

PR #5845, opened 2026-08-27
PMM-7 Fix packages.dot drift and formatter conflict

ademidoff:
Ticket number: PMM-7

Two related bits of maintenance around `packages.dot` and the formatter chain that produced it.

---

## ⚡ `make format` goes from **~4.5 minutes to under 4 seconds**

Dropping `goimports` from the chain makes formatting the whole tree **effectively instant**. Measured back to back on one machine in a single sitting, whole tree, 1289 Go files:

| chain | wall time |
|---|---|
| **before** — gofumpt → goimports → gci (+ buf) | **3m 38.6s** |
| **after** — gofumpt → gci (+ buf) | **3.07s** |

That is a **~71x speedup**, and it lines up with the ~4.5 minutes the target cost in practice before this change. `goimports` was essentially the entire runtime of `make format`, and — as part 2 explains — the work it was doing got undone by `gci` immediately afterwards.

**`make format` is now cheap enough to run unconditionally**, which is what motivated dropping the partial `format-fast` target further down.

---

## 1. `packages.dot` drifted silently

`TestImports` rewrote `packages.dot` on every run and never asserted on it, so the committed graph drifted whenever an import changed without being regenerated.

`pmm-managed-init` gained an import of `services/clickhouse` in #5414 (`managed/cmd/pmm-managed-init/main.go:25`) and the graph was never regenerated, so anyone running the tests in that package ended up with a dirty working tree and could sweep the line into an unrelated commit.

This regenerates `managed/cmd/pmm-managed/packages.dot` and makes `TestImports` build the graph in memory and assert it matches the committed file instead of overwriting it. `pmm-agent` has the same test against `agent/packages.dot`, so both get the assertion.

After an intentional import change, regenerate with:

```
go test ./managed/cmd/pmm-managed -run TestImports -update
go test ./agent -run TestImports -update
```

The failure message names that exact command, so a stale graph is now a self-explaining test failure instead of a silent working-tree change:

```
--- FAIL: TestImports (1.64s)
        	            	--- Expected
        	            	+++ Actual
        	            	+	"/cmd/pmm-managed-init" -> "/services/clickhouse";
        	Messages:   	packages.dot is stale; regenerate it with 'go test ./managed/cmd/pmm-managed -run TestImports -update'.
```

The `-update` flag follows the existing golden-file pattern in `agent/agents/mysql/slowlog/parser/parser_test.go`. Sorting already made the output order-independent, so the dedup map became redundant once the graph was built in a buffer and was replaced with `slices.Compact`.

## 2. The formatter list had a deadlock and a redundancy

While checking the linter on the touched files, `goimports` turned out to permanently report `managed/cmd/pmm-managed/main.go` and `agent/client/client.go` as unformatted, with no way to fix it:

| state of the line | golangci verdict |
|---|---|
| no blank line before the comment (as committed) | `File is not properly formatted (goimports)` |
| blank line added | `File is not properly formatted (gci)` |

Both were enabled as formatters, so no source state satisfied the linter. Adding the blank line was not durable either: `make format` ran gci last, which stripped it straight back out, and splitting the blank import into its own `import` declaration just gets merged back by gofumpt. The committed layout was the chain's fixed point.

So this drops `goimports` and lets `gci` own import sectioning, which is what `make format` already produced in practice. Measured behaviour across deliberately mis-grouped import blocks:

| import defect | gofmt | gofumpt | goimports | gci |
|---|---|---|---|---|
| unsorted within a group | FLAG | FLAG | FLAG | FLAG |
| stdlib mixed into third-party | ok | FLAG | FLAG | FLAG |
| no blank line between stdlib/third-party | ok | FLAG | FLAG | FLAG |
| percona local mixed into third-party | ok | ok | FLAG | FLAG |
| stray extra sub-group inside a section | ok | ok | **ok** | **FLAG** |
| comment-preceded import, no blank line | ok | ok | **FLAG** | ok (forbids it) |
| correctly grouped | ok | ok | ok | ok |

`gci` catches everything `goimports` catches plus stray sub-grouping. The one rule unique to `goimports` is the last row, the one `gci` forbids, and it fires only when `local-prefixes` is set - the very feature `gci`'s `prefix(github.com/percona/pmm)` section duplicates. Net effect: enforcement gets slightly stricter, and two unsatisfiable findings go away.

`gofmt` goes too, for a simpler reason: `gofumpt` is a strict superset of it. golangci enables `gofmt`'s `simplify` by default, so the honest comparison is against `gofmt -s`, and `gofumpt` flags every case it does:

| gofmt -s simplification | gofmt (plain) | gofmt -s | gofumpt |
|---|---|---|---|
| `s[1:len(s)]` | ok | FLAG | FLAG |
| `[]T{T{1}, T{2}}` | ok | FLAG | FLAG |
| `for _ = range m` | ok | FLAG | FLAG |
| plain bad spacing | FLAG | FLAG | FLAG |

No `gofmt` settings were configured, so no custom `rewrite-rules` are lost, and `make format` never invoked it. That leaves `gci` + `gofumpt`, one tool per concern.

The `format-fast` target goes as well. It ran gofumpt alone, skipping gci and the api `buf format` pass, and existed only because plain `make format` cost ~4.5min — which, as the banner above shows, was **entirely goimports**. Now that the full pass is ~3s, a partial one saves almost nothing. Measured with real work to write rather than a clean-tree no-op:

| scenario | format-fast | format |
|---|---|---|
| clean tree (no-op) | 0.76s | 3.25s |
| 74 generated `.pb.go` dirty | 1.01s | 3.29s |
| all 1289 `.go` files dirty | 1.28s | 3.85s |

The gap is flat at ~2.5s and is fixed overhead (the `find`, process startup, buf's ~1.6s), not throughput. So the target now saves a couple of seconds in exchange for leaving imports ungrouped and protos unformatted — a state `make check` rejects. Nothing referenced it: no CI workflow, make target, script or doc.

## 3. `packages.Load` failures read as success (from review)

`packages.Load` returns a non-nil error only for catastrophic failures. A pattern that matches nothing or fails to build reports through `p.Errors`, which the test never inspected - so a broken pattern loaded "successfully" and contributed nothing:

```
github.com/percona/pmm/managed                 err=<nil>  npkgs=1 PrintErrors=1
github.com/percona/pmm/agent/agents/mongodb    err=<nil>  npkgs=1 PrintErrors=1
github.com/percona/pmm/managed/models          err=<nil>  npkgs=4 PrintErrors=0
```

Previously that meant a silently incomplete graph. With the freshness assertion from part 1 it would instead read as "packages.dot is stale" and invite `-update`, committing a wrong graph. So the load is now checked both ways - `NotEmpty` catches a `/...` pattern matching nothing (zero packages *and* zero errors), which `PrintErrors` alone does not:

```go
pkgs, err := packages.Load(config, path)
require.NoError(t, err)
require.NotEmpty(t, pkgs, "pattern %s matched no packages", path)
require.Zero(t, packages.PrintErrors(pkgs), "failed to load %s", path)
```

Sweeping all 33 patterns across both constraint lists found exactly two broken, both fixed:

- **`github.com/percona/pmm/managed`** - no Go files of its own, so this "just to add them to packages.dot" entry added no edges (the graph has no `/` node). Dropped.
- **`agents/mongodb`** - no package of its own, only subpackages, so its entire `denyPrefixes` list was unenforced. Switched to `agents/mongodb/...`. The constraint key doubles as the import-path prefix for the own-subpackage allowance, so the prefix is now derived with `strings.TrimSuffix(path, "/...")`; otherwise every intra-mongodb import would be denied. The subpackages pass the independence rules with no allowlist needed, and `agent/packages.dot` gains the 31 edges it should have had.

### Verification

- Both `.dot` files are in sync with `main`, so the new assertion passes as-is and running the tests no longer modifies the working tree.
- Dropping an edge from each `.dot` makes the matching `TestImports` fail with a diff of the missing line, and `-update` restores the file byte-for-byte.
- The internal import edges under `./managed/...` and `./agent/...` are identical for `GOOS=linux` and `GOOS=darwin` (314 edges each, no diff), so the graph does not depend on whether the tests run in CI or on a developer's machine.
- **No source churn from the formatter changes:** all 1289 Go files are already gofumpt-, gci- and `gofmt -s`-clean, and running the reduced `make format` chain over the whole tree produces zero source changes. The diff here is config-only.
- `golangci-lint` no longer reports any formatter finding on `managed/cmd/pmm-managed/...` or `agent/client/...`. The remaining findings in those packages are pre-existing and untouched.
- `go mod tidy` drops `golang.org/x/telemetry`, which `goimports` was the only consumer of. `golang.org/x/tools` itself stays - `TestImports` uses `golang.org/x/tools/go/packages`; only the `cmd/goimports` tool directive goes.
- The load guard was confirmed to fire by putting the dead pattern back: `failed to load github.com/percona/pmm/managed`. Adding the mongodb subpackages produced no deny-rule violations, only the new edges.
- End-to-end check of the reduced formatter chain: a fixture with stdlib, third-party, `github.com/percona/*`-but-not-`pmm`, and `github.com/percona/pmm` imports scrambled into one block came out of `make format` correctly split into the three sections, and byte-identical to what the old chain (with goimports) produced from the same input.
- Also fixed in passing: a pre-existing `noctx` finding at `managed/cmd/pmm-managed/main_test.go:38` (`exec.Command` -> `exec.CommandContext(t.Context(), ...)`).


managed/cmd/pmm-managed/main_test.go
```diff
@@ -181,19 +183,25 @@ func TestImports(t *testing.T) {
...
+	if *updateF {
+		require.NoError(t, os.WriteFile(dotFile, []byte(graph.String()), 0o644))
+		t.Logf("%s updated.", dotFile)
+		return
 	}
-	_, err = fmt.Fprintf(f, "}\n")
+
+	expected, err := os.ReadFile(dotFile)
 	require.NoError(t, err)
+	assert.Equal(t, string(expected), graph.String(),
```
JiriCtvrtka, line 204: packages.Load errors are never checked — a failed load now reads as "packages.dot is stale"

The real site is line 136. packages.Load returns err == nil when a pattern matches nothing or fails to build; the failure lands in p.Errors, which this test never inspects. Before, that produced a silently-wrong rewrite. Now it fails with a message telling the developer to run -update — which would commit the wrong graph. Most likely in the managed CI job, inside the pmm-server container.
```
pkgs, err := packages.Load(config, path)
require.NoError(t, err)
require.Zero(t, packages.PrintErrors(pkgs), "failed to load %q", path)
```
Two patterns already hit this today:
github.com/percona/pmm/managed (line 119) — the managed root dir has no .go files, so this "just to add them to packages.dot" entry adds nothing.
agent/agents/mongodb (agent/main_test.go:114) — no Go files, only subpackages. Its whole denyPrefixes list (agentlocal, agents/, client, config) is unenforced and it adds no edges. The rule probably wants agents/mongodb/... (mongolog, profiler, realtimeanalytics), which supervisor imports.

ademidoff: Good catch, all three confirmed. Fixed in e8191a37c.

I checked the `packages.Load` behaviour directly rather than taking it on trust, and it is exactly as you describe:

```
github.com/percona/pmm/managed                 err=<nil>  npkgs=1 PrintErrors=1
github.com/percona/pmm/agent/agents/mongodb    err=<nil>  npkgs=1 PrintErrors=1
github.com/percona/pmm/managed/models          err=<nil>  npkgs=4 PrintErrors=0
```

So both patterns loaded "successfully" while contributing nothing, and you are right that the freshness assertion turned that from a silent bad rewrite into a misleading "stale" message pointing at `-update`. Worse failure mode than before, so thanks for catching it.

I added your check plus a `NotEmpty` guard, because `PrintErrors` does not catch the other half of the problem: a `/...` pattern matching nothing at all returns zero packages *and* zero errors, which is reachable now that I use one:

```go
pkgs, err := packages.Load(config, path)
require.NoError(t, err)
require.NotEmpty(t, pkgs, "pattern %s matched no packages", path)
require.Zero(t, packages.PrintErrors(pkgs), "failed to load %s", path)
```

Verified the guard fires by putting the dead pattern back: `failed to load github.com/percona/pmm/managed`.

I also swept every pattern in both constraint lists rather than only the two you spotted - those two are the only broken ones out of 33.

**`github.com/percona/pmm/managed`** - dropped. The graph has no `/` node, so it really did add nothing.

**`agents/mongodb`** - switched to `agents/mongodb/...` as you suggested. One wrinkle: the constraint key doubles as the import-path prefix for the "allow own subpackages" check, so a literal `/...` in the key breaks self-allowance and every intra-mongodb import gets denied. Derived the prefix by trimming the suffix:

```go
own := strings.TrimSuffix(path, "/...")
```

With that, the subpackages pass the independence rules with no allowlist needed - the only outside-mongodb import is `agent/agents` itself, which the `agent/agents/` deny prefix does not match because of the trailing slash. `agent/packages.dot` gains the 31 edges it should have had, including the `supervisor -> mongolog/profiler/realtimeanalytics` edges you mentioned.

JiriCtvrtka (approved):

4nte (approved):

PR #5848, opened 2026-08-28
PMM-15274: Remove "PMM updates are disabled" error toast

mattiasimonato:
Ticket number: PMM-15274

Feature build: SUBMODULES-4549

When updates are turned off, PMM showed a red "PMM updates are disabled" toast on whatever page you happened to be on, for example Inventory. The UI asks the server to check for updates on every page load, even when the server is set up to refuse. The server replies with a 400, and our global error handler turns any 400 into a red toast. The UI then quietly retries and asks only for the installed version, which works, so nothing was really broken. The toast was just noise from a request we should never have sent.

### The fix

Stop sending that request. When updates are disabled the UI now asks only for the installed version, which the server always answers. Nothing changes when updates are enabled.

A second commit fixes a small bug in the same spot. When no check has ever run the server sends no `last_check`, and the footer printed "Last checked: Invalid Date NaN, NaN, NaN:NaN UTC" on every page. Now it just shows the version.

The "Upgrades are disabled" card on the Updates page is left alone on purpose. Making that message accurate, and greying out the toggle in Settings, needs a new API field and is covered by PMM-9042.

### Why this is not an HA bug

The ticket is filed under HA, but this happens whenever updates are off, and any deployment can do that, either with `PMM_ENABLE_UPDATES=0` or by an admin unticking **Updates** in PMM Settings. There is no HA specific code anywhere in this path.

HA is only where we found it, because the Helm chart used to set `PMM_ENABLE_UPDATES: "0"` by default, so every HA tester ran into it. The same problem was reported on plain Docker in PMM-9042 back in 2021, before PMM had HA at all.


ui/apps/pmm/src/hooks/api/useUpdates.tsx
```diff
@@ -14,14 +14,32 @@ import {
...
+        return checkForUpdates({ force: false, onlyInstalledVersion: true });
+      }
+
       try {
-        return await checkForUpdates({ force: true });
+        // 400 is "updates are disabled", which the fallback below handles. Any
+        // other failure is real and the user needs to hear about it.
+        return await checkForUpdates(
+          { force: true },
+          { disableNotifications: (error) => error.response?.status === 400 }
```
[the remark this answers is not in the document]
mattiasimonato: fixed in 1c04b03b9.

Confirmed it is reachable: this endpoint returns two different 400s. `?force=true` on a disabled server gives `code: 9` (FailedPrecondition, "PMM updates are disabled"), while a bad query param gives `code: 3` (InvalidArgument, "parsing field \"force\"..."). Matching on the status alone would have silenced both.

The predicate now checks the gRPC code in the body as well, and the test covers 400/code 3, 400 with no code, 500 and 503 all returning false.

fabio-silva (approved):

ui/apps/pmm/src/components/footer/Footer.tsx
```diff
@@ -9,16 +9,25 @@ export const Footer: FC = () => {
 
   if (!versionInfo) return null;
 
+  const { lastCheck } = versionInfo;
```
matejkubinec: I think it would be a bit cleaner to create a util for this rather than do the assignment inside the component.

Something like:
```tsx
const getCheckStatus = (versionInfo, inProgress) => {
  if (inProgress) {
    return Messages.inProgress;
  }

  if (versionInfo?.lastCheck) {
    return Messages.checkedOn(formatCheckDate(versionInfo.lastCheck));
  }

   return null;
}

// Footer.tsx
const checkStatus = getCheckStatus(versionInfo, inProgress);
```

mattiasimonato: @matejkubinec, done in 562dd1245. `getCheckStatus` now lives in `ui/apps/pmm/src/components/footer/Footer.utils.ts` with the signature you sketched, and the component is back to a single const.

Since it is a pure function now, I also added `ui/apps/pmm/src/components/footer/Footer.utils.test.ts` covering the three branches, plus a case for `formatCheckDate`, which had no direct test before.

matejkubinec (approved):

PR #5851, opened 2026-08-28
PMM-15360: Add the OpenManager settings flag

plebioda:
## What

Adds `Settings.OpenManager.Enabled` to pmm-managed and an Advanced Settings toggle for it, default off like every other technical-preview flag (Azure Discover, Access Control). `PMM_ENABLE_OM` seeds it the same way `PMM_ENABLE_NOMAD`/`PMM_ENABLE_AZURE_DISCOVER` do, and reuses `validateChangeSettingsRequest`'s existing env-var-wins precedence check rather than adding a new mechanism.

This only exposes and stores the flag; nothing reads it yet to gate OpenManager's own behavior. That's the follow-up PR (#TBD), once this and PMM-15326-om-backend / PMM-15326-om-ui-nav are all available to stack on.

Ticket: [PMM-15360](https://perconadev.atlassian.net/browse/PMM-15360)

## First of three PRs for PMM-15360

1. **This one** — the flag itself (base: this branch, the epic)
2. Gate OpenManager's service on the switch — base: `PMM-15326-om-backend` (#5816)
3. Gate OpenManager's nav and page on the switch — base: `PMM-15326-om-ui-nav` (#5818)

PRs 2 and 3 both merge this branch in (rather than depending on the non-PR `PMM-15326-om-integration` branch), so their diffs will show this PR's changes inline until it merges.

Feature build: N/A on its own — this only adds a setting nothing reads yet.

[PMM-15360]: https://perconadev.atlassian.net/browse/PMM-15360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5853, opened 2026-08-28
PMM-15360: Gate OpenManager's nav and page on the switch

plebioda:
## What

Adds the settings flag as a second gate alongside the MongoDB-monitored check for OpenManager's nav entry, reusing the existing pattern for hiding MongoDB-only nav entries rather than inventing a second mechanism.

`OmPage` now fails closed on the settings flag too, not just `isPMMAdmin`: a saved or shared link has to answer "switched off" rather than the API's raw `FailedPrecondition` or the unauthorized card, which would misreport a disabled feature as a permissions problem to an admin who has every right to be here.

Also adds the technical-preview banner (dismissible per browser via `localStorage`, not a setting), with an explicit color override on its close button — `@percona/peak-ui`'s `MuiAlert` theme colors the icon/message slots via `theme.palette[severity].contrastText` but not `.MuiAlert-action`, so the default close button was rendering almost invisibly against the warning background.

Ticket: [PMM-15360](https://perconadev.atlassian.net/browse/PMM-15360)

## Stacking

Third of three PRs for PMM-15360. Stacks on `PMM-15326-om-ui-nav` (#5818, needs `OmPage.tsx`/`navigation.provider.tsx`) **and** #5851 (needs `Settings.OpenManager.Enabled`) — both are merged in here since there's no single upstream branch with both yet, so this diff includes #5851's changes inline until that one merges. Only the last commit ("Gate OpenManager's nav and page on the switch") is net-new here.

Feature build: N/A on its own, same reasoning as #5851/#5852 — pairs with that PR to make the switch fully functional.

[PMM-15360]: https://perconadev.atlassian.net/browse/PMM-15360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/apps/pmm/src/om/OmPage.tsx
+++ b/ui/apps/pmm/src/om/OmPage.tsx
@@ -16,10 +16,26 @@
  */
 
 import { FC, PropsWithChildren } from 'react';
+import Alert from '@mui/material/Alert';
+import Card from '@mui/material/Card';
+import CircularProgress from '@mui/material/CircularProgress';
 import Stack from '@mui/material/Stack';
+import Typography from '@mui/material/Typography';
 import { Page } from 'components/page';
 import { useUser } from 'contexts/user';
+import { useReadonlySettings } from 'hooks/api/useSettings';
+import { useLocalStorage } from 'hooks/utils/useLocalStorage';
 import { OrgRole } from 'types/user.types';
+
+const TECHNICAL_PREVIEW_DISMISSED_KEY = 'pmm-ui.om.technicalPreviewDismissed';
+
+const Messages = {
+  switchedOff:
+    'OpenManager is switched off. A PMM admin can turn it on in Configuration → Settings → Advanced Settings.',
+  technicalPreview: 'Technical preview',
+  technicalPreviewBody:
+    'OpenManager is a technical preview. It is still under development and may change.',
+};
 
 /**
  * Host chrome for the OM page.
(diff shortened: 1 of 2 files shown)


plebioda:
@coderabbitai review

ui/apps/pmm/src/om/OmPage.tsx
```diff
@@ -37,17 +52,66 @@ import { OrgRole } from 'types/user.types';
...
+  );
 
   return (
     <Page
       maxWidth="full"
       roles={user?.isPMMAdmin ? undefined : [OrgRole.Admin]}
     >
       <Stack gap={3} sx={{ flex: 1 }}>
-        <div>{children}</div>
+        {settings && !settings.omEnabled ? (
```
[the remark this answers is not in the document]
plebioda: Confirmed real bug, fixed in eb0ae4e7e. Added an explicit loading state (`isLoading` from `useReadonlySettings`) and require `settings.omEnabled` explicitly, instead of the previous `settings && !settings.omEnabled` which fell through to rendering `OmApp` while `settings` was still undefined.

fabio-silva (approved):

ui/apps/pmm/src/om/OmPage.tsx
```diff
@@ -16,11 +16,27 @@
...
+import Typography from '@mui/material/Typography';
 import { Page } from 'components/page';
 import { useUser } from 'contexts/user';
+import { useReadonlySettings } from 'hooks/api/useSettings';
+import { useLocalStorage } from 'hooks/utils/useLocalStorage';
 import { OrgRole } from 'types/user.types';
 
+const TECHNICAL_PREVIEW_DISMISSED_KEY = 'pmm-ui.om.technicalPreviewDismissed';
+
+const Messages = {
```
matejkubinec, line 32: Please move messages to `OmPage.messages.ts`.

matejkubinec (commented):
Just one small nit.

ui/apps/pmm/src/om/OmPage.tsx
plebioda: Done in 842437598 - moved to ui/apps/pmm/src/om/OmPage.messages.ts, matching the sibling .messages.ts convention.

PR #5854, opened 2026-08-28
PMM-15322 Pass TLS flags to valkey_exporter

marcuscruz-percona:
Ticket number: PMM-15322, PMM-15278

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/4565

## Problem

`pmm-admin add valkey` accepts `--tls`, `--tls-skip-verify`, `--tls-ca`, `--tls-cert` and `--tls-key`, and the command succeeds, but none of them reached `valkey_exporter`. Only the `rediss://` scheme was encoded into `--redis.addr`. The certificates were shipped to the agent host as text files and then never referenced by any argument.

Monitoring a Valkey/Redis instance behind a private CA, with a certificate whose SAN does not match, or requiring client certificates therefore left `redis_up` at `0` while the CLI reported the service was added successfully.

## Scope

The fix spans both sides of the connection, so this PR carries more than the exporter arguments:

- **pmm-managed**: exporter TLS arguments, `TlsSkipVerify` in both Valkey preflight requests, private-key redaction, a Valkey case for template delimiters.
- **pmm-agent**: `GetValkeyTLSConfig` rewritten to accept CA-only and pair-only material and to apply skip-verify with no files at all — the agent-side half of PMM-15278.
- **Validation**: a half client key pair is now rejected in `pmm-admin` and in `models`, instead of being stored and silently degraded.
- **Refactor**: the TLS text file names are shared constants; the MySQL path uses them too.
- **Docs**: a mutual-TLS example and a note on the key-pair rule.

The pmm-agent change ships on a pmm-client release while the pmm-managed change ships with the server, so the preflight improvements only take effect once the client is upgraded. Both halves are kept in one PR because the server-side and agent-side behaviour only make sense together; PMM-15278 is closed by the agent half here.

### Exporter arguments

`valkeyExporterConfig` now emits `--tls-ca-cert-file` whenever a CA is present, the `--tls-client-cert-file` / `--tls-client-key-file` pair only when *both* halves are, and `--skip-tls-verification` when skip-verify is set. The pair is gated as a unit because the exporter's `validateTLSClientConfig` calls `log.Fatal("TLS client key file and cert file should both be present")` on half a pair, which crash-loops the process under pmm-agent; withholding both flags degrades the connection to server authentication only instead.

Everything is gated on TLS being enabled, matching the `rediss://` decision in `DSN()`, so nothing points the exporter at certificates over a plaintext link. `Agent.Files()` applies the same two rules at the source: with TLS disabled it returns no text files at all, and it ships the client key pair only as a unit, so neither a private key for a plaintext exporter nor half a pair ever lands on the agent host.

No `pmmAgentVersion` gate is needed: all four flags exist in `oliver006/redis_exporter` v1.72.1 — the build shipped with PMM 3.5.0, the first release with Valkey support — through the currently pinned v1.89.0.

### Key-pair validation

Half a pair is unusable everywhere — the exporter aborts on it, and the connection check cannot authenticate with it — so it is now rejected at the edges rather than accepted and worked around:

- `pmm-admin add valkey` fails with `--tls-cert and --tls-key must be specified together`.
- `ValkeyOptions.Validate()` returns `InvalidArgument` from `CreateAgent`, which covers the inventory API as well as the CLI.
- `ChangeAgent` validates the merged row **only when the change touches the pair**. `ChangeValkeyExporter` sets `ValkeyOptions` unconditionally, so validating on every call would make a row already holding half a pair permanently un-editable — impossible to disable, relabel, or repair. Such rows are reachable in the field: Valkey landed in 3.5.0 and nothing validated this before. The narrow gate still rejects every way of creating or moving into a half pair.

The unit rule in `Agent.Files()` and in the exporter arguments therefore only guards those pre-existing rows; for anything registered from now on the shape cannot occur.

### Adjacent defects fixed here

- `redactWords()` covered the MySQL and PostgreSQL private keys but not `ValkeyOptions.SSLKey`, so the Valkey private key could appear verbatim in exporter logs and API output.
- `TemplateDelimiters()` had no Valkey case. pmm-agent renders each text file's *content* as a Go template with `missingkey=error`, so a PEM containing the active delimiter broke exporter startup. All three certificates feed delimiter selection, not only the private key, because all three are rendered.
- The Valkey cases of `connectionRequest()` and `serviceInfoRequest()` sent `Tls`, `Dsn` and `TextFiles` to pmm-agent but never `TlsSkipVerify`, although the agent reads that field for Valkey and the MySQL and PostgreSQL cases already set it. Adding a TLS Valkey service with skip-verify therefore failed the pre-add connection check against a self-signed or SAN-mismatched certificate, and Inventory **Check connection** failed the same way.
- On the agent side, `GetValkeyTLSConfig` (formerly gated by `isEmptyTLSFiles`, which passed when *any* of the three files was set) called `tls.X509KeyPair` unconditionally, so a CA with no client certificates failed with `tls: failed to find any PEM data in certificate input`, and a client pair with no CA failed with `failed to append certs from PEM`. With no files at all it returned no dial options, so skip-verify never reached the dialer. Each piece of material is now applied only when present, and the dial options are always returned for a TLS connection.

### Shared constants

The text file names (`tlsCa`, `tlsCert`, `tlsKey`) are now shared constants in `models` and `api/agent/v1` rather than string literals duplicated between `Agent.Files()`, the argument builders, and the agent-side TLS helpers — renaming one used to silently stop the flags from resolving. The MySQL path (`Agent.Files()`, `mysql.go`, `tlshelpers/mysql.go`) uses the same constants, and both exporters build the `{{ .TextFiles.<name> }}` reference through one `textFileRef` helper.

### Docs

The Valkey/Redis connect page gained a mutual-TLS example, now that `--tls-cert` and `--tls-key` take effect, plus a note that the two must be given together. The example text says that `--tls-ca` already supplies the trust anchor, so a self-signed certificate does not need `--tls-skip-verify`. While there, the pre-existing examples were brought in line with the CLI: `pmm-admin add valkey` takes the service name and address as positional arguments (there is no `--address` flag), the inventory service example takes name, node ID, address and port positionally, and the exporter agent subcommand is spelled `valkey-exporter`. Every example was checked against `pmm-admin --help`.

## Testing

- `go test ./managed/services/agents/ ./managed/models/ ./agent/tlshelpers/ ./admin/commands/management/`, also under `-race -count=5`.
- Coverage added for mutual TLS, CA-only, client certificate without CA, an incomplete key pair in both directions, certificates supplied with TLS disabled (no argument and no text file), skip-verify with TLS disabled, unix socket connections, an unknown pmm-agent version, redaction of the private key (and *not* the CA or certificate), material surviving an `exposeSecrets` request, PEM content containing the active delimiter, and argument determinism.
- Validation is covered for `CreateAgent` (both directions of a half pair, each beside a CA, and each with TLS disabled) and for the material it accepts (CA alone, pair without a CA, both, and a complete pair with TLS disabled). Every case attaches a Valkey service, so the pair is the only reason a row is refused.
- `ChangeAgent` is covered against a real Valkey exporter row: moving into a half pair, clearing either half of a stored pair, clearing the pair as a unit (the CA survives), and a stored half-pair row staying editable — an unrelated change, a CA change, then the repair — which is the legacy-row case the `ChangeAgent` gate exists for.
- api-tests cover the rule on both REST surfaces. Inventory: every half-pair shape rejected with `InvalidArgument`, the accepted shapes added, and `ChangeAgent` refusing to clear one half while accepting both. Management: a half pair rejected before the connection check runs, with the following add of the same service name proving the transaction left nothing behind.
- `connectionRequest()` and `serviceInfoRequest()` are covered for Valkey with TLS and skip-verify set, asserting both flags, the `rediss://` scheme and the shipped CA reach the request; also for a legacy half-pair row, which reaches pmm-agent as the CA alone, and for material supplied with TLS off, which ships nothing and keeps the `redis://` scheme.
- `GetValkeyTLSConfig` is covered against a real TLS listener: skip-verify with no certificates accepting an untrusted server, an untrusted server rejected without it, a CA alone verifying the server, and a client key pair satisfying a server that requires mutual TLS. Added: half a client key pair failing against a server that requires one, skip-verify not standing in for a client certificate, a pair without a CA still verifying the server, and a CA for a different authority being refused. The two cases that need the server's alert during the dial pin the stub to TLS 1.2, since under 1.3 it arrives on the following read; against a real Valkey that is a failed command instead, so `redis_up` stays at `0` either way.
- Full-package failures unchanged against the base commit; `golangci-lint` clean on every changed file, and one pre-existing `exhaustive` error is resolved by the new `TemplateDelimiters` case.

Run end to end against live Valkey instances (`valkey/valkey:8.1.10`) on a dev PMM Server with a stock `percona/pmm-client:3`, hot-swapping only `pmm-managed`. The exporter binary shipped in `pmm-client:3` does accept all four flag spellings.

Before the patch, every TLS service reported `redis_up 0` and no `--tls-*` flag appeared in the exporter's argv. After it:

| Service | Valkey configuration | Emitted flags | `redis_up` |
|---|---|---|---|
| `valkey-ca` | CA-signed, `tls-auth-clients no` | `--tls-ca-cert-file` | 1 |
| `valkey-mtls` | CA-signed, `tls-auth-clients yes` | all three file flags | 1 |
| `valkey-skip` | SAN `wrong.example`, skip-verify | `--skip-tls-verification` | 1 |
| `valkey-halfpair` | CA + client certificate, no key | `--tls-ca-cert-file` only | 1 |
| `valkey-plain` | no TLS | none | 1 |
| `valkey-notls` | certificates passed, `--tls` omitted | none, `redis://` | 1 |

`valkey-mtls` runs with `tls-auth-clients yes`, so its `redis_up 1` shows the client certificate is genuinely presented and accepted, not merely that the CA verified. `valkey-skip` serves a certificate whose SAN does not match, so its `redis_up 1` shows skip-verify takes effect. `valkey-halfpair` reproduced the crash-loop described above before the pair gating (`level=fatal ... should both be present`, `AGENT_STATUS_INITIALIZATION_ERROR`, restart, repeat) and starts cleanly after it; that shape is now rejected at registration, so the run predates the validation and stands as evidence for the `Files()` guard on legacy rows.

Certificates land on the agent host as `0600` files with real PEM content. No `BEGIN PRIVATE KEY` appears in pmm-agent output, `pmm-managed.log`, or `GET /v1/inventory/agents` — though neither daemon logs exporter argv at the default level, so `redactWords` itself stays covered by unit tests rather than by this run.

## Related work

- Supersedes #5763, which reported and first fixed the same root cause. This version follows the `mysqld_exporter` pattern in `managed/services/agents/mysql.go`, avoids calling `Files()` twice, and adds the redaction and delimiter fixes.
- Closes PMM-15278: both halves of the Valkey connection-check problem are fixed here — `TlsSkipVerify` never being sent by pmm-managed, and the agent-side helper rejecting CA-only and pair-only material while dropping skip-verify when no files are set. Its acceptance criteria should be read as covering the CA-only and pair-only cases too.

## Follow-up suggestions (deliberately not addressed)

1. The inventory API (`AddValkeyExporter` / `ChangeValkeyExporter`) accepts `tls_ca` / `tls_cert` / `tls_key` with `tls=false`. The material is stored and shipped, then silently ignored — the same failure mode as this ticket, through a path the CLI cannot produce (`add_valkey.go` only reads the PEM files when `--tls` is set). Rejecting it needs the same care as the key-pair rule: `ChangeValkeyExporter` always sends `ValkeyOptions`, so a naive check would break unrelated `Change` calls on rows already stored in that shape. Left alone here; `CertificatesIgnoredWhenTLSDisabled` documents the current contract.
2. Nothing in this repository verifies an exporter's flag spellings against the shipped binary — not for `valkey_exporter`, not for `mysqld_exporter`. A typo in any of these four strings passes every unit test and reproduces `redis_up 0`. The spellings were checked by hand against the binary in `pmm-client:3` for this PR, but that check is manual; automating it needs an e2e test against a live TLS Valkey, which belongs in pmm-qa.


documentation/docs/install-pmm/install-pmm-client/connect-database/valkey-redis.md
```diff
@@ -141,6 +141,23 @@ You can add your Valkey or Redis service to PMM either through the user interfac
...
+        ```sh
+        pmm-admin add valkey \
+          --address=valkey-server.example.com:6379 \
+          --username=pmm \
+          --password=StrongPassword123! \
+          --tls \
+          --tls-ca=/path/to/ca.pem \
+          --tls-cert=/path/to/client-cert.pem \
+          --tls-key=/path/to/client-key.pem \
+          Valkey-mTLS
```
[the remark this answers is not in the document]
marcuscruz-percona: Confirmed and fixed in e855081. `AddValkeyCommand.Address` is `arg:"" optional:""`, so `--address` does not exist — only `--service-name`, `--host`, `--port`. Verified against a built binary: the old form exits with the usage dump at flag parsing, the positional form parses and reaches the agent connection.

The same bug exists in the "With remote monitoring" and "With TLS connection" examples above, and in the `inventory add service valkey` examples. Those are outside this diff; noted as a follow-up in the PR description.

documentation/docs/install-pmm/install-pmm-client/connect-database/valkey-redis.md
```diff
@@ -141,6 +141,23 @@ You can add your Valkey or Redis service to PMM either through the user interfac
...
+        ```
+
+        Add `--tls-skip-verify` if the server presents a self-signed certificate or one whose SAN does not match the address you connect to. Only do this in development or testing environments.
```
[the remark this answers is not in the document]
marcuscruz-percona: Agreed, fixed in e855081. The note now reads:

> `--tls-ca` already supplies the trust anchor, so a self-signed certificate does not need `--tls-skip-verify`. Add that flag only when a validation failure cannot be corrected — a SAN mismatch, for example — and only in development or testing: it disables server authentication entirely.

managed/services/agents/valkey.go
```diff
@@ -40,13 +40,33 @@ func valkeyExporterConfig(node *models.Node, service *models.Service, exporter *
...
+		}
+
+		// The flag names come from oliver006/redis_exporter, shipped as valkey_exporter;
+		// all four have been stable since v1.72.1, the build the first Valkey release used.
+		tlsFileFlags := []struct{ file, flag string }{
+			{models.TLSCaFileName, "--tls-ca-cert-file"},
+			{models.TLSCertFileName, "--tls-client-cert-file"},
+			{models.TLSKeyFileName, "--tls-client-key-file"},
+		}
+		for _, f := range tlsFileFlags {
```
JiriCtvrtka: The three file flags are emitted independently, so a half-configured key pair produces --tls-client-cert-file with no --tls-client-key-file (or the reverse). redis_exporter's validateTLSClientConfig does log.Fatal("TLS client key file and cert file should both be present"), so the exporter exits at startup and pmm-agent crash-loops it.

This is reachable from the normal add flow: pmm-admin add valkey --tls --tls-cert=client.pem with no --tls-key — commands.ReadFile("") returns "" without error, and nothing on the managed side validates the pair. The new ClientCertificateWithoutKey / PrivateKeyWithoutCertificate tests currently assert exactly this crash-looping output as expected behaviour.

Suggested fix: emit cert and key as a unit, and only when both are present:
```
if _, ok := textFiles[models.TLSCaFileName]; ok {
    args = append(args, "--tls-ca-cert-file="+tdp.Left+" .TextFiles."+models.TLSCaFileName+" "+tdp.Right)
}
_, hasCert := textFiles[models.TLSCertFileName]
_, hasKey := textFiles[models.TLSKeyFileName]
if hasCert && hasKey {
    args = append(args,
        "--tls-client-cert-file="+tdp.Left+" .TextFiles."+models.TLSCertFileName+" "+tdp.Right,
        "--tls-client-key-file="+tdp.Left+" .TextFiles."+models.TLSKeyFileName+" "+tdp.Right)
}
```
and flip the two tests to assert neither flag is emitted. Rejecting the incomplete pair at add time in pmm-admin / the managed API would be better still, but degrading to server-auth-only here is enough to stop the crash-loop.

marcuscruz-percona: Confirmed and fixed in 1a0be656b — you were right, and it is worse than a theoretical crash-loop. Reproduced on a live stack (dev PMM Server with the patched `pmm-managed`, stock `percona/pmm-client:3`, `valkey/valkey:8.1.10`):

```
$ pmm-admin add valkey --tls --tls-ca=/certs/ca.crt --tls-cert=/certs/client.crt \
    --skip-connection-check valkey-halfpair valkey-tls:6379
Service added.

# pmm-client log:
level=fatal msg="TLS client key file and cert file should both be present"
Process: exited early: exit status 1.
Sending status: AGENT_STATUS_INITIALIZATION_ERROR (port 42009).
... restarts on 42010, same fatal, repeat
```

Took your suggestion as written: the CA flag stays independent, the client pair is emitted only when both halves are present, so an incomplete pair degrades to server authentication instead of crash-looping. `ClientCertificateWithoutKey` and `PrivateKeyWithoutCertificate` now assert neither flag is emitted, and I added `CertificateAuthorityWithIncompleteKeyPair` to pin that the CA flag still survives that case.

One knock-on: the `DelimitersAvoidCertificateContent` subtests for `cert` and `key` each set only one half, so under the new gating they produced no flag at all. They now carry a complete pair, with the delimiter-triggering content in whichever file the case is about.

Verified on the same stack after the rebuild — the crash-loop stops, the exporter starts with `--tls-ca-cert-file` only, and `redis_up` reports `1`. Full before/after matrix is in the PR description.

On rejecting the incomplete pair at add time: I agree it is the better fix and left it out on purpose. `commands.ReadFile("")` returning `""` without error is shared by the MySQL, PostgreSQL and MongoDB paths, and the managed API would need the same guard to stay consistent, so it is a wider change than this ticket. Happy to do it here if you would rather not split it.

managed/services/agents/agents.go
```diff
@@ -118,10 +118,18 @@ func redactWords(agent *models.Agent) []string {
 	if s := agent.PostgreSQLOptions.SSLKey; s != "" {
 		words = append(words, s)
 	}
+	if s := agent.ValkeyOptions.SSLKey; s != "" {
```
JiriCtvrtka: Blocker:  the new ValkeyOptions.SSLKey line in redactWords is unconditional, so the PEM private key is sent to the agent host in SetStateRequest.RedactWords in exactly the two cases Files() deliberately withholds it. Proved it in a worktree at the PR head:


TLS off:        TextFiles=map[]         RedactWords=[secret key-pem]
CA + key only:  TextFiles=map[tlsCa:…]  RedactWords=[secret key-pem]
The guarantee the PR states at managed/models/agent_model.go:963 ("keep the private key off the agent host") is not actually held. The redaction test at managed/services/agents/valkey_test.go:326 only covers the TLS-on, full-pair case, so nothing catches it.

Fix — mirror the Files() condition:


if a := agent.ValkeyOptions; agent.TLS && a.SSLCert != "" && a.SSLKey != "" {
    words = append(words, a.SSLKey)
}
and extend PrivateKeyIsRedacted with the TLS-off and half-pair cases.

marcuscruz-percona: Confirmed and fixed in 8fefea287. You were right that the guarantee was not held — reproduced both leak cases at the previous head:

- TLS off, all three certificates stored: `TextFiles=map[]`, `RedactWords=[secret key-pem]`
- TLS on, CA + key only: `TextFiles=map[tlsCa:ca-pem]`, `RedactWords=[secret key-pem]`

`redactWords` now mirrors the `Agent.Files` condition:

```go
// The key only reaches the agent host as part of a complete pair over TLS, matching
// Agent.Files, so redacting it unconditionally would ship it in RedactWords instead.
if v := agent.ValkeyOptions; agent.TLS && v.SSLCert != "" && v.SSLKey != "" {
    words = append(words, v.SSLKey)
}
```

`PrivateKeyIsRedacted` is now table-driven over four cases — `MutualTLS` (redacted), `TLSDisabled`, `PrivateKeyWithoutCertificate` and `ClientCertificateWithoutKey` (not redacted). Checked that it actually bites: stashing the `agents.go` change makes `TLSDisabled` and `PrivateKeyWithoutCertificate` fail, and the whole set passes with it restored. `go test ./managed/services/agents/ -run TestValkey` is green, and `golangci-lint` on the package reports nothing beyond the three pre-existing hits in untouched files.

marcuscruz-percona:
@coderabbitai The outside-diff finding on `valkey-redis.md:159-160` — *"Propagate `TlsSkipVerify` to both Valkey preflight requests"* — is already fixed and predates the review.

Commit 583fc2c79 (2026-09-17) added the field to both Valkey request messages:

- `managed/services/agents/connection_checker.go:221` — `CheckConnectionRequest`, `models.ValkeyServiceType` case
- `managed/services/agents/service_info_broker.go:131` — `ServiceInfoRequest`, `models.ValkeyServiceType` case

Both now set `TlsSkipVerify: agent.TLSSkipVerify`, so pmm-agent gets the real value in `GetValkeyTLSConfig` and a server with a certificate that cannot validate no longer fails preflight before the exporter sees `--skip-tls-verification`. The add path with `--tls --tls-skip-verify` was verified end to end on a live stack; the run is recorded in the PR description.

No change needed for this one.

managed/models/agent_model.go
```diff
@@ -952,16 +960,22 @@ func (a Agent) Files() map[string]string { //nolint:gocognit
 
 		return nil
 	case ValkeyExporterType:
+		// Nothing consumes the material over a plaintext link, so keep the private key off
+		// the agent host in that case.
+		if !a.TLS {
```
JiriCtvrtka, line 971: Severity: Blocker
Problem: CA-only TLS is broken end-to-end
Desc: managed/models/agent_model.go:965 ships a TextFiles map containing only tlsCa, and GetValkeyTLSConfig in agent/tlshelpers/valkey.go unconditionally calls tls.X509KeyPair("", "") on the empty cert/key, so the dial fails with tls: failed to find any PEM data in certificate input. I reproduced this. It's a blocker not because of severity in the abstract but because it's the PR's own primary documented example and its new test fixture — the feature's headline path doesn't work as shipped. Fix is one guard: build Certificates only when both cert and key are non-empty.

marcuscruz-percona: Confirmed and fixed in 5245bd7f9.

`GetValkeyTLSConfig` built both the client certificate and the root pool from whatever the text files happened to hold, so two shipped configurations failed to dial:

- CA only — `tls.X509KeyPair("", "")` → `tls: failed to find any PEM data in certificate input`
- client key pair without a CA — `AppendCertsFromPEM("")` returns false → `failed to append certs from PEM`

Each piece is now applied only when present: `Certificates` when both cert and key are non-empty, `RootCAs` when the CA is non-empty. `isEmptyTLSFiles` still short-circuits when nothing is set, so the no-material path is unchanged, and the existing bad-PEM cases still error.

Added `ca only returns dial options` and `client key pair without ca returns dial options` to `TestGetValkeyTLSConfig`. Both fail on the previous code with exactly the two errors above.

marcuscruz-percona:
## QA pass on this PR

Ran a focused audit of the Valkey TLS path in this PR, wrote the findings up, then fixed them. Seven findings; six fixed here, one deferred.

### What was audited

The full TLS material path end-to-end: `pmm-admin add valkey` flag handling → `managed/models` registration and `Files()` → `valkeyExporterConfig` argument/text-file generation → the connection check and service-info request builders → `agent/tlshelpers.GetValkeyTLSConfig` → what `redigo` actually does with the options it is handed. Behaviour of `valkey_exporter` (`oliver006/redis_exporter`) and `gomodule/redigo` v1.9.2 was read from their sources rather than assumed.

### Findings and fixes

| # | Sev | Finding | Fix |
|---|-----|---------|-----|
| 1 | High | `GetValkeyTLSConfig` returned no dial options when no TLS files were present. `redigo` derives `DialUseTLS` from the `rediss://` scheme by itself but takes skip-verify **only** from the caller, so a TLS connection with no certificates silently fell back to full verification — `tls_skip_verify` was ignored in exactly the self-signed case it exists for | Early-return now keys off `useTLS`, not file presence. Each piece of material is applied only if actually present |
| 2 | High | Docs told users to pass `--skip-connection-check` for a situation that no longer applies | Replaced with the real constraint: `--tls-cert` and `--tls-key` come as a pair |
| 3 | Medium | A half client key pair was accepted at registration. The service got registered and then monitored with weaker authentication than was asked for | Rejected in `pmm-admin add valkey`, and server-side in `ValkeyOptions.Validate()` wired into `CreateAgent` and `ChangeAgent` — one seam covering the management, inventory and direct-API paths. `ChangeAgent` validates the merged row, not just the parameters |
| 4 | Medium | `tlsCa` / `tlsCert` / `tlsKey` were string literals duplicated across agent and managed. These names are part of the agent wire protocol — exporter args reference them as `{{ .TextFiles.<name> }}` — so a typo or rename breaks deployed agents | Defined once in `api/agent/v1` next to the `TextFiles` message; `managed/models` aliases them; `agent/tlshelpers` uses them (Valkey and MySQL) |
| 5 | Low | Both `--tls-client-cert-file` and `--tls-client-key-file` could be emitted from a half pair. The exporter's `validateTLSClientConfig` calls `log.Fatal` on that | Neither flag is emitted unless both text files shipped |
| 6 | Low | `valkeyExporterConfig` took a `*logrus.Entry` solely to warn on a half pair, on every state push | Parameter removed along with the warning, since finding 3 moves the check to registration |
| 7 | — | Out of scope, tracked separately | Not touched |

### Test coverage added

The gap worth calling out: `redis.DialOption` wraps an unexported func, so the options are not inspectable — the old tests could only assert option-slice *shape*, which is exactly why finding 1 survived review. Added a suite that dials a real loopback TLS listener through the produced options, so `redigo` itself is the assertion:

- skip-verify with no certificates accepts an untrusted server (this is the finding-1 regression guard)
- without skip-verify an untrusted server is rejected
- a CA alone verifies the server
- a client key pair satisfies a server set to `RequireAndVerifyClientCert`

I confirmed this guard actually bites by temporarily restoring the old early return: it fails with `x509: certificate signed by unknown authority`.

Also added: `TlsSkipVerify` forwarded with zero certificates on both the connection-check and service-info request builders; a `ValkeyOptions.Validate()` table; `CreateAgent`/`ChangeAgent` rejection tests; CLI flag-pair rejection tests. `agent/tlshelpers` passes `-race -count=5`.

### Known tradeoff

Dropping the per-push warning (finding 6) means a stored half pair produces no operator signal until someone edits the agent. Near-theoretical — Valkey client certificates never reached a release, so such rows can only originate from this branch — but it is a real reduction versus the previous revision. Happy to add a one-shot warning if reviewers would rather keep it.

Two things deliberately left alone: `--tls-cert` without `--tls` is still silently ignored, consistent with how every other `add` command treats its flags; and there is no positive-path CLI test, which would need a live server.

JiriCtvrtka (approved):

managed/services/agents/mysql.go
```diff
@@ -119,12 +119,12 @@ func mysqldExporterConfig(
 	if textFiles != nil && !pmmAgentVersion.IsFeatureSupported(version.MysqlExporterV0_17_2) {
 		for k := range textFiles {
 			switch k {
-			case "tlsCa":
-				args = append(args, "--mysql.ssl-ca-file="+tdp.Left+" .TextFiles.tlsCa "+tdp.Right)
-			case "tlsCert":
-				args = append(args, "--mysql.ssl-cert-file="+tdp.Left+" .TextFiles.tlsCert "+tdp.Right)
-			case "tlsKey":
-				args = append(args, "--mysql.ssl-key-file="+tdp.Left+" .TextFiles.tlsKey "+tdp.Right)
+			case models.TLSCaFileName:
```
ademidoff, line 122: I wonder if anybody would test MySQL on a Valkey-related ticket.

managed/models/agent_helpers.go
```diff
@@ -1308,6 +1324,11 @@ func ChangeAgent(q *reform.Querier, agentID string, params *ChangeAgentParams) (
 		if params.ValkeyOptions.SSLKey != nil {
 			row.ValkeyOptions.SSLKey = *params.ValkeyOptions.SSLKey
 		}
+
+		err = row.ValkeyOptions.Validate()
```
ademidoff: This validates the merged row on **every** `ChangeValkeyExporter` call, not just ones that touch the key pair.

`ChangeValkeyExporter` sets `params.ValkeyOptions` unconditionally (`managed/services/inventory/agents.go:820`), even when the caller only flips `enable` or edits custom labels — so `params.ValkeyOptions != nil` is always true and this runs against whatever is already stored.

For a row already holding half a pair, that makes the agent permanently un-editable: you can't disable it, can't change its labels, can't change its log level. And such rows are reachable in the field today — `AddValkeyExporterParams` ships `tls_ca` / `tls_cert` / `tls_key` as plain strings (`api/inventory/v1/agents.proto`), Valkey landed in 3.5.0, and there was no validation before this PR.

This is the same failure mode follow-up 1 in the description explicitly declines to introduce: *"Rejecting it with `InvalidArgument` would make later, unrelated `Change` calls fail on any row already stored in that shape."* The reasoning applies verbatim here.

Narrowest fix — only validate when the change actually touches the pair:

```go
if params.ValkeyOptions.SSLCert != nil || params.ValkeyOptions.SSLKey != nil {
	if err := row.ValkeyOptions.Validate(); err != nil {
		return nil, err
	}
}
```

That still rejects every way of *creating* or *moving into* a half pair, while leaving a legacy row editable (including editable back into a valid shape). `CreateAgent` is fine as written.

---

agent/tlshelpers/valkey.go
```diff
@@ -24,40 +24,46 @@ import (
 	agentv1 "github.com/percona/pmm/api/agent/v1"
 )
 
-// GetValkeyTLSConfig returns TLS config for Valkey connections.
+// GetValkeyTLSConfig returns the dial options for a Valkey connection, or nil when TLS is off.
 func GetValkeyTLSConfig(files *agentv1.TextFiles, useTLS, tlsSkipVerify bool) ([]redis.DialOption, error) {
-	var opts []redis.DialOption
+	if !useTLS {
```
ademidoff, line 29: Anchoring a PR-level point here because this file is the clearest evidence of it: **the description has drifted from the code over the last 24 commits, and three of its claims are now false.**

1. It says the agent half stays on PMM-15278 (follow-up 3) — but this file is rewritten in this PR, and that rewrite *is* the agent half.
2. It describes a warning-based design: `ValkeyClientKeyPairIncomplete` on the model, and *"pmm-managed logs a warning naming the agent, so the downgrade is visible in `pmm-managed.log`."* Neither exists at head; registration now rejects instead, via `ValkeyOptions.Validate()`. The QA comment confirms the warning was dropped deliberately (finding 6).
3. The docs note is described as telling users to pass `--skip-connection-check`; it was replaced by the key-pair note.

Worth a rewrite before merge — PMM squash-merges, and this is what reviewers read first.

Same place, a scope note: this is no longer just "pass TLS flags to valkey_exporter." It also carries the PMM-15278 agent-side fix, new validation across the CLI and models, and the MySQL constant refactor — and the agent half ships on a different vehicle (a pmm-client release) than the managed half. Keeping them together is defensible, but the description currently says the opposite, so the ticket links should reflect what's actually here.

---

ademidoff (commented):
Reviewed at `14a46d7`. The core fix is correct — the exporter-argument gating, `TlsSkipVerify` propagation, the `TemplateDelimiters` Valkey case and the MySQL constant refactor all check out, the docs examples match the Kong structs, and the branch merges cleanly into current `main` despite the "behind" marker. Two actionable items inline; everything else I had was nits.

managed/services/agents/mysql.go
marcuscruz-percona: Fair point — the constant refactor touches two MySQL paths, so I re-ran them rather than assuming.

Both are already asserted:

- `mysql_test.go:274-276` — `--mysql.ssl-ca-file={{ .TextFiles.tlsCa }}`, `--mysql.ssl-cert-file={{ .TextFiles.tlsCert }}`, `--mysql.ssl-key-file={{ .TextFiles.tlsKey }}` (the pre-v0.17.2 branch this hunk sits in).
- `mysql_test.go:475` and `:541` — the `myCnf` text file, asserted verbatim with `ssl-ca={{ .TextFiles.tlsCa }}\nssl-cert={{ .TextFiles.tlsCert }}\nssl-key={{ .TextFiles.tlsKey }}`, which covers the `myCnfParams.CaFile/CertFile/KeyFile` assignments at `mysql.go:280-286`.

`go test ./managed/services/agents/ -run 'MySQL|Mysql' -count=1` → 22 tests pass; the full package is 224 pass. A typo in any of the three constants fails those assertions, so the rename is guarded. No code change.

I have also asked QA to include MySQL TLS in the test scope on the ticket, since the shared constants and `textFileRef` now sit on the MySQL path.

managed/models/agent_helpers.go
marcuscruz-percona: Confirmed and fixed — `ChangeValkeyExporter` sets `params.ValkeyOptions` unconditionally (`managed/services/inventory/agents.go:820`), so the merged-row check ran on every call and a row already holding half a pair could never be disabled, relabeled, or repaired. That is exactly the failure mode follow-up 1 declines to introduce.

Applied your narrowest fix:

```go
// Only the change that touches the key pair is validated: rows stored before this
// validation existed may already hold half a pair, and validating the merged row on
// every call would make them permanently un-editable, including back into a valid shape.
if params.ValkeyOptions.SSLCert != nil || params.ValkeyOptions.SSLKey != nil {
	err = row.ValkeyOptions.Validate()
	if err != nil {
		return nil, err
	}
}
```

`CreateAgent` is unchanged, and the existing `ChangeAgentRejectsIncompleteValkeyKeyPair` still passes — moving *into* a half pair is still rejected.

New test `ChangeAgentKeepsStoredIncompleteValkeyKeyPairEditable` inserts a Valkey row holding `SSLCert` only (bypassing validation, as a pre-3.x row would), then:

1. flips `Enabled` with an empty `ChangeValkeyOptions{}` — what `ChangeValkeyExporter` sends for an `enable`-only call — and asserts it succeeds with the stored cert intact;
2. supplies the missing `SSLKey` and asserts the row repairs to a valid pair.

`go test ./managed/models/ -run TestAgentHelpers` → 46 pass; `./managed/services/agents/` → 224 pass. `golangci-lint ./managed/models/...` shows no new findings (`ChangeAgent` gocognit was already 128, well over the limit, and is nolint'd for cyclop/maintidx).

The description now describes the validation design that is actually at head, including this gate.

agent/tlshelpers/valkey.go
marcuscruz-percona: All three claims check out against head, and the description is rewritten.

1. `agent/tlshelpers/valkey.go` is rewritten here — `isEmptyTLSFiles` is gone, each piece of material is applied only when present, and dial options are always returned for a TLS connection so skip-verify reaches the dialer with no files. That is the agent half of PMM-15278, not a follow-up.
2. `ValkeyClientKeyPairIncomplete` does not exist (`grep` over `*.go` is empty); the rule is `ValkeyOptions.Validate()` / the unexported `clientKeyPairIncomplete()`, and registration rejects rather than warns. No warning is logged anywhere.
3. The docs note at `valkey-redis.md:159-160` is the key-pair note; there is no `--skip-connection-check` note left on the page.

The rewritten description:

- New **Scope** section up front, naming both sides: pmm-managed (exporter args, `TlsSkipVerify`, redaction, delimiters), pmm-agent (`GetValkeyTLSConfig`), validation across `pmm-admin` and `models`, the shared-constant refactor incl. MySQL, and docs. It states plainly that the agent half ships on a pmm-client release while the managed half ships with the server, and why they are kept together.
- Ticket line is now `PMM-15322, PMM-15278`, and **Related work** says PMM-15278 is closed by the agent half here (with the note that its acceptance criteria should be read as covering CA-only and pair-only).
- The warning-based paragraph is replaced by a **Key-pair validation** section describing what is actually there: CLI rejection, `CreateAgent` rejection, and the `ChangeAgent` gate from your other comment. The `Files()` unit rule is now described as a guard for pre-existing rows only.
- Follow-up 3 is dropped; follow-up 1 is reworded so it no longer leans on reasoning this PR has since superseded.
- The **Testing** section drops the warning assertions, adds the validation and legacy-row cases, and adds the `GetValkeyTLSConfig` handshake tests against a real TLS listener.

marcuscruz-percona:
## MySQL TLS live verification

@ademidoff you asked whether MySQL was tested — the FB run linked on the ticket predates the MySQL changes here (`mysql.go` at `b32719af6`, `tlshelpers/mysql.go` at `aba63c774`) and its MySQL jobs don't use `--tls`. I ran it live myself against this PR's head (`e44b2f591`).

**Setup:** local dev PMM Server (`make env-up`), `pmm-managed` and `pmm-agent` rebuilt from this branch and hot-swapped in (no image rebuild). A `percona:5.7` container with `--require_secure_transport=ON`, using its auto-generated self-signed certs. Confirmed a plain TCP connection is actually rejected (`ERROR 3159 ... insecure transport ... prohibited`) before testing PMM against it.

**Results:**

| Check | Result |
|---|---|
| `pmm-admin add mysql --tls --tls-ca --tls-cert --tls-key` | Registered. First attempt hit a genuine SAN mismatch on the self-signed cert — needed `--tls-skip-verify`, which is the scenario PMM-15322 is about |
| `mysqld_exporter` process args | `--tls.insecure-skip-verify` present; `--config.my-cnf` points at a real file with `ssl-ca` / `ssl-cert` / `ssl-key` resolving to real paths under `/srv/pmm-agent/tmp/mysqld_exporter/<id>/` |
| `mysql_up` (queried via VictoriaMetrics) | `1` |
| MySQL `Ssl_accepts` status counter | 61 — real TLS handshakes, not a plaintext fallback |
| QAN, `--query-source=perfschema` | Agent reached `AGENT_STATUS_RUNNING`; qan-api2 logged `Saving 23 MetricsBucket(s)` / `Saving 82 MetricsBucket(s)`; ClickHouse `pmm.metrics` shows 23 rows for the service |

```
$ ps aux | grep mysqld_exporter
... --tls.insecure-skip-verify ... --config.my-cnf=/srv/pmm-agent/tmp/mysqld_exporter/d6bb.../myCnf ...

$ cat .../myCnf
[client]
host=host.docker.internal
port=3307
user=pmm
ssl-ca=/srv/pmm-agent/tmp/mysqld_exporter/d6bb.../tlsCa
ssl-cert=/srv/pmm-agent/tmp/mysqld_exporter/d6bb.../tlsCert
ssl-key=/srv/pmm-agent/tmp/mysqld_exporter/d6bb.../tlsKey

$ curl .../metrics?query=mysql_up -> value: 1

$ curl clickhouse "SELECT service_name, count() FROM pmm.metrics WHERE service_name='MySQL-TLS' GROUP BY service_name"
MySQL-TLS	23
```

Not a regression, worth recording: QAN `slowlog` mode couldn't work in this rig, because it reads the slow-log file directly off disk and here MySQL was in a separate container from pmm-agent — an environment limitation of this test setup, not a code defect. Switching to `perfschema` (same TLS connection, no file access) exercises the same shared `tlsCa`/`tlsCert`/`tlsKey` constants and confirms the path end to end. Also hit an unrelated pre-existing panic in `qan-api2`'s `GetReport` from a malformed test request of mine (missing `columns`) — not part of this PR's diff, not investigated further.

Also rebuilt the FB at the current head: [pmm-submodules run 36029721705](https://github.com/Percona-Lab/pmm-submodules/actions/runs/36029721705), building `0cd77ea5f`.

Filed the same evidence on PMM-15322.

PR #5865, opened 2026-08-31
PMM-15351: Paginate state timeline panels

mattiasimonato:
Ticket number: PMM-15351

Feature build: SUBMODULES-4552

A state timeline shares the panel height between all of its rows. The more rows, the thinner each one gets, until the names on the left print on top of each other and the panel is unreadable.

Seven panels grow like that, because they draw one row per user, server, endpoint, host or node:

- **Users Activity** (MySQL User Details)
- **Servers in ... Section** and **Sessions in ... Section** (HAProxy Instance Summary)
- **Endpoint Statuses** (ProxySQL Instance Summary)
- **Backup agent status over time** (MongoDB Backup Details)
- **Replication Members State** (PostgreSQL Replication Overview)
- **Router Status** (MongoDB Router Summary)

### The fix

Each panel now sets Grafana's **Page size**, the same option PMM-15060 used on the Uptime panel. It caps how many rows one page shows, so the rows keep their height however many series show up. The caps are 20, 10, 10, 8, 8, 6 and 4, based on how tall each panel is and checked on a live server.

**Router Status** and **Replication Members State** are short panels, so even split into pages their rows stayed under 20px. Hiding the legend hands that space to the rows: 18px to 28px on Router Status, 19px to 25px on Replication Members State. Neither panel needs the legend, because both already write the state inside the bar. The other five keep theirs.

No query, layout or panel changes.

before:
state-timeline-before

after:
state-timeline-after


fabio-silva:
@mattiasimonato seems good, can we just have some screenshots (one panel is enough)?

fabio-silva (approved):

mattiasimonato:
> @mattiasimonato seems good, can we just have some screenshots (one panel is enough)?

@fabio-silva, I added a before/after of the Users Activity panel to the PR description 👍

matejkubinec (approved):

PR #5874, opened 2026-09-02
PMM-15405 Drop the Support diagnostics setup gate

nachodd:
Stacked on #`PMM-15397-hide-collect-panel-for-non-admins` — review that one first; this PR's diff is only its own commit.

## What

`ServiceNowSetupGate` replaced the whole Support diagnostics app with a setup screen unless the ServiceNow connection read as configured. Because SEP holds `GET /sep/admin/settings` to administrators (reads included), a non-admin's read was skipped and the gate fell open, while an administrator's read succeeded, returned not-configured, and the gate closed. The less-privileged role saw the complete app; the operator who could actually fix the connection could not read the history explaining why they should.

Per the decision on PMM-15381 (option 2, decoupled), the gate is gone:

- **Reads are unconditional.** Incident list, incident detail, executions' captured output and send history render whatever the connection state, for every role. Admin and non-admin now see the same reads for the same state.
- **The send control keeps its copy and stays disabled**, as it already was via the backend's `send_disabled_reasons`. Above it, a one-line notice states that sending requires a valid ServiceNow connection, with a button beside it linking to the ServiceNow settings tab.
- **A read-only session gets no notice.** `canMutate` already withholds the entire send toolbar from Viewers and Editors, so the connection state changes nothing they could do — a banner there would be noise about a capability the role never had. (This is the open question the ticket flagged; decided this way.)
- **Nothing keys on a settings read failing** any more. The notice is driven by the app's own `/config/` probe, which reports no reasons on error, leaving the POST's 503 as the real guard.

## The seam

`packages/plugins/atw` is synced one-way from percona/SEP and cannot know where the host parks its connection form — PMM has a dedicated settings tab, SEP has only a generic settings list. So the route arrives from the host: `AtwApp` gains `deliverySettingsPath`, passed down through a small context in `deliverySettings.tsx`. PMM's router passes `PMM_SERVICENOW_SETTINGS_PATH`; a host that passes nothing gets the explanation without the link. Both files document why, so a future SEP sync does not quietly revert it.

Also removed as dead once the gate went: `SUPPORT_DIAGNOSTICS_DOCS_URL`, and `useServiceNowConnection`'s `enabled` option (its only non-admin caller was the gate).

## Acceptance criteria

- [x] Administrator with no connection reaches the incident list, incident detail, executions' output and send history
- [x] Send control rendered but disabled, original copy, one-line notice above it, settings button beside the notice
- [x] Viewer/Editor see no route to Settings
- [x] Full-page setup gate gone from both paths
- [x] Admin and non-admin see the same reads for the same connection state
- [x] Gate behaviour no longer depends on a settings read failing
- [x] No read became unreachable; no write became reachable to a role that could not perform it before

## Test plan

`ui/packages/plugins/atw/tests/ResultsPane.test.tsx` — 7 new cases: notice with the host link, notice without one, silent while configured, withheld from a read-only session (which also has no send button), shown for a failed send history on an incident with zero executions, and silent when the `/config/` read itself fails while the send stays enabled.

`ui/packages/plugins/atw/tests/AtwApp.test.tsx` (new) — mounts the app the way the shell does and proves the `deliverySettingsPath` prop reaches the rendered link, so a wiring regression cannot hide behind tests that supply the context by hand.

`cd ui && pnpm lint` (0 errors), `pnpm format:check`, `pnpm build`, `pnpm test` — 110 atw + 475 pmm tests pass.

Code was reviewed for correctness and design before opening; the review's test-coverage gaps are the AtwApp wiring test and the two edge cases above.

## Screenshots

Admin, ServiceNow not configured: 
image
Hovering over the Send button shows the tooltip: "Diagnostics delivery is not configured"

Admin, ServiceNow configured: 
image
User has to select at least 1 result for the button to enable

Non Admins (either ServiceNow configured or not): 
image
Can access the list of incidents, and the detail. But can't take action (stays the same as before).


yyyyyyyan (approved):

yyyyyyyan:
@copilot+gpt-5.6-sol resolve the merge conflicts in this pull request

PR #5579, opened 2026-07-01
PMM-14912 Dynamic thresholds

matejkubinec:
PMM-14912

Grafana: https://github.com/percona/grafana/pull/912
FB: https://github.com/Percona-Lab/pmm-submodules/pull/4449

## Summary by CodeRabbit

* **New Features**
  * Added per-node alert threshold overrides—customize alert sensitivity for individual nodes without modifying templates.
  * New management UI for viewing and editing node-specific thresholds with default value restoration.
  * Introduced overridable parameters in alert templates; select existing templates (MySQL connections, CPU load) now support per-node customization.
  * Added PostgreSQL table bloat alert with configurable thresholds.
  * Exposed effective alert thresholds via debug metrics for monitoring.

* **Documentation**
  * Added implementation plan for dynamic alert thresholds architecture.

PR #5878, opened 2026-09-02
PMM-14912 Dynamic thresholds

matejkubinec:
PMM-14912

Grafana: https://github.com/percona/grafana/pull/912
FB: https://github.com/Percona-Lab/pmm-submodules/pull/4449

Replaces #5579, which reviewed the POC. This branch reimplements the feature
on top of current `main`, with a scope-based API rather than a node-centric one.

## What

Per-target alert threshold overrides. An operator can raise or lower the
threshold of an existing alert rule for one node, without editing the template
or duplicating the rule.

## How

PostgreSQL is authoritative. Two tables back the feature: a registry row per
generated rule, and override rows keyed by `(rule_id, param_name, scope, target)`.
VictoriaMetrics carries only what has been overridden, as the gauge
`pmm_alert_threshold_override`.

The Grafana rule is written once at creation and never rewritten. An overridable
parameter becomes its own query step, `T_<param>`, which reads the gauge and
falls back to the rule's own observed query for the default:

    max by (<join>) (label_replace(pmm_alert_threshold_override{...}))
      or (max by (<join>) (<observed expr>) * 0 + <default>)

Reusing the rule's own observed expression for the fallback is deliberate - it
makes the threshold share the rule's fate. If the scrape target disappears, the
threshold series disappears with it rather than outliving the data it guards.

Clearing an override writes a tombstone instead of deleting the row, so the
emitted series changes value rather than vanishing. Measured on a live server,
this takes a clear from ~5 minutes down to ~12 seconds.

Precedence is `service` -> `node` -> `cluster`, resolved in Go. It cannot be
expressed in PromQL: reducing both sides of an `or` to a common label set is what
makes `or` prefer the left operand, and that reduction destroys the scope
information needed to rank by.

## Scope of this PR

Node scope only. The schema, resolver and proto already carry service and
cluster, but those RPCs return `Unimplemented` - service scope needs a
services-by-cluster lookup that is not in place yet.

Single-expression templates are desugared into the same three-step shape at build
time, so they can carry an overridable parameter too. The path is inert until a
template opts in; only `node_high_cpu_load` does so far.

The threshold RPCs are admin-gated. The UI lives under inventory, where editors
have no access, so inheriting `viewer` from `/v1/alerting` would have been wrong.

## Testing

- Unit tests across `managed/models`, `managed/services/alerting`, `managed/pi/alert`
- API tests in `api-tests/alerting/thresholds_test.go`
- Verified on a live server: clear latency, and threshold survival across a
  21-minute simulated scrape outage
- `make prepare-pr` clean

UI browser verification is still outstanding.


managed/services/alerting/service.go
```diff
@@ -791,12 +831,100 @@ func (s *Service) CreateRule(ctx context.Context, req *alerting.CreateRuleReques
...
+		}
+
+		if len(param.Range) != 0 {
+			pMin, pMax, err := param.GetRangeForFloat()
+			if err != nil {
+				return nil, fmt.Errorf("parameter %q: failed to parse range: %w", param.Name, err)
+			}
+
+			snapshot.Min, snapshot.Max = new(pMin), new(pMax)
+		}
```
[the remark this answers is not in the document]
matejkubinec: Verified on this branch: it compiles. `new` accepting a value argument is a Go 1.26 language feature, and `go.mod` declares `go 1.26.6`.

`go build ./managed/services/alerting/` and `go vet ./managed/services/alerting/... ./api-tests/alerting/...` both exit 0 — the second compiles the test files carrying `new(true)` and `new(scope)`, the cases called out above — and CI is green on these commits.

No change made.

fabio-silva (approved):

managed/services/alerting/reconciler.go
```diff
@@ -0,0 +1,136 @@
...
+	}()
+}
+
+// ReconcileAlertRules deletes registry rows for rules that are no longer in Grafana,
+// taking their threshold overrides with them through the foreign key.
+//
+// Rules are matched by the identity label PMM stamps on them rather than by Grafana UID,
+// so a rule that was copied or renamed still counts as present.
+func (s *Service) ReconcileAlertRules(ctx context.Context) error {
+	live, err := s.grafanaClient.ListPMMRuleIDs(ctx)
```
ademidoff, line 93: This treats the caller's ruler view as the complete set of rules. `/api/ruler/grafana/api/v1/rules` called with the caller's auth headers only returns rules in the caller's current org, and only from folders they can read.

Scenario: PMM rules and their overrides live in org 1. An admin whose active org is org 2 (or an org-2 service-account token) calls `GET /v1/alerting/thresholds`. `live` then contains only org-2 rules, so every org-1 registry row older than the grace period is deleted, and the FK cascade deletes all of its overrides with it. The rules keep running on their defaults, and there is no way to get the overrides back.

The sweep needs a view that is known to be complete, e.g. listing with credentials that see every org and folder. Alternatively, it could only reap a row after confirming that its specific rule is gone.

---

managed/services/alerting/rule_builder.go
```diff
@@ -207,3 +336,186 @@ func parseAlertTemplate(yamlContent string) (*alert.Template, error) {
...
+// share the observed data's fate: it cannot go missing while the data it guards is still
+// arriving, so a rule cannot silently stop evaluating.
+//
+// `max by` is load-bearing in both clauses. It strips instance and job - the threshold is
+// scraped from pmm-managed, which can never match an observed series - reduces both
+// operands to identical label sets so `or` prefers the left, and collapses the duplicate
+// series an HA cluster emits.
+func thresholdQueryExpr(ruleID, paramName, joinLabel, observedExpr, defaultValue string) string {
+	return fmt.Sprintf(
+		`max by (%s) (label_replace(%s{%s=%q, %s=%q}, %q, "$1", %q, "(.*)")) or (max by (%s) (%s) * 0 + %s)`,
```
ademidoff, line 420: `max by (join) (observed) * 0 + default` is NaN for any target whose observed value is ±Inf, because `Inf * 0` is NaN. A target without an override then has no usable threshold: the NaN sample is dropped, or `Inf > NaN` evaluates to false. Either way the rule reports Normal for exactly the node in the worst state, e.g. a ratio that divides by zero. Before this change, `$A > 80` would have fired for it.

`group by (join) (observed) * default` yields the default whatever the observed value is, so it avoids this.

---

managed/pi/alert/overridable.go
```diff
@@ -0,0 +1,191 @@
...
+
+	binary, ok := parsed.(*parser.BinaryExpr)
+	if !ok || !binary.Op.IsComparisonOperator() {
+		return zero, errors.New("an overridable parameter must be the right-hand side of the expression's top-level comparison")
+	}
+
+	// Anything other than the bare sentinel means the token was used inside a larger
+	// expression, e.g. `foo > [[ .threshold ]] * 100`, which cannot become a threshold step.
+	number, ok := binary.RHS.(*parser.NumberLiteral)
+	if !ok || number.Val != 0 {
```
ademidoff, line 140: This accepts any literal 0 on the right-hand side, not specifically the sentinel. That lets some templates pass validation and then get desugared into a different comparison:

- `(foo - [[ .threshold ]]) > 0`: the sentinel replaces the token inside the LHS, the top-level RHS is a real `0`, and the check passes. The sliced LHS still contains the token, so with a default of 80 the rule becomes `(foo - 80) > T` with T = 80. That means `foo > 160` instead of `foo > 80`.
- `foo < -[[ .t ]]`: the parser folds the unary minus into the literal, which gives -0, and -0 equals 0. The sign is lost, and the rule compares against +80 instead of -80.

Checking that `number.PositionRange()` covers exactly the token's location in `expr` would catch both cases.

---

managed/services/alerting/service.go
```diff
@@ -791,12 +833,101 @@ func (s *Service) CreateRule(ctx context.Context, req *alerting.CreateRuleReques
...
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to register alert rule: %w", err)
+		}
+	}
+
 	err = s.grafanaClient.CreateAlertRule(ctx, req.FolderUid, req.Group, interval, &rule)
 	if err != nil {
+		// Best effort: a lost response drops the registry row of a rule that exists.
+		s.deleteRuleRegistration(ruleID)
```
ademidoff, line 856: The comment acknowledges this, but the damage is permanent. Suppose Grafana created the rule and only the response was lost (a timeout, ctx cancelled after the POST, or a connection reset). The rule is then live with a `pmm_rule_id` label but has no registry row. Every later `SetThreshold` or `BatchUpdateThresholds` call for it returns NotFound. The reconciler only ever deletes rows, so nothing repairs this.

The reconciler already reaps rows whose rule never made it into Grafana, once the grace period has passed. So it looks safe to skip the delete here and leave both cases to the reconciler, assuming the org-scoping issue in `reconciler.go` is fixed.

---

managed/services/alerting/rule_builder.go
```diff
@@ -207,3 +336,186 @@ func parseAlertTemplate(yamlContent string) (*alert.Template, error) {
...
+	}
+
+	injections := make([]thresholdInjection, 0, len(overridable))
+	for _, param := range overridable {
+		joinLabel, err := joinLabelForParam(param)
+		if err != nil {
+			return nil, fmt.Errorf("parameter '%s': %w", param.Name, err)
+		}
+
+		observed, err := template.ObservedQueryForParam(param.Name)
```
ademidoff, line 372: In a multi-expression template, `swapOverridableTokens` replaces the token in every expression. `ObservedQueryForParam`, however, stops at the first expression that mentions it, so `T_<param>` fans out over that expression's query only.

Take `C: $A > [[ .t ]]` and `D: $B > [[ .t ]]`, where `$B` has a different label set from `$A`. `D` becomes `$B > $T_t` with no matching series, so it gets no threshold. `validateOverridableParams` only resolves the first expression, so the template still loads.

This could be fixed in either of two ways:
- Reject templates where a parameter's token appears in several expressions with different observed queries.
- Build one threshold step per (param, observed query) pair.

---

managed/services/alerting/threshold_metrics.go
```diff
@@ -0,0 +1,261 @@
...
+		if !ok {
+			continue
+		}
+
+		param, ok := rule.Params[group.paramName]
+		if !ok {
+			continue
+		}
+
+		for target, value := range models.ResolveThresholds(group.overrides, param.Default, inv) {
```
ademidoff, line 158: Tombstoned targets resolve to `param.Default`, which is the snapshot taken when the rule was created. The tombstone series sits on the left of the `or`, so it wins over the rule's own default clause.

If someone edits the rule in Grafana and changes `* 0 + 80` to `* 0 + 90`, untouched nodes move to 90. Every node whose override was ever cleared, though, stays pinned at 80 through its tombstone, and `ListThresholds` keeps reporting 80 as well. That is the pinning that the `buildThresholdUpdates` comment in the UI says clearing is meant to avoid.

---

managed/models/alert_rule_helpers.go
```diff
@@ -0,0 +1,315 @@
...
+	// Raw SQL skips reform's hooks.
+	err = override.AfterFind()
+	if err != nil {
+		return nil, err
+	}
+
+	return override, nil
+}
+
+// ClearThresholdOverride tombstones an override instead of deleting it, so the emitted
```
ademidoff, line 239: The PR description says a delete took about 5 minutes to take effect and a tombstone takes about 12 seconds. I'd like to understand why the delete was that slow. PMM doesn't set `-promscrape.noStaleMarkers`, so when the series disappears from the next pmm-managed scrape, the scraper should write a staleness marker, and the left side of the `or` should drop out right then. If staleness markers aren't taking effect here, it seems worth finding out why before building on tombstones.

Tombstones have ongoing costs. A row is only removed when its node, service or rule is deleted. So every target an admin has ever reset keeps a row, and keeps emitting a `pmm_alert_threshold_override` series for the lifetime of the rule. That series is pinned to the snapshot default (see the comment on `threshold_metrics.go`). If a delete can be made to take effect at the next scrape, both problems go away.

---

managed/services/alerting/reconciler.go
```diff
@@ -0,0 +1,136 @@
...
+// maybeReconcile runs the sweep at most once per reconcileInterval, off the goroutine
+// serving the request that triggered it.
+//
+// A request handler is the trigger because it's the one place with a real, authenticated
+// context to give ListPMMRuleIDs - a background ticker never gets one (see
+// auth.GetHeadersFromContext). The metadata is copied onto a fresh, longer-lived context
+// so the sweep survives the triggering request returning its response.
+//
+//nolint:contextcheck // the sweep intentionally does not inherit ctx: it must outlive the request that triggered it
+func (s *Service) maybeReconcile(ctx context.Context) {
```
ademidoff, line 55: Some concerns with triggering the sweep from `ListThresholds`:

- Orphans are only reaped when an admin happens to open the thresholds modal. Until then, rules deleted in Grafana keep their registry rows and overrides. The collector also keeps emitting their series, because it checks only the registry and not Grafana. The comment at line 29 says it emits nothing for a rule that is gone, which isn't the case.
- A GET endpoint becomes destructive. Any tool or prefetch that calls it can trigger hard deletes, and those deletes are based on the caller's view (see the comment at line 93).
- The goroutine isn't tied to service shutdown. As the comment at line 41 notes, the DB transaction has no deadline either, so a sweep can still be running against a closing DB during shutdown.

---

managed/pi/alert/overridable.go
```diff
@@ -0,0 +1,191 @@
...
+		return zero, fmt.Errorf("parameter '%s' is not referenced in the expression", paramName)
+	}
+
+	// "0" padded with spaces to the token's exact byte length: parseable, and leaves every
+	// following position unchanged.
+	sentinel := "0" + strings.Repeat(" ", len(token)-1)
+	probe := strings.Replace(expr, token, sentinel, 1)
+
+	// Default options: templates are ordinary PromQL, so nothing experimental is enabled.
+	parsed, err := parser.NewParser(parser.Options{}).ParseExpr(probe)
```
ademidoff, line 127: Templates are evaluated as MetricsQL, but this uses the upstream Prometheus parser, so valid MetricsQL is rejected. `singleexpr_test.go` already runs into this with `pmm_mongodb_replication_lag`, whose `max(...[1m]) by (...)` it calls a pre-existing bug. That template ships and evaluates today, because MetricsQL accepts it. `WITH` expressions and the `default`/`if`/`ifnot` operators would fail the same way. Marking any such template overridable makes it fail to load with "failed to parse expression".

One option is the MetricsQL parser (`github.com/VictoriaMetrics/metricsql`), if it provides enough position information to slice the LHS. The other is to document that overridable single-expression templates must be plain PromQL, and correct the test comment.

---

ui/apps/pmm/src/components/alert-thresholds/AlertThresholds.tsx
```diff
@@ -0,0 +1,195 @@
...
+      <FormProvider {...methods}>
+        <Stack component="form" onSubmit={methods.handleSubmit(handleSubmit)}>
+          {isLoading ? (
+            <Typography variant="body2" color="text.secondary">
+              {Messages.loading}
+            </Typography>
+          ) : rows.length === 0 ? (
+            <Typography variant="body2" color="text.secondary">
+              {Messages.empty}
+            </Typography>
```
ademidoff, line 156: Query errors fall through to the empty state. When `ListThresholds` fails, `isLoading` is false and `rows` is empty, so the modal says "No overridable thresholds for this node." After a 500, a timeout or a 403 (the endpoint is admin-only), the user sees the error snackbar and then a message that isn't true. Handling `isError` separately, e.g. with an error state, would fix this.

---

ui/apps/pmm/src/components/alert-thresholds/AlertThresholds.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  // The form is seeded here rather than through useForm's defaultValues: this component
+  // is always mounted and only returns null while closed, so defaultValues sees the empty
+  // object once, at app start.
+  //
+  // `nodeId` is a dependency in its own right - handleClose clears it, so reopening, even
+  // for the same node, re-seeds and drops edits left behind by a cancelled session.
+  useEffect(() => {
+    methods.reset(initialValues);
+  }, [nodeId, initialValues, methods]);
```
ademidoff, line 79: This follows up on the earlier thread about this effect. Separating `initialValues` from `rows` fixed the rule-title case, but a change in `data` itself still resets the form.

When the modal is reopened for a node it has shown before, the cached rows appear immediately and a background refetch runs, because the query is stale. If the refetch returns different data (for example, another admin changed a threshold in the meantime), `methods.reset()` throws away whatever the user has typed so far. A refetch on reconnect does the same.

Two ways to keep the edits: reset only when `nodeId` changes, or call `reset(values, { keepDirtyValues: true })`.

---

managed/services/alerting/threshold_overrides.go
```diff
@@ -0,0 +1,553 @@
...
+// readThreshold reports a parameter as it stands for one target after a write, resolved
+// through the same precedence the collector applies.
+func (s *Service) readThreshold(
+	q *reform.Querier,
+	ruleID, paramName string,
+	param models.AlertRuleParam,
+	scope models.ThresholdScope,
+	target string,
+) (*alerting.Threshold, error) {
+	overrides, err := models.FindThresholdOverridesByRule(q, ruleID)
```
ademidoff, line 528: `readThreshold` runs once per updated row. Each time, it reloads every override on the rule, resolves inventory for all of their targets, and looks up the target node again. That comes on top of the `FindAlertRuleByID`/`FindNodeByID` that `resolveThresholdRequest` has already done. A batch of N rows on a rule with M overridden nodes therefore costs O(N·M) inside one transaction. `ListThresholds` with a target has a similar cost: it loads every override of every rule, plus inventory for every overridden target.

Resolving once per rule after all the writes in the batch, or loading only the overrides that can affect the target, would keep this linear.

---

managed/models/alert_rule_model.go
```diff
@@ -0,0 +1,98 @@
...
+
+//go:generate go tool reform
+
+// AlertRuleParam is the snapshot of one overridable parameter, taken when the rule was
+// created. The template it came from can be edited or deleted afterwards, so this is the
+// only durable record of what the rule actually evaluates against, and the only place the
+// effective default can be read back from.
+type AlertRuleParam struct {
+	Default   float64  `json:"default"`
+	JoinLabel string   `json:"join_label"`
```
ademidoff, line 34: `JoinLabel` is written when the rule is created but never read. The builder and the collector always recompute it through `joinLabelForParam`. A few other additions are in the same state:

- `models.AlertExprParamDefinition.OverrideScopes` (`template_model.go:118`) is written in `template_helpers.go` and never read.
- `FindThresholdOverridesByTarget` is only used in tests.
- `ThresholdInventory.ServicesByCluster` is read by the resolver but never populated, so cluster overrides would silently resolve to nothing once that scope is enabled.

Storing a derived value in JSONB means it can drift from the value the code actually uses. I'd drop these until the scopes that need them land.

---

managed/pi/alert/parameter.go
```diff
@@ -21,14 +21,33 @@ import (
...
-	Range   []any  `yaml:"range,flow,omitempty"`
-	Value   any    `yaml:"value,omitempty"`
+	Name           string   `yaml:"name"`                           // required
+	Summary        string   `yaml:"summary"`                        // required
+	Unit           Unit     `yaml:"unit,omitempty"`                 // optional
+	Type           Type     `yaml:"type"`                           // required
+	Range          []any    `yaml:"range,flow,omitempty"`           // optional
+	Value          any      `yaml:"value,omitempty"`                // optional
+	Overridable    bool     `yaml:"overridable,omitempty"`          // optional
+	OverrideScopes []string `yaml:"override_scopes,flow,omitempty"` // optional
```
ademidoff, line 40: Nit: AGENTS.md asks for no trailing inline comments (`code // comment`, with `//nolint` as the only exception). The rest of this struct already uses them, so these two lines match the local style. The new test code does the same, though, and those comments could move onto their own line: `threshold_overrides_test.go:322` and `api-tests/alerting/thresholds_test.go:336`.

---

managed/services/alerting/threshold_overrides_test.go
```diff
@@ -0,0 +1,501 @@
...
+	"github.com/percona/pmm/managed/models"
+	"github.com/percona/pmm/managed/utils/testdb"
+)
+
+const thresholdTestRuleID = "threshold-api-rule"
+
+func setupThresholdAPI(t *testing.T) (*Service, *reform.DB, *models.Node) {
+	t.Helper()
+
+	sqlDB := testdb.Open(t, models.SkipFixtures, nil)
```
ademidoff, line 40: `managed/AGENTS.md` asks for `go-sqlmock` in unit tests and reserves `testdb.Open` for tests that need fixtures or migrations. This file, `reconciler_test.go:38`, `service_threshold_test.go:88` and `threshold_metrics_test.go:99` all open a real DB with `SkipFixtures`. Tests that exercise the FK cascade or the `ON CONFLICT` upsert have a reason to use the real DB. The pure service-logic tests could move to sqlmock.

---

ademidoff (changes requested):
I'm requesting changes mainly because of these five, which can lose override data or give targets the wrong threshold:

- The orphan sweep deletes registry rows, and their overrides, based on the caller's Grafana ruler view. That view is limited to the caller's current org (`reconciler.go`).
- The default clause becomes NaN when the observed value is ±Inf, so those targets never alert (`rule_builder.go`).
- `SplitSingleExpr` accepts any literal 0 rather than just the sentinel. As a result, some templates pass validation and are then desugared into a different comparison (`overridable.go`).
- If the `CreateAlertRule` response is lost, the rule stays live but can never be overridden (`service.go`).
- In multi-expression templates, only the first expression that uses a token gets a correct threshold (`rule_builder.go`).

The remaining comments cover:
- Design questions: tombstones, what triggers the sweep, and PromQL vs MetricsQL parsing.
- Two UI edge cases.
- Batch-update cost.
- Unused fields and convention nits.

---

managed/services/alerting/reconciler.go
JiriCtvrtka: Blocker: Sweep trusts an unvalidated ruler reply. do() skips JSON decode on an empty body ([client.go:196](vscode-webview://14rjts7rv4e79l3md7urt3e1ar07aa4niiht372c55p8r9n2d724/managed/services/grafana/client.go#L196)), so a 200/204 with no body yields a nil folders map and a nil error; live is then empty, every registry row older than the 10-minute grace period is deleted at [reconciler.go:109](vscode-webview://14rjts7rv4e79l3md7urt3e1ar07aa4niiht372c55p8r9n2d724/managed/services/alerting/reconciler.go#L109), and ON DELETE CASCADE ([database.go:1219](vscode-webview://14rjts7rv4e79l3md7urt3e1ar07aa4niiht372c55p8r9n2d724/managed/models/database.go#L1219)) takes every threshold override with it. Unrecoverable, logged at Info. Triggered from a read handler ([threshold_overrides.go:238](vscode-webview://14rjts7rv4e79l3md7urt3e1ar07aa4niiht372c55p8r9n2d724/managed/services/alerting/threshold_overrides.go#L238)), so viewing the page is enough. Fix: skip the sweep when live is empty, and cap the fraction of rows one sweep may reap.

managed/pi/alert/overridable.go
```diff
@@ -0,0 +1,191 @@
...
+
+// validateOverridableParams checks the constraints that depend on the template's shape,
+// rather than on the parameter alone.
+func (r *Template) validateOverridableParams() error {
+	overridable := r.OverridableParams()
+
+	for _, param := range overridable {
+		if r.UsesMultipleExpressions() {
+			// Must resolve to a query, not merely be mentioned.
+			_, err := r.ObservedQueryForParam(param.Name)
```
JiriCtvrtka, line 167: High: Multi-expression templates validate the overridable param only via ObservedQueryForParam, which requires just a preceding $RefID — never that the param is a direct comparand. $A > [[ .threshold ]] * 100 passes, swapOverridableTokens splices it to $A > $T_threshold * 100, and the rule fires on value×100 while the API and annotations report the raw value. Not reachable from any shipped template (the only overridable one, node_high_cpu_load.yml, is a direct comparison) — custom templates only. Fix: apply the single-expression path's bare-NumberLiteral guard here too.

PR #5884, opened 2026-09-03
PMM-15205 Ship SEP embedded in PMM (Tech Preview)

yyyyyyyan:
Ticket number: PMM-15205

Integration branch for the **PMM + SEP Integration (Tech Preview)** epic. Every
child ticket of PMM-15205 landed here first, so `main` sees one reviewable
change with the whole Tech Preview in it rather than a half-wired feature spread
over three dozen merges.

## Composition

| Ticket | What it adds |
| --- | --- |
| PMM-15216 | The SEP frontend, vendored into the UI workspace as `ui/packages/sep/*` and mounted on `/sep` |
| PMM-15293 | The embedded UI exchanges the PMM session for a per-user SEP bearer instead of an injected static token |
| PMM-15294 | ServiceNow connection settings submitted to SEP |
| PMM-15337 | Support diagnostics gated on ServiceNow setup |
| PMM-15358 | Per-app write controls hidden from non-admin sessions |
| PMM-15359 | Failed SEP UI actions reported in SEP's own component tree |
| PMM-15281 | SEP's nav entries and pages rendered only when the integration is enabled |
| PMM-15384 | The SEP apps grouped under a Management sidebar section |
| PMM-15397 | The Collect panel hidden from Viewers and Editors |
| PMM-15405 | Support diagnostics kept readable when ServiceNow delivery is unavailable |
| PMM-15406 | The ServiceNow delivery connection testable from the settings tab |
| PMM-15407 | SEP enablement reported in PMM Server telemetry |
| PMM-15419 | ServiceNow credential handling: Disconnect stays reachable, blank credentials refused |
| PMM-15424 | Management sidebar icons and grouping |
| PMM-15448 | SEP pages rendered on the paper surface, loading state included |
| PMM-15449 | Run progress and failure reason surfaced for a backup run |
| PMM-15450 | A restore confirmed before it overwrites live data |
| PMM-15451 | A working backup fitted within one screen |
| PMM-15452 | The field a cascade is waiting on named |
| PMM-15453 | Identifiers, dates and config rendered as meaning |
| PMM-15454 | The timezone a schedule runs in stated, with a next-runs preview |
| PMM-15455 | One noun set settled for backups and restores |
| PMM-15457 | The plugin list fitted to content width |
| PMM-15458 | One helper text per form field, and only one |
| PMM-15478 | `sep_enabled` exposed on the settings payload |
| PMM-15490 | The chain-tasks input hidden when nothing can be chained |
| PMM-15509 | Why a diagnostics run failed, and when it never ran |
| PMM-15510 | The diagnostics run form speaking PMM, not the CLI |
| PMM-15279 (#5759) | The `/sep/` nginx reverse proxy, productized out of the harness overlay |
| PMM-15316 (#5762) | SEP's `SECRET_KEY` and database credentials published as secret files; the database password generated when unset |
| PMM-15414, PMM-15415, PMM-15437, PMM-15438, PMM-15439, PMM-15440 (#5886) | The vendored SEP frontend synced to SEP `c16578dfe` |

PMM-15238 (#5700) and PMM-15288 (#5728) are already on `main` and are not part of
this diff. PMM-15280 (#5755) was closed — the Grafana service account moved to
the SEP side.

## The vendored SEP frontend is synced to SEP `193cd869a`

Three sync commits brought `ui/packages/sep/*` up to SEP's `main`: `c16578dfe`
(#5886), `be2859f3a` (#5911) and now `193cd869a`. The last one matters beyond
being current: **the watermark and the side-car image this branch is exercised
against are now the same SEP commit**, so a paired-stack failure has no version
skew left to suspect.

That final sync is content-decided rather than commit-list-decided. Twenty-two
side-car commits touched `frontend/` in the range, and several are SEP catching
up to work authored here — SEP-2039 in particular must *not* be ported, because
PMM is ahead of it via PMM-15451. The specs moved purely additively (no path or
schema removed, no type changed) and PMM had made no local spec edits, so they
were copied wholesale and the clients regenerated. The one change with a
user-visible consequence here was SEP-2046: a whitespace-only entry in a
`StrippedNonEmptyStr` field now reads "cannot be only whitespace" instead of
"does not match the required format".

One thing worth recording for the next sync: **openapi-typescript promotes any
property carrying a `default` to non-optional**, so four fields the spec's
`required` array omits still generate without `?`. Reading the spec does not
predict what the generated client demands — the generated file is the signal.

## Everything here is behind `PMM_ENABLE_SEP`

With the flag unset the shipped backend delta is one inert glob `include` in
`pmm.conf`; nginx renders no drop-in, no secret files are written, and no SEP
database is provisioned. The SEP UI is additionally gated on the `sepEnabled`
setting (PMM-15281), which defaults to false, so the nav entry and its routes
stay hidden.

## Please merge with a merge commit, not a squash

The repo's default is squash-and-merge, which would collapse the whole epic —
28 tickets and several thousand reviewed lines — into a single `main` commit and
discard the per-ticket attribution, including the `Co-authored-by` trailers that
credit everyone who worked on the UI migration. A merge commit keeps the
per-ticket history bisectable on `main`.

## Feature build

**`perconalab/pmm-server-fb:PR-4500-7b10a3d`** — built by
https://github.com/Percona-Lab/pmm-submodules/pull/4500 from submodules commit
`7b10a3d3f`, pinning this branch at **`caf211b2b`** (2026-09-22). The one
commit after it, `48e2de0e5`, only rewords a code comment.

The submodules pin follows this branch's *tip* at build time, so the image tag
names the submodules commit, not the PMM one. For anything merged here since,
ask git:

```
git merge-base --is-ancestor <the PR's merge commit> caf211b2b && echo "in the FB"
```

`8c50951eb` (a one-line gate on the diagnostics live-log viewer) landed after
this build and is covered by unit tests only.

Paired side-car: `percona/percona-sep:193cd869ad5d178da3104e10452bdfc63320f1fb`.

## Testing

Each child ticket carries its own testing evidence on its own PR; this
description does not restate it.

What is specific to the integration branch, at this tip:

- UI workspace green across all six packages: `format:check`, `lint`,
  `check-types`, `test` and `build`.
- `go build ./...` clean. `go test ./managed/utils/envvars/...` passes (21), and
  `TestSepConfigFiles` in `managed/services/server` passes (4). `TestFiles` in
  that package fails outside a container because it reads `/etc/nginx/nginx.conf`
  — it fails identically on `main`, so it is neither covered by nor broken by
  this branch.
- End-to-end against the paired side-car, on the feature-build image: MySQL
  Backups and Support diagnostics driven through their normal flows, including
  a diagnostics bundle delivered to a ServiceNow development instance.

### Two defects found and fixed while preparing this merge

`pnpm check-types` was failing at the branch tip with seven `TS2353` errors in
`@sep/plugins-atw`, from PMM-15510 (#5964): `isCliUsageTrapField` declared an
inline `{ name; label? }` parameter narrower than the `SectionField` both
production call sites pass, so TypeScript's excess-property check rejected every
field literal in its tests.

It reached the tip green because **`make ci` runs `format-check`, `lint`, `test`
and `build` but not `check-types`**, and `.github/workflows/ui.yml` mirrors that.
This is the third type error this branch has taken through that gap — the
`sepEnabled`/`__mocks__/settings.ts` TS2741 fixed in `af9fb1aec` was the second.
Worth closing in `ui.yml` separately; not done here.

The second was the merge of `main` into this branch: `testUtils.tsx` conflicted
where both sides had appended distinct test helpers. Both kept.

## Before merge

- [x] A green Feature Build, exercising the shipped image path — the `/srv/sep`
      mountpoint and the rendered nginx drop-in only exist in a real build.
- [x] End-to-end verification against the paired SEP side-car.
- [x] The remaining PMM-15205 children merged into this branch.
- [x] API Docs updated — not applicable, no API endpoints added, removed or altered.


yyyyyyyan:
## End-to-end verification against the paired SEP side-car

Run on the shipped image path, not a dev server.

| | |
|---|---|
| PMM | `perconalab/pmm-server-fb:PR-4500-e2b7a40` (submodules `e2b7a40fd`, this branch @ `d92b2646d`) |
| Side-car | `percona/percona-sep:193cd869ad5d178da3104e10452bdfc63320f1fb` |
| Fingerprint confirmed in-product | `3.10.0-PMM-15216-e2b7a40fd` |

Both halves carry the same SEP commit (`193cd869a`), so nothing below is
attributable to skew between the vendored frontend and the side-car.

### The integration works

Brought up from empty volumes: three containers healthy, `/v1/readyz` 200, all
three SEP services 200, `/sep/api/apps/` 401 unauthenticated, PMM session → SEP
bearer exchange 200, both executors registered (`pmm-server`, `sep-mysql`), app
restriction exactly `inventory` / `atw` / `mysql_backups`.

Both apps produced **real** output, not empty success:

- **Mydumper** backup → `Done`, 00:00:04, 226.6 MiB, 227M verified on disk.
- **XtraBackup** backup → `Done`, `[Xtrabackup] completed OK!`, 250 MB, 251M on disk.
- **Diagnostic run** → 313 lines of real `pt-mysql-summary` output.
- **ServiceNow delivery** → CS0062206 attachments **8 → 9**, new attachment
  `08c812f30f1f8750ad20a8d530d1b2f6` (4647 B). The sys_id in the UI's success
  message is that same attachment, so both ends agree it was delivered. Target
  was the perconadev development instance throughout; zero references to the
  production instance anywhere in the side-car.

Console clean on every page of both apps.

Verified positively, by correct behaviour rather than absence of error:
**PMM-15509** (failure reason stated, with `SEP_UNLAUNCHABLE: command=sudo
node=pmm-server`, not blank), **PMM-15510** (both snippets declare a `help`
param in source; the forms render every other declared param and omit that one),
**PMM-15437** (case autocomplete resolves against the delivery provider),
**PMM-15454** (cron mode states its timezone and previews next runs),
**PMM-15450** (a restore raises a confirmation before running).

### Defects found — none blocking, all filed rather than patched here

No crash, data-loss or security issue. Ranked:

1. **Task status chip never updates** (medium). After Execute the chip beside the
   task name stays "Created, not run yet" for the whole run and after it
   finishes, while the Last-run panel in the same viewport shows Running → Done.
   `usePluginTask` never refetches and nothing invalidates it. Reload fixes it.
2. **Diagnostics logs do not stream** (medium). `GET /sep/stream-logs/<id>` is
   fetched once, only after the run reaches a terminal state; until then the
   panel reads "No logs available for this execution." The **backups** log drawer
   does stream incrementally, so this is specific to the diagnostics Results panel.
3. **Interval-schedule timezone contradiction** (low). Header says "Times shown in
   America/Sao_Paulo"; the preview lists UTC values with no unit. Cron mode is correct.
4. **Raw identifiers leak** (low). `Database Host` renders `1` rather than
   `sep-mysql (mysql)`; `Created By` on Restores shows a UUID where Backups shows `admin`.
5. **No executor/service mismatch warning** (low). A MySQL snippet aimed at
   `pmm-server` submits silently, then fails with `SEP_UNLAUNCHABLE`.
6. **Restore confirmation carries no destructive-data warning** (low). It confirms
   the target but does not mention data loss, and shows `Overwrite tables: No`,
   which reads as a safety assurance but is only the mydumper flag.

All six sit in the vendored `ui/packages/sep/**` tree. They are **not** patched in
this PR on purpose: this branch has just brought the vendored watermark and the
paired side-car image to the same SEP commit, and a PMM-local patch to
`@sep/framework` would immediately reintroduce the reverse drift that property
exists to prevent. Items 3–6 are also shortfalls in already-reviewed tickets on
this branch rather than new regressions. The route is SEP first, then a sync — the
same route this branch's frontend history already uses.

### One finding that is not a defect

The delivery-connection API returns `{"status":"available"}` with six detail rows,
while Settings → ServiceNow still renders the unconnected setup wizard. This is not
a rendering bug: `delivery-connection` appears nowhere in PMM's UI outside the
generated client, so the connected-state panel is unbuilt. That is **PMM-15218**,
deliberately not in this branch. Worth noting there: the endpoint is live and
returning usable data today, so that work is now unblocked on the side-car side.

### Harness

`percona/SEP@pmm` is repinned to this build (`1ba97c704`), with all three
`PMM_FB_TAG`-family pins moved together and `check_pmm_fb_pins.py` passing —
confirmed to be a real pass by breaking one pin and watching it report the
disagreement. `NOMAD_VERSION` stays `2.0.5`, re-read from the new image rather
than assumed.

fabio-silva (approved):

ademidoff (approved):

PR #5888, opened 2026-09-04
PMM-15360: Gate OpenManager's service on the switch

plebioda:
## What

Implements `Enabled() bool` on `om.Service`, so every `/v1/om/*` RPC and the scheduled topology collection refuse while OpenManager is off, via the same generic gRPC-service-enabled interceptor `BackupService` and the other preview features already use — no proto or per-method changes needed.

Adds `IsAvailable()`, a read against SEP's `om_inventory` app (the same call the scheduled collection already makes), and wires it into `ChangeSettings` so turning OpenManager on is refused unless the app is actually reachable — rather than accepting a setting that would do nothing. Turning it on also kicks an immediate topology collection so the page is not empty on first view.

Does **not** touch SEP's `om_inventory` app in any way — turning the switch off does not stop its periodic sweep. That's deferred to a follow-up.

Ticket: [PMM-15360](https://perconadev.atlassian.net/browse/PMM-15360)

## Stacking

Second of three PRs for PMM-15360. Stacks on `PMM-15326-om-backend` (#5816, needs `om.Service`) **and** #5851 (needs `Settings.OpenManager.Enabled`) — both are merged in here since there's no single upstream branch with both yet, so this diff includes #5851's changes inline until that one merges. Only the last commit ("Gate OpenManager's service on the switch") is net-new here.

Feature build: N/A on its own, same reasoning as #5851 — the UI half that would make this visible lands in the sibling nav/page PR.

[PMM-15360]: https://perconadev.atlassian.net/browse/PMM-15360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
_Recreated from #5852 -- GitHub Actions kept resolving GITHUB_BASE_REF to `main` for that PR instead of `PMM-15360-om-switch-flag`, surfacing pre-existing unrelated lint findings as CI failures. A fresh push and a close/reopen on #5852 did not clear it, so recreating instead._

### DIFF
--- a/managed/services/om/sep_client.go
+++ b/managed/services/om/sep_client.go
@@ -102,3 +102,50 @@ func (a sepApp) request(ctx context.Context, method, path string, query url.Valu
 	}
 	return req, nil
 }
+
+// patchConfig sends fields to the app's PATCH /config endpoint and discards the
+// response body -- callers that only want to write settings, not read the
+// SettingResponse rows PATCH returns, use this instead of building the request
+// themselves.
+func (a sepApp) patchConfig(ctx context.Context, fields map[string]any) error {
+	req, err := a.request(ctx, http.MethodPatch, "config", nil, fields, false)
+	if err != nil {
+		return fmt.Errorf("failed to build the request: %w", err)
+	}
+
+	resp, err := a.client.http.Do(req)
+	if err != nil {
+		return fmt.Errorf("PATCH %s: %w", a.endpoint("config"), err)
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("PATCH %s: unexpected status %s", a.endpoint("config"), resp.Status)
+	}
+	return nil
+}
+
+// triggerRun queues a full-estate probe sweep via the app's POST /runs endpoint and
+// discards the accepted run's body. Body is sent empty rather than omitted -- the
+// endpoint takes an optional scope object, and FastAPI rejects a bodyless POST
+// against a typed body.
+//
+// A 409 (a sweep already in flight) is not treated as a failure to log: the estate
+// is about to be swept either way, which is exactly the outcome a caller here wants.
+func (a sepApp) triggerRun(ctx context.Context) error {
+	req, err := a.request(ctx, http.MethodPost, "runs", nil, nil, true)
+	if err != nil {
+		return fmt.Errorf("failed to build the request: %w", err)
+	}
+
+	resp, err := a.client.http.Do(req)
+	if err != nil {
+		return fmt.Errorf("POST %s: %w", a.endpoint("runs"), err)
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	if resp.StatusCode != http.StatusAccepted && resp.StatusCode != http.StatusConflict {
+		return fmt.Errorf("POST %s: unexpected status %s", a.endpoint("runs"), resp.Status)
+	}
+	return nil
+}
--- a/managed/services/server/server.go
+++ b/managed/services/server/server.go
@@ -488,19 +488,9 @@ func (s *Server) validateChangeSettingsRequest(ctx context.Context, req *serverv
 		return status.Error(codes.FailedPrecondition, "Azure Discover is configured via PMM_ENABLE_AZURE_DISCOVER environment variable.")
 	}
 
-	if req.EnableOm != nil && s.envSettings.EnableOM != nil && *req.EnableOm != *s.envSettings.EnableOM {
-		return status.Error(codes.FailedPrecondition, "OpenManager is configured via PMM_ENABLE_OM environment variable.")
-	}
-
-	if req.EnableOm != nil && *req.EnableOm {
-		currentSettings, err := models.GetSettings(s.db.WithContext(ctx))
-		if err != nil {
-			return status.Errorf(codes.Internal, "failed to get server settings: %s", err)
-		}
-
-		if !currentSettings.IsOMEnabled() && (s.omService == nil || !s.omService.IsAvailable(ctx)) {
-			return status.Error(codes.FailedPrecondition, "OpenManager cannot be enabled: the OpenManager Inventory app is not available in SEP.")
-		}
+	err := s.validateEnableOm(ctx, req.EnableOm)
+	if err != nil {
+		return err
 	}
 
 	if !canUpdateDurationSetting(metricsRes.GetHr().AsDuration(), s.envSettings.MetricsResolutions.HR) {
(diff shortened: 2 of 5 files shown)


plebioda:
@coderabbitai review

managed/services/om/sep_client.go
```diff
@@ -102,3 +102,50 @@ func (a sepApp) request(ctx context.Context, method, path string, query url.Valu
...
+// is about to be swept either way, which is exactly the outcome a caller here wants.
+func (a sepApp) triggerRun(ctx context.Context) error {
+	req, err := a.request(ctx, http.MethodPost, "runs", nil, nil, true)
+	if err != nil {
+		return fmt.Errorf("failed to build the request: %w", err)
+	}
+
+	resp, err := a.client.http.Do(req)
+	if err != nil {
+		return fmt.Errorf("POST %s: %w", a.endpoint("runs"), err)
```
4nte, line 143: this prefix duplicates what the error already carries. `http.Client.Do` returns a `*url.Error`, whose `Error()` is already `Post "<url>": <cause>`, so both the verb and the full URL get printed twice. Against a dead port:

```
POST http://sep:8000/api/apps/om_inventory/runs: Post "http://sep:8000/api/apps/om_inventory/runs": dial tcp 127.0.0.1:59999: connect: connection refused
```

managed/services/server/server.go
```diff
@@ -627,11 +660,36 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverv1.ChangeSetting
...
+
 	return &serverv1.ChangeSettingsResponse{
 		Settings: s.convertSettings(newSettings, disableInternalPgQan),
 	}, nil
 }
 
+// triggerOMCollectionIfJustEnabled kicks a topology collection so OpenManager's page is
+// not empty on first view, instead of waiting out the next scheduled tick.
+func (s *Server) triggerOMCollectionIfJustEnabled(ctx context.Context, oldSettings, newSettings *models.Settings) {
+	if oldSettings.IsOMEnabled() || !newSettings.IsOMEnabled() || s.omService == nil {
```
4nte, line 674: I suggest we extract this check to a function which should make it a bit more readable on the caller site and to enforce DRY principle.

managed/services/server/server.go
```diff
@@ -627,11 +660,36 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverv1.ChangeSetting
 		}
 	}
 
+	s.triggerOMCollectionIfJustEnabled(ctx, oldSettings, newSettings)
```
4nte, line 663: **Enabling OpenManager serializes four SEP round-trips into the settings save.**

I built the real `Server.ChangeSettings` with the real `om.Service` behind a SEP stub at 3s per request:

```
SEP hit #1  GET   /api/apps/om_inventory/services   <- validateEnableOm / IsAvailable
SEP hit #2  GET   /api/apps/om_inventory/services   <- TriggerTopologyCollection
SEP hit #3  PATCH /api/apps/om_inventory/config     <- SyncInventoryEnabled
SEP hit #4  POST  /api/apps/om_inventory/runs       <- immediate sweep kick

ChangeSettings(EnableOm=true) took: 12.02s
ChangeSettings(no OM transition) took: 0s
```

`TriggerTopologyCollection` runs `s.collect(ctx)` inline, so this is a full estate collection plus VictoriaMetrics queries inside a UI settings save. If the client times out, the setting is already committed but the request failed, which is the worst combination.

The doc comments describe these as best-effort side effects. Firing #2, #3 and #4 in a goroutine with a background context would make the code match that, and leave only `IsAvailable` (bounded at 10s) on the request path.

4nte (commented):
One finding from a local build of this branch, measured rather than read off the diff.

managed/services/om/sep_client.go
```diff
@@ -102,3 +102,50 @@ func (a sepApp) request(ctx context.Context, method, path string, query url.Valu
...
+// response body -- callers that only want to write settings, not read the
+// SettingResponse rows PATCH returns, use this instead of building the request
+// themselves.
+func (a sepApp) patchConfig(ctx context.Context, fields map[string]any) error {
+	req, err := a.request(ctx, http.MethodPatch, "config", nil, fields, false)
+	if err != nil {
+		return fmt.Errorf("failed to build the request: %w", err)
+	}
+
+	resp, err := a.client.http.Do(req)
```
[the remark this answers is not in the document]
plebioda: Half of this is real, so it is fixed in half.

Rejecting non-HTTPS base URLs would disable OpenManager in every deployment there is: SEP ships as a sidecar inside the PMM container, and `--sep-url`'s own help text documents `http://127.0.0.1:8000`. Requiring TLS on loopback buys nothing and costs the feature.

The redirect half is real and cheap to close. `net/http` only drops `Authorization` when a redirect leaves the host, so an https SEP redirecting to http keeps the bearer and puts it on the wire in clear text. The shared client now refuses to follow redirects at all (`refuseRedirect`): SEP's JSON API has none, so a 3xx from it is a misconfiguration better surfaced as an unexpected status than followed with a credential attached. `TestSEPClientRefusesRedirects` asserts the redirect target never sees the request.

A token that really would cross the network in clear text - plain HTTP to anywhere but this host - is now warned about once at startup instead of being refused. c1547be50

managed/services/om/sep_client.go
plebioda: Confirmed and fixed in c1547be50. The transport error is returned bare now, since `*url.Error` already names the verb and the full URL. The status-code errors keep their prefix: an `http.Response` carries no URL, so there the verb and endpoint are the only thing naming what failed. `TestSEPClientTransportErrorIsNotDoublePrefixed` counts the URL in the message so it stays at one.

The same wrap sits in `probeSource.fetch`. #5921 is already rewriting that function for the pagination envelope, so fixing it here would only collide; it goes in there instead.

managed/services/server/server.go
plebioda: Done in 195f87938, by merging rather than extracting: `triggerOMCollectionIfJustEnabled` and `syncOMInventoryEnabledIfChanged` are now one `applyOMSwitch`. The transition is derived once instead of each function re-deriving it from `IsOMEnabled()` with a differently shaped guard, and the caller site is one line rather than two.

That also gave your other finding on this code somewhere to live, since both effects now leave the request path together.

If you would still rather see a named predicate (`omJustEnabled(old, new)`) at the guard itself, say so and I will add one.

managed/services/server/server.go
plebioda: Reproduced and fixed in 195f87938. Thanks for measuring it rather than reading it off the diff.

Both best-effort effects now run in one goroutine on a context detached from the request - cancellation dropped, values kept - bounded by a timeout of its own, so a SEP that never answers cannot keep the goroutine alive. Only `IsAvailable` stays on the request path, where it belongs: its answer decides whether the save is allowed at all, and it is bounded at 10s by `probeRequestTimeout`.

`TestApplyOMSwitch` asserts the effects still fire after the caller's context is cancelled, which is the regression that would quietly put them back on the request path.

PR #5890, opened 2026-09-04
PMM-15260 Make pmm-agent Node registration idempotent

ademidoff:
Ticket number: PMM-15260

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/4558

`pmm-agent setup` registered the Node on every run. A forced registration makes PMM Server drop the
Node together with every Service on it, because `RegisterNode` removes the existing Node with
`RemoveCascade`. That is harmless for the database operators, whose sidecars keep their configuration
in an ephemeral file and re-add their own Services on each container start. It silently deletes
user-added Services from a long-lived PMM Client, which is what PMM-14989 reports and what the
pmm-ha Helm chart needs in order to pre-provision Client pods as monitoring delegates.

This PR covers points 1 and 2 of the ticket:

- **Registration is skipped when it would be a no-op.** An Agent whose configuration already holds an
  ID and which is set up against the same PMM Server keeps its Node and its Services. `--force` still
  registers the Node again exactly as before, and pointing an Agent at a different PMM Server
  registers it there. The address is compared against the configuration file rather than the merged
  configuration, which is what keeps `pmm-admin config` against a new server working.
- **The Agent verifies that the server still knows it before skipping.** Otherwise an Agent is
  stranded whenever PMM Server no longer has its ID, e.g. after the server was reinstalled or
  restored from a backup predating the registration, and pmm-agent loops on "No Agent with ID" while
  monitoring nothing. Setup now asks the server and registers again when the answer is no. 404, 401
  and 403 all count as an answer, since registering either succeeds or reports a credentials problem
  with an actionable message. An unreachable server is not an answer: a failed check keeps the
  existing registration, so an Agent still starts while PMM Server has no elected leader.

Both commits are the prototypes from `PMM-14665-skip-non-eligible-nodes-from-add-service`, reverted in
9af8db5b5 to keep that ticket to its own scope. They are cherry-picked onto current `main` here with
no conflicts, and carry their unit tests.

### Not in this PR

Points 3 and 4 of the ticket are deliberately left out and stay open:

- **Node adoption**, so that a client reclaims a Node whose pmm-agent is not connected instead of
  failing with AlreadyExists. It weakens a guard that matters outside Kubernetes, where two hosts can
  share a hostname, and the ticket's open question on gating it needs settling first.
- **Treating "No Agent with ID" as a terminal rejection** in the connect loop. Worth doing, but it
  would turn a systemd-managed install into a restart loop, and in Kubernetes a container restart does
[...]

PR #5785, opened 2026-08-18
[DOCS] 3.10.0 release

catalinaadam:
Ticket number: PMM-0

Feature build: SUBMODULES-0

If this PR adds, removes or alters one or more API endpoints, please review and update the relevant [API documentation](https://github.com/percona/pmm/tree/main/documentation/api) as well:

- [ ] API Docs updated

If this PR is related to other PRs, contributions, or ongoing work in this or other repositories, please reference them here:

- Links to related work items (optional).

PR #5897, opened 2026-09-07
PMM-14989 Fix the Kubernetes PMM Client persistence example

ademidoff:
Ticket number: PMM-14989

Feature build: N/A - documentation only.

Based on `doc-3.10.0`, not `main`.

## The problem

This is the user-facing half of PMM-14989. The reporter followed the Kubernetes page exactly,
created the PV and PVC it asks for, and still lost every database Service they had added whenever
the pod restarted or the client image was upgraded. The page was wrong in four ways at once:

- The volume was introduced as "persistent storage for PMM Client data between pod restarts" and
  then mounted at `/usr/local/percona/pmm/tmp`, never at `config`, where the Agent identity lives.
- `PMM_AGENT_CONFIG_FILE` was the relative `config/pmm-agent.yaml`, resolving under the image's
  `WORKDIR` and so onto the container's ephemeral layer rather than the volume.
- The workload was a `Deployment`, so the pod hostname, and with it the Node name, changed on every
  rollout.
- `PMM_AGENT_SETUP_FORCE=1` was set, which makes PMM Server remove the Node together with every
  Service configured on it.

The first two mean nothing was ever persisted; the last two mean the Node would have been replaced
even if it had been.

## The example now

A `StatefulSet`, whose pod name is stable and is passed to `PMM_AGENT_SETUP_NODE_NAME`, with the
volume claimed from a `volumeClaimTemplates` entry and mounted at both directories that have to
survive a restart: `config` for the identity, `tmp` for the metrics the Client buffers on disk while
PMM Server is unreachable. `PMM_AGENT_CONFIG_FILE` is absolute so it cannot drift from the mount,
and `fsGroup: 1002` replaces the chown init container.

The force flag is gone from the manifest and documented instead as the one-off recovery for a volume
lost while its Node still exists in PMM, with the consequence spelled out.

Setup still runs on every start. That is only safe from **3.10.0**, where PMM-15260 (percona/pmm#5890)
makes `pmm-agent setup` idempotent, so the page states the requirement up front - on earlier
versions this manifest still loses Services, and quietly.

Also adds a step that deletes the pod to show the identity surviving the restart, and fixes the
troubleshooting section, which pointed at `pmm-client-pvc`, a claim the flow no longer creates.

## The sidecar example

Its deployment pattern is unchanged and correct: an ephemeral configuration file, a forced setup and
a prerun script that re-adds the database service on every start. A sidecar's lifecycle is the
database's, so it rebuilds its state rather than keeping it, which is what the Percona database
operators do too.

What it did carry was the same false promise as the standalone example. It created a 10Gi
PersistentVolume and claim for the PMM Client and declared a pod volume for them, none of which any
container ever mounted, and step 2 introduced that storage as being "for storing PMM Client and
MySQL data between pod restarts". Second commit removes the volume, the claim and the pod volume
entry, says the remaining storage is for MySQL, and states why the Client gets none - so the
removal does not read later as an omission to correct. Every volume the example declares is now
mounted by a container.

## The command pages

`pmm-agent setup` and `pmm-admin config` are what the manifests on the install pages actually run, so
the pages describing them carry the same promise. They described the idempotent registration as it
stood when they were written, and one sentence had since become wrong: `--force` does not *replace*
the Node PMM Server holds the pmm-agent on when the names differ. It registers the name you gave as a
new Node and leaves the other one, with its Services, behind. That is the state the install examples
exist to avoid, so the wording is corrected on both pages.

Both now list what a re-run does rather than running it together in one paragraph, and cover what was
missing from it:

- the configuration file being left unchanged when setup did not read it, which is the `pmm-admin config` path
- an unverifiable PMM Server keeping the registration, warning on standard error, and still succeeding
- an encrypted configuration file needing its key before setup can tell whether the Node is registered
- a registered Node address or type being reported rather than quietly ignored
- the Agent ID `--skip-registration` expects, and where it may come from

The `--force` row in the options table matches the flag's own help text again.

## Verification

No local mkdocs build - the change is confined to one page and adds no new links, so `linkspector`
in CI covers what a build would. The manifest itself was checked, since it is hand-written and
readers copy it verbatim:

- every manifest on the page parses as YAML, and `kubectl apply --dry-run=client` accepts each one
- `PMM_AGENT_CONFIG_FILE` resolves inside a mounted path, which is the specific defect being fixed
- no `PMM_AGENT_SETUP_FORCE` in the manifest
- code fences balanced across the page, and the admonition classes match those already used in
  `documentation/docs`

## Related

- percona/pmm#5890 - PMM-15260, the pmm-agent change this page depends on.
- percona/percona-helm-charts#954 - the pmm-ha chart, which already deploys its Client pods this
  way and will drop its own registration gate once 3.10.0 is out.

A release note for PMM-15260 is not yet in `documentation/docs/release-notes/3.10.0.md`; its ticket
asks for one, and it is not part of this PR.

### DIFF
--- a/documentation/docs/install-pmm/install-pmm-client/docker.md
+++ b/documentation/docs/install-pmm/install-pmm-client/docker.md
@@ -184,5 +184,6 @@ When running PMM Client in Docker, use the `PMM_AGENT_PRERUN_SCRIPT` argument to
 
 ## Tips for Docker configuration
 
+- The examples above register the Node on every start, because `PMM_AGENT_SETUP_FORCE=1` replaces the Node together with the Services on it, and the prerun script adds them again. To keep the Node and its Services across containers instead, mount a volume at `/usr/local/percona/pmm/config`, drop `PMM_AGENT_SETUP_FORCE`, and set `PMM_AGENT_SETUP_NODE_NAME`: the Node name defaults to the hostname, which a new container does not keep, and `pmm-agent setup` stops when PMM Server has the pmm-agent registered under another name. Requires PMM Client 3.10.0 or later.
 - Ensure your host's firewall and routing rules are configured to allow Docker communications. This is crucial for Docker containers to communicate properly. For more details, see the [troubleshooting checklist](../../troubleshoot/checklist.md).
 - To view available pmm-agent command-line options, run: `docker run --rm percona/pmm-client:3 --help`
--- a/documentation/docs/install-pmm/install-pmm-client/docker.md
+++ b/documentation/docs/install-pmm/install-pmm-client/docker.md
@@ -32,6 +32,11 @@ Set up PMM Client by deploying it as a Docker container and registering it with 
 
 Deploy and register PMM Client to start monitoring your node. 
 
+PMM identifies a Node by name, and the identity that PMM Server issues to the pmm-agent is stored in its configuration file. Keep both stable, a fixed `PMM_AGENT_SETUP_NODE_NAME` and a volume for the configuration, and the Node together with the Services you add survives a restart, a re-created container, or an image upgrade.
+
+!!! caution alert alert-warning "Requires PMM Client 3.10.0 or later"
+    Earlier versions register the Node again on every container start, and PMM Server answers a re-registration by removing the Node together with every Service configured on it. On those versions the Services you add to a restarted container are lost regardless of the volume you give it.
+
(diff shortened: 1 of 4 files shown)


catalinaadam (commented):
Pushed commit with a linguistic review and updated rel notes separately with https://github.com/percona/pmm/pull/5785/changes/88679ab08e5664b76bb849d2b163ff8078eef177

PR #5904, opened 2026-09-08
PMM-15431 Keep the agent's request loop unblocked.

JiriCtvrtka:
PMM-15431

## What

pmm-agent's server-request loop could deadlock on `SetState` and stop monitoring entirely, with
nothing in the process able to break the cycle. The agent stayed alive, supervisord reported it
RUNNING, and its exporters kept running, so nothing looked wrong — while no metrics or QAN reached
the server and every inventory operation was rejected with `pmm-agent with ID <id> is not currently
connected`. Observed in production: the built-in agent on a server monitoring 28 RDS PostgreSQL
services wedged and made zero reconnection attempts for 9h17m, until `supervisorctl restart
pmm-agent`.

## Why it could not recover

`processChannelRequests` is the only reader of `channel.Requests()`, and `channel.runReceiver` is
the only goroutine that reads the gRPC stream — so it is also the only one that can deliver a
response the loop is waiting for. `SetState` ran inline on that loop and waits for the agents it
replaces to stop; those waits reach the server, because a stopping agent's last status updates go
out through the same channel and the client waits for each to be acknowledged. Loop stops draining
→ 32-slot queue fills → `runReceiver` parks on a bare channel send → the acknowledgement can never
arrive. `Client.Run` then never returns, so `run.go` never redials.

This is the same cycle PMM-15310 fixed on the server side in 3.9.1; pmm-agent's loop never got the
same treatment.

## Changes

- `SetState` is applied by a worker off the request loop and answered immediately. The response now
  means what `agent/v1/agent.proto` already documents — acceptance, not application; the outcome
  still arrives as `StateChanged` requests. This also stops large states blowing the server's 5s
  `stateChangeTimeout` and being retried as failures. That worker is deliberately not counted
  towards `Done()`, because `Done()` gates the reconnect that drains what a stopping agent waits on
  — so `Supervisor.SetState` takes the connection's context and re-checks it after acquiring `s.rw`,
  which is what keeps a dead connection's state from being applied after a fresh one's.
- `Ping` has its own single-slot queue (`channel.Pings()`), answered by its own goroutine, so a pong
  is never queued behind a request the loop has not reached. Dropped on overflow, since a stale ping
  is worthless.
- The actual statuses are reported by a goroutine of its own rather than inline before the
  processors start. That report asks the supervisor for its agents list, which takes the same `s.rw`
  a `SetState` from the previous connection can still hold for the whole stop budget — and acquiring
  a lock cannot be cancelled, so doing it on a goroutine that `Done()` waits for would stall the
  reconnect itself. Ordering is unchanged: the status forwarding waits for the snapshot, or for the
  context. `SendActualStatuses` is handed the connection's channel instead of reading `c.channel`,
  which by then may be the next connection's — and was read without `s.rw`.
- `runReceiver` no longer parks indefinitely on a full queue: it can be interrupted by the channel
  closing, and gives up on the connection after `requestQueueStuckTimeout` (2 min) so the caller
  redials. The give-up is logged at error level — this outage's defining property was invisibility.
  `Channel.close` never blocks its caller, because that give-up path runs while a sender is parked
  in `c.s.Send` holding the send lock, which is exactly the lock `close` would otherwise wait for.
- `Channel.SendAndWaitResponse` takes a context, which bounds the dial handshake and
  `pmm-admin status --network-info` (both could previously hang forever on a wedged connection). A
  caller that gives up marks its request abandoned rather than dropping the subscription, so a
  response that landed just as the deadline expired is returned instead of discarded — the same
  `abandon` treatment PMM-15310 gave the server side, and what keeps a live connection from being
  reported as down. The send itself is bounded too: `c.s.Send` cannot be interrupted, so it runs off
  the calling goroutine, and the send lock is a one-slot channel rather than a mutex so a later
  caller gives up instead of queueing behind a parked sender — one retained goroutine per wedged
  connection rather than one per call. A free lock is always taken, and an expired context gets a
  short grace before the message is dropped, so a last-gasp QAN bucket or status is not lost to a
  sender that held the lock for a moment on a healthy connection.
- The supervisor's six unbounded `<-agent.done` waits are bounded by one `agentsStopTimeout` budget
  (60s) per phase of a call — processes and built-in agents each get their own, so a single hung
  exporter cannot spend what the built-in agents after it need, and a batch of stuck agents can no
  longer hold `s.rw`, and with it every `SetState` and the local status API, indefinitely. Giving up
  on an agent does not treat it as gone: its port reservation and temporary directory are given back
  by a waiter that runs once it actually stops (`releaseAgentResources`), and a restart gives the
  replacement a port of its own. Releasing a port a live exporter still listens on fails and loses
  the reservation for good, since the agent that held it is already forgotten. The directory holds
  rendered TLS certificates and keys, so it is removed once the agent has stopped — unless the ID
  has been re-created by then, and that decision and the removal happen under one hold of `s.rw`,
  since a replacement can only be started while it is held for writing.
- Each start of an agent ID gets an instance token, so an agent abandoned by the bounded wait cannot
  report on behalf of the replacement that owns its ID — a `DONE` from it used to delete the
  replacement's status and leave a live agent reported as stopped. A token is retired when its
  forwarder exits, not when a `DONE` goes out, because `handleNomadAgent` reports `DONE` for an
  agent whose forwarder is still live. That handler no longer waits for a status-channel consumer
  either: it runs with `s.rw` held for writing, and `s.changes` is only drained by a live
  connection.
- `stopAll` closes the status/QAN/RTA channels only once no forwarder can still send to them.
  Bounded waits mean a forwarder can outlive them, and closing underneath one panics the process.

`agent/AGENTS.md` records the invariant so the next handler does not land back on the loop, and
names the three that still run inline (`CheckConnection`, `ServiceInfo`, `GetVersions` — bounded
only by the user's own timeouts; making the loop a pure dispatcher is PMM-4245). It also notes how
far the receiver's give-up recovers on its own: `run.go` waits on `client.Done()` before redialing,
so a handler still has to honor the canceled context.

## Testing

Unit tests reproduce both halves of the cycle at the layer each fix lives in. Each is
mutation-verified — reverting the fix makes its test fail:

| Revert | Failure |
|---|---|
| `SetState` back inline | `TestSetStateDoesNotBlockRequestLoop` deadlocks |
| the actual statuses back under the `Done()` accounting | `TestDoneClosesWhileActualStatusesAreBlocked` — `Done()` never closes |
| the actual statuses back on the synchronous path | `TestActualStatusesDoNotBlockPings` deadlocks |
| the wait for a failed start's forwarder | `TestStartProcessFail` drops the statuses and hangs |
| the supervisor's deferred release | `TestStopKeepsPortOfRunningAgent` — the port of an agent that has not stopped is released |
| the re-creation check on that release | `TestReleaseAgentResources` — the replacement's files are removed with the directory |
| the `stopAll` guard | `TestStopAll` — `panic: send on closed channel` |
| the receiver's give-up | `TestFullRequestQueueGivesUpOnConnection` hangs |
| the send lock's give-up | `TestSendDoesNotQueueBehindAParkedSender` — the send queues behind a parked sender |
| that give-up's grace | `TestSendTakesALockFreedDuringTheGrace` — a message a healthy connection would take is dropped |
| any processor unable to exit | `TestDoneClosesAfterCancel` — `Done()` never closes |

`TestReleaseAgentResources`, `TestWaitAgentStopped`, `TestAbandon` and
`TestAbandonedInstanceCannotReportForItsReplacement` cover the new primitives directly, including
the interleavings the gRPC harness cannot force. Two limits worth knowing: the atomicity of the
re-creation check has no deterministic test, because forcing that interleaving needs a seam between
the check and the removal, so `StillRunningTakesTheLockBeforeDecidingOnTheDirectory` guards the lock
discipline instead; and `TestReleaseAgentResources` does not assert that a port was given back,
because `portsRegistry.Release` decides by binding the port and nothing retries it, so a foreign
socket on it fails the assertion for reasons unrelated to the change — the temporary-directory
assertions are deterministic, and `TestStopKeepsPortOfRunningAgent` still covers the direction that
matters.

Gates run locally on the final tree: `go build ./...`; `go test -race` (also `-count=4
-shuffle=on`) on `client`, `client/channel`, `agents/supervisor`, `agentlocal`, `agents/process`;
`golangci-lint` with CI's `--new-from-rev` scoping, 0 issues; `gofumpt` and `gci` clean;
`make check-license`; `go-sumtype`; `go mod tidy` leaving no diff. Two mocked interfaces changed —
`agent/agentlocal` `client` (`GetNetworkInformation` gained a context) and `agent/client`
`supervisor` (`SetState` gained one) — and their mocks come from `make gen`, which leaves no diff on
this tree. No `.proto` and no reform model changes. The rest of `./agent/...` passes except the
packages that need a live MySQL, PostgreSQL, MongoDB or Valkey (`make env-up`).

Ticket based on discussion in this PR: https://perconadev.atlassian.net/browse/PMM-15549


agent/client/channel/rta_channel.go
```diff
@@ -135,6 +139,7 @@ func (c *RTAChannel) Send(msg *rtav1.CollectRequest) {
 
 	select {
 	case <-c.closeWait:
+		c.sendM.Unlock()
```
JiriCtvrtka, line 142: Missed unlock here.

agent/agents/supervisor/supervisor.go
```diff
@@ -498,19 +690,32 @@ func (s *Supervisor) startProcess(agentID string, agentProcess *agentv1.SetState
 	}
 
 	done := make(chan struct{})
+	instance := s.startInstance(agentID)
```
ademidoff, line 775: The retry advances the instance counter while the previous attempt's forwarder is still draining `processWrapper.Changes()`, so that attempt's terminal statuses are silently dropped.

Attempt 1 gets instance 1; `process.toFailing()` queues `INITIALIZATION_ERROR` and spawns `toDone()`, then signals `initialized <- false`. `startProcess` returns, `tryStartProcess` retries, and `startInstance` sets `instances[id] = 2` before forwarder 1 delivers `DONE` — `storeLastStatus(id, 1, DONE)` then returns false and neither status ever reaches pmm-managed.

Reproducible on the branch:

```
GOMAXPROCS=1 go test ./agent/agents/supervisor/ -run TestStartProcessFail -race -count=40
```

It logs the dropped status and then `panic: test timed out after 2m0s`, with `TestStartProcessFail` blocked in `assertChanges` (agent/agents/supervisor/supervisor_test.go:55). `assertChanges` has no timeout of its own, so CI burns the whole 10m budget instead of failing fast.

agent/client/channel/channel.go
```diff
@@ -320,30 +346,112 @@ func (c *Channel) runReceiver() {
...
+		select {
+		case c.requests <- req:
+		case <-c.closeWait:
+			return
+		case <-time.After(requestQueueStuckTimeout):
+			// Loudly: this is a monitoring outage that recovers itself, and the whole
+			// point of PMM-15431 is that it used to be invisible. Closing on its own is
+			// only reported at debug level.
+			c.l.Errorf("Request queue full for %s, giving up on the connection.", requestQueueStuckTimeout)
+			c.close(fmt.Errorf("request queue full for %s", requestQueueStuckTimeout))
```
ademidoff, line 480: `c.close()` here is synchronous, and `close()` takes `sendM` — exactly the lock a sender parked in `c.s.Send` is holding in the scenario this timeout exists for. The give-up path then deadlocks instead of recovering.

Against a server whose request loop is wedged (PMM-15310 unfixed), the HTTP/2 window is exhausted and a goroutine sits in `c.s.Send` holding `sendM`. The queue fills, `requestQueueStuckTimeout` fires, and `runReceiver` calls `c.close(...)` → `closeOnce.Do` → `close(c.closeWait)` (fine) → `c.sendM.Lock()`, which blocks forever. `runReceiver` never reaches its deferred `close(c.requests)` / `close(c.pings)`, so `processPings` and `processChannelRequests` never exit, `firstDone` never fires, `Run` never returns, and agent/client/run.go never cancels ctx or redials — the 9h-silent-agent outcome this PR is fixing.

Closing the queues before (or independently of) taking `sendM`, or running `c.close` in a goroutine, would let the recovery complete.

agent/client/client.go
```diff
@@ -250,67 +253,75 @@ func (c *Client) Run(ctx context.Context) error {
...
 	go func() {
-		c.processSupervisorRequests(ctx)
-		c.l.Debug("processSupervisorRequests is finished")
-		oneDone <- struct{}{}
-	}()
-	go func() {
-		c.processChannelRequests(ctx)
-		c.l.Debug("processChannelRequests is finished")
-		oneDone <- struct{}{}
+		c.processSetStates(ctx, setStates)
```
ademidoff, line 291: Leaving `processSetStates` out of the WaitGroup that gates `Done()` means a stale connection's goroutine can still be entering `supervisor.SetState` when the next connection's goroutine calls it, and mutex barging can leave the stale state applied last.

Connection A queues S1; `processSetStates(A)` calls `supervisor.SetState(S1)` and blocks on `s.rw` (held by an `AgentsList` or a previous `SetState`). The connection drops, `<-client.Done()` returns in agent/client/run.go because the WaitGroup doesn't cover this goroutine, and `Run(B)` starts `processSetStates(B)`, which receives the fresh S2 and barges ahead of A's queued waiter. S2 is applied, then S1 overwrites it — agents stay configured from the dead connection until the server happens to push another SetState.

The `ctx.Err()` guard at agent/client/client.go:499 doesn't cover this: it's checked before the blocking call, not after.

agent/agents/supervisor/supervisor.go
```diff
@@ -241,28 +276,53 @@ func (s *Supervisor) SetState(state *agentv1.SetStateRequest) {
 		return
 	}
 
-	s.setAgentProcesses(state.AgentProcesses)
-	s.setBuiltinAgents(state.BuiltinAgents)
+	deadline := time.Now().Add(s.agentsStopTimeout)
```
ademidoff: One deadline is computed here and shared by `setAgentProcesses` and `setBuiltinAgents`, so a hung exporter can consume the whole `agentsStopTimeout` before built-in agents are even cancelled.

With one stuck node_exporter and five QAN built-ins, `setAgentProcesses` spends the full 60s on the exporter; `setBuiltinAgents` then cancels its agents at T+60s and calls `waitAgentStopped` with an already-expired deadline, so all five are logged as "did not report itself stopped" and abandoned even though they would have stopped in milliseconds. That's the same "no budget at all" failure the comment at agent/agents/supervisor/supervisor.go:485 describes, just across phases rather than within one.

Re-basing the deadline per phase, or computing it after the cancels, would fix it.

agent/agents/supervisor/supervisor.go
```diff
@@ -546,21 +751,27 @@ func (s *Supervisor) startProcess(agentID string, agentProcess *agentv1.SetState
...
+	instance uint64,
 	processInfo *agentProcessInfo,
 	l *logrus.Entry,
 ) {
 	done := make(chan struct{})
 	s.agentProcesses[agentID] = processInfo
 
 	status := inventoryv1.AgentStatus_AGENT_STATUS_DONE
-	s.storeLastStatus(agentID, status)
+	s.storeLastStatus(agentID, instance, status)
```
ademidoff, line 853: This records `DONE` for the instance whose forwarder is still live, and `storeLastStatus` deletes `instances[agentID]` on `DONE` — so every later status from that still-running agent is dropped.

On a host with non-writable cgroups: `startProcess` starts the forwarder for instance N, then `handleNomadAgent` calls `storeLastStatus(id, N, DONE)`, which deletes `instances[id]`. The `processInfo` is still stored in `s.agentProcesses` (the agent is tracked as running and `cancel()` is never called), but every subsequent `storeLastStatus(id, N, ...)` from the live forwarder now sees `instances[id] == 0 != N` and returns false, so no status change for that agent reaches pmm-managed again for the life of the connection.

agent/client/channel/channel.go
```diff
@@ -175,8 +208,28 @@ func (c *Channel) SendAndWaitResponse(payload agentv1.AgentRequestPayload) (agen
 		Payload: payload.AgentMessageRequestPayload(),
 	})
 
-	resp := <-ch
-	return resp.Payload, resp.Error
+	select {
```
ademidoff, line 237: `ctx` doesn't bound the send itself: `c.send` at agent/client/channel/channel.go:206 runs before control reaches this select, and `c.send` → `c.s.Send` is precisely what parks forever on a wedged connection.

`agentlocal.Status` → `Client.GetNetworkInformation(ctx)` → `getNetworkInformation` → `SendAndWaitResponse(ctx, &Ping{})` blocks in `c.s.Send`, with `ctx.Done()` never selected on. The dial path is rescued by the `streamCancel` AfterFunc in `createChannelToAgentService`, but this one has no stream-level cancel, so `pmm-admin status --network-info` can still hang indefinitely — contrary to the comment a few lines below.

agent/agents/supervisor/supervisor.go
```diff
@@ -686,22 +897,38 @@ func (s *Supervisor) startBuiltin(agentID string, builtinAgent *agentv1.SetState
...
+			close(done)
+		}()
+
 		rtaBucketLastCollectTime := timestamppb.New(time.Now()).AsTime()
 
 		for change := range agent.Changes() {
 			if change.Status != inventoryv1.AgentStatus_AGENT_STATUS_UNSPECIFIED {
-				s.storeLastStatus(agentID, change.Status)
+				if !s.storeLastStatus(agentID, instance, change.Status) {
+					// Abandoned: a replacement owns this ID now.
```
ademidoff: The `continue` below skips the rest of the `Change`, not just the status. `agents.Change` is "status change and/or QAN collect request" (agent/agents/agents.go:28), so a `Change{Status: X, MetricsBucket: [...]}` from an abandoned agent loses already-collected QAN data along with the muted status — the `MetricsBucket` and `RTAQueriesBucket` branches below are never reached.

Guarding only the status forwarding would keep the buckets.

agent/agents/supervisor/supervisor.go
```diff
@@ -365,28 +554,30 @@ func (s *Supervisor) setBuiltinAgents(builtinAgents map[string]*agentv1.SetState
...
 	}
 
 	// restart
 	for _, agentID := range toRestart {
 		agent := s.builtinAgents[agentID]
-		agent.cancel()
-		<-agent.done
+		s.waitAgentStopped(agentID, agent.done, deadline)
 
 		err := s.startBuiltin(agentID, builtinAgents[agentID])
```
ademidoff, line 660: On the restart path the replacement is started into the same `<TempDir>/<type>/<agentID>` while the agent abandoned by the bounded wait may still be running and reading those files: `waitAgentStopped`'s result is ignored, then `startBuiltin` calls `templates.RenderDSN` into that directory, overwriting the TLS certs and text files the abandoned agent has open. The process path has the same hazard (agent/agents/supervisor/supervisor.go:522 → `tryStartProcess` → `processParams`).

The PR reasons carefully about not *deleting* that directory under a running agent in `releaseAgentResources`; *rewriting* it is the same collision.

agent/agents/supervisor/supervisor.go
```diff
@@ -365,28 +554,30 @@ func (s *Supervisor) setBuiltinAgents(builtinAgents map[string]*agentv1.SetState
...
+		s.waitAgentStopped(agentID, agent.done, deadline)
 
 		delete(s.builtinAgents, agentID)
 
 		agentTmp := filepath.Join(s.cfg.Get().Paths.TempDir, trimPrefix(agent.requestedState.Type.String()), agentID)
-		err := os.RemoveAll(agentTmp)
-		if err != nil {
-			s.l.Warnf("Failed to cleanup directory '%s': %s", agentTmp, err.Error())
-		}
+		s.releaseAgentResources(agentID, agent.done, 0, agentTmp)
```
ademidoff, line 652: For an abandoned built-in this spawns a goroutine that can wait forever to do nothing: the port is 0 and `agentTmp` is deliberately dropped, so `releasePortAndTempDir(id, 0, "")` is a no-op. Meanwhile `<TempDir>/<type>/<agentID>`, which holds the rendered certs and key files, stays on disk until the whole pmm-agent process restarts (`cleanupTmp` in agent/commands/run.go only runs after `processClientUntilCancel` returns).

Worth skipping the goroutine when there is nothing to release, and deciding explicitly how an abandoned built-in's key material gets reclaimed.

agent/client/channel/channel.go
```diff
@@ -320,30 +346,112 @@ func (c *Channel) runReceiver() {
...
 	defer c.m.Unlock()
-	if c.responses == nil { // Channel is closed
-		return nil
+	if c.responses == nil { // Channel is closed, no subscriptions left
+		return false
+	}
+	if _, ok := c.responses[id]; !ok {
+		return false
 	}
+	c.responses[id] = nil
```
ademidoff, line 502: When no response ever arrives, this nil marker stays in `c.responses` for the life of the connection. On a wedged-but-open connection every `SendAndWaitResponse` whose ctx expires (repeated `pmm-admin status --network-info`, repeated dial probes on the same channel) adds one, and only `removeResponseChannel` (on an actual response) or `close()` clears them — so the map grows unbounded on exactly the connection where this path fires most.

agent/client/client_test.go
```diff
@@ -293,6 +296,249 @@ func TestUnexpectedActionType(t *testing.T) {
...
+	s.On("AgentsList").Return([]*agentlocal.AgentInfo{})
+	s.On("ClearChangesChannel").Return()
+	s.On("SetState", mock.Anything).Run(func(mock.Arguments) {
+		close(applyingState)
+		<-releaseState
+	}).Return()
+
+	r := runner.New(cfgStorage.Get().RunnerCapacity, cfgStorage.Get().RunnerMaxConnectionsPerService)
+	client := New(cfgStorage, s, r, nil, nil, nil, connectionuptime.NewService(time.Hour), nil)
+	require.NoError(t, client.Run(context.Background()))
```
ademidoff: `t.Context()` is the convention here (AGENTS.md → Testing); same at agent/client/client_test.go:448 and :503.

In the first two it also matters practically: `client.Run` gets an uncancellable ctx, so the `processSetStates` goroutine the test starts is leaked past the test instead of being torn down at cleanup.

agent/agentlocal/mock_client_test.go
```diff
@@ -44,9 +45,9 @@ func (_m *mockClient) GetConnectionUpTime() float32 {
 	return r0
 }
 
-// GetNetworkInformation provides a mock function with no fields
-func (_m *mockClient) GetNetworkInformation() (time.Duration, time.Duration, error) {
-	ret := _m.Called()
+// GetNetworkInformation provides a mock function with given fields: ctx
```
ademidoff, line 48: This is mockery output and `agent/agentlocal` `client` is listed in .mockery.yaml:99-101, so the `GetNetworkInformation` signature change in agent/agentlocal/deps.go:31 should come through `make gen` rather than a hand edit — the next regeneration may differ in comment wording or argument naming.

The PR description says "make gen inputs are untouched — no ... mocked-interface changes", which doesn't match this part of the diff.

agent/client/channel/channel.go
```diff
@@ -320,30 +346,112 @@ func (c *Channel) runReceiver() {
...
+		// the pong loop exiting is what makes client.Run return, after which its caller
+		// cancels the client context, so a handler blocked on it gives up too. Only then does
+		// the reconnect happen - it waits on client.Done() - so a handler that ignores that
+		// cancellation holds the agent off the fresh connection for as long as it runs. See
+		// "the request loop must never block indefinitely" in agent/AGENTS.md (PMM-15431).
+		select {
+		case c.requests <- req:
+		case <-c.closeWait:
+			return
+		case <-time.After(requestQueueStuckTimeout):
```
ademidoff: `time.After` is evaluated on every iteration, so each received request allocates a 2-minute timer even when the send succeeds immediately. Go 1.23+ lets the GC reclaim them, so it's allocation churn rather than a leak, but a single `time.NewTimer` hoisted out of the loop (Reset/Stop), or arming it only after a non-blocking send attempt fails, avoids it.

agent/client/channel/rta_channel.go
```diff
@@ -109,14 +109,18 @@ func (c *RTAChannel) close(err error) {
 		c.l.Debugf("Closing with error: %+v", err)
 		c.closeErr = err
 
+		// Signal the close before taking sendM: Send holds sendM across the blocking
+		// c.s.Send, and CloseAndRecv below waits for the server, so closing closeWait after
+		// either would leave Wait and the reconnect that depends on it hanging. Send re-checks
+		// closeWait under sendM, so a sender that has not started yet gives up instead.
+		close(c.closeWait)
```
ademidoff, line 116: Moving `close(c.closeWait)` ahead of `c.s.CloseAndRecv()` lets `Wait()` unblock — and `Client.Run`'s deferred `sharedConn.Close()` (agent/client/client.go:242) run — while `CloseAndRecv` is still in flight, which then fails with a transport error and logs `Failed to receive final response: ...` at error level on an otherwise clean disconnect.

The `sendM.Unlock()` fix in this file is a genuine bug fix; it's only this ordering that adds noise.

agent/client/channel/channel.go
```diff
@@ -320,30 +346,112 @@ func (c *Channel) runReceiver() {
...
+	switch sub {
+	case subscriptionWaiting:
+		ch <- resp
+		if resp.Error != nil {
+			close(ch)
+		}
+	case subscriptionAbandoned:
+		c.l.Debugf("Response for ID %d arrived after its sender stopped waiting for it.", id)
+	case subscriptionUnknown:
+		c.l.Errorf("No subscriber for ID %d.", id)
```
ademidoff, line 556: `runReceiver`'s default branch (agent/client/channel/channel.go:337) calls `c.cancel(msg.Id, ...)` with a *server request* id the agent never sent, so `removeResponseChannel` returns `subscriptionUnknown` and this logs an ERROR for every payload type the agent doesn't implement — e.g. a newer server talking to an older agent, on every such message. It was silent before; debug level, or skipping the cancel on that path, seems closer to right.

agent/agents/supervisor/supervisor.go
JiriCtvrtka: Blocker: Confirmed, and reproduced with your command — hangs in assertChanges and panics at the 2m timeout. It's a regression from this PR; instance gating doesn't exist on main. startProcess now waits (bounded, 3s) for its own forwarder to drain before returning, so the retry's startInstance can't make the terminal statuses stale. Same command now passes in 2s over 40 runs.

agent/client/channel/channel.go
JiriCtvrtka: Blocker: Confirmed — close closes closeWait first but then parks on sendM, so runReceiver never closes Requests()/Pings() and firstDone never fires. Fixed in close() itself rather than at this call site, so every caller benefits: CloseSend now runs under sendM in a goroutine.

agent/client/client.go
JiriCtvrtka: High: Confirmed. Keeping this goroutine out of the WaitGroup is deliberate — including it re-deadlocks the reconnect against forward on a full s.changes — so instead SetState now takes the connection's ctx and re-checks it after acquiring s.rw.

agent/agents/supervisor/supervisor.go
JiriCtvrtka: High: Confirmed. A budget per phase now, in RestartAgents too.

agent/agents/supervisor/supervisor.go
JiriCtvrtka: High: Confirmed. Retirement moved out of storeLastStatus's DONE branch into the forwarder's exit (retireInstance, guarded on equality) — a forwarder exiting is what ends an instance, a status value isn't.

agent/client/channel/channel.go
JiriCtvrtka: Medium: Confirmed, including that the dial path is rescued by streamCancel and this one isn't. c.send now runs off the calling goroutine so the select actually bounds it.

agent/client/channel/channel.go
JiriCtvrtka: Medium: Confirmed as new — cancel was silent on an unknown ID on main. The cancel is now made only when the message carries an Unimplemented status (where the ID really is ours); an unimplemented server request gets a redacted warning instead.

agent/agents/supervisor/supervisor.go
JiriCtvrtka: Low: Reading is right. No agent emits Status and MetricsBucket in one Change today, so nothing is lost yet, but fixed anyway — only the status is muted now.

agent/agents/supervisor/supervisor.go
JiriCtvrtka: Low: Confirmed. Skips the goroutine when port == 0, and the log and comment now say explicitly that the directory waits for the next pmm-agent start.

agent/client/client_test.go
JiriCtvrtka: Low: Fixed at 376, 448 and 503.

agent/client/channel/channel.go
JiriCtvrtka: Low: Fixed with the hoisted timer. Worth noting I tried your other suggestion first (arm only after a failed non-blocking send) and TestFullRequestQueueDoesNotWedgeReceiver caught it — preferring the enqueue keeps delivering requests after close. The plain hoist preserves the select semantics.

agent/agents/supervisor/supervisor.go
JiriCtvrtka: Medium: Agreed, and it's the same collision class as the deletion. The fix is per-instance directories, which touches processParams, RenderDSN and cleanupTmp — I'd rather not redesign that here. Filing a follow-up unless you want it in this PR.

agent/client/channel/channel.go
JiriCtvrtka: Low: Accurate, but it grows by one entry per SendAndWaitResponse that both times out and never gets a response — i.e. per manual --network-info — on a connection this PR now tears down after 2 minutes. Leaving it; any sweep or cap costs more than it saves.

agent/agentlocal/mock_client_test.go
JiriCtvrtka: You're right about the description — I've corrected it. On the diff itself: I ran go tool mockery --config .mockery.yaml and the method body is identical to generated output; only import grouping and brace style differ, and make gen runs make format after mockery, which produces the committed style. So no hidden divergence. Same-shaped edit applied to mock_supervisor_test.go for the new SetState signature.

agent/client/channel/rta_channel.go
JiriCtvrtka: I don't think this chain exists: RTAChannel.Wait() has no callers in the repo, and Run's deferred sharedConn.Close() isn't gated on it. close() is only reached from Send's error path, where the stream is already broken, so CloseAndRecv fails either way. The reorder is also what lets runHealthPing exit instead of blocking on sendM behind CloseAndRecv. Happy to revisit if I'm missing a caller.

agent/agents/supervisor/supervisor.go
ademidoff: Verified at 82f77aaf1: `GOMAXPROCS=1 go test ./agent/agents/supervisor/ -run TestStartProcessFail -race -count=40` is `ok` in 2.9s. `cancel()` ahead of the wait and `waitForwarderDrained` before returning is the right shape.

Worth keeping in mind that the 3s bound makes it a mitigation rather than an elimination: the forwarder can only drain if something is taking from `s.changes`, so on the wedged connection this PR is about, the terminal statuses are still dropped — just with a debug line and no hang. Fine by me, the comment says as much.

agent/client/channel/channel.go
ademidoff: Verified. The one-slot channel plus the goroutine in `close` is a better fix than guarding the call site.

Checked it directly with a throwaway in-package test — parked a sender on `sendM`, called `close(err)` from a goroutine and waited: it returns immediately and `closeWait` is closed, where the `sync.Mutex` version would have blocked there forever.

agent/client/client.go
ademidoff: Verified. `clientCtx` is per-connection and cancelled at agent/commands/run.go:119 before the next `Run`, so the re-check at agent/agents/supervisor/supervisor.go:287 does catch the stale caller once it gets `s.rw`. Keeping the goroutine out of the WaitGroup and bouncing the call instead is the better trade.

agent/agents/supervisor/supervisor.go
ademidoff: Verified — per-phase budgets in `SetState`, `RestartAgents` and `stopAll`.

agent/agents/supervisor/supervisor.go
ademidoff: Verified. Tying retirement to the forwarder's exit rather than to a status value is the right invariant — `retireInstance` guarded on equality also covers the replacement case cleanly.

agent/client/channel/channel.go
ademidoff: Verified — the send is off the calling goroutine now, so the select actually bounds it, and `acquireSendLock`'s free-lock fast path plus the grace keeps a healthy connection from paying for it. `TestSendDoesNotQueueBehindAParkedSender` and `TestSendTakesALockFreedDuringTheGrace` cover both halves.

agent/agents/supervisor/supervisor.go
ademidoff: Verified — only the status is muted now, the buckets still forward.

agent/agents/supervisor/supervisor.go
ademidoff: Agreed on deferring — per-instance directories are a bigger change than belongs here. Could you link the follow-up ticket in the PR description so it doesn't get lost?

One thing for that ticket rather than this PR: the new `releaseAgentResources` has the same collision in the other direction. `isAgentRecreated` releases `s.rw` before `releasePortAndTempDir` reaches `os.RemoveAll` at agent/agents/supervisor/supervisor.go:470, so a replacement created in that window loses its freshly rendered TLS files. The window is microseconds wide, so it is not worth a fix on its own — but it is the same problem as this one and both go away with per-instance directories.

agent/agents/supervisor/supervisor.go
ademidoff: The final state of this is better than the reply describes — 82f77aaf1 does not skip the goroutine on `port == 0`, it makes the goroutine useful: it re-decides with `isAgentRecreated` once the Agent stops and actually removes the directory unless a replacement owns it. That closes the key-material half of this, which "waits for the next pmm-agent start" would not have.

Residual race noted on the temp-dir thread; it belongs to the follow-up, not here.

agent/client/channel/channel.go
ademidoff: Fair enough, and `deliver`'s `subscriptionAbandoned` branch clears the marker whenever a response does arrive, so the growth needs a response that never comes at all. Leaving it is fine.

One correction to the reasoning: the 2-minute teardown keys off a full request queue, not off a wedged send lock, so it is not guaranteed to bound this. Doesn't change the conclusion at this volume.

agent/client/client_test.go
ademidoff: Verified at agent/client/client_test.go:380, :456 and :511.

agent/agentlocal/mock_client_test.go
ademidoff: Checked the substance and you're right: `go tool mockery --config .mockery.yaml` on the branch leaves agent/agentlocal/mock_client_test.go and agent/client/mock_supervisor_test.go differing only in import grouping and brace style — the same noise untouched mocks show (agent/versioner/mock_exec_functions_test.go regenerates with the same brace diff). No hidden divergence, and the `SetState(ctx, state)` body matches generated output.

The description doesn't look corrected though — it still reads "`make gen` inputs are untouched — no `.proto`, reform model or mocked-interface changes", which is the sentence in question. Maybe the edit didn't save?

agent/client/channel/channel.go
ademidoff: Verified — and good catch that the arm-on-failure variant changes the select's preference. The plain hoist is the right call.

agent/client/channel/rta_channel.go
ademidoff: You're right and I was wrong — `RTAChannel.Wait()` has no non-test callers, and `Run`'s deferred `sharedConn.Close()` isn't gated on it, so the chain I described doesn't exist. Withdrawing this one; the reorder is what lets `runHealthPing` exit, which is worth more than the hypothetical log line.

agent/client/channel/channel.go
ademidoff: Verified — the cancel is scoped to the `Unimplemented` status now, and the server-request path gets a redacted warning instead of an error per message.

4nte (approved):

agent/agents/supervisor/supervisor.go
JiriCtvrtka: Ticket is: https://perconadev.atlassian.net/browse/PMM-15549, linked in the description under Follow-up.

One thing to take off that ticket though: the residual race you describe is already fixed. You reviewed 82f77aaf1, and 49e48a6323 replaced isAgentRecreated with isAgentTrackedLocked — the decision and the removal now happen under one hold of s.rw (agent/agents/supervisor/supervisor.go:443-453), which excludes a replacement because one can only be started with s.rw held for writing. The port is released before that and off the lock, since it needs no agreement about who owns the ID.

StillRunningTakesTheLockBeforeDecidingOnTheDirectory guards the lock discipline. It does not guard the atomicity — I checked by reinstating the split-lock version and the test still passes, because the goroutine blocks on RLock until the replacement is registered either way. Forcing that interleaving needs a seam between the two halves, which I didn't want to add to production code, so the fix is correct by construction rather than by test. Flagging it so it isn't mistaken for covered.

The rewriting direction is untouched and is what the ticket is for.

agent/agentlocal/mock_client_test.go
JiriCtvrtka: You were right that it hadn't saved — the description is updated now. It names both changed interfaces (agent/agentlocal client, agent/client supervisor), says the mocks come from make gen, and drops the "no mocked-interface changes" claim. Thanks for checking the substance independently; I got the same result — agent/versioner/mock_exec_functions_test.go regenerates with the same brace diff, so it's formatter noise rather than divergence.

agent/client/client_test.go
```diff
@@ -293,6 +296,343 @@ func TestUnexpectedActionType(t *testing.T) {
...
+		err = stream.Send(&agentv1.ServerMessage{
+			Id:      msg.Id,
+			Payload: (&agentv1.Pong{CurrentTime: timestamppb.Now()}).ServerMessageResponsePayload(),
+		})
+		require.NoError(t, err)
+
+		// Idle, with the stream healthy, until the client tears it down itself.
+		_, err = stream.Recv()
+		require.Error(t, err)
+
```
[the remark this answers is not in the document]
JiriCtvrtka: Fixed in 58acd6c004, and it caught a gap in an earlier fix of mine.

ademidoff:
Pushed two small follow-ups from the review (2e4cf29, c6a55e8). Both are pre-existing on `main` rather than regressions from this PR, but they sit in code this PR rewrites and were cheap to fold in here rather than leave behind. Happy to drop either if you would rather keep the diff to PMM-15431 proper.

**`tryStartProcess` leaked a port reservation per failed attempt.** Clearing `port` before the retry is the last reference to the port that attempt reserved, and `portsRegistry` only forgets a port on `Release` — so three attempts against an Agent that cannot start burn three ports for the life of the process. `TestStartProcessFail` shows it directly: its three attempts run on 65100, 65101 and 65102, and none of them comes back.

Only a port reserved by the attempt itself is given back. One passed in by the caller belongs to that caller's Agent entry, which goes on referring to it when every attempt fails, so releasing it would leave a live entry pointing at a port the registry can hand to somebody else. `TestFailedStartAndPorts` covers both directions, and each half is load-bearing:

| Revert | Failure |
|---|---|
| the `releasePort` call | `ReleasesAPortItReserved` — 65200, 65201 and 65202 all stay reserved |
| the `reserved` guard, releasing unconditionally | `KeepsAPortItWasGiven` — the port the caller still refers to is released |

**`handleNomadAgent`'s `done` channel was dead.** Function-local, with `make` and `close` as its only uses, so it cannot affect anything. It reads as if it mattered, though, right next to the new instance and forwarder bookkeeping: the Agent's real `done` is the status forwarder's, in `processInfo`, and that one must not be closed here because the forwarder is still live.

The function had no coverage at all, so the test goes in before the deletion rather than after — `TestHandleNomadAgent`, 0% to 85.7%, pinning that invariant along with the Agent staying tracked and its instance staying current:

| Revert | Failure |
|---|---|
| the tracking assignment | `TestHandleNomadAgent` — the Agent that cannot start is no longer tracked |
| retirement back into `storeLastStatus`'s `DONE` branch | `TestHandleNomadAgent` — the live forwarder's later statuses are muted |

The obvious wrong fix, closing `processInfo.done` instead, does not compile — the field is `<-chan struct{}`.

Gates on the pushed tree: `go build ./agent/...`; `go test -race` (also `-count=4 -shuffle=on`) on `agents/supervisor`, `client`, `client/channel` and `agentlocal`; `golangci-lint` with CI's `--new-from-rev` scoping, 0 issues; `gofmt` clean. No generated files and no `.proto` touched.

---

ademidoff (approved):

agent/AGENTS.md
```diff
@@ -25,6 +25,32 @@ The `client` package maintains a persistent bidirectional gRPC stream (`Agent.Co
...
 - A separate RTA channel streams `CollectRequest` data via client-streaming RPC
 
+#### The request loop must never block indefinitely
+
+`client.processChannelRequests` is the only reader of `channel.Requests()`, and `channel.runReceiver`
+is the only goroutine that reads the stream — so it is also the only one that can deliver a response
+the loop is waiting for. A handler that waits on the server therefore forms a cycle: the loop stops
+draining the queue, the full queue stops the receiver, and the receiver can no longer deliver what
+the handler waits for. Nothing breaks that from inside; the agent goes silent while looking healthy
+(PMM-15431 on the agent, PMM-15310 on the server).
```
ademidoff: Let's not commit any references to tickets. AGENTS.md is supposed to set the rules, no matter where they came from.

This edit reads more as PR comments rather than a new rule that can be posted to AGENTS.md.

ademidoff: I updated that part, so it now looks clean.

JiriCtvrtka: Thanks for improvements :)

PR #5906, opened 2026-09-08
PMM-15478 Report SEP enablement in server settings

yyyyyyyan:
Ticket number: [PMM-15478](https://perconadev.atlassian.net/browse/PMM-15478)

## What

Populate `sep_enabled` on `/v1/server/settings` and `/v1/server/settings/readonly` from `PMM_ENABLE_SEP`.

This is the backend follow-up #5876 named in its own description ("Consumes `sep_enabled` … once the backend exposes it; until then the UI treats SEP as disabled"). It was never written, so the flag stayed permanently falsy and SEP was unreachable in every build carrying that PR: no Management section in the sidebar at all — the group holds only the SEP apps — and direct navigation to a SEP route rendered "This feature is not enabled. Contact your administrator.", with a healthy side-car. Two people hit this on the current PMM+SEP feature build.

## Changes

- Add `sep_enabled` to `Settings` (tag 21) and `ReadOnlySettings` (tag 9); additive, `buf breaking` passes
- Add `env.EnableSEP` and `env.SEPEnabled()`; take `PMM_ENABLE_SEP` out of the settings parser's stored path
- Populate the field in `convertSettings` and `convertReadOnlySettings`
- Cover `SEPEnabled` value handling and the read-only converter with tests

## Notes

**Not persisted, and not settable.** SEP enablement describes how the process was started — the entrypoint renders SEP's nginx locations and brings up the side-car only when the variable is set. Routing it through `ChangeSettingsParams` (the path `EnableUpdates` takes) would write it to the settings table via `models.UpdateSettings`, and a nil pointer on a later start means "no change", so a deployment that dropped the variable would keep reporting `sep_enabled: true` after its SEP surface was gone. `enable_internal_pg_qan` is the precedent for computing a response field from live state instead.

**`SEPEnabled` deliberately does not use `env.GetBool`.** The entrypoint's `is_enabled` accepts only `1` and `true`. Accepting the wider set `strconv.ParseBool` allows would let `PMM_ENABLE_SEP=True` report SEP as enabled while no nginx location block exists — the UI would offer links nginx cannot route. The test covers `True` and `yes` as disabled for exactly this reason.

**No UI change is required.** `sepEnabled` already exists on `ReadonlySettings`, and `axios-case-converter` maps `sep_enabled` onto it.

## Verification

`buf breaking` and `buf lint` pass; `golangci-lint --new-from-rev=origin/PMM-15205-sep-fb` reports 0 issues; license check clean; `go build ./api/... ./managed/...` and the new tests pass; `go mod tidy` is a no-op.

Not yet exercised on a live server — this needs a new feature build off `PMM-15205-sep-fb` to confirm the Management section returns with `PMM_ENABLE_SEP=1` and stays hidden without it. Flagging that rather than claiming it.

- [ ] API Docs updated

- Completes the follow-up named in #5876 (PMM-15281)
- Related proxy work: PMM-15279 / #5759 (`PMM_ENABLE_SEP`, the `/sep/` location block)

[PMM-15478]: https://perconadev.atlassian.net/browse/PMM-15478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/server/v1/json/client/server_service/change_settings_responses.go
+++ b/api/server/v1/json/client/server_service/change_settings_responses.go
@@ -983,6 +983,10 @@ type ChangeSettingsOKBodySettings struct {
 	// True if Query Analytics for PMM's internal PG database is enabled.
 	EnableInternalPgQAN bool `json:"enable_internal_pg_qan,omitempty"`
 
+	// True if the SEP integration is enabled. Read-only: it reports how this PMM
+	// Server process was started, and cannot be changed through ChangeSettings.
+	SepEnabled bool `json:"sep_enabled,omitempty"`
+
 	// advisor run intervals
 	AdvisorRunIntervals *ChangeSettingsOKBodySettingsAdvisorRunIntervals `json:"advisor_run_intervals,omitempty"`
 
--- a/api/server/v1/json/client/server_service/get_read_only_settings_responses.go
+++ b/api/server/v1/json/client/server_service/get_read_only_settings_responses.go
@@ -707,6 +707,9 @@ type GetReadOnlySettingsOKBodySettings struct {
 
 	// True if Access Control is enabled.
 	EnableAccessControl bool `json:"enable_access_control,omitempty"`
+
+	// True if the SEP integration is enabled.
+	SepEnabled bool `json:"sep_enabled,omitempty"`
 }
 
 // Validate validates this get read only settings OK body settings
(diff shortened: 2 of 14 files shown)


yyyyyyyan:
@coderabbitai full review

PR #5923, opened 2026-09-10
PMM-15379 Fix LBAC bypass and gate the VM proxy paths

ademidoff:
Ticket number: PMM-15379

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/4563

## What

Two independent gaps on the Grafana data source surface, both covered here.

**1. LBAC filters were keyed to the wrong thing, twice.** `shallAddLBACFilters` prefix-matched only `/graph/api/datasources/proxy/1/api/v1/`, while Grafana serves the same data source under four route shapes — `proxy/<id>/`, `proxy/uid/<uid>/`, `<id>/resources/` and `uid/<uid>/resources/`. The shapes left out never got the `X-Proxy-Filter` header and vmproxy forwarded the query unfiltered, so a Viewer restricted to a label set read every metric by changing the URL. The match also ran against the raw `X-Original-Uri`, which nginx passes percent-encoded, where authentication ran against the cleaned path — so a single `%2F` evaded whatever was listed. Both are fixed: the prefix is now the whole `/graph/api/datasources/` tree, covering every shape and anything Grafana registers under it, and `ServeHTTP` cleans the URI once so authentication and filtering decide on the same value.

**2. Any dashboard user could reach VictoriaMetrics' non-query endpoints.** Grafana forwards whatever sub-path it is given to the data source, and nothing gated which ones. `snapshot/create`, `/metrics`, `/flags`, `/debug/pprof/*`, `/api/v1/status/config`, `/api/v1/targets` and the `/api/v1/admin/*` surface were all reachable. vmproxy now forwards only the query endpoints and refuses the rest with 403, logging the denied path.

The second fix lives in vmproxy deliberately. It sits below Grafana, so it sees only the rewritten backend path and covers every proxy URL form — including `/graph/api/datasources/uid/<uid>/resources/...`, which shadows the older proxy route — rather than having to enumerate Grafana's URL shapes, which is the dependency that caused defect 1. It is also unconditional, where LBAC filtering only applies when access control is enabled, and that is off by default. The trade-off and the alternative (a role-aware gate in `auth_server`) are written up on the ticket.

Admins keep the full surface through the admin-gated `/prometheus/*` nginx location, which goes straight to VictoriaMetrics without crossing vmproxy. Metric ingestion is unaffected: vmagent writes to `/victoriametrics/api/v1/write`, which nginx exact-matches and sends direct to VictoriaMetrics.

## Verification

Live, against `perconalab/pmm-server:3-dev-latest` (3.10.0) with both binaries hot-swapped in.

Defect 1 — Viewer holding a role with filter `environment="nonexistent-env"`, which matches nothing:

| Request | Before | After |
|---|---|---|
| `proxy/1/api/v1/query?query=up` | 0 series | 0 series |
| `proxy/uid/<uid>/api/v1/query?query=up` | 12 series | **0 series** |
| `proxy/4/api/v1/query?query=up` (2nd Prometheus DS) | 12 series | **0 series** |
| `proxy/uid/<uid>/api/v1/export?match[]=up` | 5339 B | **0 B** |

Defect 2 — plain Viewer, all now 403 where they were 200: `snapshot/create`, `snapshot/list`, `metrics`, `flags`, `debug/pprof/heap`, `api/v1/status/config`, `api/v1/targets`, and `POST api/v1/admin/tsdb/delete_series`. Both through `/datasources/proxy/` and through `/datasources/uid/<uid>/resources/`.

No collateral damage — this was the main risk, so it was checked by driving the real UI rather than by inspection. Headless Chromium, admin session, 10 dashboards (Home, Node Summary, Nodes Overview, PostgreSQL Instance Summary, PostgreSQL Overview Extended, MySQL Instances Overview, MongoDB Instances Overview, Prometheus Exporter Status, VictoriaMetrics, Node Temperature) plus Explore:

- **zero non-2xx responses** across all `/graph/api/*` traffic
- 126 × `POST /graph/api/ds/query`, 79 × `resources/api/v1/query`, and every template-variable lookup (20 distinct `label/<name>/values` endpoints) returned 200
- vmproxy's deny log stayed empty for the whole sweep

Admin still gets 12 series on both proxy forms and 200 on `POST /graph/api/ds/query`; the restricted viewer's `/v1/qan/metrics:getFilters` still returns 200; `/prometheus/snapshot/list` still returns 200 for admin and 403 for viewers.

Review round — the wider prefix, the cleaned-path matching and the target endpoints were verified the same way, on a fresh server with both binaries rebuilt from this branch and the updated nginx config deployed (without it the admin marker is inert). Same Viewer, filter `{environment="nonexistent-env"}`:

| Request | Before | After |
|---|---|---|
| `1/resources/api/v1/query?query=up` | 12 series | **0 series** |
| `uid/<uid>/resources/api/v1/query?query=up` | 0 series | 0 series |
| `datasources%2Fproxy/1/api/v1/query?query=up` | 12 series | **0 series** |
| `proxy/1/api/v1/targets` | 200, 12 targets with `__address__` | **403** |
| `proxy/1/targets` | 200 HTML | **403** |

The escaped uid form (`datasources%2Fuid%2F<uid>/resources/...`) and `api/v1/series`/`api/v1/export` behave the same. Admin is unchanged: 12 series on every form, 200 on `/prometheus/api/v1/targets`, `/victoriametrics/targets` and `/victoriametrics/api/v1/targets` — which now rest on the admin marker alone, confirmed by `/victoriametrics/api/v1/status/config` returning 200 although it is not in `readOnlyPaths`. `/api/v1/status/tsdb` stays allow-listed because VictoriaMetrics does filter it: the Viewer sees `totalSeries: 0` against the admin's 12390. `POST /graph/api/ds/query` returns 12 frames for the admin and an empty frame for the Viewer, and vmproxy's deny log recorded nothing but the probes themselves.

Unit tests: the reported paths plus `export` and `snapshot/create` added to the table in `managed/services/grafana/auth_server_test.go`, full-access roles asserted to still emit no header on the proxy paths, and allow-list coverage in `vmproxy/proxy/proxy_test.go` including path traversal and both nginx path shapes. `golangci-lint` is clean on both packages (including pre-existing findings in the touched vmproxy files). The review round adds the `<id>/resources/` and `%2F` forms to that table — fed through `cleanPath`, as the handler does — and moves the target endpoints to the marker-gated set in `vmproxy/proxy/proxy_test.go`.

## Notes for reviewers

- Adding a path to `readOnlyPaths` widens what any dashboard user can reach — it should be a read endpoint.
- `vmproxy/AGENTS.md` is updated: the guide previously said the proxy should carry no business logic, and an access policy is business logic, so the constraint is restated rather than silently broken.


managed/services/grafana/auth_server_test.go
```diff
@@ -286,6 +286,12 @@ func TestAuthServerAddVMGatewayToken(t *testing.T) {
 			"/graph/api/ds/query":        true,
 			"/v1/qan/metrics:getFilters": true,
 			"/v1/qan/query:exists":       true,
+			// Every data source proxy form must be filtered, whatever the id and sub-path.
+			"/graph/api/datasources/proxy/1/api/v1/query":                     true,
+			"/graph/api/datasources/proxy/137/api/v1/query":                   true,
+			"/graph/api/datasources/proxy/uid/PA58DA793C7250F1B/api/v1/query": true,
+			"/graph/api/datasources/proxy/1/api/v1/export":                    true,
+			"/graph/api/datasources/proxy/1/snapshot/create":                  true,
```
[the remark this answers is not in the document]
ademidoff: Good catch on the gap, but the suggested diff doesn't work as written — adding `1339` to that loop fails the test.

The matrix asserts on `shallAdd`, which is keyed to the URI and not the user, so every `shallAdd: true` row expects a non-empty header for *every* user in the loop. User `1339` holds `roleC`, whose filter is empty (full access), so it always gets an empty header — the change would fail all nine pre-existing `true` rows as well as the new ones, not only the proxy paths.

The underlying point is valid though, so I covered it in the place that already asserts this property: the "shall not add any filters if at least one role has full access" subtest now runs user `1339` over `/prometheus/api/v1/` plus both new proxy forms, asserting an empty header for each. Fixed in 42b95fd45.

Worth noting for anyone reading later that this is a regression guard on full-access behaviour, not a bypass test — the bypass itself is covered by the `shallAdd: true` rows for users `1337`/`1338`.

managed/services/grafana/auth_server.go
```diff
@@ -148,12 +148,23 @@ var lbacPrefixes = []string{
 	// "/graph/api/v1/labels", // Note: this path appears not to be used in Grafana
 	"/prometheus/api/v1/",
 	"/v1/qan/",
-	// https://github.com/grafana/grafana/blob/146c3120a79e71e9a4836ddf1e1dc104854c7851/public/app/core/utils/query.ts#L35
-	"/graph/api/datasources/proxy/1/api/v1/",
+	// Grafana serves the data source proxy under both /proxy/<id>/ and /proxy/uid/<uid>/,
+	// so match the whole family: a narrower prefix leaves the other forms unfiltered.
+	"/graph/api/datasources/proxy/",
```
JiriCtvrtka: Blocker: the prefix family is still incomplete. Grafana serves data source
resources under both forms:

  /api/datasources/uid/:uid/resources/*   <- covered by "/graph/api/datasources/uid"
  /api/datasources/:id/resources/*        <- not covered by anything here

/graph is mapped to `none` in rules, so pmm-managed does no role check on it and
the only thing standing between a Viewer and unfiltered data is this list. A
filtered Viewer gets raw series from:

  /graph/api/datasources/1/resources/api/v1/series?match[]=up

Same class of bypass as the proxy/:id one this PR fixes, registered in the same
Grafana route block. Please add the id form (and consider matching
"/graph/api/datasources/" plus a check for the /resources/ and /proxy/ segments,
so the next route Grafana adds under that tree is covered by default).

JiriCtvrtka: Second blocker: on this list, separate from the missing resources route: these
prefixes are matched against the raw URI, so any of them can be evaded with one
%2F.

shallAddLBACFilters (line 389, unchanged by this PR) tests
strings.HasPrefix(req.URL.Path, p), and extractOriginalRequest (line 467) set
req.URL.Path to the raw $request_uri -- still percent-encoded, query string
included. authenticate unescapes first via cleanPath (line 537), so the two
disagree:

  GET /graph/api/datasources%2Fproxy/1/api/v1/query?query=up

  authenticate:        cleanPath -> /graph/api/datasources/proxy/1/... -> authorized
  shallAddLBACFilters: raw %2F    -> no prefix match -> no X-Proxy-Filter
  nginx /graph:        proxy_pass with no URI part -> raw URI forwarded as-is
  Grafana:             Go decodes URL.Path -> route matches -> unfiltered query

Since this PR is widening the prefix to close exactly this kind of gap, the
matching needs fixing in the same change or the widened prefix is still
evadable. Either clean once in extractOriginalRequest, or pass
maybeAddLBACFilters the cleanedPath authenticate already computed instead of
re-deriving from the raw value.

vmproxy/proxy/proxy.go
```diff
@@ -26,12 +26,49 @@ import (
...
+	"/api/v1/export":           {},
+
+	// Documented admin diagnostics, reached through the admin-gated /prometheus and
+	// /victoriametrics nginx locations, which also cross this proxy. They stay open
+	// because refusing them breaks a documented feature while protecting nothing: the
+	// labels they expose are the same ones every dashboard user already enumerates
+	// through /api/v1/label/<name>/values, which template variables depend on. Neither
+	// carries credentials. The scrape configuration is not among them: it names every
+	// target and its listen port, and stays admin-only.
+	"/targets":            {},
```
JiriCtvrtka: Blocker: these two entries are allow-listed for requests with no admin marker,
so they are reachable by any Viewer through the Grafana data source proxy --
Grafana sends unprefixed paths, which match these keys exactly. VM ignores
extra_filters[] on both, so LBAC does not apply: a filtered Viewer enumerates
the entire scrape-target inventory, every node and its listen port.

This also contradicts the PR description, which states api/v1/targets returns
403 for a Viewer.

The comment above (lines 56-62) argues these expose only labels dashboard users
already enumerate, but /targets is the exception it names -- it returns
discoveredLabels including __address__, i.e. the scrape configuration the
comment says stays admin-only.

Both entries look redundant: admin routes now carry X-Proxy-Admin, and PMM's own
callers (vmalert, pmm-managed) reach VM directly rather than through this proxy.
Dropping them restores the described behaviour. /api/v1/status/tsdb on line 65
can stay -- VM does apply extra filters there.

managed/services/grafana/auth_server.go
ademidoff: Confirmed and fixed in 4060e5d3e.

Reproduced on a fresh server with this branch's binaries hot-swapped in: a Viewer holding filter `{environment="nonexistent-env"}`, which matches nothing, got the full 12 series from `/graph/api/datasources/1/resources/api/v1/query?query=up` — byte-identical to what the admin gets — while `proxy/1/`, `proxy/uid/<uid>/` and `uid/<uid>/resources/` all returned 0.

Took your parenthetical rather than adding the one missing form: `lbacPrefixes` now carries `/graph/api/datasources/` instead of enumerating shapes, so `proxy/<id>/`, `proxy/uid/<uid>/`, `<id>/resources/` and `uid/<uid>/resources/` are covered at once and anything Grafana registers under that tree is filtered by default. The header is only ever consumed by vmproxy, so matching the whole tree costs a role lookup on data source CRUD calls and nothing else.

After the fix all four forms return 0 series for the same Viewer, `api/v1/series?match[]=up` and `api/v1/export` come back empty through both, the admin still gets 12, and `POST /graph/api/ds/query` still returns 12 frames for the admin against an empty frame for the Viewer.

ademidoff: Confirmed, fixed in the same commit 4060e5d3e.

Both halves check out: nginx passes the raw URI (`proxy_set_header X-Original-Uri $request_uri`, `build/ansible/roles/nginx/files/conf.d/pmm.conf:142`), and Grafana routes the decoded path — `GET /graph/api/datasources%2Fproxy/1/api/v1/query?query=up` returned all 12 series to a Viewer whose filter matches nothing, same as the unescaped form does for an admin.

Took the second option: `ServeHTTP` cleans once and passes `cleanedPath` to `authenticate` and `maybeAddLBACFilters` alike, so nothing re-derives it from the raw value.

One thing the unit test caught while doing it: `path.Clean` drops the trailing slash, so a prefix like `/prometheus/api/v1/` stopped matching a request for exactly that directory. `shallAddLBACFilters` now matches the directory itself as well as the tree below it.

Verified after: `datasources%2Fproxy/1/api/v1/query` and `datasources%2Fuid%2F<uid>/resources/api/v1/query` both return 0 series for that Viewer, and the escaped forms are in the table in `managed/services/grafana/auth_server_test.go` so the test feeds the raw URI through `cleanPath` the way the handler does.

vmproxy/proxy/proxy.go
ademidoff: Agreed on every point — both entries dropped in 3a4246ae7.

You are right that they went stale the moment the marker started granting the full surface: the commit that added them predates it, and the comment above them argued a case that no longer needed making. Measured on a fresh server with this branch in place, a Viewer restricted to a label set got 200 with 12 `activeTargets`, `discoveredLabels.__address__ = 127.0.0.1:3000` and `scrapeUrls` carrying every listen port, through `proxy/<id>/api/v1/targets`, `proxy/<id>/targets` and `<id>/resources/api/v1/targets` alike. All three are 403 now.

Admins are unaffected, which I checked rather than assumed: `/victoriametrics/api/v1/status/config` returns 200 for an admin **although it has never been in `readOnlyPaths`**, so the marker alone carries them; `/prometheus/targets` never crosses the proxy; and `/prometheus/api/v1/targets`, `/victoriametrics/targets` and `/victoriametrics/api/v1/targets` are all still 200.

`/api/v1/status/tsdb` stays, and VictoriaMetrics does filter it as you said — same Viewer sees `totalSeries: 0` where the admin sees 12390. The comment now says so instead of lumping it in with the target endpoints.

The PR description line you flagged is accurate again: `api/v1/targets` is 403 for a Viewer.

vmproxy/proxy/proxy.go
```diff
@@ -91,6 +130,77 @@ func getHandler(cfg Config) http.HandlerFunc {
...
 }
 
+// failOnDisallowedPath answers the request with 403 and reports whether it did, so the
+// caller returns instead of proxying a path VictoriaMetrics must not be asked for.
+func failOnDisallowedPath(rw http.ResponseWriter, req *http.Request, adminHeaderName string) bool {
+	if isPathAllowed(req.URL.Path, isAdminRequest(req, adminHeaderName)) {
+		return false
+	}
+
+	// Logged so a legitimate path missing from the allow-list can be found from the
```
JiriCtvrtka, line 140: Blocker: isPathAllowed validates path.Clean(req.URL.Path), but prepareRequest then rebuilds the forwarded path with target.Parse(...), which re-parses the already-decoded path — so a %23 (or %3F) becomes a real #/? and truncates everything after it. The string that gets validated is not the string that gets forwarded. I confirmed it against a recording backend: /metrics%23/../api/v1/query returns 200 with the backend seeing /metrics, and with more .. segments the same trick reaches /debug/pprof/heap, /api/v1/status/config, and /api/v1/admin/tsdb/delete_series — no admin marker needed, reachable by a Viewer through /graph/api/datasources/proxy/<id>/... since nginx's location /graph passes the raw request URI through.

ademidoff: Confirmed and fixed in d5f844539 — thank you, this was a real bypass and the diagnosis was exact.

Reproduced it twice: against a recording backend, `/metrics%23/../api/v1/query` passed the allow-list as `/api/v1/query` and the backend saw `/prometheus/metrics`; then end-to-end on a server running this branch, as a Viewer holding an LBAC role, through `/graph/api/datasources/proxy/1/...` — Grafana passes the `%23` through untouched, which I read back out of vmproxy's own debug log:

| Payload | Before | After |
|---|---|---|
| `/metrics%23/../api/v1/query` | 200, 84 KB of VM metrics | 422 `missing "query" arg` |
| `/metrics%3F/../api/v1/query` | 200, same | 422 |
| `/flags%23/../api/v1/query` | 200 | 422 |
| `/debug/pprof/heap%23/../../../api/v1/query` | 200, 25 KB heap profile | 422 |
| `/api/v1/status/config%23/../../../../api/v1/query` | 200, full scrape config | 422 |
| `/api/v1/query%23/../../metrics` | — | 403 |

The 422 is VictoriaMetrics answering `/api/v1/query` without a `query` parameter, i.e. the request now lands on exactly the path the allow-list judged.

**One thing worth adding to the record: it reached writes.** `POST .../api/v1/admin/tsdb/delete_series%23/../../../../../api/v1/query?match[]=up{job="clickhouse"}` was forwarded verbatim to `/prometheus/api/v1/admin/tsdb/delete_series`. With access control **enabled** it returned 204 having deleted nothing, but only because the injected `extra_filters[]` happened to match nothing — that is not protection. With access control **disabled**, which is the default, the series was really gone: `/api/v1/series` returned it before the request and `[]` after. After the fix the same request is a 422 and the series survives.

**The fix:** normalize once in the handler, judge that value, and forward that value — `path.Join(target.Path, cleanedPath)` with `RawPath` cleared, instead of `target.Parse` on an already-decoded string. `isPathAllowed` now takes the normalized path so there is a single normalization point, and the deny log prints the judged path.

It also removes a second-order oddity in the same lines: an incoming `/prometheus/api/v1/query` used to be forwarded as `/prometheus/prometheus/api/v1/query`, which only worked because VictoriaMetrics answers under that prefix as well.

No collateral: viewer `query`, `query_range`, `series`, `labels`, `label/<name>/values`, `metadata`, `rules`, `alerts`, `query_exemplars`, `export`, `status/tsdb` and `POST /graph/api/ds/query` all still 200 (and still filtered — 12 frames for an admin, an empty frame for the restricted viewer); `targets` and `status/config` still 403 for a viewer; admin still 200 on `/prometheus/api/v1/*`, `/victoriametrics/targets` and `/victoriametrics/api/v1/status/config`. `vmproxy/proxy/proxy_test.go` gains a table over these payloads asserting the forwarded path, plus one asserting the three nginx path shapes reach the endpoint exactly once.

managed/services/grafana/auth_server.go
```diff
@@ -259,18 +270,22 @@ func (s *AuthServer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	l := s.l.WithField("req", fmt.Sprintf("%s %s", req.Method, req.URL.Path))
 	// TODO l := logger.Get(ctx) once we have it after https://jira.percona.com/browse/PMM-4326
 
+	// Unescape and clean the URI once, and decide everything below on that one value. The
+	// original URI arrives raw, so matching it here and the cleaned path there lets a
+	// single %2F authorize a request as one path and filter it as another.
+	cleanedPath, err := cleanPath(req.URL.Path)
```
JiriCtvrtka, line 279: Blocker: path traversal in cleanPath. Hard blocker. Unauthenticated read of all metrics via a crafted query string, and it defeats the LBAC prefix match this PR exists to add. Fix is small:


```
u, err := url.Parse(uri)
if err != nil { return "", err }
u.RawQuery = ""
u.Path = path.Clean(u.Path)
```

i.e. parse → drop the query → clean only u.Path, instead of cleaning the raw string first.

ademidoff: Confirmed and fixed in 6ea5b5bf7 — and it was worse than the prefix match: the same craft gets past authentication entirely.

On a server running this branch, **unauthenticated**:

```
GET /prometheus/api/v1/query?query=up                            -> 401
GET /prometheus/api/v1/query?query=up&x=/../../../ping           -> 401
GET /prometheus/api/v1/query?query=up&x=/../../../../ping        -> 200, real metrics
GET /prometheus/api/v1/query?query=up&x=/../../../../v1/readyz   -> 200, real metrics
```

Four `..` walk `/prometheus/api/v1/query?query=up&x=` back to the root, so the cleaned path is `/ping`, which maps to `none` — pmm-managed never authenticates, and nginx forwards the original URI to VictoriaMetrics regardless. And through the data source, a Viewer whose role filter matches nothing: five `..` still returned `series=0`, six or more returned `series=12`, i.e. the full unfiltered set. Both are 401 / `series=0` again after the fix.

I took the same shape as your suggestion but cut the query off the raw string rather than parsing first:

```go
if i := strings.IndexAny(uri, "?#"); i >= 0 {
    uri = uri[:i]
}
```

`url.Parse` reads a leading `//` as an authority, so `//evil/v1/server/logs.zip` would clean to `/v1/server/logs.zip` while nginx, which merges slashes, routes `/evil/v1/server/logs.zip` — a second mismatch of the same family. Cutting the string keeps `path.Clean` collapsing `//` the way nginx does. The fragment goes with the query for the same reason.

One consequence worth flagging: `path.Clean` drops a trailing slash, and until now only a following query string preserved it, so `PUT /v1/alerting/templates/?x=1` resolved to `editor` while the same URL without a query resolved to `viewer`. The fix restores the slash after cleaning, which makes both `editor` — consistent, and on the stricter side. `TestCleanPath` in `managed/services/grafana/auth_server_test.go` keeps its existing expectations (including the ClickHouse data-source URL, which relies on that trailing slash) and gains the two traversal payloads above plus a fragment one.

Checked for collateral on the same server: admin 200 across `/v1/inventory/services`, `/v1/management/services`, `/v1/alerting/templates`, `/v1/server/settings`, `/prometheus/api/v1/query`; viewer 200 on `/v1/server/version`, `POST /v1/qan/metrics:getFilters`, and every data-source route, 403 on `/v1/inventory/services`, `/v1/server/logs.zip`, `/prometheus/api/v1/query` and `api/v1/targets`; anonymous 200 on `/v1/server/readyz` and 401 on `/v1/server/version`. The `managed/services/grafana` package passes in full against a real PostgreSQL and Grafana.

I also checked the mirror case, where the separator is encoded in the path instead (`/prometheus/api/v1/query%3Fquery=up&x=/../../../../ping`). That one is not a bypass: nginx normalizes it exactly as `cleanPath` now does, so both agree it is `/ping`, and the caller gets pmm-managed's two-byte `{}` — no metrics. vmproxy refuses the shorter variants outright, logging `path=/ping`.

managed/services/grafana/auth_server.go
```diff
@@ -573,22 +590,31 @@ func (s *AuthServer) authenticate(ctx context.Context, req *http.Request, l *log
...
-	unescaped, err := url.PathUnescape(p)
+// cleanPath returns the unescaped, normalized path of the original URI, without its query
+// string.
+func cleanPath(uri string) (string, error) {
+	// Cut the query string off before anything else. Cleaning it together with the path
+	// lets a "../" inside it walk the path segments, so
+	// /prometheus/api/v1/query?x=/../../../../ping authenticates as /ping -- no role, no
+	// filters, and nginx forwards the original URI to VictoriaMetrics regardless. The
+	// fragment goes the same way; neither is part of what nginx matched a location on.
+	if i := strings.IndexAny(uri, "?#"); i >= 0 {
```
JiriCtvrtka, line 612: Blocker: The ?# cut runs before url.PathUnescape on line 605, so a percent-encoded delimiter survives it, decodes to a literal ?, and path.Clean then walks the ../ behind it. Verified against the real function: /graph/api/datasources/proxy/1/api/v1/query%3Fa=/../../../../../../graph/x → /graph/api/graph/x, role=None, lbac=false — the LBAC prefix match escaped on the exact data-source route this PR protects. %23 behaves the same. The comment on lines 596-600 states this class is handled, so it reads as closed. I could not chain it to data egress: vmproxy's normalizePath cleans before its allow-list check and refuses every resulting path I could construct, so defense-in-depth holds. Fix: cut on ?# again after unescaping, or reject a decoded ?/# — neither can legitimately appear in a path nginx matched a location on. This is the only merge blocker.

ademidoff: Right on every point, fixed in 949c8bff0 — and the chain to data egress does exist, it just needs the tail to be an allow-listed path rather than an arbitrary one.

The two layers clean different strings: pmm-managed cleans the whole `/graph/api/datasources/proxy/1/...` path, vmproxy cleans only the sub-path Grafana hands it. So a payload that pops six segments escapes the LBAC prefix here and still lands on `/api/v1/query` there. As `viewer1`, holding a filter that matches nothing:

```
GET .../proxy/1/api/v1/query?query=up                                          -> series=0
GET .../proxy/1/api/v1/query%3Fa=/../../../../../../api/v1/query?query=up      -> series=12
GET .../proxy/1/api/v1/query%23a=/../../../../../../api/v1/query?query=up      -> series=12
admin, same plain query                                                        -> series=12
```

`series=12` is the full unfiltered set. With `/api/v1/export` as the tail it is 590267 bytes of raw samples against 0 bytes for the same viewer's plain export. Four and five `..` still return 0, eight is refused by the allow-list — six and seven are the window where both cleanings agree in the attacker's favour.

I took the second of your two options, rejecting a decoded `?`/`#` rather than cutting again, because cutting after unescaping opens the mirror case: `/graph/x%3F/../../v1/server/logs.zip` would truncate to `/graph`, which is mapped to `none`, while nginx resolves the same URI to `/v1/server/logs.zip` and forwards it — unauthenticated access to an admin endpoint. nginx does forward that one (it answers 403 from the new check, not 400 from nginx), so the exchange would have been a bypass for a bypass. Refusing is 403 plus a warn line naming the path, and `cleanPath` now returns a sentinel for it.

You were right that the comment read as closed — it claimed the class while handling only the raw delimiters. It now says what it does and why.

On the risk of refusing legitimate traffic: I drove the real UI to check rather than argue it. Headless Chromium, admin session, twelve pages (Home, Node Summary, Nodes Overview, CPU, Memory, PostgreSQL Instance Summary, MySQL and MongoDB overviews, Advanced Data Exploration, VictoriaMetrics, QAN, Explore) — 1369 responses, 622 of them to `/graph/api/*`, `/v1/*` or `/prometheus/*`, **zero non-2xx**, and not one request path carrying an encoded or literal `?`/`#`. pmm-managed logged no rejections for the whole sweep.

`TestCleanPath` in `managed/services/grafana/auth_server_test.go` now asserts the sentinel for both your payload shapes, the `%23` variant, and the `/graph%3F/../` mirror case.

JiriCtvrtka (approved):
No blockers anymore, just few medium ones. Approved.

PR #5924, opened 2026-09-10
PMM-15458 Give every form field one helper text, and only one

peter-o-addo:
Ticket number: PMM-15458

## Problem

Where schema field help existed it was painted twice or three times: as the field caption/helperText, again as a FieldHelpIcon tooltip with the same string, and for Encryption again as `SectionLayout.description` prose under the section title. BoolField was the clearest case (`labelCaption={field.description}` plus `FieldHelpIcon` with the same description); text/number/file/yaml/select shells did the same via `FieldLabelWithHelp` + `helperText`.

Options such as Hardlink full backups, FTWRL guardian, Check disk space first, Logging directory and MySQL defaults file had no helper text at all. That copy is schema content (PMM-15480), not something the FE can invent.

From Pedro Fernandes' UX pass over the MySQL Backups Tech Preview, recorded as finding P5 on PMM-15380, under the PMM-15217 umbrella. Observed on a build that predates the vendored SEP sync to c16578dfe (#5886); re-checked against the current be2859f3a watermark.

## What

Rendering half only, in the vendored SEP framework (`ui/packages/sep/framework`). No MySQL-specific form pages: `SchemaDrivenPlugin` + `SchemaFormRenderer` own the forms.

- **Field components:** `field.description` is shown on one surface only: BoolField `labelCaption`; String/Integer/Float/TextArea/Yaml/DateTime/File `helperText`; SchemaSelectShell `FormHelperText`. The same string is no longer passed to `FieldHelpIcon` / `FieldLabelWithHelp`. Schema still has a single `description` field, so there is no separate "deep tooltip" channel; until one exists, no tooltip from `description`.
- **SchemaFormRenderer:** stop rendering `section.description`, so Encryption (and any other section) cannot restate field help even if the sidecar still sends the blurb.
- **Tests:** assert helper text once, no restating help icon, and section prose not shown. `FieldHelpIcon` / `FieldLabelWithHelp` unit tests kept for the unused primitive.

## Acceptance criteria

- [x] No field renders the same text as both caption and tooltip.
- [x] The Encryption section carries no prose that restates its fields' own help.
- [x] Field help follows PMM copy conventions (plain, imperative); copy itself is schema/PMM-15480; this PR only stops double-painting.
- [ ] Once PMM-15480 lands on the sidecar, every field on the backup and restore forms that carries a description shows that helper once.

## Notes

- Framework pieces under `ui/packages/sep/` are PMM-side divergence and want upstreaming to percona/percona-sep or they return at the next sync.
- Missing helper strings for Hardlink / FTWRL / disk space / logging / defaults are PMM-15480 (sidecar schema). This ticket does not author them.
- SEP-1606 had intentionally duplicated `description` into a tooltip so help survived when a validation error replaced helperText. This PR accepts that trade-off: one helper surface, no restatement.

### DIFF
--- a/ui/packages/sep/framework/src/components/SchemaFormRenderer/SchemaFormRenderer.fieldErrors.test.tsx
+++ b/ui/packages/sep/framework/src/components/SchemaFormRenderer/SchemaFormRenderer.fieldErrors.test.tsx
@@ -81,19 +81,13 @@ describe('SchemaFormRenderer field errors', () => {
     // Integer (helperText) field also surfaces its mapped error.
     expect(screen.getByText('must be positive')).toBeInTheDocument();
     expect(screen.queryByText('Max rows')).not.toBeInTheDocument();
-    // Help icons remain while helperText shows the error; pin count + visible <label>.
+    // Description is helper text only — no restating help icon while errors show.
     expect(document.querySelectorAll('[data-help-for="Title"]')).toHaveLength(
-      2
+      0
     );
     expect(
-      document.querySelectorAll('label [data-help-for="Title"]')
-    ).toHaveLength(1);
-    expect(
       document.querySelectorAll('[data-help-for="Row Limit"]')
-    ).toHaveLength(2);
-    expect(
-      document.querySelectorAll('label [data-help-for="Row Limit"]')
-    ).toHaveLength(1);
+    ).toHaveLength(0);
     // The persistent banner is rendered.
     expect(screen.getByText(/Failed/)).toBeInTheDocument();
   });
@@ -173,7 +167,6 @@ describe('SchemaFormRenderer field errors', () => {
 
     // Empty the required title field, then resubmit: the client-side gate blocks
     // the submit, so the eager clear must not drop `limit`'s inline highlight.
-    // Prefer role+name: getByLabelText(/Title/) also matches "Help for Title".
     await user.clear(screen.getByRole('textbox', { name: /Title/ }));
     await user.click(screen.getByRole('button', { name: 'Run' }));
 
--- a/ui/packages/sep/framework/src/components/SchemaFormRenderer/SchemaFormRenderer.spotCheck.test.tsx
+++ b/ui/packages/sep/framework/src/components/SchemaFormRenderer/SchemaFormRenderer.spotCheck.test.tsx
@@ -20,9 +20,8 @@
  *
  * Direct consumers: mysql_backups (via SchemaDrivenPlugin) and atw
  * (CollectPane); SnippetExecutionAccordion covers the synthesised
- * snippet-parameter path. Confirms help icons appear only on described fields
- * and core inputs still mount — a regression guard for the framework-global
- * label change.
+ * snippet-parameter path. Confirms field descriptions appear once as helper
+ * text (never as a restating help icon) and core inputs still mount.
  */
 
 import { describe, expect, it, vi } from 'vitest';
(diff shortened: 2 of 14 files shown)


yyyyyyyan (commented):
@peter-o-addo — the field-level half of this is clean, and I checked each component rather than trusting the pattern: `helperText` survives on String/Integer/Float/TextArea/Yaml/DateTime, `helperText={error ? error.message : field.description}` on File, `labelCaption` on Bool, `FormHelperText` on the select shell, and all three `ChoiceField` branches paint it exactly once. Nothing lost its help. Deleting `FieldLabelWithHelp` outright rather than leaving an orphan was the right call — nothing in `ui/` imports it any more.

I pushed one commit, 7da8713b6, rather than leave it as a suggestion. `FieldHelpIcon` was the only emitter of `data-help-for`, so once it is deleted `document.querySelectorAll('[data-help-for="…"]')` returns empty for every selector — a misspelled label, a field that never mounted, anything. The thirteen negative assertions could no longer fail, and nine of them had no positive pairing. The `getByText(description)` assertions are the real guard: `getByText` throws when a string matches more than one node, which is exactly the double-painting this work exists to prevent. Revert it if you would rather keep them.

**Section prose and group prose now follow opposite rules** — `OneOfGroupSlot.tsx:89-91`. The block deleted from `SchemaFormRenderer` and the one still in `OneOfGroupSlot` are the same markup down to the `sx`, and both arrived in 49123fb11:

```tsx
{group.description ? (
  <Typography variant="body2" color="text.secondary" sx={{ mb: 1 }}>
    {group.description}
  </Typography>
) : null}
```

The suite pins both rules in one file now: `SchemaFormRenderer.test.tsx:328` asserts section prose is absent, `:2649` asserts group prose is present. I went looking for a rule that separates them and could not find one that holds — *sections restate field help, groups label a choice* is the natural candidate, but the Encryption fixture at `:333` ("Pick an Encryption format first; the fields below are that format's parameters") labels a choice just as much as "Choose how to specify source rows" does. This is the same ground Copilot's suppressed comment covers, and I think it is right that the split needs settling. Dropping the group block too is the smaller change; keeping both surfaces and taking the Encryption copy out at the schema is the other. Either reads fine — what I would avoid is leaving them split with nothing recording why, since the next person to touch this finds two tests disagreeing.

**The vendored spec still documents the render that just went away** — `ui/packages/sep/api/specs/sep.json`, `components.schemas.framework__FormSection`: *"Optional helper text rendered beneath the section heading."* `framework__OneOfGroup` carries the same sentence for the group label. That file is dumped from the backend and guarded for freshness, so the sentence returns on every refresh regardless of what this renderer does. The durable fix is the backend docstring in percona-sep rather than anything here.



**The Tests bullet in the description is stale** — it says the `FieldHelpIcon` / `FieldLabelWithHelp` unit tests are kept for the unused primitive, but db93ad309 removed `FieldLabelWithHelp.test.tsx` along with the module. Worth correcting before merge, since the squash body picks the description up.

**Commits carry neither the key prefix nor a sign-off** — a514daccc through db93ad309. Every non-merge commit on `PMM-15205-sep-fb` has both (b784b4545, 2b0f86c9a, c5ea2b80e, 862e1872a), and squash-merge takes the title from the PR title but will not synthesise the trailer. Not worth rewriting history for — flagging it so the next one carries it.

Nothing else caught my eye: the checks are green, `Typography` is still used for the section legend so no import was orphaned, and the `userEvent` drop in `SchemaSelectShell.test.tsx` was correct.

yyyyyyyan:
Both of the things I raised are settled now.

**Why sections drop prose and groups keep it** — f314d13e5. I said I could not find a rule separating the two surfaces; that was wrong, and the distinction is structural rather than editorial. A section holds declared fields and every one carries its own `description`, so section prose can only ever restate field-level help. A one_of group renders a segmented control built from `branches`, and `discriminator` is a form-state path that `useWatch` and `setValue` address rather than a declared field — so nothing else can carry what that control means. The behaviour you shipped is right; it just was not written down anywhere. Both mirror types now say so, which is where a reader lands coming from either surface.

**The spec clause** — percona/SEP#1541. `FormSection.description` and `SectionLayout.description` both promised the text is rendered beneath the heading, which stays true of SEP's own renderer and is no longer true here. That PR loosens both to say presentation is the renderer's choice and names which one does what, so the sentence stops returning on every refresh. The vendored copy here keeps the old wording until the next sync — that is the normal path for those fixtures, and hand-patching `specs/sep.json` would only drift `src/generated/` against it.

Two small ones still yours: the Tests bullet in the description says the `FieldLabelWithHelp` unit tests were kept, and the commits carry no key prefix or sign-off.

yyyyyyyan (approved):

PR #5932, opened 2026-09-11
PMM-15490 Hide the chain-tasks input when nothing can be chained

peter-o-addo:
Ticket number: PMM-15490

## Problem
MySQL Backups and Restore showed "Chain tasks after execution (optional)" even when nothing could be chained — a disabled "Add task to chain…" select with "No tasks available to chain". It appeared on the run confirm dialog and the schedule form, offered an action the user could not take, and read as an unfinished feature.
Observed on the schedule form in the UX pass on PMM-15454; the run-dialog case follows from the same ChainBuilder used whenever an app declares chaining and has no other task to offer.

## What
One change in `ChainBuilder` covers both consumers (run confirm in PluginDetailPage, schedule form in ScheduledTaskForm):
- Render nothing when `hasSelectableTask` is false and the chain is empty (no label, select, or helper).
- Render the add-task select only when at least one task can be chained (same app, not the current task, not already in the chain — the rule ChainBuilder already owned for disabling the select).
- Keep an existing chain editable when nothing is left to add: chips, reorder, remove, and "Continue chain on failure"; no add select. Removing a task restores the select when a slot frees up.
- Drop the unreachable "No tasks available to chain" helper and its aria wiring.
Gating in each consumer was rejected: both would recompute a rule ChainBuilder already owns.

## Acceptance criteria
- [x] Run confirm: empty chain + nothing chainable → no chain UI (label/select/helper).
- [x] Schedule create/edit: same condition → no chain UI.
- [x] When something can be chained → input appears and behaves as today (add, reorder, remove, continue on failure).
- [x] Existing full chain stays editable without the add select; remove brings the select back when a task becomes available again.
- [x] Changing the task on schedule create re-evaluates visibility.
- [x] Submit while hidden carries no chain (same as empty chain today).
- [x] Run dialog loading spinner and load-error for the chainable-task list unchanged.

## Notes
- Delivers PMM-15454's "chaining is not presented while nothing can be chained" ahead of that ticket's scheduling redesign, and extends it to the run dialog (outside 15454's scheduling scope).
- Out of scope here: schedules list "Chain" column always showing "—", and moving chaining under Advanced (both PMM-15454).
- Framework change under `ui/packages/sep/` is PMM-side divergence and wants upstreaming to percona/percona-sep or it returns at the next sync.

### DIFF
--- a/ui/packages/sep/framework/src/components/ChainBuilder/ChainBuilder.test.tsx
+++ b/ui/packages/sep/framework/src/components/ChainBuilder/ChainBuilder.test.tsx
@@ -67,8 +67,28 @@ async function pickTaskFromDropdown(taskName: string) {
 }
 
 describe('ChainBuilder rendering', () => {
-  it('renders no chip sequence and no failure toggle when chain is empty', () => {
+  it('renders nothing when the chain is empty and no task can be chained', () => {
+    render(
+      <Harness
+        availableTasks={[{ name: 'task-self' }]}
+        currentTaskName="task-self"
+      />
+    );
+    expect(screen.queryByTestId('chain-builder')).not.toBeInTheDocument();
+    expect(screen.queryByRole('combobox')).not.toBeInTheDocument();
+    expect(
+      screen.queryByText(/no tasks available to chain/i)
+    ).not.toBeInTheDocument();
+  });
+
+  it('renders nothing when availableTasks is empty and the chain is empty', () => {
+    render(<Harness availableTasks={[]} />);
+    expect(screen.queryByTestId('chain-builder')).not.toBeInTheDocument();
+  });
+
+  it('renders the add select when at least one task can be chained', () => {
     render(<Harness />);
+    expect(screen.getByTestId('chain-builder')).toBeInTheDocument();
     expect(screen.queryByTestId('chain-sequence')).not.toBeInTheDocument();
     expect(
       screen.queryByTestId('chain-on-failure-checkbox')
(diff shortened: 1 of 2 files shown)


yyyyyyyan (approved):
@peter-o-addo — nice piece of work. Putting the decision inside ChainBuilder rather than gating it at both call sites is the right home for it, and the three rewritten rendering tests plus the stale-chain removal case cover exactly the transitions that are easy to get wrong.

One thing I pushed rather than left as a comment, since it falls straight out of the change: ChainBuilder can now render nothing, but both call sites still wrapped it in a Box that carried the spacing — so the margin outlives the widget, leaving a gap in the execute-confirm dialog and between the mode-toggle link and the button row on the schedule form. ChainBuilder now takes its own `sx` and the wrappers are gone. That is the shape ActionErrorAlert already uses for the same reason: it returns null too, and its callers pass `sx={{ mb: 3 }}` straight to it rather than wrapping it (PluginDetailPage.tsx:906). I merged the base branch in first, since PluginDetailPage had moved under the dialog content.

A few small things left over, none worth holding this up:

**Commit title length** — "PMM-15490 Cover ChainBuilder hide/show states in tests" runs to 54 characters; titles here stop at 50. Same for the PR title, which matters slightly more — the squash merge makes it the permanent message.

**Sign-off** — neither commit carries a Signed-off-by trailer, where most feature commits on PMM-15205-sep-fb do.

**Feature build** — the body dropped the template's "Feature build:" line. Probably worth a sanity check on whether this epic's shared build is meant to cover it; PRs onto this branch are split on whether they carry one.

Separately, and not yours to fix here: the same shape is already in TaskLogViewer.tsx:497. LogStepTabs returns null when there are no steps, but the Box around it carries `borderTop`, so a log with no steps draws a 1px rule with nothing under it.

PR #5949, opened 2026-09-15
PMM-15347: Drive MongoDB bootstrap from pmm-managed

plebioda:
Ticket number: PMM-15347

Feature build: TBD

Backend half of MongoDB bootstrap: an HA-leader-only stepper drives SEP's om_bootstrap app through its lifecycle, TriggerHostBootstrap kicks off a run (repointed at om_bootstrap once it landed), 3-host runs with progress polling, dispatch to a host's actual Nomad executor rather than its node id, a finalize phase for enabling auth after bootstrap, per-host PMM service naming to avoid collisions, pushing a pmm-agent state update after registering a bootstrapped host, ListBootstrapRuns/GetBootstrapRun for run history and detail, waiting for inventory sync before calling a host confirmed, labeling a bootstrapped service's environment/cluster, and a proactive sweep of a bootstrap's own hosts instead of waiting for the ambient inventory ticker.

Stacked on PMM-15360's OpenManager switch and gate work (#5866, #5888, #5853, bundled for testing in `PMM-15360-om-integration`) and on #5921 (SEP's paginated /hosts and /services envelope).

If this PR adds, removes or alters one or more API endpoints, please review and update the relevant API documentation as well:

- [ ] API Docs updated

- percona/SEP#1533, percona/SEP#1534: the om_bootstrap app this drives
- #5866, #5888, #5853: the OpenManager switch and gate this is stacked on
- #5921: the pagination envelope fix this is stacked on


plebioda:
@coderabbitai review

api/swagger/swagger.json
```diff
@@ -28715,6 +29250,24 @@
...
                         },
                         "x-order": 10
+                      },
+                      "pmm_agent_connected": {
+                        "description": "Whether PMM's own agent registry currently has a connected pmm-agent for this\nnode, independent of anything the probe collected. This is the \"PMM-Client\ninstalled and healthy\" signal om_inventory has no way to answer on its own: a\nnode existing in the inventory only means PMM registered it once, not that its\nagent is alive now.",
+                        "type": "boolean",
+                        "x-order": 11
+                      },
+                      "automation_eligible": {
+                        "description": "Whether this host is eligible for OM automation (probing today; provisioning in\na later phase): pmm_agent_connected is true, and executor reports reachable and\ndriver-healthy. Computed once here so every consumer agrees on one definition\nrather than recomputing it from the fields above separately.",
```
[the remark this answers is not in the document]
plebioda: Confirmed, fixed. Updated the field comment in om.proto and regenerated -- provisioning is exactly what TriggerHostBootstrap in this same PR implements, so the stale "later phase" note is gone.

managed/models/om_bootstrap_secret_helpers.go
```diff
@@ -0,0 +1,93 @@
...
+	} else {
+		secret.MongoDBPassword = password
+	}
+	keyFile, err := handler(secret.KeyFile)
+	if err != nil {
+		logrus.Warning(err)
+	} else {
+		secret.KeyFile = keyFile
+	}
+	return secret
```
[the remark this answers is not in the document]
plebioda: Confirmed, and it's a real one -- fixed. omBootstrapSecretEncryption now stops on the first encryption/decryption failure and returns an error instead of logging and keeping the untransformed field; CreateOmBootstrapSecret propagates it rather than inserting a partially-encrypted secret. This now matches utils/encryption.EncryptItems's own all-or-nothing behavior for agents.username/password, which is the pattern this was meant to follow.

managed/models/om_bootstrap_secret_reform.go
```diff
@@ -0,0 +1,151 @@
+// Code generated by gopkg.in/reform.v1. DO NOT EDIT.
```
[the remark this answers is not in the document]
plebioda: Checked -- this is a false positive. .licenserc.yaml explicitly exempts `**/*_reform.go` (every other generated reform file under managed/models/ carries no header either -- e.g. service_model_reform.go), and `go tool license-eye -c .licenserc.yaml header check` reports 0 invalid files with this one included. No change needed.

managed/services/om/stepper.go
```diff
@@ -0,0 +1,347 @@
...
+				continue
+			}
+			s.stepBootstrapRuns(ctx)
+		}
+	}
+}
+
+// stepBootstrapRuns discovers every run worth a look this tick and drives each one.
+func (s *Service) stepBootstrapRuns(ctx context.Context) {
+	for _, runStatus := range [...]string{bootstrapRunRunning, bootstrapRunSucceeded} {
```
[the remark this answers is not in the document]
plebioda: Verified the behavior, and it's deliberate: completeSucceededRun's own commit message explains why it revisits every succeeded run each tick rather than tracking completion -- a host whose registration got stuck (e.g. no state update ever pushed, before this function existed) self-heals on its next tick with no separate recovery path needed. RequestStateUpdate itself is a non-blocking, already-debounced channel send, so the per-tick cost is small. It's also not literally unbounded: SEP's own GET /runs caps what's returned per call at 100 (crud.py's own limit), so this is bounded per tick, not growing forever with deployment lifetime, though the underlying SEP table itself does grow unbounded over time (a separate, SEP-side retention question).\n\nA real persisted-completion-state mechanism is legitimate follow-up work, but it's real design work (has to preserve the self-healing property for hosts that never got a state update, and keep recovery retries bounded per your own suggestion) -- not something to rush into an already-multiply-reviewed PR. Tracking as follow-up rather than fixing here.

managed/models/database.go
```diff
@@ -83,6 +83,15 @@ var DefaultAgentEncryptionColumnsV3 = []encryption.Table{
 			{Name: "postgresql_options", CustomEncryptHandler: EncryptPostgreSQLOptionsHandler, CustomDecryptHandler: DecryptPostgreSQLOptionsHandler},
 		},
 	},
+	{
+		Name:        "om_bootstrap_secrets",
```
4nte, line 87: **Blocker: adding this table to `DefaultAgentEncryptionColumnsV3` corrupts every existing agent credential on upgrade.**

`dbEncryption` (`database.go:1403-1443`) builds `prepared` from only the columns *not* already in the recorded state, then calls `UpdateSettings(EncryptedItems: prepared)`. `settings_helpers.go:249-250` does `settings.EncryptedItems = params.EncryptedItems`, a replace rather than a merge. On an upgraded server the `agents.*` columns are already recorded, so they get skipped and `EncryptedItems` is overwritten with just the three new `om_bootstrap_secrets` columns.

I reproduced this against real PostgreSQL 14, building this branch (`c35d3abd16dd`) and its base (`611225673a05`) and driving the real `models.SetupDB` path that `migrateDB` calls on every start:

```
Fresh install on base, seed an agent:
  EncryptedItems count: 8
  models read back  username="verify-user" password="SuperSecret123"

Control, two restarts on base:   8 entries, values unchanged

Upgrade, boot 1 on this branch:  EncryptedItems 8 -> 3
Upgrade, boot 2 on this branch:
  models read back username="AQhEGYzRumY16YibYD4URhPIG3LPb6ZZU0wccZDXKcEDS4/hY0j9PrRFzpE="
```

The application reads back ciphertext. It does not stabilise either, because the state oscillates and every other restart adds another layer:

```
boot 3: EncryptedItems=3  agents.password raw length=132
boot 4: EncryptedItems=8  agents.password raw length=220
boot 6: EncryptedItems=8  agents.password raw length=340
boot 8: EncryptedItems=8  agents.password raw length=500
```

Affected columns are `agents.username`, `password`, `agent_password`, `aws_options`, `azure_options`, `mongo_options`, `mysql_options`, `postgresql_options`. Recovery needs the exact layer count, which depends on how many times the server restarted.

**Cause isolated:** I rebuilt this branch with only these 9 lines removed (migrations 120/121 still run, the tables are still created) and reran the same upgrade on a clean database. All eight boots stayed at `EncryptedItems=8` with `password="SuperSecret123"`. So it is this entry, not the schema change.

The underlying defect is the replace-instead-of-merge in `UpdateSettings`, which this PR is simply the first change in years to trip. Could we fix `EncryptedItems` to merge (or write the full expected set) before adding a table here?

managed/services/om/inventory.go
```diff
@@ -319,6 +364,364 @@ func (s *Service) TriggerInventoryRefresh(ctx context.Context, req *omv1.Trigger
...
+// phase targets (a handful of hosts in one replica set) makes one full fetch
+// no real cost -- see ListInventoryHosts's own similar fetch-then-filter
+// shape. A host with no executor at all (never dispatched an eligibility
+// probe) is silently dropped rather than keyed on empty string.
+func (s *Service) inventoryHostsByExecutor(ctx context.Context) (map[string]sepHost, error) {
+	probe, err := s.inventoryProbe()
+	if err != nil {
+		return nil, err
+	}
+	hosts := []sepHost{}
```
4nte, line 476: **This reintroduces the bare-array decode #5921 exists to fix, on the path that registers a bootstrapped mongod with PMM.**

Against a SEP serving the paginated envelope:

```
inventoryHostsByExecutor error = rpc error: code = Internal desc =
  failed to decode the inventory app's answer:
  json: cannot unmarshal object into Go value of type []om.sepHost
```

And the consequence, driving `completeSucceededRun` with a SUCCEEDED run and a stored secret:

```
services registered after a SUCCEEDED run: 0 (expected 1)
```

So a bootstrap completes on SEP's side, and PMM registers nothing while logging a warning every 15 seconds. #5951 converts this to `fetchAllPages`; could it move down here so this PR is functional on its own, or is the merge-order dependency on #5951 intentional?

managed/services/om/stepper.go
4nte, line 86: **Every succeeded run is reworked on every tick, forever.**

This lists `succeeded` runs with no limit and no "already registered" cut-off, so `completeSucceededRun` reruns indefinitely. The doc comment says "until that registration exists", but nothing implements the *until*.

Measured with 20 finished runs over 5 simulated ticks:

```
GET /runs/{id}      = 100     => 20 per 15s tick
full estate fetches = 100     => 20 per 15s tick
```

Each revisit also costs a secret lookup and decrypt, a run-config lookup, and per host a `FindPMMAgentsRunningOnNode`, a `FindServices` and a `RequestStateUpdate`. The same run emitted 100 warning lines across those 5 ticks, which in production is a continuous log flood. This grows monotonically with bootstrap history and will look fine in a PoC with one run.

Could a run be marked complete once every host is registered, so finished runs drop out of the sweep?

4nte (commented):
Reviewed this against a local build of the branch at `c35d3abd16dd`, with the findings reproduced empirically rather than read off the diff. Three inline comments below, plus one that has no good line to anchor to:

---

**Blocker: `.turbo/cache/` build cache is committed.**

30 files, 52.4 KiB, including 10 binary `.tar.zst` archives. `git check-ignore` confirms it is not ignored at the repo root; `ui/.gitignore:19` has `.turbo` but only covers `ui/`. CI does not flag it, and it is carried into #5950 and #5951 as well.

Worth dropping from the branch and adding `.turbo/` to the root `.gitignore` while it is still cheap. Once merged it is permanent in history.

---

The inline comments are, in order of severity: the encryption-list change (data loss on upgrade, cause isolated with an A/B build), the reintroduced bare-array decode, and the stepper's unbounded rework of finished runs.

plebioda:
Request from review on the SEP side (percona/SEP#1534, om_bootstrap's finalize phase): add a pre-flight Nomad connectivity check to `TriggerHostBootstrap` here.

Today this function validates each host's `os_id` and that it has a usable `ExecutorHost` (`inventory.go`'s per-host loop, ~L462-482, via `resolveBootstrapHostOSID`), but doesn't check whether that executor is actually *reachable* right now. An unreachable Nomad client only surfaces once `pre_check` -- itself dispatched through Nomad -- fails as a step, several seconds after the run already exists and is visible in the UI as "in progress".

The data's already there for the taking: `sepHost.Observed` (`inventory_client.go:80`) is the same `observed` dict SEP's own `_executor_usable` reads (`app/sep/apps/om_inventory/api_routes.py:158-172`) -- `observed["executor"]["reachable"]` and `observed["executor"]["driver_healthy"]`, both bools. `host` is already fetched in the same per-host loop `resolveBootstrapHostOSID` runs in, so this is a same-loop addition, not a new SEP call: type-assert `host.Observed["executor"]` the same way `resolveBootstrapHostOSID` type-asserts `host.Observed["os_id"]`, and reject with `codes.FailedPrecondition` (matching the existing style) when either is false, before any run is created.

Not blocking -- flagging it so it doesn't get lost, since it was raised as a real gap on the SEP PR and the right place to close it turned out to be here, not there.

managed/models/database.go
plebioda: Confirmed, and fixed by making the write mean what the field says. `dbEncryption` now adds what a pass encrypted to what was already recorded (and removes what it decrypted) rather than replacing the set with the columns it happened to touch: cf45d9a9b. I left `UpdateSettings` replacing, since decrypting has to be able to remove entries, and put the arithmetic in a pure function so it is covered without a database: `TestRecordedEncryptedItems`.

I reproduced your oscillation against real PostgreSQL 14 before changing anything, then reran the models suite on the fix - `EncryptedItems` stays at 8 across restarts with the new table's three added, so the ciphertext layering is gone.

Thank you for isolating it to those nine lines; that is what made it obvious the defect was in the write and not in the schema change.

managed/services/om/inventory.go
plebioda: Confirmed and fixed in ab6d68b08 - `inventoryHostsByExecutor` goes through `fetchAllPages` now, like `ListInventoryHosts` and the probe path.

Worth noting what let it through: the `/hosts` fixtures in `inventory_test.go` and `stepper_test.go` were bare arrays, so the tests were passing against a shape SEP does not serve. They are envelopes now, which is what actually fails if this regresses.

The copy of this fix that was sitting further up the stack in `PMM-15347-bootstrap-abort-per-member` is redundant as a result, and drops out when that PR restacks.

managed/services/om/stepper.go
plebioda: Yes - done in 8f14fa4cf, along the lines you suggested. A run whose every host is registered records that (`OmBootstrapRunConfig.RegisteredAt`) and leaves the sweep for good.

Persisted rather than remembered, because a new leader after a failover, or the same one after a restart, has to know it too. The run-config row already existed for a run triggered with an environment or cluster, so the marker joins it there and the row is upserted for a run that had none; one query per tick reads the whole registered set rather than one per run.

Two things I deliberately left as they were:

- A run with a host the inventory app has no row for yet stays in the sweep and keeps warning. That run is genuinely unfinished, which is what the warning is for - your 100 lines came from finished runs, and those are gone.
- The scoped inventory refresh is not part of the completion condition. It only asks SEP to re-probe sooner than its own schedule would, so a missed kick costs freshness, not correctness, and is not worth another pass over the estate every 15 seconds for the life of the server.

`TestCompleteSucceededRun` gained the two cases that matter: the marker is written and readable through `FindRegisteredOmBootstrapRunIDs`, and a run with an unregistered host does not get marked.

plebioda:
@4nte Both of your unanchored findings are fixed, along with the three inline ones.

**`.turbo/cache/`** - confirmed, 30 files that came in with this branch's first commit. Removed in ffa2956de. `ui/.gitignore` covers `ui/.turbo`, but turbo writes to the repo root when a task runs from there and nothing was ignoring that, so the root `.gitignore` now does.

Two other things landed in this round:

- The pre-flight Nomad connectivity check I flagged above is in (5b4b709a9): `TriggerHostBootstrap` now reads the same `observed.executor` sub-document SEP's own `_executor_usable` reads, in the loop that already fetches each host, and refuses with `FailedPrecondition` before a run is created. Only an explicit `false` rejects - a host whose sub-document is missing is left to SEP, so a PMM talking to an older SEP can still bootstrap.
- `WithBootstrapSource` was building its own `http.Client`, so the om_bootstrap client still followed redirects after the om_inventory one stopped (#5888's review). Both go through `newSEPClient` now (4b681d3be), which is the point: a third source cannot arrive without the guard.

The branch is also restacked onto the refreshed `PMM-15360-om-integration`, which now carries the four restacked PMM-15326/PMM-15360 PRs.

Everything above was verified against real PostgreSQL 14 rather than mocks: `managed/models` and `managed/services/om` both pass in full.

PR #5950, opened 2026-09-15
PMM-15347: Add bootstrap UI to the Hosts and Operations pages

plebioda:
Ticket number: PMM-15347

Feature build: TBD

UI half of MongoDB bootstrap: an Automation column and a Bootstrap dialog on the Hosts page, background-tab polling that keeps run status current even while the tab isn't focused, an Operations tab with bootstrap run history, and a fix so polling keeps going until confirm_monitoring actually finishes rather than stopping early.

Stacked on #5949 (PMM-15347: Drive MongoDB bootstrap from pmm-managed).

If this PR adds, removes or alters one or more API endpoints, please review and update the relevant API documentation as well:

- [ ] API Docs updated

- #5949: the backend this UI is built against


plebioda:
@coderabbitai review

ui/packages/plugins/om/src/HostsPage.tsx
```diff
@@ -503,6 +542,135 @@ const ForgetDialog = ({
...
+    return null;
+  }
+
+  const handleSubmit = async () => {
+    const accepted = await bootstrap.mutateAsync({
+      nodeId: row.node_id,
+      replicaSetName,
+      mongodbVersion,
+    });
+    setResult(accepted);
```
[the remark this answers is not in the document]
plebioda: Confirmed, and fixed: added currentNodeIdRef, read at resolution time rather than closed over at submit time, so a late response only calls setResult when the dialog is still showing the host that submitted it.

ui/packages/plugins/om/src/OperationsPage.tsx
```diff
@@ -0,0 +1,133 @@
...
+        title="Operations"
+        subtitle={
+          <Typography variant="body2" color="text.secondary">
+            Every bootstrap run this server has driven, newest first. Expand a
+            row to watch it live or review what happened.
+          </Typography>
+        }
+      />
+
+      {error && (
```
fabio-silva, line 104: If there's an error, we're still rendering things below, namely the table. We should return early

ui/packages/plugins/om/src/OperationsPage.tsx
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Bootstrap run history: every run this PMM server has driven, newest first,
+ * each expandable in place to its own step-by-step progress.
+ *
+ * Exists so a run started from the Hosts page is never lost by navigating away
+ * from it -- starting a run is one thing, and reading its progress back later
+ * is another that has to survive leaving the page it started on.
+ */
+export const OperationsPage = () => {
+  const { data: runs, isLoading, error } = useOmBootstrapRuns();
```
fabio-silva, line 89: there's a limit argument on this fn but none is used here. Is this expected?

ui/packages/plugins/om/src/HostsPage.tsx
```diff
@@ -591,6 +776,23 @@ export const HostsPage = () => {
             </Button>
           </Box>
         </Tooltip>
+        <Tooltip
+          title={
+            row.original.automation_eligible
+              ? 'Install MongoDB on this host and initialize a single-member replica set (PoC).'
+              : row.original.automation_blocked_reasons.join('; ')
```
fabio-silva, line 783: Some files before, I see:

```
title={
    row.automation_blocked_reasons.join('; ') ||
    'Not eligible for automation.'
}
```

ui/packages/plugins/om/src/api.ts
```diff
@@ -97,3 +101,47 @@ export async function request<T>(path: string, init?: RequestInit): Promise<T> {
...
+  }
+  if (run.status === 'running') {
+    return true;
+  }
+  if (run.status !== 'succeeded') {
+    return false;
+  }
+  return run.hosts.some((host) =>
+    host.finalize_steps.some(
+      (step) => step.name === 'confirm_monitoring' && step.status === 'running'
```
fabio-silva, line 131: if finalize steps dispatch after run succeeds, isn't there a time window where the run succeeded but `confirm_monitoring` hasn't started yet, increasing the interval to 60s and leaving stale data for a longer period?

fabio-silva (changes requested):

ui/packages/plugins/om/src/OperationsPage.tsx
plebioda: Right, and it mattered more than it looks: with no rows the empty table reads as "there are no bootstrap runs", which is a different statement from "we could not find out". Loading and error are early returns now, the same shape `HostsPage` already uses for those two states. d440988d0

ui/packages/plugins/om/src/OperationsPage.tsx
plebioda: Not expected, no - the argument was dead and the page was quietly getting the server's default of 20 runs with pagination disabled, so a longer history was cut off with nothing saying so.

It now asks for 100, which is the most the API will serve (`ListInventoryRunsRequest.limit` is validated `lte: 100`, and pmm-managed's `maxInventoryRunLimit` caps it again), with the reason at the call site. d440988d0

ui/packages/plugins/om/src/HostsPage.tsx
plebioda: Good catch - the two had drifted. The Automation cell falls back to "Not eligible for automation." when SEP gives no reasons, the Bootstrap button's tooltip did not, so a host that reads ineligible with an empty reason list gave MUI an empty title and no tooltip rendered at all: a disabled button with no explanation.

Both go through one `automationBlockedTitle` now, so they cannot drift again. e3821cd69

ui/packages/plugins/om/src/api.ts
plebioda: Fair question, and it would be a real gap if the step were dispatched - but it is not. `confirm_monitoring` is synthesized by pmm-managed on every read from the run's own status (`confirmMonitoringStep` in `managed/services/om/inventory.go`, and the field's own proto comment): "pending" until the run succeeds, "running" from there until the estate shows a service, "succeeded" once it does. So the same response that first reports `succeeded` already reports the step as `running`, and there is no window between them for polling to fall back into.

What the 60s interval does cover is the wait for PMM's stepper to register the host, which is exactly what the fast poll is for.

Since the question is a reasonable one to ask again in six months, I widened the predicate to accept `pending` too and put the reasoning next to it with a test (1c1c86e7e). Costs nothing today, and keeps the behaviour right if confirm_monitoring ever becomes a step something actually dispatches.

ui/packages/plugins/om/src/api.ts
```diff
@@ -97,3 +101,57 @@ export async function request<T>(path: string, init?: RequestInit): Promise<T> {
...
+ * polling to see it flip to "Succeeded" a few seconds later.
+ *
+ * A succeeded run counts as active while that step is either `running` or
+ * `pending`. Nothing dispatches confirm_monitoring -- pmm-managed synthesizes it
+ * on every read from the run's own status (confirmMonitoringStep, and the field's
+ * own proto comment), so today a payload cannot report `succeeded` with the step
+ * still `pending`, and there is no window between the two for polling to fall
+ * back to the slow interval in. Accepting `pending` keeps that true if the step
+ * ever becomes something actually dispatched.
+ */
```
fabio-silva:
```suggestion
```

ui/packages/plugins/om/src/api.ts
```diff
@@ -97,3 +101,57 @@ export async function request<T>(path: string, init?: RequestInit): Promise<T> {
...
+
+/**
+ * True once any of a host's rollback steps has been dispatched.
+ *
+ * Every rollback step is planned `pending` up front alongside the host's
+ * forward steps (om_bootstrap's own HostBootstrapState doc comment) and stays
+ * that way for the entire life of a run that never needed rollback - so "not
+ * still all pending" is "rollback actually started," the same test
+ * PMM-15347's stepper itself uses (bootstrap_decision.go's hostRollbackStarted).
+ */
```
fabio-silva:
```suggestion
```

ui/packages/plugins/om/src/HostsPage.tsx
```diff
@@ -153,6 +155,42 @@ const ExecutorCell = ({ row }: { row: OmHostRow }) => {
...
+ * when it cannot.
+ *
+ * Deliberately separate from `ExecutorCell`, not a duplicate of it:
+ * `automation_eligible` is `executor` plus `pmm_agent_connected`, a signal
+ * `executor` alone cannot carry - a host can have a perfectly healthy Nomad
+ * executor while its pmm-agent has disconnected, which `executor` would read
+ * as "Ready" and this cell would not. The blocked reasons are the server's
+ * own explanation, not re-derived here, so this cell can never disagree with
+ * why the row actually says what it says.
+ */
```
fabio-silva:
```suggestion
```

ui/packages/plugins/om/src/types.ts
```diff
@@ -422,6 +422,24 @@ export interface OmInventoryHost {
   freshness: OmInventoryFreshness;
   /** The services on it. Empty is a meaningful answer, not a gap. */
   services: OmInventoryService[];
+  /**
+   * Whether PMM's own agent registry currently has a connected pmm-agent for
+   * this node - independent of `executor`, which is Nomad/SEP's own signal.
+   * This is the "PMM-Client installed and healthy" half of automation
+   * eligibility that `executor` alone cannot answer: a node can be a known,
+   * probed host in OM's estate while its agent has since disconnected.
+   */
```
fabio-silva:
```suggestion
```

ui/packages/plugins/om/src/types.ts
```diff
@@ -422,6 +422,24 @@ export interface OmInventoryHost {
...
+  pmm_agent_connected: boolean;
+  /**
+   * Whether this host is eligible for OM automation (a probe today;
+   * provisioning in a later phase): `pmm_agent_connected` is true and
+   * `executor` reports reachable and driver-healthy. Computed server-side so
+   * every consumer agrees on one definition - see `automation_blocked_reasons`
+   * for why not, when this is false.
+   */
```
fabio-silva:
```suggestion
```

fabio-silva (changes requested):

PR #5953, opened 2026-09-15
PMM-15505 Add documentation rules to AGENTS.md

ademidoff:
Ticket number: PMM-15505

Feature build: not needed — no shipped artifact changes.

## What

`documentation/` was the only component directory without an `AGENTS.md`, and it is the folder with the most rules. This adds `documentation/AGENTS.md` plus the `documentation/CLAUDE.md` pointer, so the rules load automatically whenever the folder is touched — nothing to install or trigger.

The guide covers scope and ownership, publishing behaviour, the page rules (voice, structure, headings, Markdown conventions, word choices) and the release-note entry format.

## Reconciliation

The ticket asks for one set of rules in exactly one place, reconciled against [#5334](https://github.com/percona/pmm/pull/5334) and the `references/` writer persona in PMM-15498. Nothing was carried over on faith — every rule was measured against the published pages under `documentation/docs/` first.

**The guide itself carries no counts.** Nobody is going to re-run them on a docs PR, so they would drift and then quietly mislead; each rule states the mechanism that justifies it instead, which is both stronger and permanently true. The measurements are recorded here, once, as the reason these calls went the way they did.

Where the corpus settled it:

| Rule | Decision | Measured |
|---|---|---|
| Admonition form | Plain Material (`!!! note ""`), no `alert alert-*` classes | Nothing in `docs/css/` or `overrides/` defines those classes — a pre-Material leftover. The reworked `install-pmm` chapter dropped them, 58 of its 75. Existing 95 left alone; both forms render identically. |
| Hard wrap at 80 chars (#5334) | No wrapping, at any width | 4310 lines exceed 120 chars. Applying it would rewrite every paragraph anyone touches. |
| `{.power-number}` above every ordered list | Required | 294 uses across 128 pages, and `docs/css/design.css` keys the stepped-list styling off `.power-number+ol` — without it you get a plain `<ol>`. Missing from every other copy of the rules. |
| Release-note entry format | `- [PMM-XXXXX](url): …` | #5334 has `- Fixed an issue where … ([PMM-XXXXX](url))`, which no published entry uses. |
| Relative `.md` links | Relative, with the extension | 862 relative against 71 absolute; `use_directory_urls` is `false`. |
| Front matter | None | Not one of the pages has any. |
| Check the Percona Software Support Lifecycle page (#5334) | Dropped as a per-change step | Zero pages reference it; it is an external percona.com page, and only a platform-support change touches it. |
| Release scaffolding (#5334 step 0) | Stays with the `pmm-rn-create` skill | The guide describes the entry format; the version page is scaffolded and written on the release branch. |

Two rules deliberately depart from what is published today. Both say so in the guide, and both tell you to fix pages you are already editing rather than sweep the corpus:

| Rule | Decision | Why, given the corpus disagrees |
|---|---|---|
| `## Before you start` vs `## Prerequisites` | **Before you start** | The corpus leans the other way, 29 headings to 12, with 22 of those in the reworked chapter. Overruled on house-voice grounds: `WRITERS-NOTES.md` has always said to avoid manual-speak, and that rule stands unchanged. |
| "the following table" vs "the table below" | **the following table** | "table below" leads 8 to 2, but that was only ever an argument against a mass rewrite, not that it reads well. Position words stop being true in the PDF, in a narrow viewport, and to a screen reader. |

## Other changes

- `documentation/docs-contributing.md` → `documentation/CONTRIBUTING.md`, matching `managed/`, `dashboards/` and `qan-api2/`. The published site links to this file by absolute GitHub URL from `documentation/overrides/partials/edit.html`, which renders on every page, so that link moves with it.
- `documentation/WRITERS-NOTES.md` keeps the reference tables it owns (admonition colours, icons, symbols) and loses the two sections that now live in `AGENTS.md`; the note about `alert alert-*` classes being inert is added in place.
- `documentation/CONTRIBUTING.md` gains a one-line pointer to the style guide it previously only alluded to.
- Root `AGENTS.md`: `documentation/AGENTS.md` added to the component-guide table and the testing decision tree; the User documentation section now routes through it and states that a merge to `main` publishes live.

## Not included

The ticket asks for the release-note fragment convention from PMM-15470 (`documentation/docs/release-notes/next/PMM-12345.md`, one file per ticket). Left out deliberately: PMM-15470 has not landed, so nothing assembles or clears that directory, and a rule pointing developers at it would send them to a folder nothing consumes. The guide documents the entry format instead, which is true today and is what a fragment would carry.

## Review notes

- Not verified: `make doc-build`. No local mkdocs, and the amd64 build image was not cached on this machine. The only config touched is `documentation/overrides/partials/edit.html` (one URL).
- This branch carries a merge of `main` (`e2cf9520d`), so the squash will fold that in too.

## Related

- PMM-15470 — release-note fragments (defines `next/` and the assembly stage)
- PMM-15498 — the tooling that drafts documentation from a ticket
- PMM-15502 — the check that requires the docs and the release-note line
- [#5334](https://github.com/percona/pmm/pull/5334) — the `pmm-writer` skill this reconciles against

- [ ] API Docs updated (n/a)


ondrejpatocka (approved):

documentation/AGENTS.md
```diff
@@ -0,0 +1,203 @@
...
+|--------------|---------|
+| Hard-wrap at 80 characters | The corpus does not wrap at all. Applying it would rewrite every paragraph you touch. |
+| Say "the following table", never "above" or "below" | "table below" outnumbers "the following table" 8 to 2, and "above"/"below" appear 170+ times. Not worth a mass rewrite. |
+| Release-note bug entries read `- Fixed an issue where … ([PMM-XXXXX](url))` | Published entries open with the linked key: `- [PMM-XXXXX](url): …`. |
+| Use "Before you start", not "Prerequisites" | Reversed since [`WRITERS-NOTES.md`](WRITERS-NOTES.md) was written; `Prerequisites` now leads 29 to 12. |
+| Check the Percona Software Support Lifecycle page on every change | Zero pages reference it. It is an external percona.com page, and only a platform-support change touches it. |
+| Scaffold a release-notes page as part of writing docs | Owned by the `pmm-rn-create` skill and the release branch. |
+| Add `alert alert-*` classes to admonitions | Inert — no stylesheet in this repo defines them. |
+| Read a `doc-style-guide/` sibling repo before writing | No such repo is checked out or referenced by this one. This guide is the style guide. |
```
travagliad: This whole section seems confusing, we're writing rules that claude has to ignore? Isn't better to make actual rules table and tell what it should not do instead? 

Also, mentioning the PR seems unnecessary.

I think this needs alignment with @catalinaadam

ademidoff: Yes, I left it here to remind ourselves that we already made an attempt to write the rules in that document. 

However, I do agree with a single source of truth - we need to make it happen with this PR.

ademidoff: Removed :)

documentation/WRITERS-NOTES.md
```diff
@@ -212,25 +208,4 @@ Section titles are deliberately short. For example, in Setting up/Server/Docker,
...
-2. Item
-3. Item
-   ...
-```
-
-Other advantages:
-
-- contents can be reused in source code comments by developers;
-- encourages authors to pay attention to the order and number of steps in a recipe.
+Number items explicitly (`1.`, `2.`, `3.`) rather than repeating `1.`, and put `{.power-number}` above the list — see [AGENTS.md](AGENTS.md#markdown-conventions). Explicit numbers keep the raw Markdown readable, let developers reuse the steps in source code comments, and make authors pay attention to the order and number of steps in a recipe.
```
travagliad, line 211: AGENTS.md contains same set of instructions, small duplication here.

travagliad (approved):
Apart from the small finding, I think this looks good, might want to have a quick check from Catalina from the writing instructions pov.

catalinaadam (approved):

PR #5958, opened 2026-09-16
PMM-15454 State the timezone a schedule runs in

nachodd:
Ticket number: PMM-15454

**Based on `PMM-15453-render-as-meaning` (#5933), not `main` and not `PMM-15205-sep-fb`.** That branch touches the same files and establishes the timestamp rule this work extends, so stacking keeps the diff here to what is actually new. It retargets to `PMM-15205-sep-fb` once #5933 lands.

---

## Where this ticket's scope came from

PMM-15454 was written up from Pedro's UX finding P7 (PMM-15380) by AI, and Yan flagged that. Pedro has since asked that **the Problem section be treated as the source and the Scope and ACs as provisional** — the read-only Schedules list and the combined-view AC were the write-up's expansion, not his finding.

He split the ticket himself. This PR is the half he called cheap and unambiguous:

- state the timezone actually in force, on the form and in the list — *"that was the core of my finding, the rest was supporting detail"*
- hide the Chain column while nothing can be chained (conditional vs. unconditional left to our call)
- drop two findings that a current build shows are already fixed

**Parked for a design pass after 3.10, at Pedro's request** — he wants to come back with the flow as visuals rather than settle it in Jira comments:

- one Schedules view vs. two (provisional AC2)
- where edit and delete live (implied by provisional AC1)
- dialog vs. drawer for creation (provisional AC1)
- whether interval mode stays in the primary flow

**Dropped as stale.** The ticket reports the create form "injected as a row under table headers" and the cron field having "no validation". Neither survives on a current build: the form renders below the table in its own container, and the cron field validates inline through `cronstrue` with a humanised caption. The ticket itself says to re-check against a current build before starting. The timezone and duplication points stand.

## Problem

Two different zones share the schedules screen and neither was named:

- the zone a schedule **fires in** — the crontab's own, or UTC for an interval
- the zone its timestamps are **rendered in** — always the reader's, via `toLocaleString()`

A cron schedule written in `Europe/Lisbon` and read from a browser in `America/New_York` therefore showed times that were correct and unreadable. Interval mode asked no timezone at all and silently converted its start time between local wall clock and UTC in both directions.

## What

**State both zones.** Each row names the zone it fires in as `Runs in <zone>`, taken from the backend's computed `timezone` (PMM-15480) rather than re-derived. The table names the zone it renders timestamps in, once in its header, and `formatTimestamp` now appends that zone to the hover title every timestamp already carried.

**Interval start time stops converting behind the user's back.** It reads and writes UTC wall clock and its label says `(UTC)`. Per Pedro: *"Don't convert local wall-clock behind the user's back; if the field is UTC, label it UTC."* Interval is UTC because the backend fixes it there — an interval's cadence is an absolute `timedelta` with no wall-clock anchor and `start_time` is a `UTCDatetime` — so a timezone picker in interval mode would be decorative. It is stated, not asked.

**Preview the next runs.** The side-car computes upcoming runs through the scheduler's own schedule objects and serves them for an unsaved schedule at `/sep/periodic-tasks/schedule/preview/` (PMM-15480). The form asks rather than re-deriving a cron expression in the browser, so the preview and the saved schedule cannot disagree about what "next" means. Debounced, idle while the form describes nothing valid, and the response's own `timezone` supersedes the form's assumption.

**Hide the Chain column while nothing is chained.** These apps expose a single chainable task, so no schedule can carry a chain and the column was an unbroken run of em dashes. It returns on its own the moment any schedule carries one. Complements #5932, which hides the chain *input*; different files, no conflict.

## Reconciling with #5933 — and a scope question for @pedro-fernandes

#5933 introduced a shared `RunTime` cell and the rule that `formatTimestamp` is *"the one place a timestamp becomes text"*. An earlier revision of this PR added a second wrapper beside it; that is gone, and the zone now lives inside the shared rule.

**This has a consequence worth calling out explicitly, because it partly answers a question Pedro has not been asked yet.** Because the zone rides on `formatTimestamp`'s title, it reaches every timestamp in the app, not only the schedules table — including **`ScheduleCell`** on the backups list and **`ScheduleSummary`** on the backup detail, which show a next-run time and named no zone.

Pedro's wording was *"on the schedule form and in the list"*. I read "the list" as the Schedules list and scoped the deliberate work there. The two surfaces above now answer the zone **on hover only** — they gained it as a consequence of sharing the helper, not as a designed statement. So:

- If that is what was meant, this is done and no further work is needed.
- If those surfaces should state the zone **visibly** the way the schedules table does, that is a small follow-up and not in this PR.

This is also why five sibling test files change here: they assert the hover title verbatim, and the title gained the zone.

## Bugs found on the way

**`Intl.supportedValuesOf('timeZone')` cannot offer the zones the backend stores.** It reports canonical zones only — it carries `Etc/UTC` but not the bare `UTC` link, which is the backend's crontab default and the zone it reports for every interval schedule. MUI *displays* an off-list value while refusing to offer it, so editing any UTC cron schedule left the picker holding a value the user could not restore after changing it: a one-way door, and silent. Generalised after review — the backend also accepts aliases such as `US/Eastern` that the runtime omits — so the picker now carries whatever zone the schedule actually uses.

**Editing an unrelated field rounded a schedule's first fire down.** The start-time field is minute-precision, so round-tripping a stored `start_time` through it dropped any seconds, and saving a change to the interval wrote the truncated instant back. Pre-existing — the local conversion this replaces truncated identically — but fixed here rather than filed, since not changing a schedule behind the user's back is what this ticket is about.

## Review

Dual local review (CodeRabbit skips any PR not based on `main`): `codex-reviewer-pro` and `copilot-reviewer-pro`. Both independently flagged the start-time truncation, fixed above. Copilot additionally caught that the per-row zone caption sat beside three browser-zone columns and invited the very misreading this work removes — hence `Runs in X` wording, matching the form, plus the per-timestamp hover. Its column-count point is addressed too: header and row `colSpan` now derive from one `scheduleColumnHeaders` list instead of a parallel hand-maintained count.

Codex returned malformed output and covered only one of five requested areas, so its silence elsewhere is not a clean bill of health.

## Test plan

- `cd ui && make lint` — 0 errors (444 warnings, unchanged from base)
- `cd ui && make format-check` — clean
- `cd ui && make test` — green: `@sep/framework` 961, `ui` 572, `@sep/api` 106, `@sep/plugins-atw` 122
- `pnpm --filter @sep/framework run check-types` — clean
- New coverage: `timezones.test.ts` (UTC round trip, DST-independent by construction, zone resolution and fallbacks), plus panel cases for the zone statements, the Chain column appearing and disappearing with header/body width held in sync, start-time precision, an off-list zone staying selectable, and the next-runs preview (three of four shown, backend zone winning, idle while invalid, failure state)
- Both new guards were checked to **fail** without their fix — the first version of the off-list-zone test passed either way and was rewritten

## Not verified

**Not yet run against a live PMM server.** `AGENTS.md` asks for that on any user-visible change and this is pure UI. A Feature Build is next.

## Note

Everything here is under `ui/packages/sep/` — PMM-side divergence that wants upstreaming to `percona/percona-sep`, or the next vendored-frontend sync reverts it.


ui/packages/sep/framework/src/components/ScheduledTasksPanel/ScheduledTaskForm.tsx
```diff
@@ -472,6 +536,37 @@ export function ScheduledTaskForm({
...
+            color={previewFailed ? 'error' : 'text.secondary'}
+            sx={{ display: 'block', mt: 0.5 }}
+            data-testid="sched-form-next-runs"
+          >
+            {previewFailed
+              ? 'Could not work out the next runs for this schedule.'
+              : nextRuns.length > 0
+                ? `Next runs: ${nextRuns
+                    .map((r) => formatTimestamp(r)?.display ?? r)
+                    .join(', ')}`
```
mattiasimonato: Every value here is in the future and almost always inside `formatTimestamp`'s 7 day relative window, so `display` is a relative phrase. Three of them render as:

`Next runs: in 8 hours, tomorrow, in 2 days`

That is verbatim from a test I ran on this branch, with runs at +8h, +32h and +56h. There is no clock time, and the `title` that `formatTimestamp` builds is dropped, so hover does not recover it either. It leaves `Runs in Europe/Lisbon` right above with nothing to qualify, and the preview cannot answer "is this really 02:00", which is what the ticket asks it to answer.

`shows the next three runs the backend reports` does not catch this because its fixtures are months away, so they fall outside the window and render absolute.

```suggestion
                ? `Next runs: ${nextRuns
                    .map((r) => formatAbsoluteTime(r))
                    .join(', ')}`
```

That gives a clock time, but in the reader's zone while the caption above names the firing zone. Rendering in `zoneInForce` instead, with a small `toLocaleString(undefined, { timeZone: zoneInForce, dateStyle: 'medium', timeStyle: 'short' })` helper, is the honest version. Either way that test's assertion needs updating, since it asserts whatever `formatTimestamp` produces.

ui/packages/sep/framework/src/components/ScheduledTasksPanel/ScheduledTaskForm.tsx
```diff
@@ -35,8 +35,18 @@ import {
...
 } from '../ChainBuilder';
+import {
+  INTERVAL_TIMEZONE,
+  TIMEZONES,
+  defaultPickerTimezone,
+  utcInputToIso,
+  utcIsoToUtcInput,
+} from './timezones';
+import { useSchedulePreview } from './hooks';
+import { formatTimestamp } from '../../utils/formatTimestamp';
```
mattiasimonato: `formatTimestamp` is then unused in this file.

```suggestion
import { formatAbsoluteTime } from '../../utils/formatTimestamp';
```

mattiasimonato (changes requested):
The next-runs preview never shows a time. `formatTimestamp` renders anything within a week relatively, and an upcoming run is within a week by definition, so the line reads "Next runs: in 8 hours, tomorrow, in 2 days". I reproduced it on the branch. The rest looks right to me and the framework suite is green (979).

ui/packages/sep/framework/src/components/ScheduledTasksPanel/ScheduledTaskForm.tsx
nachodd: Fixed in d10ef7c81. Next runs now render as an absolute clock time in `zoneInForce` (the schedule's own zone, matching the `Runs in {zone}` caption above) instead of `formatTimestamp`'s relative phrase — no ambiguity about which zone the clock time is in, or whether it's really 02:00. Updated the test to use near-future fixtures (+8h/+32h/+56h) so it actually exercises the case your months-away fixtures missed.

ui/packages/sep/framework/src/components/ScheduledTasksPanel/ScheduledTaskForm.tsx
nachodd: Fixed in d10ef7c81. Dropped the unused `formatTimestamp` import — the next-runs preview no longer uses it (see the sibling reply on the other thread).

mattiasimonato (approved):

PR #5959, opened 2026-09-16
PMM-15456 Bring SEP app controls and layouts onto Peak UI

nachodd:
Finding **P9** from the UX pass over the MySQL Backups Tech Preview (PMM-15380), under the PMM-15217 umbrella. Finishes the SEP framework's conversion onto Peak UI and settles the app on one list pattern and one edit pattern.

Based on `PMM-15454-state-the-timezone` (#5958), which rewrites the same three scheduling files — retarget to `PMM-15205-sep-fb` once that merges.

## What changed

**Controls onto Peak UI**
- `SchemaSelectShell` now renders Peak's `SelectInput`. The `displayEmpty` / `notched` / forced-`shrink` combination and the `Select…` placeholder are gone, so a select's label sits in the empty field and floats on selection — the same label behaviour `TextInput` and `AutoCompleteInput` already had. That is what collapsed the two select styles into one. `SelectInput` owns the Controller, so `ChoiceField` and `MultiChoiceField` dropped theirs.
- `DateTimeField` and the schedule start time moved off `<input type="datetime-local">` onto Peak's `DateTimePickerInput`, behind a new `DateTimeInput` wrapper.
- `AlertOnFailField` became a `SwitchInput` — it was the form's one checkbox among a dozen toggles.
- `ScheduledTaskForm` rebuilt on `TextInput` / `SelectInput` / `AutoCompleteInput` / `SwitchInput` / `DateTimeInput`, and the lowercase `change to cron mode` link became a labelled `RadioGroup`.

**One list pattern** — `SepTable/sepTableProps()` holds the presentation all three lists share, and the schedules panel is a `MaterialReactTable` like the other two instead of a hand-written `<Table>` in a card. `Add new` moved from a footer strip to a header action, matching `PluginListPage`.

**One edit pattern** — schedules are created and edited on a page (`schedule/new`, `schedule/:id/edit`), the pattern plugin tasks already used. `ScheduledTaskRow` is gone; its cells are MRT column defs.

**Width** — `FORM_CONTENT_MAX_WIDTH` caps both forms. SEP pages stay `maxWidth="full"` so lists keep the screen.

## Decisions taken — please confirm

Two calls that were mine to make and are worth a second opinion. @pmcf-percona for the UX half, @yyyyyyyan for the engineering half.

1. **Editing is a full-page route**, not a dialog or a drawer. Plugin task edit was already a page, so a page is the pattern that wins on count; a dialog would have kept the schedules panel self-contained but left "page here, dialog there" — still two patterns.
2. **The raw-MUI sweep stops at the surfaces the ticket names.** See *Out of scope*.

## Out of scope

Still on raw MUI, deliberately: `TaskLogViewer`'s Select and Switch, `ExecutionEventsPanel`'s search field, `FreeSoloSelect`, `RemoteChoiceSelector`, `StandaloneHostSelector`, and `ChainBuilder`'s add-to-chain Select and *Continue chain on failure* Switch. The three autocompletes need free-solo semantics that Peak's `AutoCompleteInput` does not obviously preserve, the log viewer's controls are viewer chrome rather than form fields, and both `ChainBuilder` controls are driven by local props/state rather than a react-hook-form `control` — `SelectInput`/`SwitchInput` bind through the Controller, so there's no host form for them to bind to here. Worth a follow-up ticket if we want them converted (AC 1 is not fully met because of this).

**AC 2 stays open.** Peak UI itself would have to fix this:
- `SelectInput` declares `isRequired` but never destructures it, so the asterisk has to come via `formControlProps={{ required }}`.
- `LabeledContent` / `RadioGroup` render the required marker as a bare `<span>*</span>`, while `TextInput` and friends render MUI's ` *` asterisk. Same glyph, different spacing and a11y.

Filed as [percona/peak-ui#61](https://github.com/percona/peak-ui/issues/61); not covered by peak-ui's #15, #42, or #52.

## Notes for review

- **The date picker cannot render a wall clock inside the reader's own spring-forward gap.** `AdapterDateFns` reports `isTimezoneCompatible = false` and PMM's `App.tsx` uses it for every picker, so the bridge goes through a local `Date` and `02:30` on a spring-forward date displays as `03:30`. It is display-only: mounting does not write back (pinned in `DateTimeInput.test.tsx`), and `startTimeToSubmit` now keys on whether the field is dirty rather than comparing its value, so an untouched start time is resent byte-identically. Regression test covers a stored `02:30Z` edited from `America/New_York`.
- `ScheduledTasksPanel` is a root export of `@sep/framework` and gained optional `onCreate` / `onEdit`; omitting them hides those actions. The only in-repo consumer is `PluginSchedulePage`. Flagging in case percona-sep has others.
- `@sep/framework` gains `@mui/x-date-pickers` and `date-fns` as peer + dev deps. Both already resolve in the workspace via `apps/pmm`. `DateTimeInput` mounts its own `LocalizationProvider` so no host or test has to.
- All of this is in vendored SEP code; a percona-sep re-sync would revert it.

## Test plan

- `cd ui && make lint && make format-check && make test` — 999 framework tests, whole workspace green.
- Framework suite re-run under `TZ=UTC`, `America/New_York`, `Europe/Lisbon`, `Australia/Sydney`.
- `npx turbo run check-types` clean.
- Not yet exercised on a live server — needs a Feature Build for the visual half of the ACs.


ui/packages/sep/framework/src/components/DateTimeInput/wallClockValue.ts
```diff
@@ -0,0 +1,117 @@
...
+ * before these fields moved onto Peak UI.
+ *
+ * Keeping the shape is not inertia. Both callers mean a *wall clock with no
+ * zone*: a schema `datetime` field submits the string verbatim, and a
+ * schedule's start time is UTC wall clock that
+ * `ScheduledTasksPanel/timezones.ts` reads back as UTC (PMM-15454). Handing
+ * either an adapter `Date` would attach the reader's zone to a value that has
+ * none.
+ */
+const WALL_CLOCK = /^(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2})(?::(\d{2}))?$/;
```
mattiasimonato: The rollover guard below compares year, month and day, and the hour is deliberately excluded for the DST gap, so nothing catches minutes: `2026-01-01T10:61` parses to 11:01 and renders as a value nobody wrote, which is the opposite of what the guard's own comment promises.

Cheapest fix is the pattern rather than the guard. A minute comparison would misfire in half-hour DST zones (`Australia/Lord_Howe` shifts by 30 minutes, so a wall clock in its gap moves the minute legitimately), while a range check is zone-independent:

```suggestion
const WALL_CLOCK = /^(\d{4})-(\d{2})-(\d{2})T([01]\d|2[0-3]):([0-5]\d)(?::([0-5]\d))?$/;
```

Same group count and order, so the destructuring below is unchanged. Low severity since it needs malformed stored data, but it keeps the invariant true.

ui/packages/sep/framework/src/components/ChainBuilder/ChainBuilder.tsx
```diff
@@ -270,15 +270,19 @@ export function ChainBuilder({
           <FormControlLabel
             sx={{ mt: 1 }}
             control={
-              <Checkbox
+              // A toggle, like every other boolean in these forms. Peak UI's
+              // SwitchInput binds itself to react-hook-form by name, and this
```
mattiasimonato, line 274: The reasoning here is right and the comment is the right place for it, but the PR body reads differently:

> `AlertOnFailField` became a `SwitchInput` [...] `ChainBuilder`'s *Continue chain on failure* went the same way.

This one went Checkbox to raw MUI `Switch`, not to `SwitchInput`. Same for the add-to-chain `Select` a few lines up. Since AC 1 is "no form control is a raw MUI or native element where a Peak UI equivalent exists", could both move to the **Out of scope** list with this reason? Otherwise a reader takes AC 1 as fully met when two controls in this file are the exception.

ui/packages/sep/framework/src/components/SchemaFormRenderer/SchemaSelectShell.tsx
```diff
@@ -16,83 +16,81 @@
...
-    </FormControl>
+    <SelectInput
+      name={name}
+      control={control}
+      controllerProps={{ rules }}
+      label={
+        (<FieldLabelWithHelp label={label} description={tooltip} />) as never
+      }
+      helperText={message ?? inline}
+      formControlProps={{ fullWidth: true, size: 'small', required }}
```
mattiasimonato, line 90: Confirmed both Peak UI gaps you flagged, on peak-ui `main`:

- `src/components/form/inputs/select/select.types.ts` declares `isRequired`, and `select.tsx` never destructures it, so this workaround is needed.
- `src/components/labeled-content/labeled-content.tsx:63` renders `{isRequired && <span>*</span>}` while `TextInput` passes MUI's `required`, so `RadioGroup` and `ChoiceField` really do carry a different marker from the rest of the form.

Neither is filed in percona/peak-ui (the open issues are #15, #42 and #52). Could you open one? AC 2, one required marker, stays open until it lands, and it is worth saying that in the PR body rather than as an aside, so the ticket is not closed as fully met.

Extra reason to file: percona/peak-ui#60 ships a consumer-facing `AGENTS.md` that documents `isRequired` as adding the asterisk, which will be wrong for `SelectInput` until this is fixed.

mattiasimonato (commented):
Checked the branch against the ticket's ACs. Lists, edit pattern and width all hold; the two that do not are inline, plus one real bug and one description fix:

- `wallClockToDate` accepts out-of-range minutes
- the body says ChainBuilder's toggle became a `SwitchInput`, it is still raw MUI
- **one required marker** cannot be met from this repo, and neither Peak UI gap is filed yet

On the free-solo exclusion, your reasoning holds and the mechanism is sharper than "not obviously preserves": `AutoCompleteInput` spreads `{...field}` and wires only `onChange`, never `onInputChange`, so with `freeSolo` the typed text reaches the form only when MUI commits it. Worth putting in the follow-up ticket. (Read on peak-ui `main`; we ship 1.0.25.)

Merge Gatekeeper is red on a `context deadline exceeded` timeout, not a real failure. The UI workflow ran format-check, lint, test and build, all green.

ui/packages/sep/framework/src/constants.ts
```diff
@@ -18,3 +18,15 @@
...
+ * Widest a form in a SEP app may draw its inputs, in pixels.
+ *
+ * PMM's page container caps content at `DEFAULT_PAGE_MAX_WIDTH` (1000px), but a
+ * SEP page opts out of that cap so its *lists* can use the screen. A form has
+ * no such need: an input stretched across 1250px of window is harder to scan,
+ * not easier, and the single-column stretch was the most visible way these
+ * screens read as not-PMM (PMM-15456). Narrower than the page cap on purpose —
+ * this is measured against line length, not against the container.
+ */
+export const FORM_CONTENT_MAX_WIDTH = 800;
```
mattiasimonato: AC 5 words it as "PMM's standard content width". PMM has no form-width constant of its own, and the page cap comes from Peak's `DEFAULT_PAGE_MAX_WIDTH = 1000`, so 800 is a new number for this app.

The comment argues it well, I just want it to be a decision rather than a drift: worth a yes from @pmcf-percona alongside your two other calls.

pmcf-percona: This is a topic worth exploring as it's still grey. We already applied a width cap in PMM Setting page for text at 750px. The reasoning was reading comfort. Doing it for the form controls also makes sense, but if we apply to every piece of content then we might as well just define the layout width to 800px or 750px and call it a day. 1000px was more to not make the layout too narrow (so tabs don't appears scrunched up when the viewport has margin for maneuvering). So, still not sure what the final solution is here. Maybe let's explore this for future iterations.

Let's define a 750px max width for reading comfort and have that defined into SEP now, and that would be consistent with Settings. Then apply the same constant for setting as well and then anyone can opt in (as a design choice) to have that reading width. Meaning, anyone can choose to opt out and have all content stretch to the layout.

In future iterations we will fine tune: layout width may increase/decrease; reading width the same. Meaning, let's have that constant defined, because it's useful, and match now to 750px to inherit the rationale from Settings.

pmcf-percona:
Thanks @nachodd and @mattiasimonato!

- Full-page edit for schedules: yes. One edit pattern across the app and a page gives the form room and a clear back path.
- Required marker: from the UX side, the small difference in spacing between the two asterisks is acceptable for the Tech Preview, as long as every required field shows one. Agree with Mattia that the PR body should say this AC stays open until Peak UI fixes it.

pmcf-percona (approved):

ui/packages/sep/framework/src/components/DateTimeInput/wallClockValue.ts
nachodd: Fixed in 2876fba895 — took your suggested range-bounded pattern verbatim (`[01]\d|2[0-3]` for hours, `[0-5]\d` for minutes/seconds), zone-independent.

ui/packages/sep/framework/src/components/ChainBuilder/ChainBuilder.tsx
nachodd: Confirmed — that line was wrong, both the Switch and the add-to-chain Select are still raw MUI. Fixed the PR body: removed the false claim, and moved both into Out of scope with the reason (they're driven by local props/state, not a react-hook-form `control`, so `SwitchInput`/`SelectInput` have nothing to bind to there). AC 1 now correctly reads as not fully met for this file.

ui/packages/sep/framework/src/components/SchemaFormRenderer/SchemaSelectShell.tsx
nachodd: Filed percona/peak-ui#61 covering both gaps (dead isRequired prop on SelectInput, and the RadioGroup/ChoiceField asterisk mismatch). Updated the PR body to state plainly that AC 2 stays open until it lands, rather than as an aside.

ui/packages/sep/framework/src/constants.ts
nachodd: Went with @pmcf-percona's call in 2876fba895: FORM_CONTENT_MAX_WIDTH is now 750px, matching the reading-width cap PMM's Settings page already uses for text, and the comment says so. Left Settings itself untouched — the follow-up to point it at this shared constant is future-iteration work per that reply.

nachodd:
Both landed in 2876fba895: full-page edit for schedules stays as implemented, and the required-marker AC is now stated explicitly in the PR body as staying open until percona/peak-ui#61 lands.

mattiasimonato (approved):

PR #5962, opened 2026-09-17
PMM-15509 Say why a diagnostics run failed, and when it never ran

nachodd:
## What

Support diagnostics (ATW) told an operator a run had failed but gave them nothing to act on: a failed row expanded onto stdout regardless of which stream held the error, "No output yet." lied about a finished run, tab labels clipped, an unlaunchable run showed only a raw sentinel code, and there was no way to retry a run or reopen the form that produced it.

**Shared framework** (`ui/packages/sep/framework`, used by every log-viewer consumer including MySQL Backups):
- `TaskLogViewer` opens on whichever of stdout/stderr the active step actually wrote to; sticks to an explicit tab choice; resets when switching runs.
- Wrap defaults on; tab labels no longer truncate.
- Empty pane reads "No output" once the run has finished, "No output yet." while it may still produce something.
- Step tabs render above the log body, hidden entirely for a single step.
- Execution events move into a collapsed "Technical details" section, with step scoping independent of the log's own steps.
- A non-failure terminal status (stopped/lost/stale/unlaunchable) renders as a plain sentence, via a `terminalRunNotes` module shared with `TaskRunDetailDrawer` (which drops its own copy to avoid a duplicate wherever the log viewer is also on screen).
- The 410 "executor gone" block leads with the sentence and host name; raw job/evaluation identifiers move behind an "Error details" expand.
- New `useHasDownloadableFiles` hook backs both `TaskHistoryTable`'s existing Download button (extracted, behavior-preserving) and ATW's Files button (previously unconditional).

**ATW plugin** (`ui/packages/plugins/atw`):
- Every execution row offers **"Edit parameters and run again"** (reopens the Collect form, pre-filled) and, once this browser tab remembers the batch it dispatched, **"Run again"** (replays it unchanged). Nothing re-runnable rides on the wire today (`masked_args` is a display string with credentials masked out), so remembering is session-local; without one, the form falls back to reselecting the snippet by exact-name search and reports plainly if that search finds nothing.
- "Files" only renders once a non-empty listing is confirmed.

## Why

Ticket: PMM-15509, from Pedro Fernandes' UX pass over the Tech Preview (finding P1 on PMM-15381, under the PMM-15217 umbrella). Builds on PMM-15415, which gave an unlaunchable run its own terminal status — this is the operator-facing half.

## Deliberately out of scope

Each blocked on a wire field that does not exist yet:
- First line of a failed row's error → **PMM-15524**
- Host-selection warning for scripts needing elevation (mandatory-only) → **PMM-15519**
- Associating an error with the parameter that caused it → **PMM-15523**

## Open questions for Pedro / Yan

1. Which tab opens when a failed run wrote to **both** stdout and stderr? Implemented: stdout keeps priority, stderr wins only when stdout is empty. A "failed run always opens on stderr" reading is a defensible alternative.
2. "Run again" only appears for a run **this browser tab** dispatched (nothing reusable is on the wire to survive a reload). Acceptable that the button appears on some rows and not others?
3. Does the raw unlaunchable/executor-gone detail belong in the same "Technical details" disclosure as Execution events, or is a separate "Error details" expand (as implemented) the right split?
4. Should the log viewer's Wrap toggle persist per user, like the tail-lines choice already does?

## Testing

- `pnpm --filter @sep/framework test` — 948 passing
- `pnpm --filter @sep/plugins-atw test` — 134 passing
- `pnpm --filter @sep/framework check-types` / `@sep/plugins-atw check-types` — clean
- `oxlint` / `oxfmt --check` on all touched files — clean
- Reviewed with reviewer agent; fixed a duplicated non-failure note between `TaskRunDetailDrawer` and the embedded `TaskLogViewer`, added no-match/error handling to the rerun-by-search fallback, and added a test for the shared-mutation cross-row disabling on "Run again".

## Base branch

Branched from `PMM-15205-sep-fb` per the SEP frontend-build convention (matches PMM-15451/15452/15453). Since this isn't based on `main`, CodeRabbit will not review it automatically.


mattiasimonato (commented):
One fix needed before this can go in. The rerun resolution reads the previous search's results, so reopening the form a second time picks the wrong snippet. Everything else I looked at holds up, and both suites pass on my machine (framework 948, atw 134).

ui/packages/plugins/atw/src/CollectPane.tsx
```diff
@@ -281,6 +326,70 @@ export function CollectPane({
...
+    }
+    if (rerunResolveQuery.isError) {
+      setResolvingRerunNonce(null);
+      setRerunResolveTerm(null);
+      setRerunResolveFailed(rerunRequest.snippetFilename);
+      return;
+    }
+    if (!rerunResolveQuery.isSuccess) {
+      return;
+    }
```
mattiasimonato, line 373: `useAtwSnippetSearch` sets `placeholderData: keepPreviousData`, so the moment `rerunResolveTerm` changes, the query is already `isSuccess` while still holding the previous term's items. This guard lets that stale page through, the `find` below misses, and the resolution gives up with "Could not find ...".

To see it: reload the page so nothing is remembered, click "Edit parameters and run again" on one row, then on a row with a different snippet. The second one warns and keeps the first snippet selected. After a reload every row takes this path, so it is the normal one, not an edge case.

You already use the precise signal for the picker at line 403.

```suggestion
    if (!rerunResolveQuery.isSuccess || rerunResolveQuery.isPlaceholderData) {
      return;
    }
```

With this plus the dep below, the 135 atw tests still pass.

ui/packages/plugins/atw/src/CollectPane.tsx
```diff
@@ -281,6 +326,70 @@ export function CollectPane({
...
+    if (match) {
+      setSelected([match]);
+      setRerunResolveFailed(null);
+    } else {
+      setRerunResolveFailed(rerunRequest.snippetFilename);
+    }
+  }, [
+    rerunResolveQuery.data,
+    rerunResolveQuery.isError,
+    rerunResolveQuery.isSuccess,
```
mattiasimonato, line 388: Keep the dep list in step with the guard above.

```suggestion
    rerunResolveQuery.isSuccess,
    rerunResolveQuery.isPlaceholderData,
```

pmcf-percona:
1. Open stderr whenever the run failed and stderr has content. Fall back to stdout only when stderr is empty. Successful runs keep opening on stdout.

2. Not acceptable as shipped. A button that exists on some rows and vanishes from all of them after a reload looks like a bug, and the operator has no way to learn the rule. Please hide "Run again" entirely for now and keep "Edit parameters and run again" on every row, since that one always works. When the backend can send back the real, re-runnable arguments, we bring "Run again" back on every row at once.

3. Merge them into a single collapsed "Technical details" section, with the raw error identifiers at the top and execution events below. The plain sentence (for example "This run never started: it needed sudo on pmm-server and the executor there does not allow it") must stay visible without expanding anything. Two collapsed sections next to each other make the user guess which one holds what, and both hold the same kind of content: raw material for a support engineer, not for the operator.

4. Persist the Wrap toggle per user: Yes. It should behave exactly like the tail-lines choice. A user who turns wrap off once expects it to stay off, and having one viewer setting remembered while the neighbouring one is not feels broken.

ui/packages/plugins/atw/src/CollectPane.tsx
nachodd: Fixed in 0359a3b30: guarded on `rerunResolveQuery.isPlaceholderData` (and added it to the effect's deps) so a stale page from the previous term's search can no longer stand in for the new one.

ui/packages/plugins/atw/src/CollectPane.tsx
nachodd: Fixed in 0359a3b30: added `rerunResolveQuery.isPlaceholderData` to the dependency list alongside the guard above.

nachodd (commented):
Fixed in 0359a3b30 — see the inline replies for the `isPlaceholderData` fix.

nachodd:
Fixed in 0359a3b30:

1. `TaskLogViewer` now opens on stderr whenever the run's terminal status is `failed` and the active step's stderr has content, falling back to stdout only when stderr is empty. A successful run still prefers stdout, and the existing "stdout empty, stderr not" fallback is unchanged for a run in progress.
2. "Run again" is now hidden on every row, everywhere — only "Edit parameters and run again" remains, which always works. The underlying mutation, error alert and remembered-dispatch plumbing feeding "Run again" were removed from `ResultsPane`; `IncidentWorkspacePage`'s remembered-dispatch map stays, since "Edit parameters and run again" still uses it to prefill exact values.
3. Merged: the executor-gone error's raw identifiers (job id, evaluation id, resource) now render inside the log viewer's own "Technical details" section, above execution events, instead of a second separate "Error details" expand. The plain summary sentence is unaffected and stays visible outside any collapse.
4. The Wrap toggle now persists to `localStorage` (`sep.taskLogViewer.wrap`) exactly like the tail-lines choice, and is restored on the next mount.

mattiasimonato (approved):

PR #5964, opened 2026-09-17
PMM-15510: Make the diagnostics run form speak PMM, not the CLI

peter-o-addo:
Ticket number: PMM-15510

**Based on `PMM-15205-sep-fb`.** Collect / `@sep/framework` only; not based on `main`.

Finding **P2** from Pedro Fernandes' UX pass on the Support diagnostics Tech Preview (PMM-15381), under PMM-15217 ("Users can't complete their actions alone"). This PR is the **UI rendering half**: make the Collect form feel like PMM, not like script `--help`. Schema text and field-type updates stay on **PMM-15518**.

---

## What changed

**No repeated help.** `fieldHelp` hides the icon/helper when the description is only the same as the label (ignore case and spaces). Real dual help from PMM-15458 is unchanged.

**Execution Host label outside the field.** `HostSelector` / `FreeSoloSelect` use Peak `LabeledContent` for the label and required `*` above the box; Peak's `label` stays for a11y. Hiding the floating label left a gap in the top border; fixed with `notchedOutline legend { maxWidth: 0 }`.

**Required numbers get a real start value.** `fieldDefault` seeds required integers/floats from `ge` or `1` so Minutes is not empty on first paint. Placeholders stay grey ghost text only; they are never turned into submitted values (an earlier bridge that did that was removed after review, because example copy like Issue Time's `e.g. 2026-01-01…` would pass required checks and reach the host).

**Datetime round-trip.** `DateTimeField` converts ISO to/from `datetime-local` so a real `datetime` field gets a picker and sends ISO back. Live Collect still waits on PMM-15518 to change those six timestamps from `string` to `datetime`.

**Hide CLI usage traps.** CollectPane removes fields named `help` / `show_help` / `h` / `usage` (and "Show help message" labels) before render, including inside one-of groups. Output option wording stays on PMM-15518.

## Decisions taken (please confirm)

1. **Shared UI in `@sep/framework`, traps in CollectPane.** Help, host label, number seeds, and datetime are shared SchemaFormRenderer behaviour. Hiding CLI help/usage stays in CollectPane so other plugins do not get that filter.
2. **No placeholder-to-default promotion.** Placeholders can be real defaults or examples; the FE cannot tell them apart. Promoting them submitted example text and skipped required validation. Visible defaults belong in schema `default` (PMM-15518). Until then, helper text can say what happens when a field is left blank without inventing a value.
3. **Datetime stays on `datetime-local` here**, not Peak `DateTimePickerInput`. That Peak swap belongs with the larger Peak form work; this ticket only makes the existing `datetime` type safe when schemas start using it.

## Out of scope

- Rewriting the ~89 script schemas (labels, output options, `placeholder` to `default`, timestamps as `datetime`): **PMM-15518**.
- Peak `DateTimePickerInput` / schedule Peak conversion (other PRs).
- A live timestamp picker on Collect until a schema uses `type: datetime`. Covered in tests; production AC waits on 15518.
- Filling text fields from placeholders so "submit unchanged" always matches a hidden CLI default. That needs real schema `default`s.

## Notes for review

- All under `ui/packages/sep/` and `ui/packages/plugins/atw/`. PMM-only changes that need upstreaming to `percona/percona-sep`, or a later vendored sync will wipe them.
- `ScheduledTaskForm` gets the same FreeSoloSelect label layout where it shares that control.
- On Collect with old schemas, timestamps can still look like plain text. That is 15518, not a bug in this PR.
- Log-path and similar fields may still show a grey placeholder with an empty value; required empty fields (e.g. Issue Time) refuse submit instead of sending the example.

## Test plan

- `cd ui && make lint && make format-check`
- New/updated tests: `fieldHelp` label echo, `fieldDefault` does not promote placeholders + required number seed, `datetimeLocal` ISO round-trip, HostSelector label + no border gap, CollectPane CLI trap hide (including one-of)
- `cd ui && make test` (or `@sep/framework` + `@sep/plugins-atw`)
- Live Collect: Host label outside; no duplicate Save-samples help; no Show-help / usage toggle; placeholders stay ghost text; empty required fields block Run; Minutes seeds to a number
- Feature Build for the visual ACs once PMM-15518 lands `datetime` timestamps and real schema defaults


ui/packages/sep/framework/src/components/SchemaFormRenderer/utils/fieldDefault.ts
```diff
@@ -43,10 +44,69 @@ export function emptyFieldValue(field: PluginField): unknown {
...
+    // `datetime-local` cannot display a UTC ISO string; convert here so the
+    // picker mounts with a real wall-clock value (PMM-15510).
+    if (field.type === 'datetime') {
+      return toDatetimeLocalValue(field.default);
+    }
+    return field.default;
+  }
+  if (isTextLike(field) && field.placeholder) {
+    return field.placeholder;
+  }
```
mattiasimonato: This turns any placeholder into a value the form submits, and nothing here can tell a real default from an example. The ticket has both cases on the same screen: the log file path placeholder *is* the effective default (your good case), while Issue Time's is "a date example" the user is meant to replace.

I rendered a form with those two fields on this branch and clicked Run without touching anything. The payload:

`{"issue_time":"e.g. 2026-01-01 00:00:00","extra_args":"--databases mydb --tables t1,t2"}`

So a made-up timestamp, `e.g. ` prefix included, goes to a script that runs on the customer's host. Worse, the required check now passes, because the field is no longer empty. `toArgs` in `ui/packages/plugins/atw/src/CollectPane.tsx` only drops empty values, so these reach the batch payload unchanged. And it is not only Collect: `SchemaFormRenderer` also backs `PluginCreatePage` and `PluginTaskEditPage`, so backup config forms inherit the same behaviour.

PMM-15518, the ticket that gives fields real `default`s, is already Pending Release, so I would drop the bridge rather than ship it. If the AC needs an answer before those schemas land, saying it in the helper text ("Defaults to /var/log/mysql/error.log when left blank") states the default without inventing a value.

```suggestion
/**
 * The value a field starts at.
 *
 * Order:
 * 1. schema `default` when set
 * 2. for required integer/float with no default — `ge`, else `1`, so a required
 *    Minutes field is not empty on first paint
 * 3. otherwise {@link emptyFieldValue}
 */
export function fieldDefault(field: PluginField): unknown {
  if (field.type === 'file') {
    return undefined;
  }
  if (hasSchemaDefault(field)) {
    // `datetime-local` cannot display a UTC ISO string; convert here so the
    // picker mounts with a real wall-clock value (PMM-15510).
    if (field.type === 'datetime') {
      return toDatetimeLocalValue(field.default);
    }
    return field.default;
  }
```

Removing it also restores the guard I would want here: with Issue Time empty the form refuses to submit instead of sending an example. Three tests assert the promoted behaviour and would need updating: `seeds defaults as real values, not grey placeholders`, `promotes placeholder to the starting value when default is unset`, and `hides the placeholder when it was promoted into the value`.

ui/packages/sep/framework/src/components/SchemaFormRenderer/utils/fieldDefault.ts
```diff
@@ -43,10 +44,69 @@ export function emptyFieldValue(field: PluginField): unknown {
...
+ * only ghost what the input already shows. A distinct `default` leaves the
+ * placeholder free to be a format hint.
+ */
+export function fieldPlaceholder(field: PluginField): string | undefined {
+  if (!isTextLike(field) || !field.placeholder) {
+    return undefined;
+  }
+  if (!hasSchemaDefault(field)) {
+    return undefined;
+  }
```
mattiasimonato: With the promotion gone the widget should ghost the placeholder again, and the note above no longer applies.

```suggestion
/** Grey placeholder text for the widget. */
export function fieldPlaceholder(field: PluginField): string | undefined {
  if (!isTextLike(field) || !field.placeholder) {
    return undefined;
  }
```

mattiasimonato (commented):
Answering decision 2 in the description rather than flagging something missed: the placeholder-to-default bridge submits example text as a real script argument. With the ticket's own Issue Time case, a form sends `e.g. 2026-01-01 00:00:00` when the user submits without touching anything, and the required check no longer stops them because the field is not empty any more. Detail and a suggestion inline. Everything else I looked at holds up.

ui/packages/sep/framework/src/components/SchemaFormRenderer/utils/fieldDefault.ts
peter-o-addo: Agreed, dropped the placeholder-to-default bridge, placeholders stay ghost text, empty required fields block submit again, and real defaults wait on PMM-15518.

ui/packages/sep/framework/src/components/SchemaFormRenderer/utils/fieldDefault.ts
peter-o-addo: Agreed, `fieldPlaceholder` shows the placeholder as grey hint text again, and I removed the old comment.

mattiasimonato (approved):

yyyyyyyan:

PR #5968, opened 2026-09-18
PMM-14976 Split QAN collect requests by encoded size

theTibi:
## Summary

- **Bug**: a QAN collection interval was dropped in full when it exceeded a gRPC message limit, on either hop. Reported as `failed to send CollectRequest to QAN: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (51547812 vs. 20971520)`.
- **Root cause**: buckets were batched by count, never by bytes.
  - pmm-managed sliced into groups of 25000 before calling qan-api2, which accepts 20MiB. A bucket is dominated by its fingerprint and example, each capped at the service's `max_query_length` rather than at any fixed size, so the guard never held — even at the default 2048 a full slice is ~22MB, which is what PMM-10029 reported on v2.36.0.
  - The agent packed every bucket of an interval into one `QANCollectRequest` with no size check; pmm-managed accepts 100MiB. Crossing that also reset the agent stream, taking metrics, actions and status down with it.
- **Fix**: both sides measure encoded size and split accordingly. A bucket too large to travel even alone is dropped on its own, with a warning naming the query and service, rather than taking the rest of the interval with it. **No limit is raised on either side** — qan-api2's own comment forbids raising its limit, and this change does not need it.

## Ticket

- PMM-14976

## Test plan

Unit:

- [x] `go test ./agent/client/ ./managed/services/qan/` — pass
- [x] New regression tests fail on `main` and on mutation: removing either split rebuilds an oversized request (`61687890 > 20971520` server side, `109398531 > 104857600` agent side)
- [x] `golangci-lint`, `gofumpt`, `gci`, `license-eye`, `go mod tidy -diff` — all clean

Live, MongoDB with `--max-query-length=20480`, ~20KB queries:

- [x] 2000 distinct queries (~2007 buckets, ~81MB) — before: `ResourceExhausted (81209890 vs. 20971520)`, 0 rows stored; after: 2007 rows
- [x] 3500 distinct queries (~3507 buckets, ~142MB) — before: `larger than max (107890621 vs. 104857600)` plus an agent stream reset, 0 rows; after: no error on either side, all rows stored
- [ ] **Feature Build** — needed before QA can test an image, since this changes both pmm-agent and pmm-managed

## Notes

- `make gen` was not run: no `.proto` is touched by this change.
- The failure is still only visible in `pmm-managed.log`; there is no metric for dropped buckets. Tracked as a follow-up.
- The RTA path (`agent/agents/supervisor/supervisor.go`) builds its `CollectRequest` the same unbounded way, on the same connection. Out of scope here, tracked as a follow-up.


theTibi:
@coderabbitai full review

4nte (approved):

PR #5976, opened 2026-09-21
PMM-15512: Read a diagnostics result as a report

nachodd:
Sits under the PMM-15217 umbrella. Addresses finding **P4** from Pedro Fernandes' UX pass over the Support diagnostics Tech Preview, recorded on PMM-15381.

Based on `PMM-15205-sep-fb`, not `main`.

## What was wrong

The executions list named each run by its script filename, so a page read as variations of `pt-mysql-summary.sh`, with one row printing a raw command line beside it. No row said when a run started, how long it took, or how long it had been going.

A successful run's report opened in a panel about fifteen lines tall, scrolled to its last line, with no expand, no full screen and no Copy — 137 lines of pt-summary output read through two scrollbars, where the only way out was Download. A "Last 1000" dropdown carried no label, the search bar announced "0 matches" before anyone had searched, and a disabled "Send to support case" was blocked by two conditions at once without saying which.

### Shared framework (`ui/packages/sep/framework`)

Used by every log-viewer consumer, **including MySQL Backups**.

- **Sizes to content.** The pane fits its output up to the caller's `height` rather than always filling it, and an expand toggle raises that ceiling to about a screen. A caller that passes a CSS string has sized the pane itself and is left alone.
- **Full screen.** The whole viewer moves into a dialog, so its tabs, search and selected step go with it rather than being re-found in a second pane overlaid on the first.
- **Opens at the top.** Tailing is now driven by the stream, not the caller's `taskStatus` prop — that prop is optional, and `SnippetExecutionAccordion` mounts the viewer against a task it has just launched without one. Reading the stream also releases the tail the moment a `finish` event lands, which a frozen status prop never would.
- **Copy beside Download**, falling back to `execCommand` where the clipboard API is absent. PMM Server is routinely reached over plain HTTP, which is not a secure context — the one deployment shape where an operator is most likely to be pasting a report into a ticket.
- **The line-cap select carries a visible label.** "Last 1000" on its own named neither what was counted nor that it could be changed.
- **No match count before a search.** LazyLog always renders `0 matches` and offers no prop to suppress it; the input's own emptiness is the only available signal.

### Support diagnostics (`ui/packages/plugins/atw`)

- Every row carries display name, host, start time, duration and a status chip in **one layout**, the duration counting up while the run is going off a single shared clock that stops when nothing is running.
- The **filename and command line move behind the expand** — they identify the script, not the run.
- A disabled **"Send to support case" names every condition** holding it back. An unconfigured receiver and an empty selection are both true on a fresh pane, and naming only one left the button just as grey after the user fixed it.

## Acceptance criteria

- [x] Every row shows display name, host, start time and duration, in one layout.
- [x] A filename or command line appears only behind the expand.
- [x] A completed report opens at the top, wraps, and can be expanded and copied without downloading.
- [x] The line-cap control is labelled or hidden, and no match count shows before a search.
- [x] A disabled send control names the condition that disabled it.
- [x] No regression in MySQL Backups, which shares the log viewer.

## One thing the backend still owes this screen

`snippet_title` and `executor_host` are read from the execution payload as **optional** fields, falling back to the filename and to nothing respectively. Neither can be reached from this pane any other way:

- The title lives on a different resource, and no endpoint that serves titles yields a complete filename-to-title map — the category listing is filtered by a presentation tag and the search endpoint is paginated, so an execution can name a snippet neither returns. The pane that renders this list fetches neither snippet collection.
- The host is recorded on the batch, not the execution, and `/api/sep/task-history/` filters only by task name and status.

The title is tracked as **PMM-15520**. **That ticket needs to carry `executor_host` too**, or the host column stays empty — this was not called out in the original scope. Start time and duration needed nothing: `started_at`, `finished_at` and `task_status` were already served, so that part was purely a rendering gap.

## Test plan

- `cd ui && make lint` — 0 errors.
- `cd ui && make format-check` — clean.
- `cd ui && make test` — 1846 passing (framework 1023, atw 145, api 106, pmm 572).
- New coverage: a finished report opens at the top and wraps; a live run tails; the stream's `finish` event releases the tail; a run with no declared status still tails; expand is offered only once output outgrows the pane; full screen moves the viewer rather than copying it; Copy works and reports itself, and degrades without crashing when the clipboard is refused; the line-cap control is reachable by its label; rows name title/host/start/duration; the filename stays out of the collapsed row; a duration counts up; a queued run is not given an invented start time; the send tooltip names both blockers.
- Three existing ATW tests asserted the command line appeared in the collapsed summary — rewritten to assert it now appears only behind the expand.

Manual verification against a live PMM Server with the side-car is still outstanding; the title and host columns cannot be seen populated until PMM-15520 lands.


fabio-silva:
@nacho conflics with base branch

ui/packages/plugins/atw/tests/ResultsPane.test.tsx
```diff
@@ -1614,3 +1621,165 @@ describe('ResultsPane live log', () => {
...
+
+    renderPane(<ResultsPane incidentId="inc-1" />);
+
+    await waitFor(() => {
+      expect(screen.getByText('pt-mysql-summary.sh')).toBeTruthy();
+    });
+  });
+
+  it('counts up while a run is still going', async () => {
+    vi.useFakeTimers({ shouldAdvanceTime: true });
```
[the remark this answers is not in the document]
nachodd: Checked this one by running it: `shouldAdvanceTime: true` is load-bearing. Dropping it exactly as suggested makes the test time out at 5000ms before the first assertion — the row never renders, because the query behind it needs the clock to move for its fetch to settle, and `await act(async () => { await Promise.resolve(); })` does not move it.

The residual flake risk you describe is real but bounded: the assertions read a duration rounded to one decimal, so it takes 50ms of real time between `setSystemTime` and the first `getByText` to shift `5.0s` to `5.1s`. Leaving as is rather than trading a rare flake for a test that never passes.

ui/packages/sep/framework/src/components/TaskLogViewer/TaskLogViewer.tsx
```diff
@@ -603,16 +783,20 @@ export function TaskLogViewer({
...
       </Box>
 
-      <Box sx={{ flex: 1, minHeight: 0 }}>
+      <Box sx={{ flex: fullScreen ? 1 : 'none', minHeight: 0 }}>
         <LogOutputPane
           text={currentPaneText}
           wrap={wrap}
-          height={height}
+          height={paneHeight}
+          follow={!hasFinished}
```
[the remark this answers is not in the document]
nachodd: Not taking this one. The caller you point at is the drawer branch that renders when there is **no run record** (`!run && showLog`), so it has no basis to assert the run finished — passing `follow={false}` there would hard-code an assumption the caller cannot make, and would be wrong for a task that is still going.

Keying on `hasFinished` is deliberate, and `taskStatus` being optional is the reason for it rather than a gap in it: `SnippetExecutionAccordion` mounts the viewer against a just-launched task without a status, and a `follow` that keyed on the status prop would leave that live log sitting still while output arrived. The stream `finish` frame is the signal that actually arrives for both callers, and it already releases the tail.

The edge that remains — an already-finished run opened without a status tails briefly before its `finish` frame lands, so it settles at the bottom rather than the top — is real and worth its own ticket. It is not something this PR introduced.

ui/packages/sep/framework/src/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,125 @@
...
+    if (globalThis.isSecureContext && navigator.clipboard) {
+      try {
+        await navigator.clipboard.writeText(text);
+        ok = true;
+      } catch {
+        ok = false;
+      }
+    }
+    if (!ok) {
+      ok = copyViaExecCommand(text);
```
fabio-silva, line 100: We do not use `execCommand` on PMM. This command has been marked as insecure and deprecated years ago.
When clipboard is not available, we just point that to users (see `QueryAnalytics.tsx` or `QanHeaderActions.tsx`)

ui/packages/plugins/atw/src/ResultsPane.tsx
```diff
@@ -107,6 +124,77 @@ const SEND_STATUS_COLORS = {
...
+  if (Number.isNaN(started)) {
+    return null;
+  }
+  if (execution.finished_at) {
+    const finished = new Date(execution.finished_at).getTime();
+    if (!Number.isNaN(finished)) {
+      return (finished - started) / 1000;
+    }
+  }
+  if (!isRunning(execution)) {
```
fabio-silva, line 167: this check should come in the beginning of the function, together with `if (!execution.started_at)`

ui/packages/plugins/atw/src/ResultsPane.tsx
```diff
@@ -613,6 +714,8 @@ function ExecutionRow({
   onEditParameters,
 }: {
   execution: AtwIncidentExecution;
+  /** How long this run has taken, in seconds; `null` when it never started. */
```
fabio-silva, line 717: the variable name already states what's its purpose

ui/packages/plugins/atw/src/ResultsPane.tsx
```diff
@@ -643,6 +750,18 @@ function ExecutionRow({
...
 
+  // The name the user picked in the Collect pane, which is what they are
+  // looking for here. The filename is the fallback, not the label: it is the
+  // script's identity, not the run's, and every row would otherwise read as
+  // some variation of `pt-mysql-summary.sh`.
+  const displayName = snippet_title?.trim() || snippet_filename;
+
+  // A run that has not started has no start time, so the row falls back to when
+  // the execution was recorded — the moment the operator pressed Run — and says
+  // so, rather than leaving the column blank for a queued run.
```
fabio-silva, line 761: again, this three-liner comment is stating what can be infered by the relatively simple code

ui/packages/plugins/atw/src/ResultsPane.tsx
```diff
@@ -671,30 +790,64 @@ function ExecutionRow({
...
-                }}
-              >
-                {masked_args}
-              </Typography>
-            )}
           </Box>
+          <Typography
+            variant="body2"
+            color="text.secondary"
+            title={executor_host ?? undefined}
```
fabio-silva, line 806: `title` is already `undefined` by default

ui/packages/plugins/atw/src/ResultsPane.tsx
```diff
@@ -671,30 +790,64 @@ function ExecutionRow({
...
+          </Typography>
+          <Typography
+            variant="body2"
+            color="text.secondary"
+            title={
+              startStamp
+                ? startIsQueue
+                  ? `Queued ${startStamp.title}; not started yet`
+                  : startStamp.title
+                : undefined
```
fabio-silva, line 825: same as above

ui/packages/plugins/atw/src/ResultsPane.tsx
```diff
@@ -107,6 +124,77 @@ const SEND_STATUS_COLORS = {
...
+
+/**
+ * How long the run took, in seconds, or `null` when it cannot be known.
+ *
+ * A finished run is measured between the two recorded instants, so it never
+ * moves again. A running one is measured against `now`, which the caller ticks
+ * — the wire carries no elapsed time, and a row that reports nothing while a
+ * pt-summary works through a large instance is the complaint this answers.
+ *
+ * `null` for a run that never started: `formatDuration` renders that as an
```
fabio-silva, line 147: The `runSeconds` fn (or the comment, in this case) doesn't need to know what `formatDuration` renders. It only cares to return null, everything else is non of its business.

ui/packages/plugins/atw/src/ResultsPane.tsx
```diff
@@ -737,6 +890,30 @@ function ExecutionRow({
           )}
         </Stack>
 
+        {/*
+          The filename and the command line live here, not in the row: they
+          identify the script rather than the run, and a list of them is what
+          made the Results pane unreadable. An operator who needs either — to
+          reproduce the run in a terminal, or to check what it was given — is
+          already looking at one execution.
+        */}
```
fabio-silva, line 899:
```suggestion

```

ui/packages/plugins/atw/tests/ResultsPane.test.tsx
```diff
@@ -178,19 +184,22 @@ describe('ResultsPane recorded arguments', () => {
...
     );
 
     renderPane(<ResultsPane incidentId="inc-1" />);
 
     await waitFor(() => {
-      expect(screen.getByText(MASKED_ARGS)).toBeTruthy();
+      expect(screen.getByText('diag/mongo.sh')).toBeTruthy();
     });
+    // The filename is the row's fallback label here (this fixture carries no
+    // title), but the command line has no business in a list of runs at all.
```
fabio-silva, line 198:
```suggestion
```

ui/packages/plugins/atw/tests/ResultsPane.test.tsx
```diff
@@ -1614,3 +1621,165 @@ describe('ResultsPane live log', () => {
...
+    expect(screen.getByText('db-node-1')).toBeTruthy();
+    // 10:00:00 to 10:00:42, held by the wire and not by a ticking clock.
+    expect(screen.getByText('42.0s')).toBeTruthy();
+    expect(
+      screen.getByTitle(
+        `${new Date('2026-07-22T10:00:00Z').toLocaleString()} (${browserTimezone()})`
+      )
+    ).toBeTruthy();
+    // The script's filename identifies the script, not the run — it belongs
+    // behind the expand, and the pane opened with every row collapsed.
```
fabio-silva, line 1672:
```suggestion
```

ui/packages/sep/framework/src/components/TaskLogViewer/__tests__/TaskLogViewer.test.tsx
```diff
@@ -84,6 +96,12 @@ describe('TaskLogViewer', () => {
     vi.unstubAllGlobals();
     vi.clearAllMocks();
     globalThis.localStorage.clear();
+    // The copy cases redefine this; leaving one test's stub in place would let
+    // a later one assert against a clipboard it never installed.
```
fabio-silva, line 100:
```suggestion
```

ui/packages/sep/framework/src/components/TaskLogViewer/__tests__/TaskLogViewer.test.tsx
```diff
@@ -1250,4 +1268,204 @@ describe('TaskLogViewer', () => {
...
+  }
+
+  it('opens a finished report at the top and only tails a live run', async () => {
+    await renderWithOutput('501', 'SUCCESS');
+    expect(screen.getByTestId('log-output')).toHaveAttribute(
+      'data-follow',
+      'false'
+    );
+    // Wrapping is what makes a 137-line summary readable without a horizontal
+    // scrollbar, so it is on from the first paint, not after a toggle.
```
fabio-silva, line 1308:
```suggestion
```

ui/packages/sep/framework/src/components/TaskLogViewer/LogOutputPane.tsx
```diff
@@ -20,12 +20,39 @@ import Box from '@mui/material/Box';
...
   text: string;
   wrap: boolean;
   enableSearch?: boolean;
   height?: number | string;
   emptyLabel?: string;
+  /**
+   * Whether the pane sticks to the last line as text arrives. Off by default:
+   * a finished report is read from the top, and only a run still producing
+   * output wants tailing.
+   */
```
fabio-silva, line 54:
```suggestion
```

ui/packages/sep/framework/src/components/TaskLogViewer/TaskLogViewer.tsx
```diff
@@ -113,6 +146,13 @@ function logTailChoiceToParam(choice: LogTailLineChoice): number | undefined {
 export interface TaskLogViewerProps {
   taskHistoryId: number | string;
   taskStatus?: string;
+  /**
+   * Tallest the output pane gets before the user expands it. A number is a
```
fabio-silva, line 150: if this is true, why is it called `height` and not `maxHeight`?

ui/packages/sep/framework/src/components/TaskLogViewer/TaskLogViewer.tsx
```diff
@@ -419,6 +481,32 @@ export function TaskLogViewer({
...
+  const handleCopy = () => {
+    clipboard.copy(currentPaneText);
+  };
+
+  // A caller that passed a CSS string has sized the pane itself; only a numeric
+  // ceiling is something this viewer may fit to content or raise.
+  const sizable = typeof height === 'number';
+  const baseMaxHeight = sizable ? height : 0;
+  // Expanding never shrinks a pane whose caller already asked for more than the
+  // expanded cap.
```
fabio-silva, line 494:
```suggestion
```

ui/packages/sep/framework/src/components/TaskLogViewer/TaskLogViewer.tsx
```diff
@@ -577,6 +714,49 @@ export function TaskLogViewer({
               </IconButton>
             </span>
           </Tooltip>
+          {/*
+            Hidden in full screen, where the pane already takes the whole
+            viewport and the ceiling it raises no longer applies.
+          */}
```
fabio-silva, line 720:
```suggestion
```

fabio-silva (changes requested):

PR #5977, opened 2026-09-21
PMM-15511: Make a diagnostics run visible where it starts

nachodd:
## What

UX finding **P3** on PMM-15381 (Support diagnostics Tech Preview), under PMM-15217.

With one or two scripts selected the Collect column ran past two screen heights — every script section opened expanded, so **Execute batch** ended up at the very bottom while the Results column beside it stayed empty. Pressing it changed nothing where the user was looking: no toast, no scroll, no change on the button. The new executions appeared off-screen at the top of the other column, and the form stayed open and filled as if nothing had been submitted. A user who does not think to scroll up concludes the click did nothing, and runs again.

## Changes

**Collect form**
- Script sections open **collapsed**, each header naming its current values on one line (`Minutes: 30 · Verbose: on`, or `Using default values`).
- The run button is **pinned to the top of the form**, offset by PMM's own sticky app bar, so it is reachable at any number of selected scripts.
- While the batch is in flight the button reads **Starting…** and is disabled.

**After a run starts**
- The form gives way to a confirmation — *"Started 3 scripts. Their progress is in Results."* — with **Change parameters and run again**, which brings the form back filled in with what was just submitted.
- A toast names how many scripts started.
- The Results pane scrolls into view, resets to the first page, and outlines the new executions for 8 seconds.

**Sections that cannot safely collapse.** A collapsed shell mounts none of its children (`unmountOnExit`) while the renderer seeds every leaf in the schema regardless. So a section opens anyway when it holds a required field with no default (never validated, never seen), a `one_of` group (its own slot is what unregisters the branches you did not pick, so every branch's default would ship at once), or a gated field (`requires` / `forbidden` are evaluated by the field's own slot). ATW snippet schemas declare none of the last two today.

**Shared form renderer** (`ui/packages/sep/framework/`, also used by MySQL Backups) gains three opt-in props: `submitPlacement`, `stickySubmitOffset`, and `sectionValueSummary`. All three default to today's behaviour, so no other plugin changes.

## Columns versus stacking — decision

The ticket asked to settle this and record it. **Collect and Results stay side by side above `md`.** Stacking answered the symptom — a Collect column taller than two screens — which is now fixed at its source, and side by side is the only arrangement where a run is watched while the form that started it is still there. Stacking would put Results below the fold permanently, for every run. Recorded in a comment on the grid in `IncidentWorkspacePage.tsx`.

## Acceptance criteria

- [x] The run button is reachable without scrolling past the form, at any number of selected scripts.
- [x] Pressing it produces a visible change at the point of click — the button's own running state, then the confirmation in the same column.
- [x] After a run starts, the form is collapsed and the new executions are in view.
- [x] Script sections open collapsed with a readable summary of their values.
- [x] The column-versus-stack decision is recorded and the implementation matches it.

## Test plan

Reviewed the diff and fixed four issues it turned up — inactive `one_of` branches shipping from a collapsed section, a conditionally-required field staying hidden, a late batch result overwriting a selection that had moved on, and a second batch missing the pagination reset. Each has a test that fails without its fix.

New coverage:
- `SchemaFormRenderer.collapsedForm.test.tsx` — summary content, only while collapsed, tracks edits, opt-in; sticky placement, offset, and that it still submits.
- `sectionValueSummary.test.ts` — the formatter, including choice labels, off switches, and the overflow count.
- `CollectPane.runFeedback.test.tsx` — collapsed sections and the three that stay open; the running label; the confirmation and its count; values surviving a reopen; a failed batch keeping the form; a late result not pinned onto a new selection.
- `ResultsPane.test.tsx` — the new-execution outline and the page reset, including a second batch inside the first one's highlight window.
- `IncidentWorkspacePage.test.tsx` — the toast, the scroll, and the marked-out rows.

Green locally: `cd ui && make lint` (0 errors), `make format-check`, `make test` — 628 app + 1061 framework + 164 ATW + 106 api tests.

**Feature build:** to follow.


fabio-silva:
@nachodd some conflicts with base branch

ui/packages/plugins/atw/src/IncidentWorkspacePage.tsx
```diff
@@ -74,27 +79,81 @@ export function IncidentWorkspacePage() {
...
           if (
             item.task_history_id !== null &&
             item.task_history_id !== undefined
           ) {
+            startedIds.push(item.task_history_id);
             next.set(item.task_history_id, { snippets, values });
           }
         }
         return next;
       });
```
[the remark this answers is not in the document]
nachodd: Fixed in 5da9ee377 — the list is derived from the response before the updater, and the updater is left pure.

The failure mode is the empty one rather than the doubled one: React only evaluates an updater eagerly when the fiber has nothing queued, so with an update already pending the list is read empty and the toast, the highlight and the scroll to Results are all skipped on a run that did start scripts. No test came with the fix, because the eager path is the one that runs in practice — a test for it passes against the old code too — and forcing the other path means driving the update queue into a state no public API exposes.

ui/packages/plugins/atw/src/CollectPane.tsx
```diff
@@ -171,6 +169,52 @@ export function omitCliUsageTrapFields(fields: SectionField[]): SectionField[] {
...
+ * - a required field with no default is neither validated nor visible, and the
+ *   batch goes out missing an argument nobody was shown;
+ * - a `one_of` group's inactive branches are unregistered by the group's own
+ *   slot, so a closed section submits every branch's seeded default at once;
+ * - a gated field's `requires` / `forbidden` predicates are evaluated by the
+ *   field's own slot, so a gate that would make it required does nothing.
+ *
+ * ATW snippet schemas declare neither groups nor gates today — the gated-field
+ * warning below fires loudly if one appears — so in practice this is the
+ * required-field rule.
```
fabio-silva: AI generated, search for unuseful comments: this closing paragraph narrates the *current* state of ATW snippet schemas ("declare neither groups nor gates today") and cross-references "the warning below" — it'll go stale silently the day a schema adds a group or gate. The opening two sentences (mount-vs-seed insight) and the `one_of`/gated bullets above already carry the load-bearing information; consider dropping this paragraph.

ui/packages/plugins/atw/src/CollectPane.tsx
```diff
@@ -330,6 +374,15 @@ export function CollectPane({
   const [rerunResolveFailed, setRerunResolveFailed] = useState<string | null>(
     null
   );
+  // How many scripts the last batch started, set only while its confirmation
+  // is standing in for the form. Non-null is what collapses the form: a
+  // dispatch that changed nothing where the reader was looking is the whole
+  // defect this replaces (PMM-15511).
```
fabio-silva: AI generated, search for unuseful comments: "a dispatch that changed nothing where the reader was looking is the whole defect this replaces (PMM-15511)" narrates the bug this PR fixes rather than documenting current behavior — that belongs in the PR description, not here. Worth keeping: "Non-null is what collapses the form" — that's the real, non-obvious invariant.

ui/packages/plugins/atw/src/IncidentWorkspacePage.tsx
```diff
@@ -213,6 +272,15 @@ export function IncidentWorkspacePage() {
         </Box>
       )}
 
+      {/*
+        Two columns above `md`, decided on PMM-15511 against stacking Collect
+        over Results at the standard content width. The complaint stacking
```
fabio-silva: AI generated, search for unuseful comments: "decided on PMM-15511 against stacking Collect over Results" cites the ticket rather than documenting the decision itself. The rest of the comment (why stacking was rejected, why side-by-side) is genuinely useful context worth keeping — just the ticket citation reads as changelog rather than code documentation.

PR #5983, opened 2026-09-22
PMM-15459: Keep the parent nav item active on nested SEP routes

peter-o-addo:
## What

UX finding P10 on PMM-15380 (MySQL Backups Tech Preview), under PMM-15217.

On a backup detail page (and the same pattern on Support diagnostics detail routes) the sidebar highlighted Home page and collapsed Management. The user was inside a SEP app and the nav said otherwise.

- `NAV_SEP_ATW` and `NAV_SEP_MYSQL_BACKUPS` in `ui/apps/pmm/src/contexts/navigation/navigation.constants.ts` set matches to their own url.
- In `isActive` (`ui/apps/pmm/src/utils/navigation.utils.ts`) a matches entry other than `*` is compared with end true, so those entries only repeat the exact match the url already gives.
- Only the literal `*` widens to path + `/*`. Both apps mount at PATH/*, so every nested route left no nav item active.
- Every other PMM nav item with nested routes already uses matches `*`. These two were the only ones repeating their own url.

## Changes

- Replace the self-referential matches entry on both SEP nav items with `*`.
- No change to `isActive` itself.

## Acceptance criteria

- A nested MySQL Backups route keeps MySQL Backups active and Management expanded.
- A nested Support diagnostics route does the same.
- Unit coverage in `navigation.provider.test.tsx` pins the nested-route case for both entries (root and nested paths). Expectation updates in `navigation.utils.test.ts`.

## Test plan

- `cd ui` then run the navigation Vitest files covering provider, utils, and findActiveNavItem.
- `make lint` and `make format-check`.


peter-o-addo:
image

nachodd (approved):
Reviewed and verified locally — correct, minimal fix. Approving; the notes below are non-blocking.

## Verification

Applied the branch onto a clean tree and ran the real suites rather than reading only:

- `vitest run src/components/sidebar src/contexts/navigation src/utils/navigation.utils.test.ts` → 49 passed / 6 files.
- Negative control: reverted only `navigation.constants.ts` while keeping the new tests → 4 failures, including both nested cases (`expected undefined to be 'sep-atw'`). The added tests genuinely pin the regression.
- `oxfmt --check` clean on the touched directory. The `oxlint` warnings in that directory (`consistent-type-imports`) are pre-existing and in files this PR does not touch.

## The diagnosis holds

- In `matchesUrl` (`ui/apps/pmm/src/utils/navigation.utils.ts`), a `matches` entry other than `*` is compared with `end: true`, so `matches: [SEP_ATW_PATH]` only repeated the exact match the `url` already produces.
- Only `*` widens to `path + '/*'`, and the router mounts both SEP apps as splats (`router.tsx`, `${relativeToNav(SEP_ATW_PATH)}/*` and the MySQL Backups equivalent), so every nested route left no nav item active.
- The consistency claim checks out: every other nav item with nested routes already uses `*`; these two were the only self-referential ones.
- One refinement to the description's mechanism: `findActiveNavItem` returns `undefined` for a nested SEP path, and it is `Sidebar.tsx`'s "keep previous item active if there isn't a match" plus `useState(navTree[0])` that leaves Home page highlighted on a fresh deep link. Symptom as described, cause is the fallback.

No over-match risk: `/sep/atw/*` requires a segment boundary, the two SEP paths are disjoint, and both items sit below Management, so the deepest-match rule in `findActiveNavItem` keeps them ahead of any root-level `*`. MySQL Restore shares the same `routeBase`, so it is covered too.

## Non-blocking notes

1. **DCO** — only `00062907e` carries a `Signed-off-by` trailer; the other three commits do not. Worth re-signing (`git commit -s`) before merge.
2. **PR title** — the repo convention is `PMM-XXXX Short summary` (no colon, summary up to ~50 chars). Squash merge takes the PR title, so this one becomes the commit subject as-is.
3. **Description wording** — "Expectation updates in `navigation.utils.test.ts`" is ambiguous: two files share that basename. The changed one is `contexts/navigation/navigation.utils.test.ts`; the `utils/` one already covers `*` semantics generically and is untouched.
4. **Assertion value** — both `matches: ['*']` expectation updates just re-assert the constant against itself. That is the existing pattern in those files, so no change requested; the deep-link tests in `navigation.provider.test.tsx` are the ones carrying the weight.

## Possible follow-up (out of scope)

`matches` has a quiet footgun: any entry other than `*` is exact-only, so a self-referential entry type-checks and does nothing — which is how this bug got in. A follow-up could either skip an entry equal to `item.url` in `isActive`, or support a uniform per-entry `/*` suffix.

fabio-silva (approved):

